PR #721, opened 2021-03-05
feat: widget front-end container components

handorff:
**Asana task**: [Container components](https://app.asana.com/0/1185117109217413/1200023663996429)

This PR defines components for the screens and flex zones.

### DIFF
--- a/assets/src/components/v2/widget.tsx
+++ b/assets/src/components/v2/widget.tsx
@@ -1,6 +1,18 @@
 import React from "react";
 
-const TYPE_TO_COMPONENT = {};
+import Normal from "Components/v2/screen/normal";
+import Takeover from "Components/v2/screen/takeover";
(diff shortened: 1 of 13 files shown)


assets/src/components/v2/widget.tsx
```diff
@@ -10,7 +22,7 @@ const Widget = ({ data }) => {
     return <Component {...props} />;
   }
 
-  return <></>;
+  return <>{type}</>;
```
jzimbel-mbta, line 25: Just for debug purposes for now, right?

jzimbel-mbta (approved):

assets/src/components/v2/widget.tsx
handorff: Yeah. I'll file a (very small) task to clean it up, just so that we don't forget.

PR #739, opened 2021-03-10
fix: small widget backend fixes

handorff:
**Asana task**: [Verify backend](https://app.asana.com/0/1185117109217413/1200038952107753)

- Clean up some types in `Screens.V2.ScreenData`
- Rename candidate generator `candidate_templates` -> `candidate_template`
- Define `candidate_template` for `Screens.V2.CandidateGenerator.BusShelter`
- Add header and footer widget instances to `BusShelter.candidate_instances`
- Add missing `Template.slot_combinations` call to `pick_instances`

### DIFF
--- a/lib/screens/v2/candidate_generator.ex
+++ b/lib/screens/v2/candidate_generator.ex
@@ -3,7 +3,7 @@ defmodule Screens.V2.CandidateGenerator do
 
   alias Screens.V2.ScreenData
 
-  @callback candidate_template() :: Screens.V2.Template.template()
+  @callback screen_template() :: Screens.V2.Template.template()
 
   @callback candidate_instances(ScreenData.config()) :: ScreenData.candidate_instances()
 end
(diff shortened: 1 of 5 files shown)


lib/screens/v2/candidate_generator.ex
```diff
@@ -3,7 +3,7 @@ defmodule Screens.V2.CandidateGenerator do
 
   alias Screens.V2.ScreenData
 
-  @callback candidate_templates() :: Screens.V2.Template.template()
+  @callback candidate_template() :: Screens.V2.Template.template()
```
jzimbel-mbta, line 6: I don't want to make you do another rename, but is it still a candidate if there's only one and it's always used? Maybe `screen_template`?

jzimbel-mbta (approved):
One optional naming suggestion, otherwise good to go.

lib/screens/v2/candidate_generator.ex
handorff: Good point, made the change.

PR #740, opened 2021-03-10
feat: Normal header widget

jzimbel-mbta:
**Asana task**: [normal header](https://app.asana.com/0/1185117109217413/1199987812749015/f)

Basic normal header widget with expected props matching the fields of the `NormalHeader` widget instance struct on the backend. It displays the ISO8601 timestamp directly for now rather than formatting it in a human-readable way.

### DIFF
--- a/assets/css/v2.scss
+++ b/assets/css/v2.scss
@@ -1,3 +1,5 @@
+@import "v2/header/normal";
+
 body {
   margin: 0px;
 }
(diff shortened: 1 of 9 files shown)


handorff (changes requested):
I think you also need to add this to `TYPE_TO_COMPONENT` in `widget.tsx`. (e.g. see the container components in #721.)

handorff (approved):
Thanks!

PR #775, opened 2021-03-24
feat: setup frontend for widgetized screen apps

handorff:
**Asana task**: [Frontend setup](https://app.asana.com/0/1185117109217413/1200050362904579)

This PR updates the webpack config to build a JS and CSS bundle for each v2 screen app and defines the top-level TSX and SCSS files for each. It also defines a `Widget` switchboard component for each screen type (which doesn't contain any widgets yet).


jzimbel-mbta (commented):
What do you think about defining one base Widget component that takes as props the `TYPE_TO_COMPONENT` mapping? Then the screen-specific widget components could just pass that into it rather than duplicating all of the logic.

```tsx
// v2/widget.tsx
type WidgetData = ...;

interface Props {
  data: WidgetData;
  typeToComponent: Record<string, React.ComponentType<any>>;
}

const Widget: React.ComponentType<Props> = ({ data, typeToComponent }) => ...;
```

```tsx
// v2/bus_shelter/widget.tsx
import Widget, { WidgetData } from "v2/widget";

const TYPE_TO_COMPONENT: Record<string, React.ComponentType<any>> = {...};

interface Props {
  data: WidgetData;
}

const BusShelterWidget: React.ComponentType<Props> = ({ data }) => {
  return <Widget data={data} typeToComponent={TYPE_TO_COMPONENT} />;
};
```

handorff:
As discussed offline, I switched to a single `Widget` component which uses React context to set `TYPE_TO_COMPONENT` mapping.

jzimbel-mbta (approved):

PR #786, opened 2021-03-29
Fix discrepancies between frontend/backend DUP override config

jzimbel-mbta:
**Asana task**: [BUG: DUP override config mismatched with React component props/CSS classes](https://app.asana.com/0/1185117109217413/1200059432166862/f)

There are a few breaking changes in the API response shape for DUP alert screens from from this. While it's not possible to predict disruptions, we should try to coordinate with Outfront to deploy the new frontend package during a period where there are no active disruptions at the relevant stations.

- [ ] Needs version bump?


assets/css/dup/free_text.scss
```diff
@@ -95,6 +109,13 @@
   }
 }
 
+.free-text__text-pill__text {
+  display: inline-block;
+  font-family: Helvetica, sans-serif;
+  font-weight: 700;
+  transform: translateY(6px);
```
handorff, line 116: I'd bump this up a bit -- to me it doesn't look centered right now (e.g. look at http://localhost:4000/screen/DUP-SullivanSquare/0).

assets/css/dup/free_text.scss
```diff
@@ -39,8 +39,13 @@
   font-family: Helvetica, sans-serif;
   font-weight: 700;
   text-align: center;
+  transform: translateY(-20px);
```
handorff, line 42: Why are these necessary? I don't think the Helvetica leading issue would affect the positions of these, is there a separate problem?

assets/css/dup/free_text.scss
```diff
@@ -60,15 +65,24 @@
...
+  font-weight: 700;
+  text-align: center;
+  transform: translateY(6px);
+}
+
 .free-text__text-pill {
   display: inline-block;
   font-family: Helvetica, sans-serif;
   font-weight: 700;
+  transform: translateY(-20px);
```
handorff, line 85: Same as above.

handorff (changes requested):
In addition to these, we may want to add support for inline icons in a separate PR (and add the circular GL branch icons as new icons).

assets/css/dup/free_text.scss
jzimbel-mbta: Yeah, the pill itself sat a bit low on the line, so I broke the text element out into a child div and adjusted both the pill and the text's vertical alignment to look correct.

handorff (approved):

PR #788, opened 2021-03-30
feat: support for inline icons in free text

handorff:
**Asana task**: [Inline icons in Free Text](https://app.asana.com/0/1185117109217413/1200131817716379)


lib/screens/config/dup/override/free_text.ex
```diff
@@ -7,6 +7,7 @@ defmodule Screens.Config.Dup.Override.FreeText do
           | %{route: route_pill}
           | %{color: color, text: String.t()}
           | %{special: special}
+          | %{icon: Screens.Config.Dup.Override.FreeTextLine.icon()}
```
jzimbel-mbta, line 10: Would it make sense to remove the B/C/D/E options for the `route_pill` type below now, since we have the circular icon versions instead? Same thing for the frontend `RoutePill` component.

handorff: We sometimes use GL·B pills, so maybe it should be that instead? I can change the text and make the font size smaller for those.

jzimbel-mbta (approved):

PR #794, opened 2021-04-01
feat: widget frontend templates

handorff:
**Asana task**: [Frontend templates](https://app.asana.com/0/1185117109217413/1200050362904589)

This PR also cleans up a lot of react components and styles associated with our proof-of-concept implementation. It also adds our simplified bus shelter screen config (stop_id only) to the admin tool.


jzimbel-mbta (approved):
As discussed on slack, consider merging all of the `TakeoverScreen` components into one until they need to have different behavior per screen type. Otherwise, looks good to go.

PR #797, opened 2021-04-06
fix: Use display: block for DUP stacked route pill parts

jzimbel-mbta:
**Asana task**: ad hoc

`display: flow` is not a valid property on Chrome. This fix makes the stacked slashed bus route pills display correctly on both Chrome and Firefox. (And hopefully on the DUP screens)

We'll need to make a new zip to get this change in.

- [ ] Needs version bump?


handorff (approved):
✅

PR #805, opened 2021-04-08
feat: automatically handle multiple westbound shuttles at Kenmore

handorff:
**Asana task**: [Support for multiple shuttles at Kenmore](https://app.asana.com/0/1185117109217413/1200132302343957)

In addition to the partial and full-screen alerts specified in the linked Asana task, this PR sets a headway message for the Green Line when all three branches have shuttling (since, in this case, Kenmore is a temporary mid-route terminal and won't have any predictions). This is consistent with other situations where we have DUPs at mid-route shuttle endpoints.

Note: This PR assumes that each line's shuttle has its own alert. I believe this has been true in the past (when working on Beacon Junction outside of Kenmore, the C and D had their own shuttles), and Alerts UI doesn't seem to provide a way to specify stops on multiple routes as a part of the same alert.

### DIFF
--- a/lib/screens/dup_screen_data/special_cases.ex
+++ b/lib/screens/dup_screen_data/special_cases.ex
@@ -118,38 +118,85 @@ defmodule Screens.DupScreenData.SpecialCases do
          %{effect: :shuttle, region: :boundary, routes: ["Green-B"], headsign: "Boston College"},
          %{effect: :shuttle, region: :boundary, routes: ["Green-C"], headsign: "Cleveland Circle"}
        ]) do
+    kenmore_bc_partial_alert_text()
+  end
+
+  defp kenmore_partial_alert_text([
+         %{effect: :shuttle, region: :boundary, headsign: "BC/Clev. Circ."}
+       ]) do
+    kenmore_bc_partial_alert_text()
+  end
+
+  defp kenmore_partial_alert_text([
+         %{effect: :shuttle, region: :boundary, routes: ["Green-B"], headsign: "Boston College"},
+         %{effect: :shuttle, region: :boundary, routes: ["Green-D"], headsign: "Riverside"}
+       ]) do
+    kenmore_bd_partial_alert_text()
+  end
+
+  defp kenmore_partial_alert_text([
+         %{effect: :shuttle, region: :boundary, headsign: "BC/Riverside"}
+       ]) do
+    kenmore_bd_partial_alert_text()
+  end
+
+  defp kenmore_partial_alert_text([
+         %{
+           effect: :shuttle,
+           region: :boundary,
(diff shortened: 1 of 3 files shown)


lib/screens/dup_screen_data/special_cases.ex
```diff
@@ -0,0 +1,337 @@
...
+      ) do
+    alerts = Request.fetch_alerts(stop_ids, route_ids)
+
+    interpreted_alerts =
+      alerts
+      |> Enum.map(fn alert ->
+        Map.put(
+          Data.interpret_alert(alert, stop_ids, pill),
+          :routes,
+          Data.alert_routes_at_station(alert, stop_ids)
```
jzimbel-mbta, line 34: Just to make this slightly clearer could you use pipes? `alert |> Data.interpret_alert(...) |> Map.put(...)`

lib/screens/dup_screen_data/special_cases.ex
```diff
@@ -0,0 +1,337 @@
...
+
+  def handle_special_cases(_, _), do: nil
+
+  defp render_kenmore_alerts(interpreted_alerts, primary_departures, "0") do
+    current_time = DateTime.utc_now()
+
+    case kenmore_partial_alert_text(interpreted_alerts) do
+      {:ok, text} ->
+        line = %FreeTextLine{icon: :warning, text: text}
+        override = %PartialAlertList{alerts: [%PartialAlert{color: :green, content: line}]}
```
jzimbel-mbta, line 50: Nice, I like the idea of using an automatically generated "manual override".

lib/screens/dup_screen_data/special_cases.ex
```diff
@@ -0,0 +1,337 @@
...
+  defp kenmore_fullscreen_alert_text(_), do: nil
+
+  defp kenmore_bc_fullscreen_alert_text do
+    [
+      "No",
+      %{icon: :green_b},
+      %{format: :bold, text: "Boston Coll"},
+      "or",
+      %{icon: :green_c},
+      %{format: :bold, text: "Cleveland Cir"},
```
jzimbel-mbta, line 298: If I remember correctly, you might need an explicit line break (`%{special: :break}`) to keep these two parts on the same line. Feel free to ignore if you already checked all of these cases.

jzimbel-mbta (commented):
Looks good. I'm as unsure as you are about using the exact custom headsign strings, but the only alternative approach I can think of right now is to change the headsign matcher to produce a list of atoms or something, and then keeping a mapping from those to the custom headsign string. Then in this Kenmore special case code, you could work off of the atom list (which is less likely to change) rather than the "rider-facing" strings.

jzimbel-mbta (approved):

PR #826, opened 2021-04-27
feat: create v2 config

handorff:
**Asana task**: [Create V2 config for all screen types](https://app.asana.com/0/1193769697792450/1200233815812785)

Notes: This PR removes the v1 bus shelter config, combines the v2 GL e-Ink Single and Double into one GL e-Ink app, and doesn't yet define a v2 DUP app.


lib/screens/v2/candidate_generator/gl_eink.ex
```diff
@@ -0,0 +1,90 @@
...
+       normal: [
+         :header,
+         :main_content,
+         :medium_flex,
+         :footer
+       ],
+       bottom_takeover: [
+         :header,
+         :main_content,
+         :bottom_screen
```
jzimbel-mbta, line 24: Another thought: as we build these templates out, I'm wondering what the best way is to keep track of the slot IDs available in the templates and returned by the widgets' `slot_names`, in one place. For example, looks like this `:bottom_screen` slot only exists on the e-ink screen templates.

lib/screens/v2/candidate_generator/solari_large.ex
```diff
@@ -26,7 +26,10 @@ defmodule Screens.V2.CandidateGenerator.SolariLarge do
   end
 
   defp header_instances(config, now) do
-    %Screen{app_params: %Solari{station_name: header_text}} = config
-    [%NormalHeader{screen: config, text: header_text, time: now}]
+    %Screen{
+      app_params: %SolariLarge{header: %Header{type: :current_stop_name, stop_name: stop_name}}
```
jzimbel-mbta, line 30: Hmm. Should assertions like `type: :current_stop_name` happen here? I can't think of a better way to do it if we want to have one shared header config but only support one variant of it for Solari screens, but it feels iffy to have these subtle constraints on the config spread about.

lib/screens/v2/screen_data.ex
```diff
@@ -212,7 +209,7 @@ defmodule Screens.V2.ScreenData do
   def resolve_paging(layout_and_instances, refresh_rate, now \\ DateTime.utc_now())
 
   def resolve_paging(layout_and_instances, nil, _now) do
-    layout_and_instances
+    Tuple.append(layout_and_instances, %{})
```
jzimbel-mbta, line 212: Oops, thanks for catching this.

test/screens/v2/candidate_generator/solari_large_test.exs
```diff
@@ -1,17 +1,20 @@
...
+    config = %Screen{
+      app_params: %V2.SolariLarge{
+        departures: %V2.Departures{sections: []},
+        header: %V2.Header{type: :current_stop_name, stop_name: "Ruggles"}
+      },
+      vendor: :gds,
       device_id: "TEST",
       name: "TEST",
-      app_id: :solari_large
+      app_id: :bus_eink_v2
```
jzimbel-mbta, line 17: I think this should be `:solari_large_v2`

jzimbel-mbta (commented):
Would you mind also updating the test config at test/fixtures/config.json with some examples of the new config shapes? I believe the test that uses it can remain unchanged since it just generally checks serialization/deserialization.

lib/screens/v2/candidate_generator/gl_eink.ex
handorff: We could use types for this purpose. I'm also not sure we need all of the different configurations that currently exist (for instance, we've never actually supported or used this `bottom_takeover` layout).

test/screens/v2/candidate_generator/bus_shelter_test.exs
```diff
@@ -1,17 +1,21 @@
...
+      app_params: %V2.BusShelter{
+        departures: %V2.Departures{sections: []},
+        header: %V2.Header.CurrentStopId{stop_id: "1216"},
+        footer: %V2.Footer{stop_id: "1216"}
+      },
+      vendor: :gds,
       device_id: "TEST",
       name: "TEST",
-      app_id: :bus_shelter
+      app_id: :bus_eink_v2
```
jzimbel-mbta, line 18: `:bus_shelter_v2`, right?

test/screens/v2/candidate_generator/solari_test.exs
```diff
@@ -1,17 +1,20 @@
...
+    config = %Screen{
+      app_params: %V2.Solari{
+        departures: %V2.Departures{sections: []},
+        header: %V2.Header.CurrentStopName{stop_name: "Ruggles"}
+      },
+      vendor: :gds,
       device_id: "TEST",
       name: "TEST",
-      app_id: :solari
+      app_id: :bus_eink_v2
```
jzimbel-mbta, line 17: `:solari_v2`

jzimbel-mbta (approved):
Good to go after fixing the app IDs in your tests!

PR #830, opened 2021-04-30
Pass all screen templates through Builder.build_template/1

jzimbel-mbta:
**Asana task**: ad hoc

This updates all screens to pass their "draft" templates through `build_template/1` before returning from `screen_template/0`. This has no effect on any template besides the bus shelter one for now, but if/when we add backend paging to more screen types, it will be necessary.


handorff (approved):
✅

PR #851, opened 2021-05-15
feat: Add Screens.Config.Struct and use it throughout config modules

jzimbel-mbta:
**Asana task**: [Reduce code duplication in backend config modules](https://app.asana.com/0/1185117109217413/1200260123173553/f)

A PR with more code deleted than added 😌

This adds the `Screens.Config.Behaviour` behaviour to informally define config modules as having `from_json/1` and `to_json/1` definitions.

It also defines a `use`-able `Screens.Config.Struct` module that takes most of the work out of defining struct-based parts of the config. I attempted to [document](https://github.com/mbta/screens/pull/851/files#diff-e2c917994e0191f0d115dd625745214f0771ff8a05b4e22aa1c407ba78c44fb5) it for ease of use.

This change is purely a refactor. No config shapes have changed (hopefully).

- [ ] Needs version bump?

### DIFF
--- a/lib/screens/config.ex
+++ b/lib/screens/config.ex
@@ -2,7 +2,6 @@ defmodule Screens.Config do
   @moduledoc false
 
   alias Screens.Config.{Devops, Screen}
-  alias Screens.Util
 
   @type t :: %__MODULE__{
           screens: %{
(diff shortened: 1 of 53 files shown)


handorff (changes requested):
Overall this looks great! Just a couple small things.

handorff (approved):

PR #861, opened 2021-06-01
feat: Alert config + widget struct + accessor functions

jzimbel-mbta:
**Asana task**: [define struct and accessor functions for widget](https://app.asana.com/0/1185117109217413/1200350197926295/f)

This defines the struct with the expectation that a future change will supply it with the necessary API data, and the `WidgetInstance` callbacks will be fleshed out using the accessor functions defined in this PR.

- [ ] Needs version bump?


lib/screens/v2/widget_instance/alert.ex
```diff
@@ -2,85 +2,210 @@ defmodule Screens.V2.WidgetInstance.Alert do
...
+  @spec seconds_to_next_active_period(t(), DateTime.t()) :: integer() | :infinity
+  def seconds_to_next_active_period(t, now \\ DateTime.utc_now())
+
+  def seconds_to_next_active_period(
+        %__MODULE__{alert: %Alert{active_period: active_periods}},
+        now
+      )
+      when not is_nil(active_periods) do
+    next_active_period =
+      Enum.find(active_periods, fn {start, _} ->
```
handorff, line 46: Is `active_periods` guaranteed to be sorted?

lib/screens/v2/widget_instance/alert.ex
```diff
@@ -2,85 +2,210 @@ defmodule Screens.V2.WidgetInstance.Alert do
...
+      stop in context.downstream_stops -> [:downstream]
+      true -> []
     end
   end
 
-  defp effect_priority(_), do: 100
+  # Only route is not nil (route type ignored)
+  defp informed_entity_to_zone(%{stop: nil, route: route}, context) do
+    # Should we consider whether the stop is a terminus?
+    # If so, an alert that informs a route through the stop might not always be all 3 zones.
```
handorff, line 167: I don't think this would change priority ranking, but it could potentially matter to `slot_names` on Green Line e-Ink screens, where we check whether the home stop is inside a suspension or shuttle (not on the boundary). In this case, we'd want a whole-route alert to be full-screen -- I think that's consistent with what you have here already.

lib/screens/v2/widget_instance/alert.ex
```diff
@@ -2,85 +2,210 @@ defmodule Screens.V2.WidgetInstance.Alert do
...
+  end
 
-  defp serialize_text(%T{alert: _alert}) do
-    ["Dummy alert text"]
+  # Both stop and route are not nil (route type ignored)
+  defp informed_entity_to_zone(%{stop: _stop, route: route} = informed_entity, context) do
+    if route in context.active_routes do
+      informed_entity_to_zone(%{informed_entity | route: nil}, context)
+    else
+      []
```
handorff, line 176: It might be worth checking with Laina + design about checking whether the route is in `active_routes`. It makes sense that we'd want to de-prioritize routes which aren't running on the given day, but I don't think we ever explicitly discussed that case. (Same comment applies to line 168 above.)

handorff (approved):
Left a few comments, but overall this looks good to me!

lib/screens/v2/widget_instance/alert.ex
jzimbel-mbta: Thanks for catching this! I updated the alert parser to sort active periods up front!

PR #908, opened 2021-06-23
feat: Alert widget priority logic

jzimbel-mbta:
**Asana task**: [determine widget priority](https://app.asana.com/0/1185117109217413/1200354915858975/f)

Background for non-screens-team reviewer:
Widget priority determines the order in which the framework attempts to place candidate widgets into the layout.
It is represented as a list of positive integers, because Elixir/Erlang term ordering compares values in a list one by one—this allows us to have arbitrary numbers of "tie-breakers" when comparing two widgets by priority.
```
widget A's priority = [1, 1, 2, 4, 3]
widget B's priority = [1, 1, 2, 5, 1]

[1, 1, 2, 4, 3] < [1, 1, 2, 5, 1]
          ^                 ^
tie is broken here; all values afterward have no effect on ordering

widget A is considered for placement before widget B
```

I did my best to split the work in this PR into self-explanatory/self-contained commits, so it might be helpful to review it commit-by-commit. Though a few commits are purely to address merge conflicts from rebasing after another PR was recently merged—sorry! You can safely ignore any of the merge-conflict-addressing commits before 9736f28 ("Add mix_test_watch dependency").

The priority logic I implemented is described in the "Rules for displaying alerts" section of [this document](https://mbta.sharepoint.com/:w:/s/CTD/EVnlDHsD7B9AiC14j9fxMKsBfwAgKM9_Z-shAK6cCaQ3_A?e=yqsjah).


lboyarsky (approved):
Looks good to me, I've checked and the logic here seems to correspond with what is described in the requirements doc. 

However, as I mentioned in slack, I had very little interaction w/ elixir during the last two years, so I'm not really qualified to say whether the code needs improvement or not. 

I would suggest soliciting additional reviews.

lib/screens/alerts/alert.ex
```diff
@@ -352,13 +364,12 @@ defmodule Screens.Alerts.Alert do
...
   # defined as: some active period contains the current time
-  defp happening_now(%{active_period: aps}) do
-    now = DateTime.utc_now()
+  defp happening_now(%{active_period: aps}, now \\ DateTime.utc_now()) do
     if Enum.any?(aps, &in_active_period(&1, now)), do: 1, else: 0
   end
 
-  def happening_now?(alert) do
-    case happening_now(alert) do
+  def happening_now?(alert, now \\ DateTime.utc_now()) do
```
alicia-mbta, line 371: Would it make sense to consolidate the `happening_now` logic into one function? I was thinking something like:
```elixir
def happening_now?(%{active_period: aps}, now \\ DateTime.utc_now()) do
  if Enum.any?(aps, &in_active_period(&1, now)), do: true, else: false
end
```

lib/screens/v2/widget_instance/alert.ex
```diff
@@ -17,30 +17,59 @@ defmodule Screens.V2.WidgetInstance.Alert do
...
-          routes_at_stop: list(%{route_id: route_id(), active?: boolean()})
+          routes_at_stop: list(%{route_id: route_id(), active?: boolean()}),
+          now: DateTime.t()
         }
 
-  def active?(%__MODULE__{alert: alert}, happening_now? \\ &Alert.happening_now?/1) do
-    happening_now?.(alert)
+  @base_priority 2
+
+  @effect_priorities ~w[shuttle stop_closure suspension station_closure detour stop_moved snow_route elevator_closure]a
```
alicia-mbta, line 26: [might be just me getting a little lost in translation but] a quick question/suggestion: would it make sense to rename this to `@event_priorities` or `@alert_type_priorities`? (if I'm understanding this right). Please disregard if not.

lib/screens/v2/widget_instance/alert.ex
```diff
@@ -17,30 +17,59 @@ defmodule Screens.V2.WidgetInstance.Alert do
...
-  def seconds_from_onset(_t, _now), do: :infinity
+  def active?(%__MODULE__{alert: alert, now: now}, happening_now? \\ &Alert.happening_now?/2) do
+    happening_now?.(alert, now)
+  end
 
-  @spec seconds_to_next_active_period(t(), DateTime.t()) :: integer() | :infinity
-  def seconds_to_next_active_period(t, now \\ DateTime.utc_now())
+  @spec seconds_from_onset(t()) :: integer()
+  def seconds_from_onset(%__MODULE__{alert: %Alert{active_period: [{start, _} | _]}, now: now})
+      when not is_nil(start) do
```
alicia-mbta, line 64: Is there a fallback in case `start` is `nil` or does Elixir have something like that by default?

alicia-mbta (approved):
I'd feel more comfortable if someone from the screens teams took a quick look as well 😅 
But overall it looks good to me! The comments in certain parts of the code clarifying the logic were really helpful.
I left a couple of comments but please ignore if they don't make that much sense.

lib/screens/v2/widget_instance/alert.ex
jzimbel-mbta: Those would also make sense, but the official terminology in alert data for these values is `effect`. E.g. `{..., "effect": "STOP_MOVE", ...}` for [this alert](https://api-v3.mbta.com/alerts/384443). For consistency's sake I tried to stick with the terminology used by alerts.

lib/screens/v2/widget_instance/alert.ex
jzimbel-mbta: There is not (the function would raise an exception in that case), but when I checked all alerts in the system, their active periods always had a non-null start value.

The API documentation also seems to indicate that the start value is non-nullable. I can't deep-link it, but if you scroll to the "Models" section at the bottom of the [V3 API docs](https://api-v3.mbta.com/docs/swagger/index.html), the ActivePeriod type shows only the end date as nullable:
image

lib/screens/alerts/alert.ex
jzimbel-mbta: Unfortunately I can't fully remove the other function—it was defined first and we need a numeric value in some places, a boolean value in others. I tried to clear up what's going on there, though.

PR #910, opened 2021-06-29
feat: Implement WidgetInstance.slot_names/1 for Alert widget

jzimbel-mbta:
**Asana task**: [determine widget slot name](https://app.asana.com/0/1185117109217413/1200354915858977/f)

Possibly helpful explanation of the big `reduce_while` inside `informs_all_active_routes_at_home_stop/1`: It starts with the set of routes that are actively serving the home stop on the current day as the accumulator. It looks at each informed entity one at a time, and removes any route IDs from the set.

If the set is empty after checking through all the informed entities, then the alert informs all active routes serving the home stop. If not, it does not inform all of them.

Note: I also updated `priority/1` to depend on `slot_names/1`. If `slot_names/1` returns `[:full_screen]`, then `priority/1` returns `[1]` in order to have the alert be placed before other widgets that might preclude it from getting the full-screen slot.

- [ ] Needs version bump?

### DIFF
--- a/lib/screens/v2/widget_instance/alert.ex
+++ b/lib/screens/v2/widget_instance/alert.ex
@@ -15,7 +15,7 @@ defmodule Screens.V2.WidgetInstance.Alert do
   @type route_id :: String.t()
 
   @type t :: %__MODULE__{
-          screen: Screens.Config.Screen.t(),
+          screen: Screen.t(),
           alert: Alert.t(),
           stop_sequences: list(list(stop_id())),
           routes_at_stop: list(%{route_id: route_id(), active?: boolean()}),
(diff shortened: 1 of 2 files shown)


lib/screens/v2/widget_instance/alert.ex
```diff
@@ -37,15 +37,34 @@ defmodule Screens.V2.WidgetInstance.Alert do
...
 
-  def slot_names(_t) do
-    [:medium_left, :medium_right]
+  def slot_names(%__MODULE__{screen: %Screen{app_id: app_id}} = t)
+      when app_id in [:bus_shelter_v2, :bus_eink_v2] do
+    if active?(t) and effect(t) in [:stop_closure, :stop_move, :suspension, :detour] and
+         informs_all_active_routes_at_home_stop?(t) do
+      [:full_screen]
+    else
+      [:medium_left, :medium_right]
```
handorff, line 57: I think `bus_eink_v2` just has `:medium_flex`.

handorff (changes requested):
One small thing, otherwise looks good!

lib/screens/v2/widget_instance/alert.ex
jzimbel-mbta: Good catch, thanks!

handorff (approved):
Thanks!

PR #937, opened 2021-07-06
feat: Alert widget serialize function; factored-out route pill logic

jzimbel-mbta:
**Asana task**: [serialize widget](https://app.asana.com/0/1185117109217413/1200354915858979/f)

I tried to factor out the logic for serializing route pills so that we don't have duplicated code for departures, alerts, and other places that use route pills.

There's also some questionably-readable module attributes at the top of the alert widget module now, where I tried to link all of the effect-related hardcoded values together. I'm totally open to making those less of an eyesore if the current approach seems bad.

- [ ] Needs version bump?

### DIFF
--- a/lib/screens/v2/widget_instance/serializer/route_pill.ex
+++ b/lib/screens/v2/widget_instance/serializer/route_pill.ex
@@ -81,7 +81,7 @@ defmodule Screens.V2.WidgetInstance.Serializer.RoutePill do
 
   @spec serialize_for_alert(Route.id()) :: t()
   def serialize_for_alert(route_id) do
-    route = do_serialize(route_id, %{gl_branch: true, cr_abbrev: true, ferry_abbrev: true})
+    route = do_serialize(route_id, %{gl_branch: true, cr_abbrev: true})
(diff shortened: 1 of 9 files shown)


lib/screens/v2/widget_instance/serializer/route_pill.ex
```diff
@@ -0,0 +1,166 @@
...
+  end
+
+  defp do_serialize("CR-" <> _line, _) do
+    %{type: :icon, icon: :rail}
+  end
+
+  defp do_serialize("Boat-" <> line, %{ferry_abbrev: true}) do
+    case line do
+      "F1" -> %{type: :slashed, part1: "HNG", part2: "HUL"}
+      "F4" -> %{type: :text, text: "CTN"}
```
handorff, line 126: Where did these abbreviations come from?

handorff (approved):
Looks good to me, and thanks for factoring out route pill serialization.

Not blocking, but I'm curious where the ferry line abbreviations came from -- I don't think I've seen those before.

PR #942, opened 2021-07-07
feat: Log new Solari screen data requests

jzimbel-mbta:
**Asana task**: [Log data requests from new Solari screens](https://app.asana.com/0/1185117109217413/1200074998406104/f)

`is_screen_conn?` now takes the screen ID as well. Solari user agents are just standard Ubuntu + latest/recent Firefox now, so I added a check on the requested screen ID to try and filter out as many non-screen requests as possible.

We still will probably get a few false-positive logs (e.g. if Ryan looks at the Solari app on his Ubuntu machine) but there isn't much we can do about that.

- [ ] Needs version bump?


handorff (approved):
✅

PR #953, opened 2021-07-15
feat: Bus shelter flex zone alert widget component

jzimbel-mbta:
**Asana task**: [react component with styles for alerts widget on bus shelter screens](https://app.asana.com/0/1185117109217413/1200356261737391/f)

Mostly straightforward. This only adds support for the bus shelter flex zone version of alerts—still need to add styles (and JS, if necessary, though it should be possible to have a single component) for bus shelter full-screen, as well as e-ink medium flex and full-screen.

Screenshot 2021-07-15 at 16-46-45 Screens

- [ ] Needs version bump?


lib/screens/v2/widget_instance/alert.ex
```diff
@@ -107,6 +107,7 @@ defmodule Screens.V2.WidgetInstance.Alert do
         _ -> route_id
       end
     end)
+    |> Enum.take(2)
```
handorff, line 110: Just checking: is this something you agreed on with Laina and design?

assets/src/components/v2/alert.tsx
```diff
@@ -0,0 +1,120 @@
...
+      className={classWithModifier(
+        "alert-widget__content__body-text",
+        isSmall ? "small" : "regular"
+      )}
+      ref={ref}
+    >
+      {children}
+    </div>
+  );
+};
```
handorff, line 118: I like this pattern of making the sizer a wrapper; I wonder if there's a way to generalize this to make it easier to make new ones.

handorff (approved):
Really exciting to see this come to life! Left a couple comments; this also makes me think we might want to file a task to organize our `/static/images` directory a bit, but that's definitely out of scope for this PR.

lib/screens/v2/widget_instance/alert.ex
jzimbel-mbta: No, and after checking Adam's designs again I just noticed that the route pills are small enough to fit 3—I'll update styles so that we aren't using _exactly_ the same route pills as what's shown in departure rows.
Will also double check with Laina and design on what should happen if there are more than 3 pills to show on a single flex zone alert.

assets/src/components/v2/alert.tsx
jzimbel-mbta: Should be doable for simple cases! Though if we need to generalize it to handle any number of sizing tiers (versus just 2, or some other fixed number) that would probably be harder. Next time we need to do text sizing somewhere, I'll see if I can factor it out.

lib/screens/v2/widget_instance/alert.ex
jzimbel-mbta: After checking with Adam on this, here's what we'll do:
- up to 2 affected routes → full-size route pills
- 3 affected routes → small route pills
- more than 3 affected routes → collapse to single full-size mode pill
- affected entity is whole stop, or whole mode → single full-size mode pill

jzimbel-mbta: (Originally he had cases of >3 pills wrap onto a second line)

lib/screens/v2/widget_instance/alert.ex
```diff
@@ -96,18 +96,26 @@ defmodule Screens.V2.WidgetInstance.Alert do
...
-        _ -> route_id
-      end
-    end)
-    |> Enum.map(&RoutePill.serialize_for_alert/1)
+    routes = informed_routes(t)
+
+    if MapSet.size(routes) <= 3 do
+      routes
+      |> Enum.to_list()
+      |> Enum.sort_by(fn route_id ->
```
handorff, line 104: You can consider this out of scope for this PR, but routes do have a sort order which represents the order in which they should be shown to riders. (I don't think we're parsing this from the routes API response, currently.)

jzimbel-mbta: Oh, good to know. I'll create a followup task.

PR #961, opened 2021-07-21
feat: bus shelter not-quite-fullscreen ("full-body") alert widget

jzimbel-mbta:
**Asana task**: [react component with styles for alerts widget on bus shelter screens](https://app.asana.com/0/1185117109217413/1200356261737391/f) (second subtask)

This adds the bus shelter alert widget variant that replaces all content on the screen except the header.
I needed to tweak the screen template to support this since we previously only allowed a full-screen takeover.
I also needed to update priority values to prevent the subway status widget from competing with takeover widgets in the placement order.

The widget itself has all the same logic and receives the same API data as the flex zone version—I just factored out a few things in places where they differ.

image

- [ ] Needs version bump?


handorff (approved):
✅ 

Looks good to me!

PR #979, opened 2021-08-05
fix: Visual tweaks for bus shelter screen

jzimbel-mbta:
**Asana task**: [Updates from departures implementation review](https://app.asana.com/0/1185117109217413/1200480000015510/f)

Changes:
- [x] **Put crowding icons into a consistent column offset from the right by the width of the "ARR/BRD/Now" time div.**
  - [x] **bus shelter**
  - [x] **bus e-ink**
  - [ ] **bus shelter XL (single route)**: I don't think this is a thing anymore?
- [x] **Smaller font size for timestamps on e-ink**
  - [x] **bus e-ink**
  - [x] **GL e-ink**
- [x] **Suppress crowding info on departure rows with a timestamp (HH:MM) time.**
- [x] **GL e-ink: Shorten line between departures to 794px**: :warning: Appears to already be 794px. See [this line](https://github.com/mbta/screens/blob/master/assets/css/v2/gl_eink/departures/time.scss#L12).
- [x] **Shrink bus shelter header by 8px**: This originally called for an 8px bottom padding to be added to the header with the departures list background color, but given that the content below the header isn't always a departures list (e.g. a takeover alert), I decreased the header slot size and increased the main content slot size instead, and added top padding to the departures widget.
  - I also fixed the footer, which had a height greater than its slot height.
- [x] **Update flex zone slot with darker background/"sunken" look.**
  - I also gave the bus shelter screen container element `overflow: hidden` to ensure that the screen is always 1080x1920, and to hide parts of the main content and footer box shadows that are cast on the flex zone.

- [ ] Needs version bump?

### DIFF
--- a/assets/css/bus_shelter_v2.scss
+++ b/assets/css/bus_shelter_v2.scss
@@ -2,6 +2,8 @@
 
 @import "colors";
 
+@import "v2/bus_shelter/screen_container";
+
 @import "v2/bus_shelter/screen/normal";
 @import "v2/bus_shelter/screen/takeover";
 
--- a/assets/css/v2/bus_eink/departures/time.scss
+++ b/assets/css/v2/bus_eink/departures/time.scss
@@ -8,6 +8,10 @@
   &:not(:first-child) {
     margin-top: 32px;
   }
+
+  .departure-time {
+    width: 264px;
+  }
 }
 
 .departure-time {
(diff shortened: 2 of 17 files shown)


handorff:
FYI: Bus Shelter XL doesn't exist but might in the future -- it's the version for when there's only a single route, which won't happen on Columbus Ave but might on Summer Street.

handorff (approved):

PR #1022, opened 2021-09-03
feat: switch from HTTPoison to Mojito for fetching

paulswartz:
**Asana task**: [title](url)

Description

- [ ] Needs version bump?


losvedir:
If you do `mix deps.unlock --unused` does that remove `hackney`? Otherwise, I'm thinking it might still be in there via some other dependency and therefore might still have the problem.

paulswartz:
No, ExAws -> HTTPoison -> Hackney.

losvedir:
Ah, but it looks like `api_v3_get_json_error` is what generated all the checkout_timeout errors, so this would still help with that.

lib/screens/vendor_data/fetch.ex
```diff
@@ -4,13 +4,13 @@ defmodule Screens.VendorData.Fetch do
...
   def make_and_parse_request(url, headers \\ [], parse_fn, vendor_name, opts) do
-    with {:http_request, {:ok, response}} <- {:http_request, HTTPoison.get(url, headers, opts)},
+    with {:http_request, {:ok, response}} <- {:http_request, Mojito.get(url, headers, opts)},
          {:response_success, %{status_code: 200, body: body}} <- {:response_success, response},
          {:parse, {:ok, parsed}} <- {:parse, parse_fn.(body)} do
       {:ok, parsed}
     else
       {:http_request, {:error, e}} ->
-        log_fetch_error(vendor_name, :http_fetch_error, %{message: HTTPoison.Error.message(e)})
+        log_fetch_error(vendor_name, :http_fetch_error, %{message: e.message, reason: e.reason})
```
losvedir, line 13: It looks like the [spec](https://hexdocs.pm/mojito/Mojito.Base.html#t:error/0) for `e.reason` is `any()` (not sure what it is in practice). Maybe `inspect()` it either here or in the `log_fetch_error` function?

losvedir (approved):
👍  other than the bit about the error type.

paulswartz:
Not going to merge for now: will deploy this version to prod directly from the console.

paulswartz:
Closing this PR as it's 1) out of date with the current default branch 2) using Mojito which is itself deprecated: https://hex.pm/packages/mojito

PR #1038, opened 2021-09-21
feat: Schedules for Evergreen content

cmaddox5:
**Asana task**: [Add ability to schedule evergreen content](https://app.asana.com/0/1185117109217413/1200928169910248/f)

Evergreen content can now have a schedule. Schedule will be a list of objects with a start and end date. Existing Evergreen content will have a default schedule of always (nil for start and end).

- [ ] Needs version bump?

### DIFF
--- a/lib/screens/config/v2/evergreen_content_item.ex
+++ b/lib/screens/config/v2/evergreen_content_item.ex
@@ -12,8 +12,8 @@ defmodule Screens.Config.V2.EvergreenContentItem do
         }
 
   @enforce_keys ~w[slot_names asset_path priority]a
-  defstruct slot_names: [],
-            asset_path: "",
+  defstruct slot_names: nil,
+            asset_path: nil,
             priority: nil,
             schedule: [%Schedule{}]
 
--- a/lib/screens/config/v2/evergreen_content_item.ex
+++ b/lib/screens/config/v2/evergreen_content_item.ex
@@ -1,18 +1,23 @@
 defmodule Screens.Config.V2.EvergreenContentItem do
   @moduledoc false
 
+  alias Screens.Config.V2.Schedule
   alias Screens.V2.WidgetInstance
 
   @type t :: %__MODULE__{
           slot_names: list(WidgetInstance.slot_id()),
           asset_path: String.t(),
-          priority: WidgetInstance.priority()
+          priority: WidgetInstance.priority(),
+          schedule: list(Schedule.t())
         }
 
   @enforce_keys ~w[slot_names asset_path priority]a
-  defstruct @enforce_keys
+  defstruct slot_names: [],
+            asset_path: "",
+            priority: nil,
+            schedule: [%Schedule{}]
 
-  use Screens.Config.Struct
+  use Screens.Config.Struct, children: [schedule: {:list, Schedule}]
 
   defp value_from_json("slot_names", slot_names) do
     Enum.map(slot_names, &String.to_existing_atom/1)
--- a/lib/screens/config/v2/schedule.ex
+++ b/lib/screens/config/v2/schedule.ex
@@ -1,0 +1,31 @@
+defmodule Screens.Config.V2.Schedule do
+  @moduledoc false
+
+  @type t :: %__MODULE__{
+          start_dt: DateTime.t(),
+          end_dt: DateTime.t()
+        }
+
+  defstruct start_dt: nil,
+            end_dt: nil
+
+  use Screens.Config.Struct
+
+  defp value_from_json(_, nil), do: nil
+
+  defp value_from_json("start_dt", iso_string) do
+    {:ok, dt, _offset} = DateTime.from_iso8601(iso_string)
+    dt
+  end
+
+  defp value_from_json("end_dt", iso_string) do
+    {:ok, dt, _offset} = DateTime.from_iso8601(iso_string)
+    dt
+  end
+
+  defp value_to_json(_, nil), do: nil
+
+  defp value_to_json(_, datetime) do
+    DateTime.to_iso8601(datetime)
+  end
+end
--- a/lib/screens/config/v2/schedule.ex
+++ b/lib/screens/config/v2/schedule.ex
@@ -2,8 +2,8 @@ defmodule Screens.Config.V2.Schedule do
   @moduledoc false
 
   @type t :: %__MODULE__{
-          start_dt: DateTime.t(),
-          end_dt: DateTime.t()
+          start_dt: DateTime.t() | nil,
+          end_dt: DateTime.t() | nil
         }
 
   defstruct start_dt: nil,
--- a/lib/screens/v2/widget_instance/evergreen_content.ex
+++ b/lib/screens/v2/widget_instance/evergreen_content.ex
@@ -30,7 +30,28 @@ defmodule Screens.V2.WidgetInstance.EvergreenContent do
 
   def widget_type(_instance), do: :evergreen_content
 
-  def valid_candidate?(_instance), do: true
+  def valid_candidate?(%__MODULE__{schedule: schedule, now: now}) do
+    schedule
+    |> Enum.map(fn x ->
+      case x do
+        %Schedule{start_dt: nil, end_dt: nil} ->
+          true
+
+        %Schedule{start_dt: start_dt, end_dt: nil} ->
+          DateTime.compare(start_dt, now) in [:lt, :eq]
+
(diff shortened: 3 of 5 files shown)


lib/screens/config/v2/evergreen_content_item.ex
```diff
@@ -1,18 +1,23 @@
...
-          priority: WidgetInstance.priority()
+          priority: WidgetInstance.priority(),
+          schedule: list(Schedule.t())
         }
 
   @enforce_keys ~w[slot_names asset_path priority]a
-  defstruct @enforce_keys
+  defstruct slot_names: [],
+            asset_path: "",
+            priority: nil,
```
jzimbel-mbta, line 17: The values here are fine but won't actually be used since their keys must be supplied when creating this struct—if you want you can just make all 3 of them `nil`.

lib/screens/v2/widget_instance/evergreen_content.ex
```diff
@@ -22,7 +30,28 @@ defmodule Screens.V2.WidgetInstance.EvergreenContent do
...
+          DateTime.compare(end_dt, now) == :gt
+
+        %Schedule{start_dt: start_dt, end_dt: end_dt} ->
+          DateTime.compare(start_dt, now) in [:lt, :eq] and DateTime.compare(end_dt, now) == :gt
+
+        _ ->
+          false
+      end
+    end)
+    |> Enum.any?(fn x -> x end)
```
jzimbel-mbta, line 53: If you want, you can pass the anonymous function that was passed to `Enum.map/2` (which transforms each value to a boolean) directly to `Enum.any?/2` here! No need to map the values to booleans and then check if any are true in two separate steps.

lib/screens/v2/widget_instance/evergreen_content.ex
```diff
@@ -22,7 +30,28 @@ defmodule Screens.V2.WidgetInstance.EvergreenContent do
...
+        %Schedule{start_dt: nil, end_dt: end_dt} ->
+          DateTime.compare(end_dt, now) == :gt
+
+        %Schedule{start_dt: start_dt, end_dt: end_dt} ->
+          DateTime.compare(start_dt, now) in [:lt, :eq] and DateTime.compare(end_dt, now) == :gt
+
+        _ ->
+          false
+      end
+    end)
```
jzimbel-mbta, line 52: `fn` allows you to provide multiple clauses in the same way that `def`/`defp` do—you can remove one level of indentation here like so:
```suggestion
    |> Enum.map(fn
      %Schedule{start_dt: nil, end_dt: nil} ->
        true

      %Schedule{start_dt: start_dt, end_dt: nil} ->
        DateTime.compare(start_dt, now) in [:lt, :eq]

      %Schedule{start_dt: nil, end_dt: end_dt} ->
        DateTime.compare(end_dt, now) == :gt

      %Schedule{start_dt: start_dt, end_dt: end_dt} ->
        DateTime.compare(start_dt, now) in [:lt, :eq] and DateTime.compare(end_dt, now) == :gt

      _ ->
        false
    end)
```

Also, it should be safe to remove the `_ ->` fallback clause—if we get a value that isn't at all what we were expecting, we'd rather just let this process crash, rely on automatic Splunk logging to track down where things went wrong, and fix the problem. In general Elixir development encourages a ["let it crash" or "fail fast"](https://elixir-lang.org/getting-started/mix-otp/supervisor-and-application.html) methodology.

Otherwise, great use of pattern matching!

test/screens/v2/widget_instance/evergreen_content_test.exs
```diff
@@ -4,14 +4,27 @@ defmodule Screens.V2.WidgetInstance.EvergreenContentTest do
...
+      },
+      widget_old_schedule: %EvergreenContent{
+        screen: %Screen{app_params: nil, vendor: nil, device_id: nil, name: nil, app_id: nil},
+        slot_names: [:medium_left, :medium_right],
+        asset_url: "https://mbta-screens.s3.amazonaws.com/screens-dev/videos/some-video.mp4",
+        priority: [2, 3, 1],
+        schedule: [
+          %Schedule{start_dt: DateTime.add(DateTime.utc_now(), 30, :second), end_dt: nil}
+        ],
+        now: DateTime.utc_now()
```
jzimbel-mbta, line 27: This works since both datetimes are relative to now, but in tests it's generally encouraged to use a static datetime so that the test runs on exactly the same data every time.

Could you pass explicit datetimes in these two places?
You can use the [`~U` sigil](https://elixir-lang.org/getting-started/mix-otp/supervisor-and-application.html) to quickly create a UTC-timezone datetime, like `~U[2021-09-21T16:00:00Z]`

test/screens/v2/widget_instance/evergreen_content_test.exs
```diff
@@ -44,8 +57,12 @@ defmodule Screens.V2.WidgetInstance.EvergreenContentTest do
   end
 
   describe "valid_candidate?/1" do
-    test "returns true", %{widget: widget} do
+    test "returns true with blank schedule", %{widget: widget} do
       assert WidgetInstance.valid_candidate?(widget)
     end
+
+    test "returns false with old schedule", %{widget_old_schedule: widget_old_schedule} do
+      assert !WidgetInstance.valid_candidate?(widget_old_schedule)
```
jzimbel-mbta, line 65: This is fine, but just FYI there's also a [`refute`](https://hexdocs.pm/ex_unit/ExUnit.Assertions.html#refute/1) assertion available for things like this.

lib/screens/config/v2/schedule.ex
```diff
@@ -0,0 +1,31 @@
+defmodule Screens.Config.V2.Schedule do
+  @moduledoc false
+
+  @type t :: %__MODULE__{
+          start_dt: DateTime.t(),
+          end_dt: DateTime.t()
```
jzimbel-mbta, line 6: I think I stated differently during our last meeting, but I changed my mind. 😁
Could you include that these are nullable in the type def? You can just add `| nil`:
```suggestion
          start_dt: DateTime.t() | nil,
          end_dt: DateTime.t() | nil
```

jzimbel-mbta (commented):
Just some small suggestions, but looks great overall. Thanks!

In case you haven't checked it out yet, there is some great documentation available for both Elixir's core library and the testing library that ships with it (ExUnit).
https://hexdocs.pm/elixir/Kernel.html
https://hexdocs.pm/ex_unit/ExUnit.html

You might also want to bookmark the Phoenix documentation as well while you're at it. That's the web framework we use: https://hexdocs.pm/phoenix/Phoenix.html

test/screens/v2/widget_instance/evergreen_content_test.exs
```diff
@@ -4,14 +4,27 @@ defmodule Screens.V2.WidgetInstance.EvergreenContentTest do
...
   setup do
     %{
       widget: %EvergreenContent{
         screen: %Screen{app_params: nil, vendor: nil, device_id: nil, name: nil, app_id: nil},
         slot_names: [:medium_left, :medium_right],
         asset_url: "https://mbta-screens.s3.amazonaws.com/screens-dev/videos/some-video.mp4",
-        priority: [2, 3, 1]
+        priority: [2, 3, 1],
+        schedule: [%Schedule{}],
+        now: DateTime.utc_now()
```
jzimbel-mbta, line 17: Could you use an explicit datetime instead of `DateTime.utc_now()` for this?

test/screens/v2/widget_instance/evergreen_content_test.exs
```diff
@@ -4,14 +4,27 @@ defmodule Screens.V2.WidgetInstance.EvergreenContentTest do
...
+      },
+      widget_old_schedule: %EvergreenContent{
+        screen: %Screen{app_params: nil, vendor: nil, device_id: nil, name: nil, app_id: nil},
+        slot_names: [:medium_left, :medium_right],
+        asset_url: "https://mbta-screens.s3.amazonaws.com/screens-dev/videos/some-video.mp4",
+        priority: [2, 3, 1],
+        schedule: [
+          %Schedule{start_dt: ~U[2021-01-01T00:00:00Z], end_dt: ~U[2021-01-02T00:00:00Z]}
+        ],
+        now: DateTime.utc_now()
```
jzimbel-mbta, line 27: Same as previous comment.

jzimbel-mbta (commented):
Just one small thing, otherwise good to go.

jzimbel-mbta (approved):

PR #1041, opened 2021-09-22
temp feat: Only show here and now alerts on bus v2 screens

cmaddox5:
**Asana task**: [Only show alerts now/here on bus screens](https://app.asana.com/0/1185117109217413/1200992643788400/f)

For the time being, v2 bus alerts that do not actively apply to a screen's location will be excluded from candidacy. This is temporary until a better display strategy is decided on for other alerts.

- [ ] Needs version bump?

### DIFF
--- a/lib/screens/v2/widget_instance/alert.ex
+++ b/lib/screens/v2/widget_instance/alert.ex
@@ -168,8 +168,16 @@ defmodule Screens.V2.WidgetInstance.Alert do
     end
   end
 
-  def valid_candidate?(t) do
-    priority(t) != :no_render
+  def valid_candidate?(%__MODULE__{screen: %Screen{app_id: screen_type}} = t) do
+    cond do
+      screen_type in [:bus_shelter_v2, :bus_eink_v2] ->
+        priority(t) != :no_render and
+          location(t) == :inside and
+          active?(t)
+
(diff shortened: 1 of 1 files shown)


lib/screens/v2/widget_instance/alert.ex
```diff
@@ -168,8 +168,16 @@ defmodule Screens.V2.WidgetInstance.Alert do
...
+    cond do
+      screen_type in [:bus_shelter_v2, :bus_eink_v2] ->
+        priority(t) != :no_render and
+          location(t) == :inside and
+          active?(t)
+
+      screen_type == :gl_eink_v2 ->
+        priority(t) != :no_render
+    end
   end
```
jzimbel-mbta, line 181: This works! Just as an exercise though, see if you can split the behavior into two separate function clauses based on the `app_id` value. Hint: you can check if a variable from the function args is in a list of allowed values using a guard like so
```ex
def f(%{key: value} = t) when value in [:some, "list", true] do ...
```

Often if you find yourself checking equality in an `if` or `cond` expression, it could likely be replaced by pattern matching/guards in a `case` expression or the function head.

jzimbel-mbta (commented):
Looks good, just sharing some thoughts on best practices.

jzimbel-mbta (approved):
✅

PR #1050, opened 2021-09-28
feat: Modify typedefs for departures

cmaddox5:
**Asana task**: [Update type defs to support "notice rows" in departures list](https://app.asana.com/0/0/1201067686861561/f)

Changed up typedefs. This will allow the front end to display a single `section` with a list of objects that will either be one `notice`, one `notice` and one `departure`, or two `departures`.

A couple of discussions to start here:

1. The section attribute name `departures` doesn't make sense anymore. What would be a better name? `data`? `rows`?
2. Is `type` necessary anymore? Could there be a future need for it? I don't think so but wanted to double check. I know there are conditionals in the front-end that check for it, but they may need to be removed because of the typedefs changes anyway.

- [ ] Needs version bump?

### DIFF
--- a/lib/screens/v2/widget_instance/departures.ex
+++ b/lib/screens/v2/widget_instance/departures.ex
@@ -11,18 +11,16 @@ defmodule Screens.V2.WidgetInstance.Departures do
   defstruct screen: nil,
             section_data: []
 
-  @type normal_section :: %{
+  @type section :: %{
           type: :normal_section,
-          departures: list(Departure.t())
+          departures: list(Departure.t() | notice())
         }
 
-  @type notice_section :: %{
-          type: :notice_section,
-          icon: atom() | nil,
-          text: FreeText.t()
+  @type notice :: %{
+          destination: String.t(),
+          headway: non_neg_integer()
         }
 
-  @type section :: normal_section | notice_section
   @type t :: %__MODULE__{
           screen: Screen.t(),
           section_data: list(section)
--- a/lib/screens/v2/widget_instance/departures.ex
+++ b/lib/screens/v2/widget_instance/departures.ex
@@ -46,7 +46,7 @@ defmodule Screens.V2.WidgetInstance.Departures do
   end
 
   def serialize_section(%{type: :notice_section, icon: icon, text: text}, _screen) do
-    %{type: :notice_section, icon: icon, text: FreeTextLine.to_json(text)}
+    %{type: :notice_section, icon: icon, text: text}
   end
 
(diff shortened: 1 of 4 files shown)


jzimbel-mbta (commented):
Two things:
1. Could you keep the `notice_section` type around for now?*
2. Could you change the fields for `notice` to be simply `text: FreeTextLine.t()`? And update `notice_section`'s `text` field to `FreeTextLine.t()`, as well. (our mistake)**

I agree that `departures` could be named better now. `rows` seems like a good fit since that's what each element of the list really describes.

---

*There are cases we'd like to support in the future where a whole section should be replaced with a notice centered in a space multiple rows tall, e.g. when we need to temporarily suppress departures for a certain mode due to inaccurate prediction data. This has happened a handful of times when the V3 API started receiving bad data from a vendor.

**In general we have tried to keep as much of our logic and assumptions as possible on the backend—the departures widget is used by all screen types, so we don't want to be limited to only showing headway info in "text" rows. We've used the `FreeTextLine` structure in several places as a common way to communicate custom/formatted text with an optional leading icon to the client.

cmaddox5:
Both suggestions make sense so changes have been made. Will the `notice_section` still need the `icon`, or will the `icon` in `FreeTextLine` be enough?

jzimbel-mbta:
> Both suggestions make sense so changes have been made. Will the `notice_section` still need the `icon`, or will the `icon` in `FreeTextLine` be enough?

Good question. Just the one in `FreeTextLine` is enough—you can remove the `icon` field in `notice_section`.

lib/screens/v2/widget_instance/departures.ex
```diff
@@ -42,11 +44,11 @@ defmodule Screens.V2.WidgetInstance.Departures do
     def valid_candidate?(_instance), do: true
   end
 
-  def serialize_section(%{type: :notice_section, icon: icon, text: text}, _screen) do
-    %{type: :notice_section, icon: icon, text: FreeText.to_json(text)}
+  def serialize_section(%{type: :notice_section, text: text}, _screen) do
+    %{type: :notice_section, text: text}
```
jzimbel-mbta, line 48:
```suggestion
    %{type: :notice_section, text: FreeTextLine.to_json(text)}
```

jzimbel-mbta (approved):
This looks good for an intermediate step in this task! 👍

You will likely need to adjust the logic in `Departures.group_departures/1` to handle a mixed list of departures and notice rows, but as long as you write tests for your changes in the final PR, that and any other issues resulting from this change should be reasonably simple to track down and address.

Feel free to keep using this branch for the rest of the work if that's easier.

PR #1070, opened 2021-10-14
feat: make v2 line map work at heath st

handorff:
**Asana task**: [[E-Ink feature request] GL line map update for terminals](https://app.asana.com/0/1185117109217413/1200881723138823)

This PR also stops generating alerts as candidates (since they're not ready to be displayed to riders, and Heath St screens are supposed to be installed Monday).

It also includes a couple of unrelated formatting changes to audio and normal departures, sorry! We don't actually have any checks in CI for front-end stuff, so there's no enforced standard (but maybe there should be).


lib/screens/v2/widget_instance/line_map.ex
```diff
@@ -52,29 +81,98 @@ defmodule Screens.V2.WidgetInstance.LineMap do
...
+        reverse_stops
+        |> Enum.reverse()
+        |> Enum.drop(1)
+        |> Enum.with_index()
+        |> Enum.map(fn {%Screens.Stops.Stop{name: label}, i} ->
+          %{
+            label: label,
+            downstream: false,
+            current: false,
+            terminal: i == length(reverse_stops) - 2
```
jzimbel-mbta, line 112: Could you explain why it's `length(reverse_stops) - *2*`? (Also would this ever happen? I think all of our lines are long enough that you wouldn't see the whole length from one terminus to the other on the e-ink line map)

lib/screens/v2/widget_instance/line_map.ex
```diff
@@ -3,25 +3,35 @@ defmodule Screens.V2.WidgetInstance.LineMap do
 
   alias Screens.Config.{Screen, V2}
   alias Screens.Predictions.Prediction
+  alias Screens.Trips.Trip
```
jzimbel-mbta, line 6: Doesn't look like `Trip` is used anywhere.

test/screens/v2/widget_instance/line_map_test.exs
```diff
@@ -4,6 +4,7 @@ defmodule Screens.V2.WidgetInstance.LineMapTest do
   alias Screens.Predictions.Prediction
   alias Screens.Schedules.Schedule
   alias Screens.Stops.Stop
+  alias Screens.Trips.Trip
```
jzimbel-mbta, line 7: Doesn't look like `Trip` is used anywhere.

handorff: I think it is (e.g. here: https://github.com/mbta/screens/pull/1070/files#diff-57482df2b19067549da87f0f7d0d88f13aa39dc250a41473aa1ade5103c2b610R110).

PR #1072, opened 2021-10-15
feat: V2 ScreenAudioData module and controller endpoint to produce dynamic audio readouts

jzimbel-mbta:
**Asana task**: [Add v2 audio to serialize](https://app.asana.com/0/1185117109217413/1201071462192297/f)

Not a lot of changes for this, actually! The reason for the huge diff is that I had to rename all the audio view modules for them to be compatible with Phoenix's expected view/template naming practice.

The meat of the logic is in `ScreenAudioData` and the new functions in `AudioController`.

Essentially `ScreenAudioData.by_screen_id` uses the existing `ScreenData.by_screen_id` to do all of the usual widget generation and placement, then it takes the selected widgets just before they get cut down to a single page of content and serialized. From there, it calls the `WidgetInstance` audio callbacks to prepare them for SSML rendering via the various views.

The top level audio view uses no render function—all the logic lives in the template at lib/screens_web/templates/v2/audio/index.ssml.eex.

** I also fixed a small issue where the server would produce a 500 error if we asked for a nonexistent V2 screen.

- [ ] Needs version bump?


lib/screens_web/controllers/screen_controller.ex
```diff
@@ -54,12 +54,12 @@ defmodule ScreensWeb.ScreenController do
   end
 
   defp screen_ids(target_app_id) do
-    screen_ids =
-      for {screen_id, %Screen{app_id: app_id}} <- State.screens(), app_id == target_app_id do
+    ids =
+      for {screen_id, %Screen{app_id: ^target_app_id}} <- State.screens() do
```
cmaddox5, line 58: Nifty change right here.

lib/screens_web/controllers/v2/screen_controller.ex
```diff
@@ -32,6 +33,10 @@ defmodule ScreensWeb.V2.ScreenController do
     assign(conn, :environment_name, environment_name)
   end
 
+  defp v2_layout(conn, _) do
```
cmaddox5, line 36: Any reason for this change other than cleanup?

cmaddox5 (approved):
Really great stuff! Poked around a bunch locally and don't see any issues. So close to having the backend done!

PR #1075, opened 2021-10-18
feat: Define audio config

cmaddox5:
**Asana task**: [Define and check Audio config module for bus shelter screens](https://app.asana.com/0/1185117109217413/1201071462192299/f)

Audio config is now defined and expected in bus_shelter_v2 configs. I did not enforce the audio key in bus shelters. I felt this would be a little cleaner considering `by_screen_id` requires the presence of an audio config for BusShelter. 

- [ ] Needs version bump?

### DIFF
--- a/lib/screens_web/controllers/v2/screen_controller.ex
+++ b/lib/screens_web/controllers/v2/screen_controller.ex
@@ -4,7 +4,7 @@ defmodule ScreensWeb.V2.ScreenController do
   require Logger
 
   alias Screens.Config.{Screen, State}
-  alias Screens.Config.V2.{Audio, BusShelter}
+  alias Screens.Config.V2.Audio
   alias Screens.V2.ScreenData.Parameters
 
   @default_app_id :bus_eink


lib/screens/v2/screen_audio_data.ex
```diff
@@ -1,25 +1,54 @@
...
+         %Audio{
+           start_time: start_time,
+           stop_time: stop_time,
+           days_active: days_active
+         },
+         now
+       ) do
+    Date.day_of_week(now) not in days_active or
+      DateTime.compare(start_time, now) == :gt or
+      DateTime.compare(stop_time, now) == :lt
```
jzimbel-mbta, line 52: `now` is in UTC timezone, so you'll need to shift it to Eastern time before checking against the values from configuration which we're also expecting to be in Eastern time, in order for DST changes to take effect automatically in the readout schedule.

You'll also want to use `Time.compare` instead of `DateTime.compare` since we're only interested in the time of day, not the specific date.
```ex
{:ok, now_eastern} = DateTime.shift_zone(now, "America/New_York")

Date.day_of_week(now_eastern) ...
Time.compare(now_eastern, ...) == ...
```

test/screens/v2/screen_audio_data_test.exs
```diff
@@ -1,11 +1,79 @@
...
+        vendor: :lg_mri,
+        device_id: "TEST",
+        name: "TEST",
+        app_id: :bus_shelter_v2
+      },
+      config_audio_inactive: %Screen{
+        app_params: %V2.BusShelter{
+          audio: %V2.Audio{
+            start_time: ~U[2021-02-01T00:00:00Z],
+            stop_time: ~U[2021-02-01T01:00:00Z],
```
jzimbel-mbta, line 26: Use `~T` for a time-only value (no date or timezone), here and elsewhere in the setup data.

lib/screens/config/v2/bus_shelter.ex
```diff
@@ -11,7 +11,8 @@ defmodule Screens.Config.V2.BusShelter do
           header: CurrentStopId.t() | CurrentStopName.t(),
           alerts: Alerts.t(),
           evergreen_content: list(EvergreenContentItem.t()),
-          survey: Survey.t()
+          survey: Survey.t(),
+          audio: Audio.t()
```
jzimbel-mbta, line 15:
```suggestion
          audio: Audio.t() | nil
```

jzimbel-mbta (approved):
Looks great! Approved assuming you make these two small changes before merging.

PR #1083, opened 2021-10-21
feat: Hook up audio to front end

cmaddox5:
**Asana task**: [Add audio support to v2 frontend, and configure for bus shelter app](https://app.asana.com/0/1185117109217413/1201228986897061/f)

Audio has been hooked up for bus shelters. The 5 minute interval I choose was arbitrary and up for discussion. 

- [ ] Needs version bump?

### DIFF
--- a/assets/src/hooks/v2/use_audio_readout.tsx
+++ b/assets/src/hooks/v2/use_audio_readout.tsx
@@ -11,7 +11,7 @@ const useAudioReadout = ({
   id,
   config
 }: UseAudioReadoutArgs): void => {
-  if (config === null || config === undefined) {
+  if (config === null || config === undefined || config.readoutIntervalMinutes === 0) {
     return;
   }
 


assets/src/components/v2/screen_container.tsx
```diff
@@ -71,10 +83,15 @@ const ScreenLayout: ComponentType<ScreenLayoutProps> = ({
 
 const ScreenContainer = ({ id }) => {
   const blinkConfig = useContext(BlinkConfigContext);
+  const audioConfig = useContext(AudioConfigContext);
   const [showBlink, setShowBlink] = useState(false);
 
   const { apiResponse, requestCount } = useApiResponse({ id });
 
+  if (audioConfig != null) {
```
bfauble, line 91: Ideally you don't want to conditionally run hooks within a component since it can alter the order of hooks are called in subsequent renders. https://reactjs.org/docs/hooks-rules.html#only-call-hooks-at-the-top-level

You could pass the full `audioConfig` object to the hook itself and do the null check there and return early if the config is null. 

Additional question on this line, should this be `!==` null since this is checking for the default value which is `null`

PR #1111, opened 2021-11-03
tweak: Added headsign to second arrival departures audio.

cmaddox5:
**Asana task**: [Update the second bus arrival in departures audio](https://app.asana.com/0/1185117109217413/1201290145138216/f)

- [ ] Needs version bump?

### DIFF
--- a/lib/screens_web/views/v2/audio/departures_view.ex
+++ b/lib/screens_web/views/v2/audio/departures_view.ex
@@ -59,7 +59,8 @@ defmodule ScreensWeb.V2.Audio.DeparturesView do
     ~E|<s><%= content %></s>|
   end
 
-  defp render_time_with_crowding({%{crowding: crowding, time: time}, _}, route, _headsign) do
+  defp render_time_with_crowding({%{crowding: crowding, time: time}, _}, route, headsign) do
+    route_headsign_rendered = render_route_headsign(route, headsign)
     crowding_rendered = render_crowding_level(crowding)
     preposition = preposition_for_time_type(time.type)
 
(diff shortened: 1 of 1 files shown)


lib/screens_web/views/v2/audio/departures_view.ex
```diff
@@ -59,7 +59,8 @@ defmodule ScreensWeb.V2.Audio.DeparturesView do
     ~E|<s><%= content %></s>|
   end
 
-  defp render_time_with_crowding({%{crowding: crowding, time: time}, _}, route, _headsign) do
+  defp render_time_with_crowding({%{crowding: crowding, time: time}, _}, route, headsign) do
+    route_headsign_rendered = render_route_headsign(route, headsign)
```
jzimbel-mbta, line 63: Looks like we only wanted the route + vehicle type for subsequent times, not the full route + vehicle + headsign.
E.g. "A later 22 bus arrives ...", not "A later 22 bus to Jackson Square arrives ..."

This list should give that result when passed to `build_text/1`:
```ex
[
  prefix,
  {route.route_text, &render_route/1},
  route.vehicle_type || "trip",
  ...rest of the existing items
]
```
---

I also noticed a small bug/inconsistency in my own code—in `render_route_headsign/2` could you change this line: `{vehicle_type, fn v -> ~E|<%= v %>| end}` to just `vehicle_type || "trip"`?

cmaddox5: Good catch. Was so busy listening for the route that I didn't notice the headsign.

Also, on the screen I'm testing on (1401), I noticed that Polly is saying S-T instead of street. Are we ok with this or should we add some logic to get Polly to say the full word?

jzimbel-mbta: Thanks for catching that! I added a lexicon entry to fix it. (For the record, the instance was "Magazine St @ Green St")

jzimbel-mbta (approved):

PR #1112, opened 2021-11-04
doc: Update DUP packaging README

jzimbel-mbta:
**Asana task**: ad hoc

This bit about packaging for local testing was missing.

- [ ] Needs version bump?


assets/src/components/dup/README.md
```diff
@@ -9,6 +9,7 @@
 - Create priv/static/dup-app.html if it doesn’t already exist. Copy paste contents as directed below.
```
bfauble, line 9: while you're in here, do you think it'd be possible to move the dup-app.html code sample to underneath this bullet? it's kind of a weird flow of steps to hop down beyond the end of the instructions then back up to continue.

bfauble (changes requested):

bfauble (approved):

PR #1114, opened 2021-11-10
tweak: Update hardcoded B-branch stops in subway status widget

cmaddox5:
**Asana task**: [Update hardcoded B-branch stops in subway status widget](https://app.asana.com/0/1185117109217413/1201338926200103/f)

- [ ] Needs version bump?


bfauble (approved):
any idea where Pleasant St. is? ticket calls it out to be removed but not seeing it in the code

cmaddox5:
> any idea where Pleasant St. is? ticket calls it out to be removed but not seeing it in the code

I checked the project and couldn't find it. Was thinking it was assumed to be hardcoded because it is referenced in the change but was never added to the code. I can throw a question about that in the task so we know for sure we handled this properly.

cmaddox5:
> > any idea where Pleasant St. is? ticket calls it out to be removed but not seeing it in the code
> 
> I checked the project and couldn't find it. Was thinking it was assumed to be hardcoded because it is referenced in the change but was never added to the code. I can throw a question about that in the task so we know for sure we handled this properly.

Checked with Paul and Pleasant was actually removed a while ago. So only two stops need to be removed at this time.

PR #1116, opened 2021-11-12
feat: Add Sentry

cmaddox5:
**Asana task**: [Set up Sentry](https://app.asana.com/0/1185117109217413/1201331890574433/f)

Until SENTRY_DSN is added as an environment variable and `record_sentry` is set to true in the config, issues will not report to Sentry. DSNs can be found [here](https://sentry.io/settings/mbtace/projects/screens-frontend/keys/) if you ever want to test this locally.

PR to add new variables: https://github.com/mbta/devops/pull/201

- [ ] Needs version bump?

### DIFF
--- a/assets/src/apps/v2/dup.tsx
+++ b/assets/src/apps/v2/dup.tsx
@@ -2,7 +2,6 @@ declare function require(name: string): string;
 // tslint:disable-next-line
 require("../../../css/dup_v2.scss");
 
-import * as Sentry from "@sentry/react";
 import React from "react";
 import ReactDOM from "react-dom";
 import { BrowserRouter as Router, Route, Switch } from "react-router-dom";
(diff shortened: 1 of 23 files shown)


assets/src/apps/v2/dup.tsx
```diff
@@ -14,6 +15,13 @@ import Placeholder from "Components/v2/placeholder";
 import NormalHeader from "Components/v2/dup/normal_header";
 import NormalDepartures from "Components/v2/departures/normal_departures";
 
+const sentryDsn = document.getElementById("app")?.dataset.sentry;
```
bfauble, line 18: a way to cut down on the repeat of this init code would be to move it into its own custom hook `useSentry()` or something along those lines. 

You should be able to move all this logic into the hook definition and I believe it should work the same way.

cmaddox5: Good idea. Re-tested after the change and all works as expected.

assets/src/hooks/use_sentry.tsx
```diff
@@ -0,0 +1,13 @@
+import * as Sentry from "@sentry/react";
+
+const useSentry = () => {
+  const sentryDsn = document.getElementById("app")?.dataset.sentry;
+  console.log(sentryDsn);
```
bfauble, line 5: make sure to remove this. Also I think the only thing I would add to this just to avoid re-initializing is doing something like....
```js
useEffect(() => {
  if (sentryDsn) {
    Sentry.init( ... )
  }
}, [sentryDsn]);
```

bfauble (approved):

PR #1117, opened 2021-11-15
feat: E-Ink updated page layouts + alert widget

jzimbel-mbta:
**Asana task**: [Styles for alerts widgets on e-Ink screens](https://app.asana.com/0/1185117109217413/1200356261737393/f)

This turned out to be a very large change 😬

This adds alerts to both e-ink apps, but in the process I needed to rearrange the screen templates for both apps to add paging and allow for "filler" content to take over the bottom screen when there's an active takeover alert.

There were also a few challenging aspects of the alert widget designs, e.g. the special card background and the SVG arrow that has a different length in the flex zone alert vs the takeover alert.

I'm still testing all of my changes and need to take a different approach for the link arrow component, as setting SVG attributes from CSS turns out to be unsupported on most browsers besides Chrome, but wanted to get this PR posted for the sake of accountability, at least.

I expect to make a few more updates to this PR over the next day or two, but the bulk of it is done.

**Remaining**:
- [x] hard-code alert link arrows in a similar way to the card backgrounds
- [x] double-check alert styles, ask design about spacing between multiple pills + small pill variant
- [x] test changes


cmaddox5:
This looks good. Only thing I noticed is the new `bottom_screen_filler` widget instance does not have a test file. There is also not an audio view file (although I know this is eink so it won't use it).

jzimbel-mbta:
> This looks good. Only thing I noticed is the new `bottom_screen_filler` widget instance does not have a test file. There is also not an audio view file (although I know this is eink so it won't use it).

The audio view is [here](https://github.com/mbta/screens/pull/1117/files#diff-3b61afe00e97494a279315f793b2fa871915c6927bc00de4db5db13b77b3cce5) but I did miss adding a test file, thanks!

cmaddox5 (approved):

PR #1133, opened 2021-12-02
fix: Prevent text wrapping on DUP partial alert

cmaddox5:
**Asana task**: [DUP partial alert text for Ashmont/Braintree suspension and shuttle](https://app.asana.com/0/1185117109217413/1201010702999849/f)

The word `trains` caused bad wrapping for Ashmont/Braintree alerts. That word has been removed for that case. Also added CSS to prevent other cases from causing the same issue.

- [ ] Needs version bump?

### DIFF
--- a/lib/screens/dup_screen_data/response.ex
+++ b/lib/screens/dup_screen_data/response.ex
@@ -60,8 +60,16 @@ defmodule Screens.DupScreenData.Response do
     end
   end
 
+  defp partial_alert_specifier(%{headsign: {:adj, "Ashmont/Braintree" = headsign}}) do
+    {headsign, ""}
+  end
+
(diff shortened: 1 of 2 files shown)


jzimbel-mbta (approved):

lib/screens/dup_screen_data/response.ex
```diff
@@ -60,10 +60,18 @@ defmodule Screens.DupScreenData.Response do
     end
   end
 
+  defp partial_alert_specifier(%{headsign: {:adj, "Ashmont/Braintree" = headsign}}) do
```
bfauble, line 63: Seems like this signature only ever matches Kenmore "westbound" (via quick search for the `:adj` tag. So I don't think this is needed.

PR #1151, opened 2021-12-14
fix: name used for non-numerical route pills

cmaddox5:
**Asana task**: [Fix handling of non-numerical bus route pills](https://app.asana.com/0/1185117109217413/1201507620913022/f)

- [ ] Needs version bump?

### DIFF
--- a/test/screens/v2/departure_test.exs
+++ b/test/screens/v2/departure_test.exs
@@ -196,7 +196,7 @@ defmodule Screens.V2.DepartureTest do
 
   describe "route_name/1" do
     test "returns prediction route short_name when present" do
-      prediction = %Prediction{route: %Route{type: :bus, short_name: "214/216"}}
+      prediction = %Prediction{route: %Route{id: "34E", type: :bus, short_name: "214/216"}}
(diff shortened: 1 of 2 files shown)


test/screens/v2/departure_test.exs
```diff
@@ -196,15 +196,15 @@ defmodule Screens.V2.DepartureTest do
...
+      prediction = %Prediction{route: %Route{type: :bus, short_name: "214/216"}}
       schedule = %Schedule{route: %Route{id: "1", short_name: "1"}}
       departure = %Departure{prediction: prediction, schedule: schedule}
 
       assert "214/216" == Departure.route_name(departure)
     end
 
     test "returns schedule route short_name when no prediction is present" do
-      schedule = %Schedule{route: %Route{id: "214216", short_name: "214/216"}}
+      schedule = %Schedule{route: %Route{type: :bus, short_name: "214/216"}}
```
bfauble, line 207: can you add the breaking case that was causing this bug? 'sl2' or something like that for the id

bfauble (approved):

bfauble (approved):

PR #1153, opened 2021-12-17
feature: Create backend for pre fare screens

cmaddox5:
**Asana task**: [Set up pre-fare screen back-end](https://app.asana.com/0/1185117109217413/1201522871607744/f)

Left a comment in code but will reiterate here. This will initially use the standard fixed-page builder for flex. We will need to create a new sticky builder for variable page length.

- [ ] Needs version bump?

### DIFF
--- a/lib/screens/config/v2/pre_fare.ex
+++ b/lib/screens/config/v2/pre_fare.ex
@@ -1,17 +1,16 @@
 defmodule Screens.Config.V2.PreFare do
   @moduledoc false
 
-  alias Screens.Config.V2.Header.CurrentStopName
-
   @type t :: %__MODULE__{
-          header: CurrentStopName.t()
+          station_id: Screens.Stops.Stop.id()
         }
 
-  @enforce_keys [:header]
-  defstruct header: nil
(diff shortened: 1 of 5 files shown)


lib/screens/config/v2/pre_fare.ex
```diff
@@ -0,0 +1,16 @@
+defmodule Screens.Config.V2.PreFare do
+  @moduledoc false
+
+  @type t :: %__MODULE__{
+          station_id: Screens.Stops.Stop.id()
```
jzimbel-mbta, line 5: Following the convention from [bus shelter](https://github.com/mbta/screens/blob/master/lib/screens/config/v2/bus_shelter.ex) config, this struct can just be empty for now! We will add keys to it corresponding to each widget that can appear on pre-fare screens. (So probably, `header: CurrentStopId.t()` and `elevator_status: ElevatorStatus.t()` pretty soon)

jzimbel-mbta (approved):
Just two minor considerations. Approved assuming you respond as you see fit!

PR #1161, opened 2021-12-28
Update screens repo readme file

PaulJKim:
Description:

Encountered a few hurdles when setting up and running the screens application locally for the first time. This PR contains some amendments to the readme that should help future engineers get past those hurdles quickly.

Quick summary of the changes:
- Specify install of autoconf version 2.69 as it looks like latest versions of autoconf are preventing asdf from building erlang
- Include a workaround (a bit hacky) for MacOSX and Erlang configure error
- Include mention of `signs_ui_config.json` file needed for running the app

View a preview of the rendered readme [here](https://github.com/mbta/screens/tree/pk/update-readme#readme)


jzimbel-mbta (approved):
✅ Thanks! Hopefully if I can get the Devcontainer approach working this will be the last time you or another dev on the team needs to worry about building Erlang from source.

PR #1169, opened 2022-01-03
feat: Pre-Fare frontend

cmaddox5:
**Asana task**: [Set up pre-fare screen front-end](https://app.asana.com/0/1185117109217413/1201522871607749/f)

These changes line up with the changes made in https://github.com/mbta/screens/pull/1153. Any changes made during review of that PR will need to be tested with this PR before this is merged.

- [ ] Needs version bump?


jzimbel-mbta (approved):
Good to go! I had a comment about double checking heights of the header/body but you've already done that since then.

PR #1171, opened 2022-01-04
fix: Split long stop names on line map

cmaddox5:
**Asana task**: [[GL e-Ink] Long station names on line map wrap onto 2 lines](https://app.asana.com/0/1185117109217413/1201527717994173/f)

SVGs made this tricky. Final solution was to split the word on the space closest to the middle. Each part is then displayed in a `tspan` with the top `tspan` being 1.2em above the bottom. Only happens when the label has more than 15 characters. That number is completely arbitrary and is open for discussion.

- [ ] Needs version bump?


assets/src/components/v2/gl_eink_double/line_map.tsx
```diff
@@ -24,6 +25,22 @@ const abbreviateStop = (stop) => {
...
+  let before = labelText.lastIndexOf(" ", middle);
+  let after = labelText.indexOf(" ", middle + 1);
+
+  if (middle - before < after - middle) {
+    middle = before;
+  } else {
+    middle = after;
+  }
+
+  return [labelText.substring(0, middle), labelText.substring(middle + 1)];
```
jzimbel-mbta, line 41: Will we always find a space character? If one isn't found on either side of `middle` then this returns `["", labelText]`.

I did some checking of the possible results of this on paper and overall it looks right, but you might want to explicitly handle cases where a space isn't found on the left or right side, or both, just to be safe.

jzimbel-mbta (approved):
Looks good, one optional suggestion.

hannahpurcell (changes requested):
Seems that the new split stop names aren't centered around the stop-dot.

Screen Shot 2022-01-12 at 4 54 11 PM
 
RE:  the MAXIMUM_STOP_LABEL_LENGTH:  it's tricky to base formatting off of a char-length, especially when the number is so small (15) where variable character sizes can make a big impact.  But I bet it's definitely trickier working with calculated widths, etc.  I guess we should check with design on the label length to make sure it's ok that some names that weren't broken to 2 lines will now be broken onto 2 lines.  For example, Fairbanks St and Washington Sq:

Screen Shot 2022-01-12 at 5 01 12 PM Screen Shot 2022-01-12 at 5 01 05 PM

cmaddox5:
@jzimbel-mbta I tried out some of our longer names and they seem to wrap as expected with this solution. Screenshots below.

Screen Shot 2022-01-21 at 8 21 45 AM
Screen Shot 2022-01-21 at 8 21 32 AM

@hannahpurcell According to design, the bottom line should be lined up with the dot. I think this is to prevent crowding in the event of two long station names on top of each other. As for the max char length, I did bump that up just a bit. I played around with a few of our longest names and those wrap like we expect. Some names that really don't need to wrap shouldn't with the new length. 

Besides keeping a map of names we want to wrap and how they should wrap, I think this may be the best solution when it comes to doing it programmatically. Keeping this map seems tedious but I'm open to your opinions on it.

assets/src/components/v2/gl_eink_double/line_map.tsx
```diff
@@ -15,6 +15,7 @@ const TEXT_LEFT_MARGIN = 18;
 const TEXT_TOP_MARGIN = 10;
 const VEHICLE_ICON_SIZE = 44;
 const SCHEDULED_DEPARTURE_SIZE = 192;
+const MAXIMUM_STOP_LABEL_LENGTH = 18;
```
hannahpurcell, line 18: This const is only used to determine whether or not the line should wrap, but it's still possible to have a stop label exceed this max on a single line.  (For example, `abcdefg hi jklmnopqr stu vwxyz` will wrap like `abcdefg hi jklmnopqr\n stu vwxyz`, giving the first line 20chars.)  Maybe add clarifying comment:  `Max length before text wrapping`)

hannahpurcell (approved):
> According to design, the bottom line should be lined up with the dot.

Oh gosh, sorry I missed that.  Looks good!

PR #1176, opened 2022-01-07
feat: elevator status widget instance

cmaddox5:
**Asana task**: [Define elevator status WidgetInstance module (with stubs)](https://app.asana.com/0/0/1201621878856246/f)

- [ ] Needs version bump?


jzimbel-mbta (approved):
Looks good for the initial definition! We might need to add another key to the struct to hold info (stop name at least; possibly lat/long as well) about the home stop and any stops informed by the alerts, but those and other things can be added as they become necessary.

PR #1178, opened 2022-01-13
feat: elevator status widget instance

cmaddox5:
**Asana task**: [Elevator status widget module](https://app.asana.com/0/1185117109217413/1201626469173588/f)

This is a first draft of the WidgetInstance. Right now, we are not doing any paging based on pixel heights. It will be easier to pick pixel height values once the front end is started.

- [ ] Needs version bump?

### DIFF
--- a/lib/screens/v2/widget_instance/elevator_status.ex
+++ b/lib/screens/v2/widget_instance/elevator_status.ex
@@ -298,13 +298,6 @@ defmodule Screens.V2.WidgetInstance.ElevatorStatus do
     end)
   end
 
-  # defp parse_facility_data(facilities) do
-  #   facilities
-  #   |> Enum.map(fn %{"attributes" => %{"long_name" => long_name}, "id" => id} ->
-  #     %{name: long_name, id: id}
-  #   end)
-  # end
-
   defp serialize_detail_page(
          %Alert{header: header, informed_entities: entities} = alert,
          %__MODULE__{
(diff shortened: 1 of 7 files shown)


lib/screens/v2/widget_instance/elevator_status.ex
```diff
@@ -0,0 +1,432 @@
...
+
+    upcoming_on_connecting_lines =
+      t
+      |> get_upcoming_on_connecting_lines()
+      |> Enum.map(&serialize_detail_page(&1, t))
+
+    # first show detail pages for each closure at this station
+    # then if there is still space, show list pages for _all_ active closures across the system
+    # then if there is still space, show detail pages for upcoming closures at this station
+    # then if there is still space, show detail pages for active closures along lines serving this station
```
hannahpurcell, line 388: Couple questions I have for this section: 

Where did the points for 385-88 come from?  Is it coming from the "Content Logic" page of the mocks?

Line 388 says "show detail pages for active closures along lines serving this station" (which matches the "Content Logic" page).  But on line 395, we're passing in `upcoming_on_connecting_lines` instead of an `active_on_connecting_lines` value.  I'm not sure the design ever uses an "upcoming on connecting lines" value

hannahpurcell (commented):
Getting there on review... first, a question

lib/screens/v2/widget_instance/elevator_status.ex
```diff
@@ -0,0 +1,432 @@
...
+      _ -> false
+    end)
+  end
+
+  # defp parse_facility_data(facilities) do
+  #   facilities
+  #   |> Enum.map(fn %{"attributes" => %{"long_name" => long_name}, "id" => id} ->
+  #     %{name: long_name, id: id}
+  #   end)
+  # end
```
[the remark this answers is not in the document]
cmaddox5: No. That and the commented code in `serialize` shouldn't be there anymore. Removing it now.

lib/screens/v2/widget_instance/elevator_status.ex
cmaddox5: Good catch. I'm not really sure what I was thinking here, but luckily the fix is easy. Just pushed up the changes.

PR #1194, opened 2022-01-24
fix: Single alert GL multi branch edge case

cmaddox5:
**Asana task**: [[Subway Status] Handle case where a single alert affects multiple branches](https://app.asana.com/0/1185117109217413/1201514401502521/f)

A single alert on the GL with stops on multiple lines selected was causing the screen to crash. Added a case statement to handle this edge case.

Screen Shot 2022-01-24 at 8 44 44 AM
Screen Shot 2022-01-24 at 8 44 49 AM

- [ ] Needs version bump?


hannahpurcell (approved):
Woohoo!!  Simple and works like a charm

PR #1196, opened 2022-01-26
Persistent Widget

hannahpurcell:
**Asana task**: [Implement sticky component and HoC](https://app.asana.com/0/1185117109217413/1201621878856258)

A little rough with console logs still there for ease of testing (need to remove those logs, add a couple comments, run prettier), but otherwise!  Ready for review, at last!


assets/src/components/v2/bus_shelter/normal_body.tsx
```diff
@@ -1,25 +1,78 @@
...
+import useInterval from "Hooks/use_interval";
+import FlexZonePageIndicator from "Components/v2/flex/page_indicator";
 
 interface Props {
   main_content: WidgetData;
   flex_zone: WidgetData;
   footer: WidgetData;
 }
 
+const ProofOfConceptElevatorWidget = ({ data, lastUpdate, onFinish }) => {
```
cmaddox5, line 14: Might be a good idea to put this file back to normal now that the ElevatorStatus widget skeleton has been started. Either that or move it to PreFare because BusShelter wouldn't use the widget anyway.

hannahpurcell: Mmm definitely, will shift it to PreFare

cmaddox5:
Besides moving the POC widget to PreFare, this looks like what we need! I have a feeling most issues (and there might not be any) will come up after integrating this with the actual elevator widget.

cmaddox5 (approved):

assets/src/components/v2/bus_shelter/persistentHOC.tsx
```diff
@@ -0,0 +1,38 @@
+import React, { useContext, useEffect, useState } from "react";
```
cmaddox5, line 1: I missed this before, but this should live in the level above the apps. The best place for this is probably `src/apps/components/v2`. I feel like multiple apps would need to use this in the future. Can you also change the name of the file to match the name of the "component"? In this case, `persistent_wrapper.tsx`.

cmaddox5 (changes requested):

assets/src/components/v2/bus_shelter/normal_body.tsx
```diff
@@ -28,4 +28,4 @@ const NormalBody: React.ComponentType<Props> = ({
   );
 };
 
-export default NormalBody;
+export default NormalBody;
```
hannahpurcell, line 31:
```suggestion
export default NormalBody;

```

hannahpurcell:
There we go, moved `persistent_wrapper.tsx` up a level. I don't believe we have JS/TS tests, is that right?

PR #1208, opened 2022-02-01
feat: Finalize pre-fare candidate_generator

cmaddox5:
**Asana task**: [Pre-fare template:  update the candidate_generator with finalized template](https://app.asana.com/0/1185117109217413/1201736552436953/f)

Changed some slots names, added body_takeover for left screen, added flex_zone, moved SubwayStatus instance functions to a neutral folder.

- [ ] Needs version bump?

### DIFF
--- a/lib/screens/v2/candidate_generator/pre_fare.ex
+++ b/lib/screens/v2/candidate_generator/pre_fare.ex
@@ -16,21 +16,23 @@ defmodule Screens.V2.CandidateGenerator.PreFare do
           %{
             screen_normal_left: [
               :header_left,
-              :main_content_left
+              {:body_left,
+               %{
+                 body_normal_left: [:main_content_left],
+                 body_takeover_left: [:full_body_left]
+               }}
             ],
             screen_takeover_left: [
               :full_screen_left
-            ],
-            body_takeover_left: [:full_body_left]
+            ]
           }},
          {:right,
           %{
             screen_normal_right: [
               :header_right,
-              {:body,
+              {:body_right,
                %{
-                 body_normal: [
-                   :lower_right,
+                 body_normal_right: [
                    Builder.with_paging(
                      {:upper_right,
                       %{
@@ -38,7 +40,8 @@ defmodule Screens.V2.CandidateGenerator.PreFare do
                         two_medium: [:medium_left, :medium_right]
                       }},
                      2
-                   )
+                   ),
+                   :lower_right
                  ],
                  body_takeover_right: [:full_body_right]
                }}
--- a/lib/screens/v2/candidate_generator/pre_fare.ex
+++ b/lib/screens/v2/candidate_generator/pre_fare.ex
@@ -20,7 +20,8 @@ defmodule Screens.V2.CandidateGenerator.PreFare do
             ],
             screen_takeover_left: [
               :full_screen_left
-            ]
+            ],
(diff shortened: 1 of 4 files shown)


lib/screens/v2/candidate_generator/pre_fare.ex
```diff
@@ -20,16 +20,27 @@ defmodule Screens.V2.CandidateGenerator.PreFare do
...
               :full_screen_left
-            ]
+            ],
+            body_takeover_left: [:full_body_left]
           }},
          {:right,
           %{
             screen_normal_right: [
               :header_right,
               {:body,
```
jzimbel-mbta, line 30: If you agree with the above suggestion, this should probably be renamed to `:body_right`, as well as the branch label `body_normal` -> `body_normal_right`.

(I wonder if there's a less cumbersome way to do this than appending `_left` and `_right` all over the place :thinking:)

lib/screens/v2/candidate_generator/pre_fare.ex
```diff
@@ -20,16 +20,27 @@ defmodule Screens.V2.CandidateGenerator.PreFare do
             ],
             screen_takeover_left: [
               :full_screen_left
-            ]
+            ],
+            body_takeover_left: [:full_body_left]
```
hannahpurcell, line 24: This still needs the header, I think.  Could be consolidated with `screen_normal_left`, maybe?

jzimbel-mbta (approved):
✅

lib/screens/v2/candidate_generator/pre_fare.ex
cmaddox5: I've thought about that a bunch and can't think of a better way given the direction we went with both screens using the same URL. If each screen had its own URL, we wouldn't need `_left` or `_right` but would be a big change in the frontend.

lib/screens/v2/candidate_generator/pre_fare.ex
cmaddox5: Good catch. Changed up the left template to allow for that.

lib/screens/v2/candidate_generator/pre_fare.ex
hannahpurcell: 🤔  Why are the appends needed?  

I see why `:main_content_left` and `:main_content_right` need to be differentiated because they map to different components, but not so sure about the containers, like `:body_left` and `:screen_takeover_right`.  

As I've interpreted it so far, the containers are kinda just branches that don't need to keep track of whether they are on the right screen or left screen.  It's redundant to say `:body_left` because we've already gone down the `screen_normal_left` branch.  Just `:body` could work.  Could you explain what I'm missing?

jzimbel-mbta: That's true Hannah—it's possible that only the "leaf node" slots (that actually get widgets dropped into them) need to have unique names across the whole template, since those names are used by the widgets themselves to indicate exactly which slot(s) they can be placed in.

I can try testing this out after the branch is merged! Don't want to burden Christian with the extra work.

hannahpurcell (approved):
Looks good!  Might consider simplifying slot names to more generic ones -- the complexity of `body_normal_left` is not needed, while the complexity of `main_content_left` (which represents a component) is needed

PR #1211, opened 2022-02-02
feat: Add elevator instance to pre-fare candidate_generator

cmaddox5:
**Asana task**: [Update pre-fare app candidate generator](https://app.asana.com/0/1185117109217413/1201621878856254/f)

~~This PR has a dependency on https://github.com/mbta/screens/pull/1178 so that will need to be reviewed first.~~

Original PR: https://github.com/mbta/screens/pull/1190


lib/screens/v2/candidate_generator/widgets/elevator_closures.ex
```diff
@@ -0,0 +1,102 @@
...
+      |> Enum.map(fn %{stop: stop_id} -> stop_id end)
+      |> Enum.filter(&String.starts_with?(&1, "place-"))
+    end)
+  end
+
+  defp routes_to_icons(routes) do
+    routes
+    |> Enum.map(fn
+      %Screens.Routes.Route{type: :subway, id: id} -> id |> String.downcase() |> String.to_atom()
+      %Screens.Routes.Route{type: :light_rail} -> :green
```
hannahpurcell, line 89: Not necessarily :green.  The Mattapan line is a light rail segment of the RL.  There is at least one station (Ashmont) on Mattapan that has an elevator

lib/screens/v3_api.ex
```diff
@@ -5,7 +5,7 @@ defmodule Screens.V3Api do
 
   @default_opts [timeout: 2000, recv_timeout: 2000, hackney: [pool: :api_v3_pool]]
 
-  def get_json(route, params \\ %{}, extra_headers \\ [], opts \\ []) do
+  def get_json(route, params \\ %{}, extra_headers \\ [], opts \\ [], include_headers \\ false) do
```
hannahpurcell, line 8: Hmm, I think a term like `include_headers` usually flags whether or not the request is passed the headers, but here it's used to flag whether the returned data should have the header.  What do you think of calling it `return_with_headers`?

lib/screens/v2/candidate_generator/widgets/elevator_closures.ex
cmaddox5: Good to know. I added `id` to the pattern to only look at GL ids.

lib/screens/v3_api.ex
cmaddox5: That's a good idea. Especially not that we are dealing with request and response headers now.

lib/screens/v2/candidate_generator/widgets/elevator_closures.ex
hannahpurcell: Which of these conditionals will catch the Mattapan route?  Looks like it will fall to `%Screens.Routes.Route{type: type} -> type`, which will return the type `:light_rail` and doesn't map to a route icon

hannahpurcell: Shouldn't it be an icon option for the widget?  Or, I guess it would be captured ok by the RL icon

cmaddox5: Good question. I will ask Mindy about this.

cmaddox5: Yep, Mattapan should have an icon! Went ahead and added it here. Mindy will work on an icon in the meantime, but that isn't needed for this.

hannahpurcell (approved):
nailed it!  Thanks, Christian 👍

PR #1214, opened 2022-02-08
feat: Elevator Status widget DetailPage

cmaddox5:
**Asana task**: [Elevator status widget component - detail view](https://app.asana.com/0/1185117109217413/1201621878856264/f)

Ended up being a little bigger than we thought. Some backend changes were needed for this. Mostly because I thought of a more concise way to represent a DetailPage.

- [ ] Needs version bump?


lib/screens/v2/widget_instance/elevator_status.ex
```diff
@@ -182,16 +188,18 @@ defmodule Screens.V2.WidgetInstance.ElevatorStatus do
          %Alert{effect: :elevator_closure, informed_entities: entities} = alert,
          %__MODULE__{now: now, stop_sequences: stop_sequences} = t
        ) do
-    stations = get_stations_from_entities(entities)
+    stations = for %{stop: stop} <- entities, do: stop
```
jzimbel-mbta, line 191: My bad, this code will likely give us a list with a lot of `nil` elements because each entity always has a `stop` key (which is potentially `nil`). Feel free to confirm.

That's not going to break anything since we just use this list in equality checks, but it could slightly impact efficiency.

Suggestion:
```suggestion
    stations = for %{stop: stop} when is_binary(stop) <- entities, do: stop
```
or even more specific since we're only looking for platform stop IDs:
```suggestion
    informed_platforms = for %{stop: stop} when is_binary(stop) <- entities, match?({_n, ""}, Integer.parse(stop)), do: stop
```

assets/css/v2/pre_fare/elevator_status.scss
```diff
@@ -0,0 +1,154 @@
...
+      position: relative;
+
+      .detail-page__closure-alert-icon-container {
+        position: absolute;
+        left: -68px;
+        height: 45px;
+        width: 45px;
+
+        .detail-page__closure-alert-icon {
+          object-fit: fill;
```
jzimbel-mbta, line 120: Same here
```suggestion
          object-fit: contain;
```

jzimbel-mbta (commented):
Just leaving comments on some things I noticed while we were pairing on this and either didn't have time to bring up or hadn't thought of a suggested alternative yet.

assets/src/components/v2/elevator_status.tsx
```diff
@@ -0,0 +1,273 @@
...
+
+  const page = pages[pageIndex];
+  let pageToRender;
+  if (page == null) return null;
+  if (instanceOfDetailPage(page)) {
+    pageToRender = <DetailPageComponent {...(page as DetailPage)} />;
+  } else if (instanceOfListPage(page)) {
+    pageToRender = <ListPageComponent {...(page as ListPage)} />;
+  } else {
+    return null;
```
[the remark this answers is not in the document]
cmaddox5: Throwing an error is two birds one stone, so I'll add that.

lib/screens/v2/widget_instance/elevator_status.ex
cmaddox5: Wow, super cool. Elixir is awesome.

lib/screens/v2/widget_instance/elevator_status.ex
```diff
@@ -182,16 +188,22 @@ defmodule Screens.V2.WidgetInstance.ElevatorStatus do
...
     flat_stop_sequences =
       stop_sequences
       |> List.flatten()
 
     Alert.happening_now?(alert, now) and
-      Enum.any?(stations, fn station ->
-        station in flat_stop_sequences and station != parent_station_id
+      Enum.any?(informed_platforms, fn platform ->
+        platform != parent_station_id and platform not in platform_stop_ids and
+          platform in flat_stop_sequences
```
hannahpurcell, line 206: Trying to understand what `platform_stop_ids` is in the elevator status config, and why it is needed on pre-fare screens.  (I do understand why we'd need the ids for the elevator screens down the road)

That will help me understand what this piece is doing:  `platform not in platform_stop_ids`

cmaddox5: The reason we need them is to know what `platform_ids` are located at the `parent_station`. With these elevator alerts, the `informed_entities` list contains (most importantly) the parent station. So an elevator closure at Sullivan Square will have an entry for `place-sull` in the `informed_entities` list. There will also be an entry for each platform at Sullivan Square. These are numerical ids that do not tell us anything by looking at them. We need to know these ids when looking for alerts on connecting routes. If we do not exclude `platform_ids` at the home station, `active_at_home` alerts will show in the `active_on_connecting_routes` list. So if an alert affects a stop in the `stop_sequences` list and does not have an entry for any associated ids for the parent station, we know it is on the connecting route and not for the station itself.

hannahpurcell: Do we need to have `platform_stop_ids` be a configuration, though?  Is there any way to get the `platform_stop_ids` from the API given the parent station so that we don't have to hardcode the `parent_station_id` _and_ the `platform_stop_ids` for each screen?

lib/screens/v2/widget_instance/elevator_status.ex
```diff
@@ -270,18 +283,25 @@ defmodule Screens.V2.WidgetInstance.ElevatorStatus do
         serialize_closure(alert, facility, now)
       end)
 
+    icons =
+      station_id_to_icons
+      |> Map.fetch!(parent_station_id)
+      |> Enum.take(3)
```
hannahpurcell, line 289: out of curiosity, are there any situations we know of where the icon list is longer than 3?

lib/screens/v2/widget_instance/elevator_status.ex
cmaddox5: I don't remember the exact reason we went this way originally. I think it was to do with how complicated the query was getting. We went to Paul and asked how he felt about it being in the config and he felt that was reasonable. Mostly because these IDs are not going to be changing.

lib/screens/v2/widget_instance/elevator_status.ex
cmaddox5: Hmm, good question. I did not encounter one in testing, but I threw the `take` on there just in case because design gave that maximum.

lib/screens/v2/widget_instance/elevator_status.ex
hannahpurcell: Mmm, I guess thats true, ids won't be changing.  Just out of curiosity, I'll ask it in slack.  But otherwise, you've got my approval 👍

hannahpurcell (approved):

PR #1217, opened 2022-02-09
tweak: Pre-Fare frontend changes to match backend template

cmaddox5:
**Asana task**: [Pre-fare template: implement frontend changes to the screen template](https://app.asana.com/0/1185117109217413/1201736552436957/f)

Some components were missing/not correctly named because we adjusted the template in the backend. This should have all of the pieces we need to plug and play new widgets.

- [ ] Needs version bump?

### DIFF
--- a/assets/css/v2/pre_fare/flex/one_large.scss
+++ b/assets/css/v2/pre_fare/flex/one_large.scss
@@ -1,0 +1,12 @@
+.flex-one-large {
+  position: relative;
+  z-index: 2;
+}
+
+.flex-one-large__large {
+  position: absolute;
+  top: 24px;
+  left: 28px;
+  width: 1024px;
+  height: 576px;
+}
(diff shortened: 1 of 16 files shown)


assets/css/v2/pre_fare/flex/one_large.scss
```diff
@@ -0,0 +1,12 @@
+.flex-one-large {
+  position: relative;
+  z-index: 2;
```
jzimbel-mbta, line 3: We might not need `z-index` values since the pre-fare flex zone doesn't involve the "sunken tray" situation that the bus shelter one uses, but we can come back to this during final review.
(I had to set `z-index`es to have drop-shadows display correctly from both the containing tray and the cards "floating" inside the tray)

jzimbel-mbta (approved):
Looks good! Just a thought about the `z-index` styles, but I'm keeping a reminder to check back on that during final polish.

assets/css/v2/pre_fare/flex/one_large.scss
cmaddox5: I removed them and watched them for a bit and saw no change. So I think you are right, we shouldn't need them.

PR #1219, opened 2022-02-09
 Pre-Fare Headers

hannahpurcell:
**Asana task**: [Implement header widgets](https://app.asana.com/0/1185117109217413/1201753620238503)

This doesn't perfectly match the steps outlined in the Asana task.  Since it's my first time getting a widget set up, it was a great opportunity to learn by trial and error and figure out how to set up a widget on my own by mimicking other parts of the repo.  

I made some choices that seemed more convenient and _seemed_ logical, but might violate some of the "normal ways" of doing things.  For example, I decided _not_ to make two separate widgets when I could re-use a configurable widget.  I prefer that, but please do push back if that was the wrong choice!

[EDIT] Still writing a couple tests.  [Update] tests done!

- [ ] Needs version bump?


lib/screens/v2/candidate_generator/pre_fare.ex
```diff
@@ -62,15 +65,26 @@ defmodule Screens.V2.CandidateGenerator.PreFare do
...
+      fn -> header_instances(config, now) end,
+      fn -> elevator_status_instances_fn.(config, now) end,
+      fn -> placeholder_instances() end
+    ]
     |> Task.async_stream(& &1.(), ordered: false, timeout: :infinity)
     |> Enum.flat_map(fn {:ok, instances} -> instances end)
   end
 
+  defp header_instances(config, now) do
+    %Screen{app_params: %PreFare{header: %CurrentStopId{stop_id: stop_id}}} = config
```
cmaddox5, line 78: Are we assuming the `stop_id` in the header config will be populated with the full name of the station? If so, it may make more sense to use `CurrentStopName.t()` instead. Either that or use `CurrentStopId.t()` and make a V3 call to get the name of the station.

cmaddox5 (commented):
This looks great! I like that your solution is able to use the common components in both the front and backend. Just had one question/suggestion before approving.

lib/screens/v2/candidate_generator/pre_fare.ex
cmaddox5: I wonder if we could combine them by fetching from V3 if the given value is an integer or contains "place-". That should cover station IDs. If it isn't one of those, we just use the string.

hannahpurcell:
Fully updated to accommodate the new dual screen view!  Wasn't too much refactoring at all!

test/screens/v2/candidate_generator/pre_fare_test.exs
```diff
@@ -55,4 +57,35 @@ defmodule Screens.V2.CandidateGenerator.PreFareTest do
...
+          icon: nil,
+          text: "Test Station",
+          time: ~U[2020-04-06T10:00:00Z]
+        },
+        %NormalHeader{
+          screen: config,
+          icon: nil,
+          text: "Test Station",
+          time: ~U[2020-04-06T10:00:00Z]
+        }
```
cmaddox5, line 78: Are two header instances still expected now that we no longer separate left vs. right?

cmaddox5 (approved):
Looks great! Just had one question for one of the tests.

test/screens/v2/candidate_generator/pre_fare_test.exs
hannahpurcell: Aha!  Thanks!
```suggestion
        %NormalHeader{
          screen: config,
          icon: nil,
          text: "Test Station",
          time: ~U[2020-04-06T10:00:00Z]
        }
```

PR #1220, opened 2022-02-10
feat: Split pre-fare screen content using CSS

jzimbel-mbta:
**Asana task**: ad hoc / might fill this in with a task afterward

Previously we achieved the screen split in JS/HTML, but this meant a strict division between the left and right sides. In some cases (e.g., the header and high-impact alerts) we want a single widget's content to span across both screens.

This changes the split to be achieved in CSS. The client code renders the entire 2160x1920 page, and based on the value of the `screen_side` URL query param, sets a CSS class on a containing div to slide the appropriate half of that content into view inside the physical screen's 1080x1920 viewport.

image

- [ ] Needs version bump?


cmaddox5 (approved):
Looks good, but I was there when most of this was written. I'll give my approval anyway.

assets/css/v2/pre_fare/header/normal_left.scss
```diff
@@ -1,3 +1,5 @@
+// TODO: remove this file and merge styles into normal_header.scss
+
```
hannahpurcell, line 2: TODO when?

assets/css/v2/pre_fare/top_level_switch.scss
```diff
@@ -0,0 +1,27 @@
+.pre-fare-screen-viewport {
+  width: 1080px;
```
hannahpurcell, line 2: Doesn't have to be here, but I might rename this file `pre-fare-dual-screen.scss`

hannahpurcell (approved):
👍

PR #1223, opened 2022-02-15
tweak: Api fetch interval changes

cmaddox5:
**Asana task**: [Synchronize left and right screen data refreshes](https://app.asana.com/0/1185117109217413/1201756843663652/f)

In order to sync up the pre-fare screens, I made a v2 change to give api fetches more of a schedule instead of just fetching on a set interval. Schedule is dependent on refresh rate of the screen. There is an offset built in to each screen so not all screens will request data at the same time.

- [ ] Needs version bump?


lib/screens_web/controllers/v2/screen_controller.ex
```diff
@@ -44,15 +44,24 @@ defmodule ScreensWeb.V2.ScreenController do
...
           Parameters.get_audio_interval_offset_seconds(config)
         )
         |> assign(:sentry_frontend_dsn, Application.get_env(:screens, :sentry_frontend_dsn))
+        |> assign(
+          :refresh_rate_offset,
+          screen_id
+          |> Base.encode16()
+          |> String.to_integer(16)
+          |> rem(refresh_rate)
+        )
```
hannahpurcell, line 64: nice nice, great solution

test/screens_web/controllers/v2/screen_controller_test.exs
```diff
@@ -3,7 +3,7 @@ defmodule ScreensWeb.V2.ScreenControllerTest do
 
   describe "index/2" do
     test "returns 200", %{conn: conn} do
-      assert %{status: 200} = get(conn, "/v2/screen/1")
+      assert %{status: 200} = get(conn, "/v2/screen/1401")
```
hannahpurcell, line 6: why the change?

hannahpurcell (approved):
Looks good!  Especially loved seeing it in action 🙌

assets/src/hooks/v2/use_api_response.tsx
```diff
@@ -110,17 +115,28 @@ const useApiResponse = ({
...
-  if (refreshMs != null) {
-    useInterval(() => {
+    // Schedule subsequent fetches
+    setTimeout(function setFetchDataInterval() {
       fetchData();
-    }, refreshMs);
-  }
+      setTimeout(setFetchDataInterval, calculateMsToNextRefresh());
+    }, calculateMsToNextRefresh());
+  }, []);
```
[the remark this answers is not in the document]
cmaddox5: At first glance, I do like your solution better. Would love to jump on a call and talk over this though. This particular issue is not one that I think I fully understand yet.

hannahpurcell (approved):
I confess that some of this is a bit over my head, but I get the gist.  I like the reusable hook, and it's good to have it applied to the audio readout.  One request for a comment, but here's a fresh approval

PR #1229, opened 2022-02-22
fix: Subway Status with GL alert affecting whole line

cmaddox5:
**Asana task**: [[bug] Subway status ignores GL alerts which affect the entire Green Line](https://app.asana.com/0/1185117109217413/1201804054848498/f)

When an alert that affecting the whole GL exists, the subway status was ignoring it. This was because we only looked for trunk vs. branch alerts. Now, when an alert comes through with all 4 branches in its informed entities, we treat it as a trunk alert. 

Screen Shot 2022-02-22 at 8 52 38 AM

Screen Shot 2022-02-22 at 8 52 43 AM

- [ ] Needs version bump?

### DIFF
--- a/lib/screens/v2/widget_instance/subway_status.ex
+++ b/lib/screens/v2/widget_instance/subway_status.ex
@@ -542,7 +542,6 @@ defmodule Screens.V2.WidgetInstance.SubwayStatus do
     alert_whole_line_stops =
       informed_entities
       |> Enum.map(fn e -> Map.get(e, :route) end)
-      |> Enum.reject(&is_nil/1)
       |> Enum.filter(fn
         "Green-" <> _ -> true
         _ -> false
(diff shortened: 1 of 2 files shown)


lib/screens/v2/widget_instance/subway_status.ex
```diff
@@ -537,7 +539,19 @@ defmodule Screens.V2.WidgetInstance.SubwayStatus do
       |> Enum.into(MapSet.new())
       |> MapSet.intersection(gl_trunk_stops)
 
-    MapSet.size(alert_trunk_stops) > 0
+    alert_whole_line_stops =
+      informed_entities
+      |> Enum.map(fn e -> Map.get(e, :route) end)
+      |> Enum.reject(&is_nil/1)
```
hannahpurcell, line 545: Is the reject needed, or does the filter below handle "nil" case?

hannahpurcell (approved):
Thanks!  Looks great

PR #1232, opened 2022-02-24
Full line map widget backend

PaulJKim:
**Asana task**: [Create Line Map Widget](https://app.asana.com/0/1185117109217413/1201781408086812/f)

Description

This PR adds a line map widget for Prefare screens.

As of 2/24, the PR only contains backend changes. Mainly just want to get some early eyes on it.

Note: files are named prefixed with `pre_fare` because there is an existing line map widget for e-ink screens where the files are just named `line_map.ex` etc.

- [ ] Needs version bump?


lib/screens/v2/widget_instance/pre_fare_line_map.ex
```diff
@@ -0,0 +1,51 @@
...
+
+  defstruct screen: nil,
+            asset_url: nil,
+            slot_names: nil,
+            priority: nil
+
+  @type t :: %__MODULE__{
+          screen: Screen.t(),
+          asset_url: String.t(),
+          slot_names: list(WidgetInstance.slot_id()),
```
cmaddox5, line 15: If you are going to define `slot_names` and `priority` directly in here (which I think is fine for this widget), they do not need to be attributes on the struct.

lib/screens/v2/widget_instance/pre_fare_line_map.ex
```diff
@@ -0,0 +1,51 @@
...
+  defimpl Screens.V2.WidgetInstance do
+    alias Screens.V2.WidgetInstance.PreFareLineMap
+
+    def priority(instance), do: PreFareLineMap.priority(instance)
+    def serialize(instance), do: PreFareLineMap.serialize(instance)
+    def slot_names(instance), do: PreFareLineMap.slot_names(instance)
+    def widget_type(instance), do: PreFareLineMap.widget_type(instance)
+    def valid_candidate?(instance), do: PreFareLineMap.valid_candidate?(instance)
+    def audio_serialize(instance), do: PreFareLineMap.audio_serialize(instance)
+    @spec audio_sort_key(Screens.V2.WidgetInstance.PreFareLineMap.t()) :: 0
```
cmaddox5, line 46: This should be safe to remove. May have been a mistake.

cmaddox5 (changes requested):
Left a couple of comments. Another thing I would consider is changing the widget name to some app agnostic. With widgets, we want to make them as general as possible so any app can use them. Naming it `PreFareLineMap` is a little too specific. How about something like `StaticLineMap`? Feel free to also take this to our Slack channel and get more opinions on.

Other than that, great work on this! This should be exactly what we need!

lib/screens/v2/widget_instance/pre_fare_line_map.ex
PaulJKim: Ah yep think just a copy paste mistake

lib/screens/v2/widget_instance/pre_fare_line_map.ex
PaulJKim: Gotcha. I'm actually curious about this one. So I was doing some testing of the instances returned in the candidate generator and I noticed that if I did something like `instance.priority()` it would return `nil` and not `[2]` like I was expecting. But if I call the function at the module level, like `Screens.V2.WidgetInstance.FullLineMap.priority()` it would return that default value.

Why exactly is that the case? I figured it might have to do with functional programming principles and pure functions, but not sure. And also, how does this play into the framework too? In other words, when exactly would those directly defined values/defaults for priority and slot_names get called?

cmaddox5: With these `WidgetInstance` protocol functions, the instance is not the object that possesses the functions. It is the parameter that you feed into `Screens.V2.WidgetInstance.priority/1`. So in the case where you have separated the `defimpl` functions from the `full_line_map` functions, you either call `FullLineMap.priority/1` to get the specific return for `FullLineMap`, or call `WidgetInstance.priority/1` with any `WidgetInstance` type as a parameter to get the priority value for that particular instance. You can see this used in `screen_data` in `select_best_placement` on line 178. We get the priority in a generic way using `WidgetInstance.priority(instance)`.

Hopefully this makes sense. If not, feel free to reach out.

lib/screens/v2/candidate_generator/widgets/full_line_map.ex
```diff
@@ -0,0 +1,19 @@
+defmodule Screens.V2.CandidateGenerator.Widgets.FullLineMap do
+  @moduledoc false
+
+  alias Screens.Config.Screen
+  alias Screens.Config.V2.{PreFare, FullLineMap}
+  alias Screens.V2.WidgetInstance.FullLineMap, as: FullLineMapWidget
+  alias Screens.Util.Assets
+
+  def line_map_instances(%Screen{app_params: %PreFare{full_line_map: full_line_maps}} = config) do
```
cmaddox5, line 9: Sorry, one last nitpicky thing. Do you mind changing this function and functions with similar names to match the widget name? Normally it wouldn't be a big deal but we don't want to confuse this with the e-ink line map.

cmaddox5 (changes requested):

lib/screens/v2/candidate_generator/widgets/full_line_map.ex
PaulJKim: Oh definitely, thanks for that catch!

cmaddox5 (approved):

hannahpurcell (approved):
Confirming, still looks good after further commits 👍

PR #1235, opened 2022-02-28
Basics for the Reconstructed Alert

hannahpurcell:
**Asana task**: [Create reconstructed alert widget](https://app.asana.com/0/1185117109217413/1201879625507572)

More robust testing should be built out in the later tasks.  Subway status addition was slightly out-of-scope, but it sits in the same screen area as the Reconstructed Alerts, and it only required a few lines to add it, so it seemed beneficial to include


lib/screens/v2/candidate_generator/pre_fare.ex
```diff
@@ -53,21 +53,27 @@ defmodule Screens.V2.CandidateGenerator.PreFare do
...
     ]
     |> Task.async_stream(& &1.(), ordered: false, timeout: :infinity)
     |> Enum.flat_map(fn {:ok, instances} -> instances end)
   end
 
   defp header_instances(config, now) do
-    %Screen{app_params: %PreFare{header: %CurrentStopName{stop_name: stop_name}}} = config
+    %Screen{app_params: %PreFare{header: %CurrentStopId{stop_id: stop_id}}} = config
+
+    stop_name = Screens.V2.CandidateGenerator.BusShelter.fetch_stop_name(stop_id)
```
cmaddox5, line 76: I think it makes sense to fetch the name, but we should probably go ahead and move the fetch function to a utility file. I think `stop.ex` would be a good place for it that already exists.

cmaddox5: Ahh, I see the comments where you note this. So I'll let you decide between now or later.

lib/screens/config/v2/reconstructed_alert.ex
```diff
@@ -0,0 +1,14 @@
+defmodule Screens.Config.V2.ReconstructedAlert do
+  @moduledoc false
+
+  @type t :: %__MODULE__{alert: String.t()}
```
cmaddox5, line 4: Do you have something in mind that this key will hold? Or is this more of a placeholder?

lib/screens/v2/widget_instance/reconstructed_alert.ex
```diff
@@ -0,0 +1,34 @@
+defmodule Screens.V2.WidgetInstance.ReconstructedAlert do
+  @moduledoc false
+
+  alias Screens.V2.WidgetInstance.ReconstructedAlert
+
+  defstruct alert: nil
+
+  @type t :: %__MODULE__{
+          alert: String.t()
```
cmaddox5, line 9: This type should be `Alert.t()`.

lib/screens/v2/widget_instance/reconstructed_alert.ex
```diff
@@ -0,0 +1,34 @@
...
+
+  defimpl Screens.V2.WidgetInstance do
+    def priority(_instance), do: [3]
+    def serialize(t), do: ReconstructedAlert.serialize(t)
+    def slot_names(_instance), do: [:large]
+    def widget_type(_instance), do: :reconstructed_alert
+    def valid_candidate?(_instance), do: true
+    def audio_serialize(_instance), do: %{}
+    def audio_sort_key(_instance), do: 0
+    def audio_valid_candidate?(_instance), do: true
```
cmaddox5, line 31: This should be false until we are ready to implement audio.

assets/css/pre_fare_v2.scss
```diff
@@ -29,6 +29,10 @@
 @import "v2/pre_fare/elevator_status";
 @import "v2/pre_fare/flex/paging_indicator";
 
+// Can this scss be generic?
```
cmaddox5, line 32: I know this comment is probably for the future, but I think yes. There shouldn't be much (if any) differences between how Bus Shelter and Pre-Fare display Subway Status.

cmaddox5 (changes requested):
Looks great! Just found a couple of things that need changing and some questions.

lib/screens/v2/widget_instance/reconstructed_alert.ex
```diff
@@ -0,0 +1,34 @@
...
+  # Temporarily using subway alert type
+  def serialize(%__MODULE__{alert: %Screens.Alerts.Alert{header: header}}) do
+    %{alert_header: header}
+  end
+
+  # Priority will either be 3 if it's a regular alert
+  #     or 1 if a takeover
+
+  # slot_names will be :large if it's a regular alert
+  #     or body_left_takeover / body_right_takeover if a takeover (how can 1 alert fill both sides??)
```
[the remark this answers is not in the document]
hannahpurcell: Both screens.  Ohhh I think I figured it out:  I was confused at how we could put the takeover-version of the alert across both screens, but I realize we have a :full_body slot.  

Still, is there a way to put one widget in multiple slots?

hannahpurcell, line 21:
```suggestion
  #     or :full_body if a takeover
```

assets/css/pre_fare_v2.scss
hannahpurcell: I might as well do that now!

lib/screens/config/v2/reconstructed_alert.ex
hannahpurcell: Placeholder.  Right now, I only want to display _something_ about an alert for proof-of-concept, and currently that something is the alert header, a string.

lib/screens/v2/widget_instance/reconstructed_alert.ex
hannahpurcell: Surprising that this compiled properly...

lib/screens/v2/widget_instance/reconstructed_alert.ex
hannahpurcell, line 31:
```suggestion
    def audio_valid_candidate?(_instance), do: false
```

lib/screens/v2/widget_instance/reconstructed_alert.ex
```diff
@@ -0,0 +1,34 @@
...
+
+  defimpl Screens.V2.WidgetInstance do
+    def priority(_instance), do: [3]
```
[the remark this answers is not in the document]
hannahpurcell: At this moment, the SubwayStatus never appears if the priority of the alerts is [2].  There are too many alerts, so it wants to fill all the space.  Can I keep it at [3] for now, since this will get reworked in the later tickets anyway?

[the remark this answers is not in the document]
cmaddox5: Hmm, that's interesting. If that's the case, we can address it later.

lib/screens/v2/widget_instance/reconstructed_alert.ex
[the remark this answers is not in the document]
cmaddox5: Pre-Fare is definitely unique in this problem, but I _think_ we would need to conditionally return a WidgetInstance for each slot we want it to occupy. Kinda like how you originally had the header before the app layout change. If we use `full_body`, I think the component will need to be created in a way that creates a takeover on each screen in one render.

cmaddox5 (approved):

lib/screens/v2/widget_instance/reconstructed_alert.ex
```diff
@@ -0,0 +1,35 @@
...
+  end
+
+  # Priority will either be 3 if it's a regular alert
+  #     or 1 if a takeover
+
+  # slot_names will be :large if it's a regular alert
+  #     or :full_body if a takeover
+
+  defimpl Screens.V2.WidgetInstance do
+    def priority(_instance), do: [2]
```
hannahpurcell, line 25:
```suggestion
    def priority(_instance), do: [3]
```

PR #1237, opened 2022-03-01
tweak: Allow variable paging on V2 apps

cmaddox5:
**Asana task**: [Pre-Fare:  Allow flex zones to have a variable page number](https://app.asana.com/0/1185117109217413/1201812714325999/f)

This change looks through the selected candidates that are to be placed and removes slots from the layout if it does not have a matching candidate. This allows us to use a maximum page approach to paged slots that use `Screens.V2.Template`.

- [ ] Needs version bump?


hannahpurcell (approved):
This took me a while to parse, so I'm impressed with how quickly you wrote it!  Logical solution to a complicated problem ✔️

PR #1238, opened 2022-03-01
Full line map widget frontend

PaulJKim:
**Asana task**: [Create Line Map Widget](https://app.asana.com/0/1185117109217413/1201781408086812/f)

PR for the frontend changes to add the Full Line Map Widget.

Mechanism for switching between the images admittedly feels a little hacky. Very open to feedback/ideas there, although this is just an initial MVP.

Note: This branch was branched off of the backend changes so it also contains those.

- [ ] Needs version bump?


assets/src/components/v2/full_line_map.tsx
```diff
@@ -0,0 +1,35 @@
...
+const FullLineMap: ComponentType<Props> = ({asset_urls}) => {
+  const [assetIndex, setAssetIndex] = useState(0);
+
+  useEffect(() => {
+  if (assetIndex === asset_urls.length - 1) {
+    setTimeout(() => {setAssetIndex(0)}, intervalInMs)
+  } else {
+    setTimeout(() => {setAssetIndex((i) => i + 1)}, intervalInMs)
+  }
+}, [assetIndex])
```
jzimbel-mbta, line 18: This will work for now because the list of asset URLs for a given screen is essentially unchanging, but in the future when the content starts being more real-time, we'll need to use the persistent carousel higher-order component instead of this `useEffect` to achieve the correct paging logic.

assets/src/components/v2/full_line_map.tsx
```diff
@@ -0,0 +1,35 @@
...
+
+  return <Image assetUrl={asset_urls[assetIndex]} />;
+}
+
+interface ProperProps {
```
jzimbel-mbta, line 23: Maybe `ImageProps`?

jzimbel-mbta (approved):
Looks good, nice work!

PR #1239, opened 2022-03-02
Hp/alert data fetch

hannahpurcell:
**Asana task**: [Add Reconstructed Alert Widget to Pre-Fare screen candidate generator](https://app.asana.com/0/1185117109217413/1201752179318946)

`candidate_generator` handles fetching the alerts relevant to this the subway lines of this station, ditches alert effects that should never appear on screens, and passes this to the WidgetInstance

- [ ] Needs version bump?


lib/screens/v2/candidate_generator/widgets/reconstructed_alert.ex
```diff
@@ -1,27 +1,35 @@
...
-        )
-
-      :error ->
-        []
+  @doc """
+  Given the stop_id defined in the header, determine relevant routes
+  Given the routes, fetch all alerts for the route
+  """
+  def reconstructed_alert_instances(
+        %Screen{app_params: %PreFare{header: %CurrentStopId{stop_id: stop_id}}},
```
jzimbel-mbta, line 18: Eventually we'll want to use a separate config definition for the reconstructed alert widget, but this makes sense for now!

jzimbel-mbta (approved):
Looks good, just some minor optional comments/suggestions! Great test coverage too 💯

lib/screens/v2/candidate_generator/widgets/reconstructed_alert.ex
hannahpurcell: Ah, why? I don't foresee other values we'd need,  so a concise config seems useful.

jzimbel-mbta: (Discussed over a slack huddle—makes sense to avoid duplicating config values just because they're used by different widgets! We will probably try to define a common "screen location context" structure that all widgets can use as needed)

PR #1240, opened 2022-03-03
Elevator status list view + factor out persistent carousel logic

jzimbel-mbta:
**Asana task**: [Elevator status widget component - list view](https://app.asana.com/0/1185117109217413/1201621878856262/f)

This adds styles for the elevator status widget's list view. Logic to split up list view pages according to row heights will be added in a separate PR to keep the diff manageable.

I also split out the "persistent carousel" logic into its own HoC that builds on top of the `makePersistent` HoC, allowing the elevator status and other future frontend-paged widgets to skip the boilerplate and state management, and just pass a presentational component to the HoC.

Please let me know if the documentation makes sense!

Screenshots:
Screenshot 2022-03-03 at 09-23-00 Screens
Screenshot 2022-03-03 at 09-22-46 Screens

An example with multiple elevator closures at stations. Ignore the row overflow at the bottom for now; that will be prevented when we add the logic to properly fit content into pages.
Screenshot 2022-03-03 at 09-47-12 Screens

### DIFF
--- a/assets/src/components/v2/persistent_carousel.tsx
+++ b/assets/src/components/v2/persistent_carousel.tsx
@@ -1,11 +1,15 @@
 import React, { ComponentType, useEffect, useState } from "react";
-import makePersistent from "./persistent_wrapper";
+import makePersistent, { WrappedComponentProps } from "./persistent_wrapper";
 
-interface Props {
-  PageRenderer: ComponentType<any>;
+interface PageRendererProps {
+  page: any;
+  pageIndex: number;
+  numPages: number;
+}
+
+interface Props extends WrappedComponentProps {
+  PageRenderer: ComponentType<PageRendererProps>;
   pages: any[];
-  onFinish: () => void;
-  lastUpdate: number | null;
 }
 
 const Carousel: ComponentType<Props> = ({ PageRenderer, pages, onFinish, lastUpdate }) => {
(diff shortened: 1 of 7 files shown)


assets/src/components/v2/persistent_carousel.tsx
```diff
@@ -0,0 +1,99 @@
...
+    <PageRenderer page={pages[pageIndex]} pageIndex={pageIndex} numPages={pages.length} />
+  );
+}
+
+const PersistentCarousel = makePersistent(Carousel as ComponentType<WrappedComponentProps>);
+
+/**
+ * Call this function on some `PageRenderer` component to wrap it in a persistent component that
+ * expects a `pages` prop containing an array of data, and will pass one element of the array at a time to
+ * `PageRenderer`.
```
hannahpurcell, line 45: Is there a reason this is formatted like this instead of as a @moduledoc?

hannahpurcell, line 45: Is there a reason this is formatted like this instead of as a @moduledoc?

assets/src/components/v2/persistent_carousel.tsx
```diff
@@ -0,0 +1,99 @@
...
+ * // in some containing component's render...
+ * const pages = [
+ *   {url: "/psa.png", caption: "It's a PSA!"},
+ *   {url: "/squirrel.jpg", caption: "Beware."},
+ *   {url: "/survey.png", caption: "Penny for your thoughts?"}
+ * ];
+ *
+ * return (
+ *   <div>
+ *     <PersistentImageCarousel pages={pages} />
```
hannahpurcell, line 88: `PersistentImageCarousel` isn't explicitly defined within the example, so consider altering line 77 to define it as a const

hannahpurcell (approved):
I think the way you've split out the Persistent work into individual functional HOCs is great and wise.

The code is dense to me and tough to comb through.  However, I can't think of a way to distill it into anything simpler, and your documentation helps a lot.  So, 👍  from me.

The CSS naming you used is also v clean and conventional.  Thanks for that!

assets/src/components/v2/persistent_carousel.tsx
jzimbel-mbta: Yup, this is just standard JSdoc which I think is borrowed from Javadoc? `@moduledoc` is Elixir's syntax for essentially the same thing, so we use that to attach documentation to modules in the backend code.

assets/src/components/v2/persistent_carousel.tsx
jzimbel-mbta: Oops, thanks!

PR #1243, opened 2022-03-09
Cm/subway status gl direction error

cmaddox5:
**Asana task**: [Subway status errors](https://app.asana.com/0/1185117109217413/1201897942452378/f)

This bug occurs when a GL alert affects the whole GL in one direction (See screenshot below). There wasn't a mapping for the whole GL in `@route_directions` so `Map.get` was returning `nil`. After adding the mapping, everything looks good.

Screen Shot 2022-03-09 at 11 21 35 AM

Screen Shot 2022-03-09 at 11 21 53 AM

- [ ] Needs version bump?


hannahpurcell (approved):
Awesome, thanks for adding a test!  👍

PR #1246, opened 2022-03-10
tweak: Increased header title width

cmaddox5:
**Asana task**: [[Header] "Downtown Crossing" is wrapping](https://app.asana.com/0/0/1201896247257224/f)

I picked 1040 to give title the illusion of `padding-right: 40px`. Can't set padding now that one screen uses width of both screens.

- [ ] Needs version bump?

### DIFF
--- a/assets/css/v2/pre_fare/normal_header.scss
+++ b/assets/css/v2/pre_fare/normal_header.scss
@@ -16,7 +16,7 @@
 }
 
 .normal-header-title {
-  width: 824px;
+  width: 1040px;
   position: absolute;
   left: 40px;
   bottom: 40px;
--- a/assets/css/v2/pre_fare/normal_header.scss
+++ b/assets/css/v2/pre_fare/normal_header.scss
@@ -16,7 +16,7 @@
 }
 
 .normal-header-title {
-  width: 1040px;
+  width: 1000px;
   position: absolute;
   left: 40px;
   bottom: 40px;


assets/css/v2/pre_fare/normal_header.scss
```diff
@@ -16,7 +16,7 @@
 }
 
 .normal-header-title {
-  width: 824px;
+  width: 1040px;
```
jzimbel-mbta, line 19: Should this be 1000px? If the left screen width is 1080px and we're going for 40px of padding on either side of this text, and the left padding is already accounted for with `left: 40px;`, then the remaining width to allow here would be 1040 - 40 = 1000.

Screen Shot 2022-03-11 at 2 02 33 PM

jzimbel-mbta (changes requested):

assets/css/v2/pre_fare/normal_header.scss
cmaddox5: So you are definitely right, but unfortunately Downtown Crossing still wraps at 1000. So while 1040 was not picked for the right reason, we might need to use it or find a smaller width that fits that station name. What do you think?

jzimbel-mbta: Hmm, there's some logic in the base header widget that's supposed to use a smaller font size if the large size wraps. If I manually change the classname to `...--small`, DTX fits. Trying to see why it's not working as expected right now.

jzimbel-mbta: Actually I'll try not to get into the weeds on this and instead create a bugfix task!
With that said, I think it's best to go with the padding we want (40px on either side, which allows for a width of 1000px) and just make sure the logic to fit text within that space works.

jzimbel-mbta:
Just realized I was not very clear on what I'm asking for here—I think in this PR, **the width should be set to 1000px**, and in another PR we can fix the logic that's supposed to auto-resize the header text to fit in that space.

jzimbel-mbta (approved):

PR #1247, opened 2022-03-10
[Subway Status] Adjust footer text and color

PaulJKim:
**Asana task**: [[Subway Status] Adjust footer text and color](https://app.asana.com/0/0/1201914758246007/f)

Adjust the background and text colors for the Subway status widget footer. Also change the font to be unbolded.

Pre-fare:
Screen Shot 2022-03-11 at 10 27 05 AM

Bus Shelter:
Screen Shot 2022-03-11 at 10 27 20 AM

- [ ] Needs version bump?

### DIFF
--- a/assets/css/v2/pre_fare/subway_status.scss
+++ b/assets/css/v2/pre_fare/subway_status.scss
@@ -1,10 +1,10 @@
 @import "../lcd_common_styles/subway_status.scss";
 
 .subway-status-footer {
-  background-color: #cccbc8 !important;
+  background-color: #cccbc8;
 }
 
 .subway-status-footer__link {
-  font-weight: unset !important;
-  color: #171f26 !important;
+  font-weight: unset;
+  color: #171f26;
 }
--- a/assets/css/v2/pre_fare/subway_status.scss
+++ b/assets/css/v2/pre_fare/subway_status.scss
@@ -1,0 +1,10 @@
+@import "../lcd_common_styles/subway_status.scss";
+
+.subway-status-footer {
+  background-color: #cccbc8 !important;
+}
+
+.subway-status-footer__link {
+  font-weight: unset !important;
+  color: #171f26 !important;
+}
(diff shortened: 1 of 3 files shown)


PaulJKim:
Realized I missed a note in the ticket saying that this should only apply to this widget on pre-fare screens, not on bus shelter screens. Will circle back to that tomorrow

cmaddox5 (approved):

cmaddox5 (changes requested):
Just gonna add a Change Request to remove the approval until you move the CSS. Sorry about not looking closer at the note you left.

PaulJKim:
@cmaddox5 Just added a file with some overrides rather than copying the file over. Let me know if you agree with that approach or not!

assets/css/v2/pre_fare/subway_status.scss
```diff
@@ -0,0 +1,10 @@
+@import "../lcd_common_styles/subway_status.scss";
+
+.subway-status-footer {
+  background-color: #cccbc8 !important;
```
cmaddox5, line 4: I don't think `!important` is needed here because it will auto override whatever styles are in the import.

assets/css/v2/pre_fare/subway_status.scss
```diff
@@ -0,0 +1,10 @@
...
+}
+
+.subway-status-footer__link {
+  font-weight: unset !important;
+  color: #171f26 !important;
```
cmaddox5, line 9: Same as above.

cmaddox5 (commented):
Looks good! Good idea importing the existing styles instead of copy and pasting. 

~~Looking at the ticket, it looks like the background color of the elevator status widget footer should match this footer. We might need to look at changing that too. I'll get a ticket created for that.~~

Edit: The ticket mentioned in your task covers that. For some reason, I thought it was another ticket.

assets/css/v2/pre_fare/subway_status.scss
PaulJKim: oh gotcha, thanks!

cmaddox5 (approved):

PR #1250, opened 2022-03-11
tweak: Elevator Status footer color

cmaddox5:
**Asana task**: [[Elevator Footer] Adjust background color](https://app.asana.com/0/0/1201914758246004/f)

Adjusted background color

- [ ] Needs version bump?


jzimbel-mbta (approved):
Side note, I wish `box-sizing: border-box` was the default box-sizing treatment so that `width` and `height` actually do what they say 🙃

PR #1262, opened 2022-03-23
feat: Detail Page alert description variable text sizing

cmaddox5:
**Asana task**: [[Elevator Detail page] Use variable text sizes for the alert description](https://app.asana.com/0/0/1201901724848538/f)

This is a proposed general solution to variable text sizing for our v2 apps. Took the existing logic and put it in a custom hook. Definitely open to feedback on whether this is the solution we want going forward.

- [ ] Needs version bump?

### DIFF
--- a/assets/css/v2/pre_fare/elevator_status.scss
+++ b/assets/css/v2/pre_fare/elevator_status.scss
@@ -238,9 +238,29 @@
 
     .detail-page__description {
       font-family: Inter, sans-serif;
-      font-size: 24px;
-      line-height: 32px;
       padding-right: 32px;
+
+      &--extra-large {
+        font-size: 36px;
+        line-height: 48px;
+      }
+
+      &--large {
+        font-size: 36px;
+        line-height: 44px;
+      }
+
+      &--small {
+        font-size: 28px;
+        line-height: 40px;
+        letter-spacing: -0.4px;
+      }
+
+      &--extra-small {
+        font-size: 24px;
+        line-height: 32px;
+        letter-spacing: -0.4px;
+      }
     }
   }
 }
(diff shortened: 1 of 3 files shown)


jzimbel-mbta (commented):
Overall looks great to me, very straightforward.

PaulJKim (commented):
Looks good to me as well!

hannahpurcell (commented):
I'm excited for this!

Noticed that the sizing does not kick in on initial render, which leads to mis-sized font until the first page turn. I think that's fine. It's only confusing/cumbersome from a developer's perspective (the dev has to wait to see whether the styling is correct.) If there's any way to avoid that, great. If not, the solution is still worth it.

cmaddox5:
> I'm excited for this!
> 
> Noticed that the sizing does not kick in on initial render, which leads to mis-sized font until the first page turn. I think that's fine. It's only confusing/cumbersome from a developer's perspective (the dev has to wait to see whether the styling is correct.) If there's any way to avoid that, great. If not, the solution is still worth it.

@hannahpurcell Weird, I don't seem to be able to reproduce that issue. Can you remake the alert you were testing with and I'll check it out?

hannahpurcell:
> Weird, I don't seem to be able to reproduce that issue. Can you remake the alert you were testing with and I'll check it out?

Just using the current prod alerts, set to Haymarket. Issue only happens if the detail view is the first page that renders. Also, on refresh, I didn't see it appear again... so not totally sure why this is happening. Maybe it's pretty infrequent.

Screen Shot 2022-03-30 at 5 05 18 PM

hannahpurcell (approved):
The first-render thing will be less of a dev concern than I thought, and it won't impact riders at all. Looks good!

PR #1266, opened 2022-03-25
fix: Add sorter for Enum functions with DateTimes

cmaddox5:
**Asana task**: [Check usages of `Enum.sort` for cases where we're incorrectly sorting a list of structs](https://app.asana.com/0/1185117109217413/1201894802937869/f)

Tested all of these after adding the sorter. The new Util function was added because some objects were strings in ISO8601 format. Those need to be converted before the sorter can work.

- [ ] Needs version bump?

### DIFF
--- a/lib/screens/audio.ex
+++ b/lib/screens/audio.ex
@@ -71,7 +71,7 @@ defmodule Screens.Audio do
   defp merge_section_departures(sections) do
     sections
     |> Enum.flat_map(& &1.departures)
-    |> Enum.sort_by(& &1.time)
+    |> Enum.sort_by(& &1.time, DateTime)
     |> Util.group_by_with_order(&{&1.route, &1.route_id, &1.destination})
     |> Enum.map(fn {key, departures} ->
       {key,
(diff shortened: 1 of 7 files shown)


hannahpurcell (commented):
Great!  I think all the changes are great.  Did just a quick look over the repo, and there are a couple more `Enum.sort`s that I think might need a sorter?  Those are: 

special_cases.ex line 35 -- because routes is probably a list, so I'm not sure how Enum.sort would sort a list of alerts by a list of routes?
subway_status.ex line 551 -- because informed_entities is an array of structs

What do you think?

cmaddox5:
> Great! I think all the changes are great. Did just a quick look over the repo, and there are a couple more `Enum.sort`s that I think might need a sorter? Those are:
> 
> special_cases.ex line 35 -- because routes is probably a list, so I'm not sure how Enum.sort would sort a list of alerts by a list of routes? subway_status.ex line 551 -- because informed_entities is an array of structs
> 
> What do you think?

I think those two cases are fine. In `special_cases`, the list is transformed into a map with `routes` being the key. And the function pattern matches the routes to only contain `"Green-B", "Green-C", "Green-D"`. In `subway_status`, the `informed_entities` has been mapped and filtered down to just a list of GL routes. In both cases, an alphabetical search will work.

hannahpurcell (approved):
Thanks for the chat in slack! Looks good

PR #1267, opened 2022-03-25
fix: Use `makePersistentCarousel` to have line map widget update in sync with rest of the page

jzimbel-mbta:
**Asana task**: [[Pre-Fare] Line map needs to render in-sync](https://app.asana.com/0/1185117109217413/1201995187651695/f)

This uses the `makePersistentCarousel` higher-order component to sync the FullLineMap widget component's updates with the rest of the components that update based on the `lastSuccess` timestamp.

- [ ] Needs version bump?

### DIFF
--- a/assets/src/components/v2/full_line_map.tsx
+++ b/assets/src/components/v2/full_line_map.tsx
@@ -1,34 +1,13 @@
-import React, { ComponentType, useState, useEffect } from "react";
+import makePersistentCarousel, { PageRendererProps } from "Components/v2/persistent_carousel";
+import React, { ComponentType } from "react";
 
-interface Props {
-  asset_urls: string[];
+interface Page {
+  asset_url: string;
 }
 
-const intervalInMs = 10000;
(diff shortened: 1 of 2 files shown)


PaulJKim (approved):
Overall, LGTM! So does this mean that the list of Pages is stored at a level above `full_line_map.tsx` and this component now represents one "page" on the carousel?

jzimbel-mbta:
> Overall, LGTM! So does this mean that the list of Pages is stored at a level above `full_line_map.tsx` and this component now represents one "page" on the carousel?

Yup! The `makePersistentCarousel` HOC abstracts away all of the following logic for you:
- receiving an array of data from a data request
- holding onto that array until we've paged through all of its elements
- rendering one element of the array per data refresh

So all you need to do is provide a "presenter" component that is able to render a single page of data.

PaulJKim:
Thank you for the explanation! Makes sense to me 👍

PR #1272, opened 2022-03-30
tweak: Change audio_sort_key return value

cmaddox5:
**Asana task**: ad-hoc

Engineers decided audio sorting logic should return a list of integers, similar to visual priority. This PR changes the return values to a list of whatever integer the sort key was before.

- [ ] Needs version bump?


hannahpurcell (approved):
Thanks!

PR #1275, opened 2022-03-31
refactor: Improve parent station fetch performance

cmaddox5:
**Asana task**: [[PreFare] Improve performance of get_parent_station_id](https://app.asana.com/0/1185117109217413/1201950331696435/f)

I noticed that we already get the data we need when fetching `stop_sequences`. I took that result and made a map of platform IDs -> parent station name. We can then do a lookup and avoid individual V3 calls.

- [ ] Needs version bump?

### DIFF
--- a/lib/screens/route_patterns/route_pattern.ex
+++ b/lib/screens/route_patterns/route_pattern.ex
@@ -55,21 +55,65 @@ defmodule Screens.RoutePatterns.RoutePattern do
 
     case get_json_fn.("route_patterns", params) do
       {:ok, result} ->
-        stop_sequences =
-          get_in(result, [
-            "included",
-            Access.filter(&(&1["type"] == "trip")),
-            "relationships",
-            "stops",
-            "data",
-            Access.all(),
-            "id"
-          ])
-
-        {:ok, stop_sequences}
+        {:ok, get_stop_sequences_from_result(result)}
 
       _ ->
         :error
     end
+  end
+
+  @doc """
+  Gets the list of stop sequences for stop and creates a map of platform IDs to parent station name.
+  Assumes that all stop sequences in result are platforms.
+  """
+  @spec fetch_parent_station_sequences_through_stop(Stop.id(), list(String.t())) ::
+          {:ok, list(list(Stop.id())), map()} | :error
+  def fetch_parent_station_sequences_through_stop(
+        stop_id,
+        route_filters,
+        get_json_fn \\ &V3Api.get_json/2
+      ) do
+    params = %{
+      "include" => "representative_trip.stops,route",
+      "filter[stop]" => stop_id,
+      "filter[route]" => Enum.join(route_filters, ",")
+    }
+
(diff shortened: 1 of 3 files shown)


lib/screens/route_patterns/route_pattern.ex
```diff
@@ -55,21 +55,66 @@ defmodule Screens.RoutePatterns.RoutePattern do
...
+          {:ok, list(list(Stop.id())), map()} | :error
+  def fetch_parent_station_sequences_through_stop(
+        stop_id,
+        route_filters,
+        get_json_fn \\ &V3Api.get_json/2
+      ) do
+    params = %{
+      "include" => "representative_trip.stops,route",
+      "filter[stop]" => stop_id,
+      "filter[route]" => Enum.join(route_filters, ",")
```
hannahpurcell, line 79: Previously, route_filters was a conditional parameter, but here it's required. Which is fine, I think? Because all the cases where we want to use it is for subway. In that case, does it need to be a parameter at all? We could set it right here in this function to :light_rail and :subway.

lib/screens/v2/candidate_generator/widgets/reconstructed_alert.ex
```diff
@@ -20,21 +20,21 @@ defmodule Screens.V2.CandidateGenerator.Widgets.ReconstructedAlert do
...
-        get_parent_station_id_fn \\ &get_parent_station/1
+        fetch_stop_sequences_by_stop_fn \\ &RoutePattern.fetch_parent_station_sequences_through_stop/2,
+        fetch_alerts_fn \\ &Alert.fetch/1
       ) do
     # Filtering by subway and light_rail types
     with {:ok, routes_at_stop} <- fetch_routes_by_stop_fn.(stop_id, now, [0, 1]),
          route_ids_at_stop = Enum.map(routes_at_stop, & &1.route_id),
          {:ok, alerts} <- fetch_alerts_fn.(route_ids: route_ids_at_stop),
-         {:ok, stop_sequences} <- fetch_stop_sequences_by_stop_fn.(stop_id, route_ids_at_stop) do
+         {:ok, stop_sequences, platform_to_station_map} <-
```
hannahpurcell, line 30: Can the data processing that happens on lines 32-40 be moved to `fetch_parent_station_sequences_through_stop` so that this function only returns the final iteration of `stop_sequences`?

hannahpurcell (commented):
One question and one recommendation

lib/screens/route_patterns/route_pattern.ex
cmaddox5: I thought about this too, and I decided to leave it because the filters passed in here are more specific than `:light_rail` and `:subway`. It's actually the routes that service the station. So to keep the V3 query as filtered as possible, I think it's best we keep it as a parameter.

hannahpurcell (approved):
Looks good to me. Thanks for this! Relief to get rid of that parent station look-up.

PR #1280, opened 2022-04-05
doc: Audio-only widgets tech spec

jzimbel-mbta:
**Asana task**: [Audio investigation: Audio-only widgets](https://app.asana.com/0/1185117109217413/1202049932203142/f)

Trying out tech specs in the project repo. Share your thoughts!

I made the basic changes necessary to accomplish this, just to give a better idea of what it looks like in code.
I can also add an example global audio widget, if that would help!

To view the markdown in rendered form, click this icon at the top right of the file view:
Screen Shot 2022-04-05 at 8 22 41 AM


jzimbel-mbta (commented):
Need to rethink how `audio_sort_key` works.

Can we define a separate protocol for audio-only widgets? Not great that we're forced to define stubbed funcs right now.

Does `widgets_snapshot` always have to get the full list? Should the candidate generator get just the data we need? (This would be analogous to the candidate generator for normal widgets fetching just the V3 API data needed for a given widget.)

Q: Can we do API queries in `insert_global_audio_instances`?
A: Yes, that's fine. Even if the data already exists in some normal widget's data, it's ok to just fetch what we need again.

Look into whether we can embed this documentation in a Notion doc, for wider team visibility.

jzimbel-mbta:
Also, add a `TEMPLATE.md` for future tech specs!

paulswartz:
something else to consider for documentation like this: Architcture Decision Records (ADRs)! Some examples from the Data Platform and Skate:
- https://github.com/mbta/data_platform/commit/f88ece21f85d41a6977b956fb0cc6aa09230fa3a
- https://github.com/mbta/data_platform/commit/fc6ca5f6773411eb98db0a15f3223662da1c09d9
- https://github.com/mbta/skate/commit/7fa7a9b0835b5a5d7d31986316f1396ad28d5803

docs/tech_specs/TEMPLATE.md
```diff
@@ -0,0 +1,26 @@
+# Template instructions
```
cmaddox5, line 1: A template is a great idea!

docs/tech_specs/audio_only_widgets.md
```diff
@@ -0,0 +1,69 @@
...
+
+### ScreenAudioData
+In `Screens.V2.ScreenAudioData`, we'll add a call to `candidate_generator.audio_only_instances/2` into this pipeline:
+```diff
+  config
+  |> fetch_data_fn.()
+  |> elem(1)
+  |> Map.values()
+  |> Enum.filter(&WidgetInstance.audio_valid_candidate?/1)
++ |> then(fn widgets_with_audio_implementations ->
```
cmaddox5, line 49: That Elixir upgrade is coming in handy already!

cmaddox5 (approved):
Awesome work!

PR #1283, opened 2022-04-06
fix: Elevator Status connecting lines logic

cmaddox5:
**Asana task**: [[Bug] Elevator Status is not showing active alerts on connecting lines ](https://app.asana.com/0/1185117109217422/1202081351164230/f)

`informed_entities` does not always contain the platforms we expected. I changed `stop_sequences` to be a list of parent station ids. This is an ID that is always in a closure alert. Using that, we can reliably tell what alerts are on a connecting line.

- [ ] Needs version bump?

### DIFF
--- a/lib/screens/v2/widget_instance/elevator_status.ex
+++ b/lib/screens/v2/widget_instance/elevator_status.ex
@@ -212,18 +212,11 @@ defmodule Screens.V2.WidgetInstance.ElevatorStatus do
     # This assumes platform-level stop IDs are always numeric strings, e.g. "70045"
     informed_platforms =
       for %{stop: stop} when is_binary(stop) <- entities,
-          match?({_n, ""}, Integer.parse(stop)),
+          match?("place-" <> _, stop),
           do: stop
 
     connecting_platform_ids =
-      stop_sequences
-      |> List.flatten()
-      |> MapSet.new()
-      |> MapSet.difference(
-        t
-        |> platform_stop_ids()
-        |> MapSet.new()
-      )
+      stop_sequences |> List.flatten() |> List.delete(parent_station_id(t))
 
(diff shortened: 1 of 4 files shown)


lib/screens/v2/widget_instance/elevator_status.ex
```diff
@@ -212,18 +212,11 @@ defmodule Screens.V2.WidgetInstance.ElevatorStatus do
...
     connecting_platform_ids =
-      stop_sequences
-      |> List.flatten()
-      |> MapSet.new()
-      |> MapSet.difference(
-        t
-        |> platform_stop_ids()
-        |> MapSet.new()
-      )
+      stop_sequences |> List.flatten() |> List.delete(parent_station_id(t))
```
PaulJKim, line 219: Is this removing the `parent_station_id` since it's redundant to say that the current platform connects with itself?

lib/screens/v2/widget_instance/elevator_status.ex
```diff
@@ -212,18 +212,11 @@ defmodule Screens.V2.WidgetInstance.ElevatorStatus do
     # This assumes platform-level stop IDs are always numeric strings, e.g. "70045"
```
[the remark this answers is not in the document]
cmaddox5: Good catch. Got that removed.

lib/screens/v2/widget_instance/elevator_status.ex
cmaddox5: That's correct. And we don't want the parent station closures to show up in the section for connecting lines.

PaulJKim: gotcha, that makes sense

PaulJKim (approved):
LGTM!

PR #1284, opened 2022-04-06
fix: Elevator Status Silver Line icon

cmaddox5:
**Asana task**: [[Pre-fare Elevator Polish] Silver Line is showing up as a bus route](https://app.asana.com/0/1185117109217413/1202019168003511/f)

We don't have a ton to go on in the data. This is one option. Another is looking at the `color` that is returned for routes in V3. Open to suggestions/opinions.

- [ ] Needs version bump?


jzimbel-mbta (approved):
Seems like the most reasonable approach to me! ✅

PR #1285, opened 2022-04-06
fix: PreFare header text size

cmaddox5:
**Asana task**: [[Bug] Pre-fare screen header text not auto-resizing](https://app.asana.com/0/1185117109217413/1201945334732260/f)

Header was actually working. The wrong `NormalHeader` component was being rendered for Pre-Fare so the `maxHeight` was greater than it should have been. I went ahead and change the header to use the text resizer hook.

Screen Shot 2022-04-06 at 1 25 29 PM

- [ ] Needs version bump?

### DIFF
--- a/assets/css/v2/pre_fare/normal_header.scss
+++ b/assets/css/v2/pre_fare/normal_header.scss
@@ -36,7 +36,7 @@
 
   &--small {
     font-size: 72px;
-    line-height: 80px;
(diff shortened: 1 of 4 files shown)


hannahpurcell (approved):
Future tech debt task: use the LCD header, but pass a parameter for maxHeight. Because otherwise, these components are nearly identical

PR #1290, opened 2022-04-07
feat: Pre-Fare NoData state

cmaddox5:
**Asana task**: [[PreFare] no_data state](https://app.asana.com/0/1185117109217413/1202049932203148/f)

I wrapped the LCD NoData in a PreFare specific one to allow both screens to show the same page. Is there a better way to do this? I don't think so but I'm open to suggestions.

- [ ] Needs version bump?


assets/src/components/v2/pre_fare/no_data.tsx
```diff
@@ -0,0 +1,23 @@
...
+  show_alternatives: boolean;
+}
+
+const NoData: ComponentType<Props> = ({
+  show_alternatives: showAlternatives,
+}) => {
+  return (
+    <>
+      <div className="no-data-left">
+        <LcdNoData show_alternatives={showAlternatives} />{" "}
```
jzimbel-mbta, line 14: Is the rendered space character (`{" "}`) necessary for some reason? Here and below.

assets/src/components/v2/pre_fare/no_data.tsx
```diff
@@ -0,0 +1,23 @@
+import React, { ComponentType } from "react";
+import { default as LcdNoData } from "Components/v2/lcd/no_data";
```
jzimbel-mbta, line 2: When importing a module's default export, you can name it whatever you want.
```suggestion
import LcdNoData from "Components/v2/lcd/no_data";
```

jzimbel-mbta (approved):
I think this is the cleanest way to do it. Just some minor comments. Looks good to me once you address them!

assets/src/components/v2/pre_fare/no_data.tsx
cmaddox5: Must have been autocomplete or something. Removed those.

assets/src/components/v2/pre_fare/no_data.tsx
cmaddox5: Not sure why I blanked on this. In my mind, the name of the import had to match the export 😬 Got that fixed.

PR #1292, opened 2022-04-07
fix: Filter audio-only widgets by `audio_valid_candidate?`

jzimbel-mbta:
**Asana task**: ad hoc

I made a small oversight in the original PR and didn't filter audio-only instances by their `audio_valid_candidate?/1` return value. This fixes that.

I also took the opportunity to make the code hopefully more readable.

- [ ] Needs version bump?


hannahpurcell (approved):
Ooo yeah, that is more readable. Thanks! (Also, ew to the nails on chalkboard widget. Thanks for making your filler data fun-to-read)

PR #1294, opened 2022-04-07
feat: Alerts intro audio-only widget

jzimbel-mbta:
**Asana task**: [Audio implementation: Audio-only widgets](https://app.asana.com/0/1185117109217413/1202035304691540/f) (Part 2 of a 3-part trilogy)

This defines the alerts intro widget.

- [ ] Needs version bump?

### DIFF
--- a/lib/screens/v2/widget_instance/audio_only/alerts_intro.ex
+++ b/lib/screens/v2/widget_instance/audio_only/alerts_intro.ex
@@ -1,0 +1,110 @@
+defmodule Screens.V2.WidgetInstance.AudioOnly.AlertsIntro do
+  @moduledoc """
+  An audio-only widget that introduces the section of the readout describing service alerts.
+  """
+
+  alias Screens.Config.Screen
+  alias Screens.V2.WidgetInstance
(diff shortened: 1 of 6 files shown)


lib/screens/v2/widget_instance/audio_only/alerts_intro.ex
```diff
@@ -0,0 +1,110 @@
...
+    no_takeover_content? and at_least_one_service_alert_widget?
+  end
+
+  def audio_valid_candidate?(_t, _slot_names_fn) do
+    false
+  end
+
+  def audio_view(_instance), do: ScreensWeb.V2.Audio.AlertsIntroView
+
+  defp service_alert_widget?(%Alert{}), do: true
```
hannahpurcell, line 90: We don't use `Alert` on PreFare, right?
```suggestion
```

hannahpurcell (approved):
Fantastic work, thank you!

lib/screens/v2/widget_instance/audio_only/alerts_intro.ex
jzimbel-mbta: Oops ok! I wasn't sure if we used `Alert` for the PIO alert fallback cases so I included a case for it just to be safe.

PR #1295, opened 2022-04-07
Alert headsigns

hannahpurcell:
**Asana task**:  Adhoc

Added Govt Center and Maverick to headsign list. Use that for reconstructed alerts on boundaries

- [ ] Needs version bump?

### DIFF
--- a/lib/screens/v2/widget_instance/reconstructed_alert.ex
+++ b/lib/screens/v2/widget_instance/reconstructed_alert.ex
@@ -140,7 +140,7 @@ defmodule Screens.V2.WidgetInstance.ReconstructedAlert do
 
     %{
       issue: FreeTextLine.to_json(issue),
-      remedy: "Please seek an alternate route",
+      remedy: "Seek alternate route",
       location: location_text,
       cause: cause_text,
       routes: get_route_pills(affected_routes),
@@ -199,7 +199,7 @@ defmodule Screens.V2.WidgetInstance.ReconstructedAlert do
 
     %{
       issue: "Station Closed",
-      remedy: "Please seek an alternate route",
+      remedy: "Seek alternate route",
       location: "",
       cause: cause_text,
       routes: get_route_pills(affected_routes),
@@ -222,7 +222,7 @@ defmodule Screens.V2.WidgetInstance.ReconstructedAlert do
 
     %{
       issue: "No trains",
-      remedy: "Please seek an alternate route",
+      remedy: "Seek alternate route",
       location: "",
       cause: cause_text,
       routes: get_route_pills(affected_routes),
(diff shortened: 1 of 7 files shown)


lib/screens/v2/widget_instance/reconstructed_alert.ex
```diff
@@ -90,14 +94,32 @@ defmodule Screens.V2.WidgetInstance.ReconstructedAlert do
...
+      informed_stop_ids = Enum.into(informed_entities, MapSet.new(), & &1.stop)
+
+      :screens
+      |> Application.get_env(:reconstructed_alert_headsign_matchers)
+      |> Map.get(stop_id)
+      |> Enum.find_value(nil, fn {informed, not_informed, headsign} ->
+        if alert_region_match?(to_set(informed), to_set(not_informed), informed_stop_ids),
+          do: headsign,
+          else: false
+      end)
```
cmaddox5, line 106: Something to consider is pulling this out into some common file for this widget and DUPs. The code looks the same so would be less overhead to have it in one place. Up to you though.

lib/screens/v2/widget_instance/reconstructed_alert.ex
```diff
@@ -90,14 +94,32 @@ defmodule Screens.V2.WidgetInstance.ReconstructedAlert do
...
     else
       @route_directions
       |> Map.get(route_id)
       |> Enum.at(direction_id)
     end
   end
 
+  defp to_set(stop_id) when is_binary(stop_id), do: MapSet.new([stop_id])
+  defp to_set(stop_ids) when is_list(stop_ids), do: MapSet.new(stop_ids)
+  defp to_set(%MapSet{} = already_a_set), do: already_a_set
```
cmaddox5, line 116: I don't think this case is necessary. Passing in a MapSet to `MapSet.new` will return a MapSet. You should be able to combine this and the clause above it by removing the guard.

```suggestion
  defp to_set(stop_ids), do: MapSet.new(stop_ids)
```

lib/screens_web/views/v2/audio/reconstructed_alert_view.ex
```diff
@@ -109,4 +109,9 @@ defmodule ScreensWeb.V2.Audio.ReconstructedAlertView do
       _ -> ""
     end
   end
+
+  defp adjust_remedy(remedy) when remedy === "Seek alternate route",
+    do: "Please seek an alternate route"
```
cmaddox5, line 114: Nothing wrong with this, but can be done without a guard.

```suggestion
  defp adjust_remedy("Seek alternate route"), do: "Please seek an alternate route"
```

cmaddox5 (approved):
Looks good as is, but left some suggestions. Your call on implementing them.

lib/screens/v2/widget_instance/reconstructed_alert.ex
hannahpurcell: Nice!

PR #1297, opened 2022-04-08
fix: Reconstructed Alert audio bug for whole GL alerts

cmaddox5:
**Asana task**: [Government Center audio error](https://app.asana.com/0/1185117109217422/1202100266982551/f)

GL alerts that affect the whole line no longer have a key for branches. Tweaked some filtering to look for key instead of checking the length of a value that doesn't exist.

- [ ] Needs version bump?


hannahpurcell (approved):
noice, thanks

PR #1298, opened 2022-04-08
fix: local admin tool cognito bug

cmaddox5:
**Asana task**: [Visiting screens admin in local development unexpectedly checks cognito](https://app.asana.com/0/1185117109217413/1201754217963829/f)

Needed to set the fake strategy to ignore CSRF.

- [ ] Needs version bump?


hannahpurcell (approved):
Yayyy!!! So happy to have this. So because the changes were only applied to `fake.ex`, this is only applied in the local env?

cmaddox5:
> Yayyy!!! So happy to have this. So because the changes were only applied to `fake.ex`, this is only applied in the local env?

Yep! We only use the fake strategy in `dev.exs`.

PR #1303, opened 2022-04-08
When fetching station sequences, get them in same direction

hannahpurcell:
**Asana task**: [Subway Status and Alert widgets don't handle Green Line suspension Lechmere - Government Center](https://app.asana.com/0/1185117109217422/1202101924522475)

- [ ] Needs version bump?

### DIFF
--- a/lib/screens/route_patterns/route_pattern.ex
+++ b/lib/screens/route_patterns/route_pattern.ex
@@ -76,6 +76,7 @@ defmodule Screens.RoutePatterns.RoutePattern do
     params = %{
       "include" => "representative_trip.stops,route",
       "filter[stop]" => stop_id,
+      "filter[direction_id]" => 0,
       "filter[route]" => Enum.join(route_filters, ",")
     }
 


lib/screens/route_patterns/route_pattern.ex
```diff
@@ -76,6 +76,7 @@ defmodule Screens.RoutePatterns.RoutePattern do
     params = %{
       "include" => "representative_trip.stops,route",
       "filter[stop]" => stop_id,
+      "filter[direction_id]" => 0,
```
PaulJKim, line 79: Is it true that we will always want the stop sequence in the same direction?

hannahpurcell: Yes, ha. 

We only need an ordered sequence of stops. Doesn't matter which way they are, EB or WB. In the case of the GL, they all need to go in the same direction as each other. So setting 0 or 1 would fix this

PaulJKim (approved):

PR #1304, opened 2022-04-08
fix: FreeTextLine audio readout bug

cmaddox5:
**Asana task**: [Error on fullscreen audio for OL fullscreen alert](https://app.asana.com/0/1185117109217422/1202102359681100/f)

- [ ] Needs version bump?


hannahpurcell (approved):
Aha! This was my bad. I renamed the attr after you created the audio

PR #1309, opened 2022-04-12
fix: Only create reconstructed alert widgets for active service alerts

jzimbel-mbta:
**Asana task**: ad hoc

This adds an additional check to the `relevant?` predicate to filter out alerts that aren't currently active.


hannahpurcell (approved):
Thanks! Quick and organized fix.

PR #1311, opened 2022-04-13
POC: Paging animations

cmaddox5:
**Asana task**: [Investigate paging transition implementation](https://app.asana.com/0/1185117109217413/1202036010364183/f)

Mostly creating this to get eyes on it and start a discussion.

https://user-images.githubusercontent.com/10713153/163214132-5c93f034-2920-4765-b375-8eaee7c0e016.mov

A couple of notes:
- I tried a bunch of different ways to centralize this code. The most logical place for this to live is either in it's own `Widget`-like component or in the `Widget` component itself. It does not animate consistently there. Sometimes one page would animate and not the others, sometimes no page would animate. 
- For the CSS class names, the four CSS classes must be present and start with the `classNames` property passed into `<CSSTransition>`.
- I moved the `<FlexZonePageIndicator>` out of the flex components so they do not animate. 

The core of this change is what I think we want, with design requirements being applied to the CSS classes.  

- [ ] Needs version bump?

### DIFF
--- a/assets/css/v2/bus_shelter/body/normal.scss
+++ b/assets/css/v2/bus_shelter/body/normal.scss
@@ -31,3 +31,22 @@
   z-index: 1;
   overflow: hidden;
 }
+
+.slide-enter {
+  opacity: 0;
+  transform: translateX(100%);
+}
+.slide-enter-active {
+  opacity: 1;
+  transform: translateX(0);
+  transition: opacity 500ms, transform 500ms;
+}
+.slide-exit {
+  opacity: 1;
+  transform: translateX(0);
+}
+.slide-exit-active {
+  opacity: 0;
+  transform: translateX(-100%);
+  transition: opacity 500ms, transform 500ms;
+}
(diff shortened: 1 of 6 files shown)


hannahpurcell (commented):
Oh neat! This was a lighter lift than I expected. I think the CSS looks good (but it will be up to the design team to define requirements). And the use of TransitionGroup basically takes care of the work! Cool.

What should come next? I think this proof-of-concept is good, but implementation will depend on a design/product decision.

cmaddox5:
@hannahpurcell I think we have done all we can for now. This PR might be good to keep around when design has decided how it should be animated. We can then tweak this instead of trying to do it from scratch again.

hannahpurcell:
@cmaddox5  Sounds good. Ping other teammates if you'd like other eyes on it at the moment. If not, you can close the Asana task. I'd also recommend closing this PR, because we'll be able to find it again and re-open when it's time to think about it again.

PR #1312, opened 2022-04-14
Marathon alerts in reconstructed alert widget

hannahpurcell:
**Asana task**: [Handle Multi-Branch Station Closure on Pre-Fare](https://app.asana.com/0/1185117109217413/1202112456216984)

`get_stations()` is only used by this one serializer, so I made it into a very specific function -- handling the string formatting and everything. It's a small enough function that I didn't really want to split it up, but lemme know if it should be.

- [ ] Needs version bump?


lib/screens/v2/widget_instance/reconstructed_alert.ex
```diff
@@ -647,6 +645,9 @@ defmodule Screens.V2.WidgetInstance.ReconstructedAlert do
     informed_entities
     |> Enum.map(fn %{stop: stop_id} -> stop_id end)
     |> Enum.filter(&String.starts_with?(&1, "place-"))
+    |> Enum.uniq()
+    |> Enum.map(&Stop.fetch_stop_name(&1))
```
jzimbel-mbta, line 649: Noted in slack: You'll need to pass a stubbed function down into this function in order to prevent it from making API requests while being unit tested. (Also, need unit tests for this change)

lib/screens/v2/widget_instance/reconstructed_alert.ex
```diff
@@ -613,10 +614,7 @@ defmodule Screens.V2.WidgetInstance.ReconstructedAlert do
     affected_routes = get_affected_routes(informed_entities)
     cause_text = get_cause_text(cause)
 
-    station =
-      case get_stations(informed_entities) do
-        [station | _] -> Stop.fetch_stop_name(station)
-      end
+    station = get_stations(informed_entities)
```
jzimbel-mbta, line 617: Based on its name, `get_stations` returning a single string is a little confusing right now—maybe still have it return a list, rename the variable it's bound to to `bypassed_stations`, and call `Util.format_name_list_to_string` down below?

So—
```ex
bypassed_stations = get_stations(informed_entities)

%{
  issue: "Trains will bypass #{Util.format_name_list_to_string(bypassed_stations)}",
  ...
}
```

jzimbel-mbta (approved):
Approved assuming you test this in a followup PR, and consider the suggested change!

lib/screens/v2/candidate_generator/widgets/reconstructed_alert.ex
```diff
@@ -74,4 +78,18 @@ defmodule Screens.V2.CandidateGenerator.Widgets.ReconstructedAlert do
...
   end
+
+  defp get_stations(alert) do
+    %{alert: alert}
+    |> BaseAlert.informed_entities()
+    |> Enum.map(fn %{stop: stop_id} -> stop_id end)
+    |> Enum.filter(&String.starts_with?(&1, "place-"))
+    |> Enum.uniq()
+    |> Enum.map(
+      case &Stop.fetch_stop_name(&1) do
```
jzimbel-mbta, line 89: You'll need to move the `&` to the front of `case` for this to work as expected!
```suggestion
      &case Stop.fetch_stop_name(&1) do
```

Currently the way this will evaluate is:
- `case` gets passed the anonymous function `&Stop.fetch_stop_name(&1)`, which will always match `name -> name`.
- The whole case expression then evaluates to `&Stop.fetch_stop_name(&1)`, which then gets passed as the callback to `Enum.map`. `fetch_stop_name` runs on each element of the list without any error handling.

This will still ~kind of work, (until there's an API error) but the `case` expression isn't actually doing any error handling since it gets passed the function, not the result of the function's evaluation.

test/screens/v2/candidate_generator/widgets/reconstructed_alert_test.exs
```diff
@@ -213,7 +226,8 @@ defmodule Screens.V2.CandidateGenerator.Widgets.ReconstructedAlertTest do
                  now,
                  fetch_routes_by_stop_fn,
                  fetch_parent_station_sequences_through_stop_fn,
-                 x_fetch_alerts_fn
+                 x_fetch_alerts_fn,
+                 get_stations_fn
```
jzimbel-mbta, line 230: Looks like one additional assertion is needed  after this one, now that there's another query that can fail:
```ex
assert [] ==
         reconstructed_alert_instances(
           config,
           now,
           fetch_routes_by_stop_fn,
           fetch_parent_station_sequences_through_stop_fn,
           fetch_alerts_fn,
           x_get_stations_fn
         )
```

jzimbel-mbta (commented):
Two minor things and a missing test assertion for the new query! Otherwise looks good. ✅

lib/screens/v2/candidate_generator/widgets/reconstructed_alert.ex
```diff
@@ -74,4 +78,18 @@ defmodule Screens.V2.CandidateGenerator.Widgets.ReconstructedAlert do
...
+        :error -> ""
```
jzimbel-mbta, line 90: You might want to use `Enum.flat_map` instead of `Enum.map`, and have the callback function return an empty list in case of `:error`.

Currently, if any `Stop.fetch_stop_name` call fails, you'll get a list that looks like `["Stop Name 1", "", "Stop Name 3"]`. When that gets passed to `Util.format_name_list_to_string`, it will produce `"Stop Name 1, , and Stop Name 3"`.

If you use `Enum.flat_map`, this error-handling clause can return an empty list, which will remove the element completely. You'll get `["Stop Name 1", "Stop Name 3"]` which formats to `"Stop Name 1 and Stop Name 3"`.

test/screens/v2/candidate_generator/widgets/reconstructed_alert_test.exs
hannahpurcell: I don't think the whole thing should fail if this query fails, though. I could make sure it fails gracefully though

lib/screens/v2/candidate_generator/widgets/reconstructed_alert.ex
```diff
@@ -74,4 +78,19 @@ defmodule Screens.V2.CandidateGenerator.Widgets.ReconstructedAlert do
...
+    |> BaseAlert.informed_entities()
+    |> Enum.map(fn %{stop: stop_id} -> stop_id end)
+    |> Enum.filter(&String.starts_with?(&1, "place-"))
+    |> Enum.uniq()
+    |> Enum.map(
+      &case fetch_stop_name_fn.(&1) do
+        :error -> []
+        name -> name
+      end
+    )
```
jzimbel-mbta, line 93: Oop, in order to have stops whose names we failed to fetch disappear completely from the list this would need to be
```suggestion
    |> Enum.flat_map(
      &case fetch_stop_name_fn.(&1) do
        :error -> []
        name -> [name]
      end
    )
```

Some examples [here in the docs](https://hexdocs.pm/elixir/Enum.html#flat_map/2).

hannahpurcell: Ope, I applied your change, and then merged from master and lost the flat_map. I didn't write line 91 correctly though. Thanks for catching both of these!

config/dev.exs
```diff
@@ -22,7 +22,7 @@ config :screens, ScreensWeb.Endpoint,
   ]
 
 config :screens,
-  default_api_v3_url: System.get_env("API_V3_URL", "https://api-v3.mbta.com/"),
+  default_api_v3_url: System.get_env("API_V3_URL", "https://api-dev-green.mbtace.com/"),
```
hannahpurcell, line 25: No! Bad commit!
```suggestion
  default_api_v3_url: System.get_env("API_V3_URL", "https://api-v3.mbta.com/"),
```

PR #1318, opened 2022-04-19
fix: DeparturesNoData server error

cmaddox5:
**Asana task**: [[Bug] DeparturesNoData throwing error for Bus Shelter screens](https://app.asana.com/0/1185117109217413/1202123170003797/f)

The DeparturesNoData widget was looking for `stop_id` in the header but we do not always put it there. In each V2 config, we place a `stop_id` in the `alerts` section, so I changed it to look there instead.

This solution does fix backend issues for eInk apps, but this widget is still not registered to those components in the front-end. That will be fixed in [this task](https://app.asana.com/0/1185117109217422/1202131702864644/f).

- [ ] Needs version bump?

### DIFF
--- a/lib/screens/v2/widget_instance/departures_no_data.ex
+++ b/lib/screens/v2/widget_instance/departures_no_data.ex
@@ -2,8 +2,7 @@ defmodule Screens.V2.WidgetInstance.DeparturesNoData do
   @moduledoc false
 
   alias Screens.Config.Screen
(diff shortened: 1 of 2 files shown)


hannahpurcell (approved):
Agh, I see why the config is so confusing! Sometimes header has stop_id, sometimes stop_name. I'd feel better with a common context config, but I guess we're waiting until Screenplay for that...

PR #1319, opened 2022-04-19
tweak: Add screen_side to request logs

cmaddox5:
**Asana task**: [Log screen side (left vs right) for Pre-fare screens](https://app.asana.com/0/1185117109217413/1202116831580952/f)

Added `screen_side` to the map we log for requests. Will only exist if `screen_side` is a param in the request, so Pre-Fare will be the only screen that logs it at the moment.

- [ ] Needs version bump?

### DIFF
--- a/lib/screens_web/controllers/v2/screen_api_controller.ex
+++ b/lib/screens_web/controllers/v2/screen_api_controller.ex
@@ -16,7 +16,7 @@ defmodule ScreensWeb.V2.ScreenApiController do
     end
   end
 
-  def show(%{params: params} = conn, %{"id" => screen_id, "last_refresh" => last_refresh}) do
+  def show(conn, %{"id" => screen_id, "last_refresh" => last_refresh} = params) do
     is_screen = ScreensWeb.UserAgent.is_screen_conn?(conn, screen_id)
     screen_side = get_screen_side(params)
 
--- a/lib/screens_web/controllers/v2/screen_api_controller.ex
+++ b/lib/screens_web/controllers/v2/screen_api_controller.ex
@@ -16,10 +16,11 @@ defmodule ScreensWeb.V2.ScreenApiController do
     end
   end
 
-  def show(conn, %{"id" => screen_id, "last_refresh" => last_refresh}) do
+  def show(%{params: params} = conn, %{"id" => screen_id, "last_refresh" => last_refresh}) do
     is_screen = ScreensWeb.UserAgent.is_screen_conn?(conn, screen_id)
+    screen_side = get_screen_side(params)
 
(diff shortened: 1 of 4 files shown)


lib/screens_web/controllers/v2/screen_api_controller.ex
```diff
@@ -16,10 +16,11 @@ defmodule ScreensWeb.V2.ScreenApiController do
     end
   end
 
-  def show(conn, %{"id" => screen_id, "last_refresh" => last_refresh}) do
+  def show(%{params: params} = conn, %{"id" => screen_id, "last_refresh" => last_refresh}) do
```
jzimbel-mbta, line 19: The second argument to this (and any other controller action function) is actually the same as `conn.params`. Would recommend binding `params` there instead, just to follow conventions outlined in [the guide](https://hexdocs.pm/phoenix/controllers.html#actions).
```suggestion
  def show(conn, %{"id" => screen_id, "last_refresh" => last_refresh} = params) do
```

lib/screens_web/controllers/v2/screen_api_controller.ex
```diff
@@ -16,10 +16,11 @@ defmodule ScreensWeb.V2.ScreenApiController do
...
     is_screen = ScreensWeb.UserAgent.is_screen_conn?(conn, screen_id)
+    screen_side = get_screen_side(params)
```
jzimbel-mbta, line 21: A helper function is fine, but if you want you could also use the [`data[key]` syntax](https://hexdocs.pm/elixir/Access.html) or `Map.get`, which both return `nil` if the key doesn't exist.
```suggestion
    screen_side = params["screen_side"]
```
```suggestion
    screen_side = Map.get(params, "screen_side")
```

lib/screens_web/controllers/v2/screen_controller.ex
```diff
@@ -35,10 +35,11 @@ defmodule ScreensWeb.V2.ScreenController do
     put_layout(conn, {ScreensWeb.V2.LayoutView, "app.html"})
   end
 
-  def index(conn, %{"id" => screen_id}) do
+  def index(%{params: params} = conn, %{"id" => screen_id}) do
```
jzimbel-mbta, line 38: Same here, let's bind the second arg to `params`.
```suggestion
  def index(conn, %{"id" => screen_id} = params) do
```

jzimbel-mbta (commented):
Looks good! Just some minor things.

lib/screens_web/controllers/v2/screen_api_controller.ex
cmaddox5: Thanks! That is very good to know!

jzimbel-mbta (approved):

PR #1321, opened 2022-04-20
feat: Log successful API responses

cmaddox5:
**Asana task**: [Log successful API responses](https://app.asana.com/0/1185117109217413/1202116831580950/f)

v1 was already _kinda_ doing this, but was always logging responses with message `[screen api response success]`. I changed that to say the status of the response instead. 

v2 was not doing response logging at all so that was a straightforward addition to the controller logic.

- [ ] Needs version bump?

### DIFF
--- a/lib/screens/log_screen_data.ex
+++ b/lib/screens/log_screen_data.ex
@@ -55,8 +55,7 @@ defmodule Screens.LogScreenData do
       data = %{
         screen_id: screen_id,
         screen_name: screen_name_for_id(screen_id),
-        last_refresh: last_refresh,
-        status: status
+        last_refresh: last_refresh
       }
 
       log_message("[screen api response #{status}]", data)
--- a/lib/screens/log_screen_data.ex
+++ b/lib/screens/log_screen_data.ex
@@ -33,29 +33,34 @@ defmodule Screens.LogScreenData do
     end
   end
 
-  def log_api_response(%{force_reload: true}, screen_id, last_refresh, is_screen) do
-    log_api_response_success(screen_id, last_refresh, is_screen)
+  def log_api_response(%{force_reload: true, status: status}, screen_id, last_refresh, is_screen) do
+    log_api_response_success(screen_id, last_refresh, is_screen, status)
   end
 
-  def log_api_response(%{success: true}, screen_id, last_refresh, is_screen) do
-    log_api_response_success(screen_id, last_refresh, is_screen)
+  def log_api_response(%{success: true, status: status}, screen_id, last_refresh, is_screen) do
+    log_api_response_success(screen_id, last_refresh, is_screen, status)
+  end
+
+  def log_api_response(status, screen_id, last_refresh, is_screen)
+      when is_atom(status) do
+    log_api_response_success(screen_id, last_refresh, is_screen, status)
   end
 
   def log_api_response(_response, _screen_id, _last_refresh, _is_screen) do
     :ok
   end
 
-  defp log_api_response_success(screen_id, last_refresh, is_screen) do
-    _ =
-      if is_screen do
-        data = %{
-          screen_id: screen_id,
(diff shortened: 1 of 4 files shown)


lib/screens/log_screen_data.ex
```diff
@@ -33,29 +33,34 @@ defmodule Screens.LogScreenData do
...
+      data = %{
+        screen_id: screen_id,
+        screen_name: screen_name_for_id(screen_id),
+        last_refresh: last_refresh,
+        status: status
+      }
 
-        log_message("[screen api response success]", data)
-      end
+      log_message("[screen api response #{status}]", data)
```
PaulJKim, line 62: Do we need to have `status` in `data` if we're logging it in the message as well?

lib/screens/log_screen_data.ex
```diff
@@ -33,29 +33,34 @@ defmodule Screens.LogScreenData do
...
   end
 
-  def log_api_response(%{success: true}, screen_id, last_refresh, is_screen) do
-    log_api_response_success(screen_id, last_refresh, is_screen)
+  def log_api_response(%{success: true, status: status}, screen_id, last_refresh, is_screen) do
+    log_api_response_success(screen_id, last_refresh, is_screen, status)
+  end
+
+  def log_api_response(status, screen_id, last_refresh, is_screen)
+      when is_atom(status) do
```
PaulJKim, line 45: Is there a specific use-case that requires us to have this `is_atom/1` check in place?

lib/screens_web/controllers/screen_api_controller.ex
```diff
@@ -32,6 +32,8 @@ defmodule ScreensWeb.ScreenApiController do
     _ = Screens.LogScreenData.log_data_request(screen_id, last_refresh, is_screen)
 
     if nonexistent_screen?(screen_id) do
+      Screens.LogScreenData.log_api_response(:nonexistent, screen_id, last_refresh, is_screen)
```
PaulJKim, line 35: I see that one of the done criteria in the ticket is to only log for "real" screens. I'm making the assumptions that nonexistent == not real and that in this case, `is_screen` would be `false`. If those assumptions are true, would we be able to remove this call to `Screens.LogScreenData.log_api_response/4` here (and in v2) and also remove the `if is_screen do` check on line 54 of `lib/screens/log_screen_data.ex`?

cmaddox5: So `is_screen` is a request parameter that is passed to the server. That parameter is set for live screens in whatever CMS it is using. So this specific response could still happen for live screens. Will be uncommon but happens mostly when we take screens out of the config temporarily for big config changes (and even that happens mostly in dev and dev-green). 

And since it is a parameter that we won't ever set when working locally (unless we need to test logging), then we want to keep the conditional so we don't log while pulling up screens in the browser.

lib/screens/log_screen_data.ex
cmaddox5: Nope, that was something I needed in my first idea. Just forgot to delete it when I went another direction. Good catch.

lib/screens/log_screen_data.ex
cmaddox5: That's a good call. Having `status` in the message should be enough.

lib/screens_web/controllers/screen_api_controller.ex
PaulJKim: Gotcha, thanks for the explainer!

So does that imply that we would want to be logging responses for nonexistent screens while they're temporarily out of the config? And if that's correct, what's the reasoning behind that?

cmaddox5: I think the main goal of the response logging is to compare the number of responses to the number of requests. So if there is a request for the nonexistent screen, we will want this log just to show the server did respond.

PR #1323, opened 2022-04-21
Headsigns for new Pre-Fare screens

hannahpurcell:
**Asana task**: [Add stations to the destination lookup table as needed](https://app.asana.com/0/1185117109217413/1202101724194527)

Added headsigns to config of new Pre-Fare screens (headsigns are needed for reconstructed alerts)

Also fixed Mattapan error that only appears for Ashmont. (We need to filter out "Mattapan" from the route ids associated with Ashmont to prevent Mattapan alerts from being returned)


jzimbel-mbta (approved):
Nice work! Your added comments are helpful.

PR #1325, opened 2022-04-22
fix: screen_side logging

cmaddox5:
**Asana task**: ad-hoc

Didn't think about how `Map.put` returns the updated map and doesn't update map in place, so it wasn't actually doing anything. Also added `screen_side` to response logging.

- [ ] Needs version bump?


hannahpurcell (approved):
Aha, yep! Looks good

PR #1328, opened 2022-04-29
Set up Sentry for Elixir

hannahpurcell:
**Asana task**: [Set up Sentry for Elixir](https://app.asana.com/0/1185117109217413/1202116831580948)

https://sentry.io/mbtace/screens-elixir/getting-started/elixir/

Terraform PR: https://github.com/mbta/devops/pull/365


config/dev.exs
```diff
@@ -35,6 +35,11 @@ config :ueberauth, Ueberauth,
     cognito: {Screens.Ueberauth.Strategy.Fake, []}
   ]
 
+config :sentry,
```
cmaddox5, line 38: I believe that this config file is only used when `Mix.env()` returns `:dev` which should only happen when working locally. So this tag won't be used as long as local development is not connected to Sentry. Might be better to set `tags` in `config.exs` the same way you set `environment_name`.

cmaddox5 (changes requested):
Looks great! Just one question and one change needed.

config/config.exs
```diff
@@ -53,6 +53,25 @@ config :ueberauth, Ueberauth.Strategy.Cognito,
...
+  root_source_code_path: File.cwd!(),
+  included_environments: [:prod, :dev],
+  tags: %{
+    env:
+      case System.get_env("SENTRY_ENVIRONMENT") do
+        nil -> Mix.env()
+        env -> String.to_existing_atom(env)
+      end
+  }
+
```
cmaddox5, line 74: I believe variables are allowed in configuration files, so you could do:

```suggestion
env_name =
  case System.get_env("SENTRY_ENVIRONMENT") do
    nil -> Mix.env()
    env -> String.to_existing_atom(env)
  end

# Centralize Error reporting
config :sentry,
  dsn: System.get_env("SENTRY_DSN_BACKEND") || "",
  environment_name: env_name,
  enable_source_code_context: true,
  root_source_code_path: File.cwd!(),
  included_environments: [:prod, :dev],
  tags: %{
    env: env_name
  }

```

cmaddox5 (approved):
Looks great! One small suggestion, but otherwise good to go!

config/config.exs
hannahpurcell: Ah gooood

cmaddox5 (approved):
Looks great!

PR #1331, opened 2022-05-02
tweak: filter delay alerts in incoming direction for terminals

cmaddox5:
**Asana task**: [[Pre-fare] Don't show alerts in direction which doesn't have service at terminal stations](https://app.asana.com/0/1185117109217413/1202176930559888/f)

If the delay affects the line in the incoming direction of the terminal, we don't want to show it. Luckily, the order of `stop_sequences` is in order so we can use that order to determine if the station is a terminal.

- [ ] Needs version bump?


lib/screens/v2/candidate_generator/widgets/reconstructed_alert.ex
```diff
@@ -79,22 +79,69 @@ defmodule Screens.V2.CandidateGenerator.Widgets.ReconstructedAlert do
...
+
+  # This function assumes that stop_sequences is ordered by direction north/east -> south/west.
+  # If the current station has only one stop sequence and the stop_id is the first or last entry in the stop_sequence,
+  # it is a terminal. Delay alerts heading in the direction of the station are not relevant.
+  defp relevant_direction?(
+         %ReconstructedAlert{
+           screen: %Screen{
+             app_params: %{reconstructed_alert_widget: %CurrentStopId{stop_id: stop_id}}
+           },
+           stop_sequences: [stop_sequence]
```
jzimbel-mbta, line 117: Hmm, I guess this code wouldn't work if we ever have a terminal stop that's also a transfer stop (served by multiple lines). In that case, there would be more than one stop sequence in the list.

E.g., imagine if the blue line ended at Government Center instead of Bowdoin. Government Center would be a terminal stop for the blue line.
image

While that assumption is safe to make now, it might not be forever, and it would be hard to catch the incorrect behavior for that specific edge case in the future.

What do you think about doing this instead, as a more general approach?
```ex
defp relevant_direction?(... stop_sequences: stop_sequences ...) do
  first_stops = MapSet.new(stop_sequences, &List.first/1)
  last_stops = MapSet.new(stop_sequences, &List.last/1)

  relevant_direction_for_terminal =
    cond do
      ...
      stop_id in first_stops -> 0
      stop_id in last_stops -> 0
      ...
    end

  ...
end
```

lib/screens/v2/candidate_generator/widgets/reconstructed_alert.ex
```diff
@@ -79,22 +79,69 @@ defmodule Screens.V2.CandidateGenerator.Widgets.ReconstructedAlert do
...
+         } = t
+       ) do
+    informed_entities = BaseAlert.informed_entities(t)
+
+    direction_id =
+      informed_entities
+      |> Enum.map(fn %{direction_id: direction_id} -> direction_id end)
+      |> Enum.uniq()
+      |> hd()
```
jzimbel-mbta, line 126: Will an informed entity with a non-nil direction_id always come before any ones with nil direction_id? If not, I think this would be a safer way to always get the first non-nil one, if it exists.
```suggestion
      informed_entities
      |> Enum.find(fn %{direction_id: direction_id} -> not is_nil(direction_id) end)
```
(`Enum.find` returns nil if no matching element is found)

lib/screens/v2/candidate_generator/widgets/reconstructed_alert.ex
cmaddox5: This was something I got caught up on during implementation, but I couldn't quite find the right solution. This is closer than I got, but I think it might be a little too generic. In this example (I know it's just a hypothetical but bear with me), BL trains from Gov't Center travel East and GL trains travel West. So they will always be filtering each other out. I think the biggest obstacle to this problem is not knowing which stop sequence belongs to which line. If we knew that, I think this approach would be great. I'll look into this, but do you know of a way to get this from V3?

lib/screens/v2/candidate_generator/widgets/reconstructed_alert.ex
cmaddox5: I don't _think_ it's possible for `informed_entities` to have nil and non-nil direction ids. So I don't think that should be a problem. I do like leaning on a default, though, so now `hd()` will never cause us issues.

lib/screens/v2/candidate_generator/widgets/reconstructed_alert.ex
```diff
@@ -79,22 +79,69 @@ defmodule Screens.V2.CandidateGenerator.Widgets.ReconstructedAlert do
...
+        relevant_boundary_alert?(reconstructed_alert)
 
       _ ->
         false
     end
   end
 
+  defp relevant_inside_alert?(%ReconstructedAlert{alert: alert} = reconstructed_alert) do
+    alert.effect != :delay or
+      (alert.severity > 3 and relevant_direction?(reconstructed_alert))
```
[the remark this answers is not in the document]
cmaddox5: This is a good idea. These functions can get confusing quickly if more conditionals are needed.

lib/screens/v2/candidate_generator/widgets/reconstructed_alert.ex
jzimbel-mbta: I think I see what you mean. I wonder if we could get around this issue by working with platform-level stop IDs in this check, rather than station IDs? That way, a delay alert on the BL only informing the BL platform would only match up with the stop sequence that contains BL platform IDs. (and vice versa for GL)

Related: do we get multiple stop sequences for any stop on the GL? I thought we got separate ones for the trunk and the different branches. If that's the case then this will also run into trouble for any terminal GL stop 😬

PR #1337, opened 2022-05-05
Fix Elixir Sentry with backend logger

hannahpurcell:
I wasn't sure about updating application.ex per the tutorial instructions. I tried it, but it errored, and I couldn't understand why. But looks like CTD has two ways it likes to add the LoggerBackend, so I'm trying the way that realtime_signs uses.


cmaddox5 (approved):

lib/screens/application.ex
```diff
@@ -26,6 +26,8 @@ defmodule Screens.Application do
     # See https://hexdocs.pm/elixir/Supervisor.html
     # for other strategies and supported options
     opts = [strategy: :one_for_one, name: Screens.Supervisor]
+    # Logger needed for Sentry; must be started after start_link
```
hannahpurcell, line 29:
```suggestion
```

PR #1338, opened 2022-05-09
Sentry config tweak

hannahpurcell:
Move configuration of Sentry from config to releases.ex, where realtime values can be used


config/releases.exs
```diff
@@ -54,6 +54,13 @@ config :screens,
   signs_ui_s3_bucket: signs_ui_s3_bucket,
   sentry_frontend_dsn: System.get_env("SENTRY_DSN")
 
+config :sentry,
+  dsn: System.get_env("SENTRY_DSN"),
+  environment_name: eb_env_name,
+  included_environments: [eb_env_name],
+  enable_source_code_context: true,
+  root_source_code_path: File.cwd!()
```
cmaddox5, line 62: Were `tags` unnecessary or should they still be here?

cmaddox5 (approved):
Looks good! Just one question.

config/releases.exs
hannahpurcell: I don't see any other team making use of them, so I'm leaving them out unless we deem them helpful

PR #1342, opened 2022-05-12
Set blank Sentry env vars for local dev

hannahpurcell:
**Asana task**: adhoc

Local dev errors were getting logged in Sentry and sending us emails! Agh! Make it stop.

I set the relevant variables to blank values in dev.exs. config.exs has a couple settings that can be set before runtime. And the newly named runtime.exs sets the variables that pull from our AWS config.


config/config.exs
```diff
@@ -57,11 +57,9 @@ config :ueberauth, Ueberauth.Strategy.Cognito,
...
   aws_region: {System, :get_env, ["COGNITO_AWS_REGION"]}
 
-env_name =
-  case System.get_env("ENVIRONMENT_NAME") do
-    nil -> Mix.env()
-    env -> String.to_existing_atom(env)
-  end
+config :sentry,
+  enable_source_code_context: true,
+  root_source_code_path: File.cwd!()
```
jzimbel-mbta, line 62: I'm a little wary about having this piece of config evaluate at compile time, in case the file structure or location that the application runs in changes between when it's compiled in GH Actions and when it runs in AWS.

At the same time, it might be better to evaluate here, during build, since we're looking for the root source code path? Hmmm.

If you're sure it will work as intended, or would be easy to catch and fix if it doesn't, looks good to me!

jzimbel-mbta (approved):
One comment, but I'm not 100% sure if it's a valid concern or not 🙃

config/config.exs
hannahpurcell: Oh uhhh hm! That's good insight. I realize this isn't needed anywhere but for the released app, so back to the releases.exs it goes.

PR #1343, opened 2022-05-13
fix: Response logging error

cmaddox5:
**Asana task**: [🐞 Protocol.UndefinedError: protocol String.Chars not implemented for %{force_reload: false, status: :success, success: false...](https://app.asana.com/0/1185117109217413/1202253017694183/f)

Removed a function clause. Only status is needed so matching to other keys is unnecessary.

- [ ] Needs version bump?

### DIFF
--- a/lib/screens/log_screen_data.ex
+++ b/lib/screens/log_screen_data.ex
@@ -41,17 +41,7 @@ defmodule Screens.LogScreenData do
   def log_api_response(response, screen_id, last_refresh, is_screen, screen_side \\ nil)
 
   def log_api_response(
-        %{force_reload: true, status: status},
-        screen_id,
-        last_refresh,
-        is_screen,
-        screen_side
-      ) do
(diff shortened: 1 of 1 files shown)


hannahpurcell (commented):
Line 53 assumes that if a response doesn't look like `%{status: string}` then the value must be a string (in order for it to be correctly printed on line 67). Is that a correct assumption?

Shouldn't  a `response` have a `status` in all scenarios?

cmaddox5:
> Line 53 assumes that if a response doesn't look like `%{status: string}` then the value must be a string (in order for it to be correctly printed on line 67). Is that a correct assumption?
> 
> Shouldn't a `response` have a `status` in all scenarios?

In the case of V2 logging, the status is represented as an atom, not a map key. So in those scenarios, it does not match to that function clause. This is due to the fact that V2 apps are able to evaluate response status in the controller.

hannahpurcell (approved):
Thanks for the explanation 👍

PR #1352, opened 2022-05-20
Prevent V3 API requests from being added to tests; fixed tests

hannahpurcell:
**Asana task**: [[tech debt] Prevent V3 API (and other external) requests during unit testing](https://app.asana.com/0/1185117109217413/1202094695278057)

Added atom `default_api_v3_url` to force an error whenever an API request is made in a test.
Investigated preventing all HTTPoison requests from occurring during tests, but seems like adding ExVCR is the better solution, which would be a different task from this one.
Fixed tests that were relying on the API.
ALSO test failures were logging to Sentry, so fixed that.

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -2,4 +2,4 @@
 
 Description
 
-- [ ] Needs version bump?
+- [ ] Tests added?
--- a/config/test.exs
+++ b/config/test.exs
@@ -9,7 +9,8 @@ config :screens, ScreensWeb.Endpoint,
 config :screens,
   config_fetcher: Screens.Config.State.LocalFetch,
   local_config_file_spec: {:test, "config.json"},
-  signs_ui_config_fetcher: Screens.SignsUiConfig.State.LocalFetch
+  signs_ui_config_fetcher: Screens.SignsUiConfig.State.LocalFetch,
+  default_api_v3_url: [:no_api_requests_allowed_during_testing]
 
 config :screens, ScreensWeb.AuthManager, secret_key: "test key"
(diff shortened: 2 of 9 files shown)


.github/pull_request_template.md
```diff
@@ -2,4 +2,4 @@
 
 Description
 
-- [ ] Needs version bump?
+- [ ] Tests added?
```
jzimbel-mbta, line 5: Thank you 😅

config/test.exs
```diff
@@ -9,7 +9,8 @@ config :screens, ScreensWeb.Endpoint,
 config :screens,
   config_fetcher: Screens.Config.State.LocalFetch,
   local_config_file_spec: {:test, "config.json"},
-  signs_ui_config_fetcher: Screens.SignsUiConfig.State.LocalFetch
+  signs_ui_config_fetcher: Screens.SignsUiConfig.State.LocalFetch,
+  default_api_v3_url: [:no_api_requests_allowed_during_testing]
```
jzimbel-mbta, line 13: Consider adding a comment above this explaining the reasoning, and/or demonstrating how to pass a `fetch_*` function in as an argument with a default value.

Main points:
- Functions that request data from an external resource outside our control cause flaky tests
- Pass request-firing functions in as arguments with the "normal" fetch function as the default value, so that we can easily pass a stubbed version in tests

jzimbel-mbta (approved):
Nice!! Glad we made this change before having way too many tests to fix.
Just one optional comment to consider. Otherwise looks good, after addressing the CI complaints.

PR #1360, opened 2022-06-02
fix: GL Location Text Bug

cmaddox5:
**Asana task**: ad-hoc

When GL alert is created that doesn't specify a branch, the location text returns nil unless a single trunk stop is affected. Added some logic that looks for endpoints in each branch `stop_sequences` before returning `nil`.

- [x] Tests added?


jzimbel-mbta (approved):
Thanks for testing!

PR #1379, opened 2022-06-24
Wellington alert bugfix

hannahpurcell:
Wellington station issue:  the code for getting the issue text for an outside suspension attempted to get a destination description, but that doesn't make sense for outside alerts. For example:

With a suspension between North Station and Back Bay, we only want to say "No trains between North Station and Back Bay" at Wellington, not "No North Station trains between..." or vice versa (unless the suspension is only happening in one direction, which is very unlikely.)

Tests updated.


hannahpurcell:
Re-writing the station_sequences portion of the test in the GL branch instead

jzimbel-mbta (approved):
Looks great, thanks for this!

PR #1380, opened 2022-06-24
fix: Recon alerts bug with alerts affecting all stops on a line

cmaddox5:
**Asana task**: ad-hoc

Alerts that affect an entire line were failing to get endpoints and causing screens to error. This changes the logic to return `nil` in those scenarios instead of looking for the endpoints.

- [ ] Tests added?


hannahpurcell (approved):
Looks good, thanks

PR #1383, opened 2022-06-27
feat: V3 retries

cmaddox5:
**Asana task**: [Re-try API requests on request failure](https://app.asana.com/0/1185117109217413/1202469674022356/f)

While doing the work ourselves might be easy, using this package is way easier. With this code, this function will retry up to 3 times with .5 seconds between each try each time `:bad_response_code` is returned. This atom is returned when we get a response from V3 that is not a 200 or 304.

- [ ] Tests added?


lib/screens/v3_api.ex
```diff
@@ -39,7 +42,10 @@ defmodule Screens.V3Api do
         :not_modified
 
       {:response_success, %{status_code: status_code}} = response ->
-        log_api_error({:bad_response_code, response}, status_code: status_code)
+        {:bad_response_code, _} =
+          log_api_error({:bad_response_code, response}, status_code: status_code)
```
jzimbel-mbta, line 46: Is this binding here because credo / dialyzer complained about not using the return value of the `log_api_error` call? If so, we've had an informal convention of just matching that to `_` in the past. Would suggest doing that here as well if we don't really care about the return value at all, and it looks like we don't.
```suggestion
        _ = log_api_error({:bad_response_code, response}, status_code: status_code)
```

lib/screens/v3_api.ex
```diff
@@ -39,7 +42,10 @@ defmodule Screens.V3Api do
...
+
+        :bad_response_code
```
jzimbel-mbta, line 48: This adds a new potential return value to the function if all retries fail, which makes me a liiiittle nervous, but I looked through our calls to `get_json` and they all wrap it in a `case` with a `_ ->` fallback, so we should be ok.

If there's any place where the function is called that doesn't have a catch-all fallback, it should also be easy to detect and fix as Sentry will report a `MatchError` if/when it receives `:bad_response_code`.

(Just laying out my thinking as to why this is change is safe)

jzimbel-mbta (approved):
One minor comment, otherwise looks good! Thanks for going with the `retry` library, it's very clear and concise.

lib/screens/v3_api.ex
cmaddox5: For peace of mind, I double-checked all references of this function and every references uses a `_ ->` fallback like you said. So I think we are good here. `log_api_error` already returns a few different values and none of those values are used for anything once the log has been captured.

PR #1388, opened 2022-06-30
fix: Add clause to departures_post_processing mapper fn to handle failed section fetches

jzimbel-mbta:
**Asana task**: [🐞 FunctionClauseError: no function clause matching in anonymous fn/1 in Screens.V2.CandidateGenerator.GlEink.departures_...](https://app.asana.com/0/1185117109217413/1202253124588511/f)

The surrounding code that calls this function already handles when a section fetch comes back as `:error`—we just needed to pass through that value when encountered.

- [ ] Tests added?

### DIFF
--- a/lib/screens/v2/candidate_generator/gl_eink.ex
+++ b/lib/screens/v2/candidate_generator/gl_eink.ex
@@ -207,6 +207,9 @@ defmodule Screens.V2.CandidateGenerator.GlEink do
 
       {:ok, departures} ->
         {:ok, departures}
+
+      :error ->
+        :error
     end)
   end
 end


lib/screens/v2/candidate_generator/gl_eink.ex
```diff
@@ -207,6 +207,9 @@ defmodule Screens.V2.CandidateGenerator.GlEink do
 
       {:ok, departures} ->
         {:ok, departures}
+
+      :error ->
```
cmaddox5, line 211: A potential suggestion: could we show the headsign here so the riders see something when this failure happens? Evaluating the headsign shouldn't need a request so we should always have that info. What do you think?

cmaddox5 (approved):
Looks great! Just one small suggestion. Totally up to you.

lib/screens/v2/candidate_generator/gl_eink.ex
jzimbel-mbta: I think that would make sense since this code is specifically for GL! Ok if it's done in a separate "Add headway mode to v2 GL e-ink screens" task, to let this fix get out faster?

cmaddox5: Yep, sounds good!

PR #1396, opened 2022-07-07
Increase checkout_timeout to double the connect_timeout

hannahpurcell:
**Asana task**: adhoc

I believe this should increase hackney's checkout_timeout threshold (which we want to be double connect_timeout)

- [ ] Tests added?

### DIFF
--- a/lib/screens/v3_api.ex
+++ b/lib/screens/v3_api.ex
@@ -5,7 +5,7 @@ defmodule Screens.V3Api do
 
   require Logger
 
-  @default_opts [timeout: 2000, recv_timeout: 2000, hackney: [pool: :api_v3_pool]]
+  @default_opts [timeout: 2000, recv_timeout: 2000, hackney: [{:pool, :api_v3_pool}, {:checkout_timeout, 4000}]]
 
   @retry with: Stream.take(constant_backoff(500), 3), atoms: [:bad_response_code]
   def get_json(
--- a/lib/screens/v3_api.ex
+++ b/lib/screens/v3_api.ex
@@ -8,7 +8,7 @@ defmodule Screens.V3Api do
   @default_opts [
     timeout: 2000,
     recv_timeout: 2000,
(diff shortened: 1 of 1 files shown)


lib/screens/v3_api.ex
```diff
@@ -5,7 +5,7 @@ defmodule Screens.V3Api do
 
   require Logger
 
-  @default_opts [timeout: 2000, recv_timeout: 2000, hackney: [pool: :api_v3_pool]]
+  @default_opts [timeout: 2000, recv_timeout: 2000, hackney: [{:pool, :api_v3_pool}, {:checkout_timeout, 4000}]]
```
jzimbel-mbta, line 8: One quick thing: `[pool: :api_v3_pool]` is syntactic sugar for `[{:pool, :api_v3_pool}]` (See [Keyword lists](https://hexdocs.pm/elixir/Keyword.html#module-examples))

I'll get back to this review soon (maybe tomorrow morning given my schedule) to double check that the option will be passed through to hackney correctly, but looks good on first glance!

hannahpurcell, line 8:
```suggestion
  @default_opts [timeout: 2000, recv_timeout: 2000, hackney: [pool: :api_v3_pool, checkout_timeout: 4000]]
```

lib/screens/v3_api.ex
```diff
@@ -5,7 +5,11 @@ defmodule Screens.V3Api do
 
   require Logger
 
-  @default_opts [timeout: 2000, recv_timeout: 2000, hackney: [pool: :api_v3_pool]]
+  @default_opts [
+    timeout: 2000,
+    recv_timeout: 2000,
+    hackney: [pool: :api_v3_pool, :checkout_timeout: 4000]
```
hannahpurcell, line 11:
```suggestion
    hackney: [pool: :api_v3_pool, checkout_timeout: 4000]
```

jzimbel-mbta (approved):

PR #1399, opened 2022-07-11
feat: endpoint for retrieving simulation data

cmaddox5:
**Asana task**: Part for [Full-screen simulation for E-ink, Prefare, Bus Shelter](https://app.asana.com/0/1185117109217413/1202505849835046/f)

This is the endpoint we will use to simulate v2 screens for Screenplay. The endpoint returns the full page, that will show the live data for the screen, and each page of the flex zone. Part 2 of this is to use this data to display the full page and each page of the flex zone at the same time.

- [ ] Tests added?


lib/screens_web/controllers/v2/screen_api_controller.ex
```diff
@@ -74,6 +74,25 @@ defmodule ScreensWeb.V2.ScreenApiController do
...
+      outdated?(screen_id, last_refresh) ->
+        json(conn, ScreenData.outdated_response())
+
+      disabled?(screen_id) ->
+        json(conn, ScreenData.disabled_response())
+
+      true ->
+        json(conn, %{
+          full_page: ScreenData.by_screen_id(screen_id),
+          flex_zone: ScreenData.paged_slots_by_screen_id(screen_id)
```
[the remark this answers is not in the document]
jzimbel-mbta: Suggested name: `V2.ScreenData.simulation_data_by_screen_id`

cmaddox5:
> Also, I'm assuming that the equivalent simulation data endpoint will be added for v1 screens in a separate PR, which is fine.

I think outside of DUPs (and maybe Solari), we are just going to use the existing endpoints in the iframes. v1 e-Ink has no plan to support overrides so I don't think there's a need for a separate endpoint for those. DUPs _might_ be an exception, but we may be able to get away with existing endpoints for those as well. I think that will need to be decided in a huddle/meeting in the near future.

jzimbel-mbta:
> I think outside of DUPs (and maybe Solari), we are just going to use the existing endpoints in the iframes.
Ah right, makes sense!

lib/screens_web/controllers/v2/screen_api_controller.ex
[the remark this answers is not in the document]
cmaddox5: Good call. I like that better.

lib/screens/v2/screen_data.ex
```diff
@@ -368,6 +392,36 @@ defmodule Screens.V2.ScreenData do
     response(data: data)
   end
 
+  def serialize_paged_slots(instance_map) do
```
jzimbel-mbta, line 395: This function can be private now

lib/screens_web/controllers/v2/screen_api_controller.ex
```diff
@@ -74,6 +74,23 @@ defmodule ScreensWeb.V2.ScreenApiController do
     end
   end
 
+  def simulation(conn, %{"id" => screen_id, "last_refresh" => last_refresh}) do
+    cond do
+      nonexistent_screen?(screen_id) ->
+        not_found_response(conn)
+
+      outdated?(screen_id, last_refresh) ->
+        json(conn, ScreenData.outdated_response())
```
jzimbel-mbta, line 83: What's the behavior if we send this back to one of the simulation iframes running inside the screenplay app? Does only the iframe refresh, or does the whole containing screenplay page refresh?

Looks like since the reload logic runs _inside_ the iframe (i.e., in the screens client code, not screenplay), it should work the way we want it to—the `window` object will refer to the iframe in that context. However if that turns out to not be the case, we can address that by changing what we call in screens: https://stackoverflow.com/questions/86428/what-s-the-best-way-to-reload-refresh-an-iframe
([currently](https://github.com/mbta/screens/blob/master/assets/src/hooks/use_api_response.tsx#L86) we call `window.location.reload`)

assets/css/v2/bus_eink/simulation.scss
```diff
@@ -0,0 +1,46 @@
...
+  .simulation__flex-zone-widget {
+    width: 300px;
+    height: 286px;
+
+    &:not(:last-child) {
+      margin-right: 24px;
+    }
+
+    .flex-one-medium {
+      //   margin: -8px;
```
jzimbel-mbta, line 37: Commented code can be removed, here and below.

assets/src/hooks/v2/use_simulation_api_response.tsx
```diff
@@ -0,0 +1,161 @@
...
+  // The request was successful.
+  {
+    fullPage: ApiResponse;
+    flexZone: ApiResponse;
+  };
+
+const FAILURE_RESPONSE: SimulationApiResponse = {
+  fullPage: { state: "failure" },
+  flexZone: { state: "failure" },
+};
```
jzimbel-mbta, line 24: Is the different shape of response data the main difference between `use_api_response` and this module? Wondering if we could avoid having a copy-pasted `useApiResponse`—I'm worried about code drift over time...

Factoring out the common code could be an interesting exercise, let me know if you'd like to pair on it 🤓

assets/src/hooks/v2/use_simulation_api_response.tsx
```diff
@@ -0,0 +1,161 @@
...
+    return { state: "success", data };
+  } else {
+    return { state: "failure" };
+  }
+};
+
+const rawResponseToSimulationApiResponse = ({
+  full_page,
+  flex_zone,
+}: {
```
jzimbel-mbta, line 42: This function, or some other logic in `useSimulationApiResponse`, needs to also handle the "screen is disabled" and "please reload" responses since it's currently possible for the endpoint to send those back, based on what's in the controller action down below:
```ex
outdated?(screen_id, last_refresh) ->
  json(conn, ScreenData.outdated_response())

disabled?(screen_id) ->
  json(conn, ScreenData.disabled_response())
```

jzimbel-mbta (commented):
Looking good! I think the backend is ready to go; I have a couple high-level questions about the frontend code but overall it looks pretty straightforward.

assets/src/hooks/v2/use_simulation_api_response.tsx
cmaddox5: Good callout. I think my response is should we care about outdated or disabled responses in these simulations? I'm actually not sure behavior of disabled screens has been brought up yet. Do we show simulations for those at all in Screenplay? We have that in the Screenplay data (although other than re-running the script, we really don't have a way to update that when a change is made in the admin tool), so we could use that to prevent iframes from being present for disabled screens. As for outdated, it kinda seems unnecessary for Screenplay. I don't really see these simulations being looked at long enough for it to get in that state. What do you think?

assets/src/hooks/v2/use_simulation_api_response.tsx
cmaddox5: The response shape was definitely what put me on this path at first. In the very beginning of this branch, I tried using `use_api_response` for simulations, but the amount of refactoring that was needed seemed dangerous given that we want the current version of the hook to be completely unchanged for existing endpoints. In my mind, I have felt like all things simulations should be siloed into their own hooks/functions/components so that we don't change something on live screens by mistake. Open to any suggestions on this, though.

lib/screens_web/controllers/v2/screen_api_controller.ex
jzimbel-mbta: 1. I think so—we should assume PIOs may keep a long-running tab open to screenplay, and make sure it doesn't get "out of sync" over time.
2. Hmm, that function should be able to return true. Every client loads with a timestamp of when it was initially loaded. It sends that timestamp with all of its data requests, and the server compares it with a timestamp kept in config for that screen ID. If we update the config with a later timestamp, then `outdated?/2` should return true and cause the server to send back a "please reload yourself" response

assets/src/hooks/v2/use_simulation_api_response.tsx
jzimbel-mbta: I think if the difference is only the response data + the logic that parses it, and we could have it be handled differently in the two cases by passing a callback function with the parsing logic, it would still be good to define the common logic around firing requests in one place. There's a clean separation between "fire requests, parse responses using [handler], handle the responses" and the "[handler]" part, as long as we strictly define the expected inputs and outputs of "[handler]".

In fact you've basically done that as far as I can tell—you call `rawResponseToSimulationResponse` vs `rawResponseToApiResponse`.

lib/screens_web/controllers/v2/screen_api_controller.ex
cmaddox5: Ahh, that admin tool request works a bit differently than I would have guessed. It adds a value into the config and then removes it once the refresh has occurred. So yeah, that's my bad.

assets/src/hooks/v2/use_simulation_api_response.tsx
jzimbel-mbta: Good question. If we have the data available in screenplay, then it's probably ok to just show "This screen is manually disabled" or something rather than loading up the iframe. If the client is outdated, then it should refresh itself, even if it's a simulation client. You're right that it would probably be a rare case, but it shouldn't be a challenge to implement.

jzimbel-mbta (approved):
Good to go, just one possible remaining change! Approving in advance.

assets/src/hooks/v2/use_api_response.tsx
```diff
@@ -124,23 +162,19 @@ const useApiResponse = ({
...
-          lastSuccess,
-          failureModeElapsedMs,
-          setApiResponse,
-          apiResponse
-        );
+        doFailureBuffer(lastSuccess, setApiResponse, apiResponse);
       } else {
         setApiResponse((prevApiResponse) => {
+          // Prevent on simulation_success
           if (prevApiResponse != null && prevApiResponse.state !== "success") {
```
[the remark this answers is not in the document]
jzimbel-mbta: If this does need updating, I would suggest defining an `isSuccess(response: ApiResponse)` helper function in case this check needs to happen somewhere else. (or actually, it's already needed [a few lines up from here](https://github.com/mbta/screens/blob/6f451ddb38fd1388b405944bd35e89dc9c9171a2/assets/src/hooks/v2/use_api_response.tsx#L94))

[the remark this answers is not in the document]
cmaddox5: Great catch. Definitely think we need to prevent these logs in simulations. They would just be redundant, or worst case they wouldn't be representative of actual issues on live screens. The helper function is a good idea so I got that added, but I wanted to run this by you one more time just for validation that I'm thinking it through correctly (it's Friday so my brain is mush). `isSuccess` as it is in my commit will for sure help us find successful responses. I think the problem is that `!isSuccess()` doesn't tell us on it's own if it was a failure. To combat this, I added a second part to just that conditional. Am I thinking about that correctly?

assets/src/hooks/v2/use_api_response.tsx
```diff
@@ -100,6 +100,9 @@ const doFailureBuffer = (
   }
 };
 
+const isSuccess = (response: ApiResponse) =>
+  response != null && response.state === "success";
```
jzimbel-mbta, line 104: I was thinking this helper func would return true for both forms of successful response: `success` and `simulation_success`. I might be missing a reason why we shouldn't lump them together like that, though.

If that does make sense to you, this also needs to check for `"simulation_success"`—either `["success", "simulation_success"].includes(response.state)` or `(response.state === "success" || response.state === "simulation_success")`

PR #1412, opened 2022-07-22
feat: Standardize on using data attributes on `#app` div for "frontend env vars"

jzimbel-mbta:
**Asana task**: ad hoc

This PR does the following:
- adds JS utility functions to easily get values from data attributes on the `#app` div, in much the same way that we use `Application.get_env` on the server
  - this includes a function, `fetchDatasetValue`, that throws an error if a given key is not present, so that we can avoid ending up with `NaN` (from `parseInt(undefined)`) and other bad values that silently cause problems
- uses the new functions throughout the client code, replacing any JS logic that looks at URL query params in the process
- has the server set new data attributes for `is_real_screen` and `screen_side`.

The overall goal is to standardize our approach to passing values from the server to the client, when the values are constant for the lifetime of the client.

Since there are a number of other large PRs that add the new screen simulation logic, I'm happy to wait on merging this until after those higher-priority changes are merged, to avoid slowdowns caused by conflicts.

- [ ] Tests added?

### DIFF
--- a/assets/src/hooks/v2/use_audio_readout.tsx
+++ b/assets/src/hooks/v2/use_audio_readout.tsx
@@ -1,5 +1,6 @@
 import { AudioConfig } from "Components/v2/screen_container";
 import useDriftlessInterval from "Hooks/use_driftless_interval";
+import { fetchDatasetValue, getDataset } from "Util/dataset";
 
 const readoutPath = (id: string) => `/v2/audio/${id}/readout.mp3`;
 const volumePath = (id: string) => `/v2/audio/${id}/volume`;
@@ -40,8 +41,7 @@ const useAudioReadout = ({
 
   const intervalPeriodMs = config.readoutIntervalMinutes * 60000;
 
-  const { refreshRateOffset } = document.getElementById("app").dataset;
-  const refreshRateOffsetMs = parseInt(refreshRateOffset, 10) * 1000;
+  const refreshRateOffsetMs = parseInt(fetchDatasetValue("refreshRateOffset"), 10) * 1000;
   const intervalOffsetSeconds = config.intervalOffsetSeconds;
   const intervalOffsetMs = refreshRateOffsetMs + (intervalOffsetSeconds * 1000);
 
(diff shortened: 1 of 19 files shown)


assets/src/util/dataset.tsx
```diff
@@ -0,0 +1,23 @@
...
+
+/**
+ * Gets the entire dataset.
+ */
+export const getDataset = () => document.getElementById("app")?.dataset ?? {};
+
+/**
+ * Gets one value from the dataset, or undefined if the key is not present.
+ */
+export const getDatasetValue = (key: string) => getDataset()[key];
```
cmaddox5, line 11: Just curious, why do we need both `getDatasetValue` and `fetchDatasetValue`? I know one function throws and the other doesn't, but wouldn't we always want to throw?

assets/src/hooks/v2/use_audio_readout.tsx
```diff
@@ -1,5 +1,6 @@
 import { AudioConfig } from "Components/v2/screen_container";
 import useDriftlessInterval from "Hooks/use_driftless_interval";
+import { fetchDatasetValue, getDataset } from "Util/dataset";
```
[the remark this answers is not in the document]
cmaddox5:
```suggestion
import { fetchDatasetValue } from "Util/dataset";
```

cmaddox5 (commented):
Looks great! Surprised we didn't think of this before now. Also really like the new util functions. Just have one question and one suggestion.

assets/src/util/dataset.tsx
jzimbel-mbta: In some cases we're ok with the value not being set, e.g. [the Sentry DSN](https://github.com/mbta/screens/blob/master/lib/screens_web/templates/v2/screen/index.html.eex#L9-L11). `fetch` is for when it has to be there, `get` is for when we expect it to sometimes be undefined. I was going off of Elixir's `Map.get` and `Map.fetch!` functions.

However if you think it makes more sense, I could remove `getDatasetValue` and have the server always set _some_ value (maybe empty string or "none") for every expected data attribute.

cmaddox5: Ahh I see. Didn't think about those cases. I think two util functions is better than always setting the value in that case. I could imagine that adding new optional attributes could cause errors until we remember that _something_ needs to be there. So I like this.

cmaddox5 (approved):
Looks great! There is a merge conflict, but I'll go ahead and throw you an approval.

PR #1413, opened 2022-07-22
fix: Use full station names where we expect them for Clev Circ, Wash Sq

jzimbel-mbta:
**Asana task**: [Add full-length "Cleveland Circle" name for reconstructed alerts](https://app.asana.com/0/1185117109217422/1202661980135928/f)

I didn't work directly on this code, so I might be missing some context around why these two stops are abbreviated in two different ways rather than the first string in the tuple being unabbreviated like the rest of the entries. Please let me know if there was a reason for this!

(:thinking: Also, should the abbreviation for Washington Square be Wash Sq? Or keep it as-is (Washington))

- [ ] Tests added?


hannahpurcell (approved):
> context around why these two stops are abbreviated in two different ways rather than the first string in the tuple being unabbreviated like the rest of the entries

This code is also used for Subway status, in case that jogs some old memory. I don't know of an obvious reason why these two would be abbreviated by default, so the change looks good with me 👍

PR #1416, opened 2022-07-28
Sentry for E-ink

hannahpurcell:
**Asana task**: no asana task

The e-ink screens (which use QtWeb browser, circa 2013) couldn't handle the `@sentry/react` or `@sentry/browser` SDK for frontend error tracking. So, while the backend error tracking was working fine, the frontend error tracking had issues on e-inks when we turned it on with the `is_real_screen` parameter.

So, this rolls us back to an older SDK, called Raven (circa 2019). Tested:
- Backend error logging still works as it always has (nothing changed there)
- Frontend captureMessage works at time of initialization ([example](https://sentry.io/organizations/mbtace/issues/3429325798/?project=6061747&query=is%3Aunresolved)) and at other times, like from`useApiResponse` ([example](https://sentry.io/organizations/mbtace/issues/3461463035/?project=6061747&query=is%3Aunresolved))
- Frontend error logging now works ([example](https://sentry.io/organizations/mbtace/issues/3459204429/?project=6061747&query=is%3Aunresolved))
- Entering / exiting the no-data state at odd times is caught correctly ([example](https://sentry.io/organizations/mbtace/issues/3459046977/?environment=screens-dev-green&project=6061747&query=is%3Aunresolved&statsPeriod=14d). this was observed on the actual screen)

Note: The polyfill was previously suspected as helpful, but it actually had no effect.

- [ ] Tests added?

Agh, I'm sure tests can/should be added, but it would take a lot of figuring out... Unsure if it's worth it?


assets/src/util/sentry.tsx
```diff
@@ -1,5 +1,5 @@
-import * as Sentry from "@sentry/react";
 import { isRealScreen } from "Util/util";
+import Raven from "raven-js";
```
hannahpurcell, line 2:
```suggestion
// Previously tried @sentry/react and @sentry/browser as the SDK, but the QtWeb browser on e-inks could not 
// use them. Raven is an older stable SDK that better works with older browsers.
import Raven from "raven-js";
```

assets/src/hooks/v2/use_api_response.tsx
```diff
@@ -57,7 +57,7 @@ const doFailureBuffer = (
       // This will trigger until a success API response is received.
       setApiResponse((prevApiResponse) => {
         if (prevApiResponse != null && prevApiResponse.state === "success") {
-          Sentry.captureMessage("Entering no-data state.");
+          Raven.captureMessage("Entering no-data state.", {level: "info"});
```
jzimbel-mbta, line 60: Oh! This is neat! I completely missed that we can set the log level like that. This is really helpful. I assumed there was no way to stop Sentry from freaking out about every error/message we sent to it.

assets/src/util/sentry.tsx
```diff
@@ -11,12 +13,11 @@ const initSentry = (appString: string) => {
...
-  });
 
-  Sentry.captureMessage(`Sentry intialized for app: ${appString}`);
+  if (sentryDsn && isRealScreen()) {
+    Raven.config(sentryDsn, {environment: env}).install();
+    Raven.captureMessage(`Sentry intialized for app: ${appString}`, {level: "info"});
+  }
 };
 
 export default initSentry;
```
jzimbel-mbta, line 23: Do you think it might also be useful to export a function (or functions) that wraps `Raven.captureMessage` so that if/when we need to change what gets called (e.g. switching back to `Sentry.captureMessage`, or something else), the change only needs to happen in one place?

Like:
```tsx
// https://docs.sentry.io/clients/javascript/usage/#raven-js-additional-context
type LogLevel = "info" | "warning" | "error";

const log = (message: string, level: LogLevel) => {
  Raven.captureMessage(message, {level});
};
const info = (message: string) => log(message, "info");
const warn = (message: string) => log(message, "warning");
const error = (message: string) => log(message, "error");

export default initSentry; // (unchanged)
export {info, warn, error};

//// in another file ////
import * as Logger from "Util/sentry";

Logger.info("something happened");
Logger.error("something bad happened");
```

jzimbel-mbta (approved):
Nice work! I just have a suggestion to simplify our event-based logging. Mainly, it would help us avoid accidentally setting a log level that Raven/Sentry doesn't understand, like `"warn"` instead of `"warning"`.

assets/src/util/sentry.tsx
hannahpurcell: Oh sure, thanks! Tested the update, and it works nicely.

PR #1422, opened 2022-08-04
experiment: Switch upper and lower and remove elevator status

cmaddox5:
**Asana task**: [Implement alert experiment: eliminate elevator widget and move subway status and alert widget](https://app.asana.com/0/1185117109217413/1202709465836411/f)

See design mock-up in task.

***This should not be deployed until go-ahead is given by design**

- [ ] Tests added?


jzimbel-mbta (approved):
I love that we can do this with relatively few changes! ✨

PR #1423, opened 2022-08-04
experiment: Switch sides for pre-fare content

cmaddox5:
**Asana task**: [Implement alert experiment: switch left and right pre-fare screens](https://app.asana.com/0/1185117109217413/1202709465836408/f)

See design mock-up in task.

***This should not be deployed until go-ahead is given by design**

- [ ] Tests added?


PaulJKim (approved):
The change makes sense to me, given that this is a temporary experiment for now and we don't need to be concerned about some of the map keys referring to the opposite side of the screen.

PR #1429, opened 2022-08-09
polish: Widget sim background

cmaddox5:
**Asana task**: [📺▶️ : 🇵🇱 | Make "speech bubble" around widgets match Figma](https://app.asana.com/0/1185117109217413/1202728175921081/f)

This one was a doozy but I think I got it. Some of the CSS in the existing background was not right/not at all written well so I corrected what I saw. The angled background is done in the `:before` part of the CSS files.

- [ ] Tests added?

### DIFF
--- a/assets/css/v2/bus_eink/simulation.scss
+++ b/assets/css/v2/bus_eink/simulation.scss
@@ -31,10 +31,10 @@
       content: "";
       position: absolute;
       right: 100%;
-      height: 216px;
-      border-top: 50px solid transparent;
-      border-right: 70px solid #2e3f4d;
-      border-bottom: 50px solid transparent;
+      // NO HEIGHT
+      border-top: 158px solid transparent;
+      border-right: 128px solid #2e3f4d;
+      border-bottom: 158px solid transparent;
     }
   }
 
--- a/assets/css/v2/bus_eink/simulation.scss
+++ b/assets/css/v2/bus_eink/simulation.scss
@@ -23,22 +23,30 @@
     margin-top: auto;
     margin-bottom: auto;
     border-radius: 8px;
+    height: 316px;
+    border-radius: 0 8px 8px 0;
   }
 
   .simulation__flex-zone-widget {
     width: 300px;
     height: 286px;
+    margin-top: auto;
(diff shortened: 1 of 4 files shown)


assets/css/v2/bus_eink/simulation.scss
```diff
@@ -8,8 +8,9 @@
   .simulation__full-page {
     height: 480px;
     width: 180px;
-    margin-right: 24px;
+    margin-right: 72px;
```
hannahpurcell, line 11: The gl_eink margin also needs to be 72px

assets/css/v2/bus_eink/simulation.scss
```diff
@@ -22,23 +23,41 @@
...
+
+    &:before {
+      content: "";
+      position: absolute;
+      right: 100%;
+      height: 216px;
+      border-top: 50px solid transparent;
+      border-right: 70px solid #2e3f4d;
+      border-bottom: 50px solid transparent;
+    }
```
hannahpurcell, line 38: A more systematic way of creating this triangle is taking the height of the projection (316px), dividing it by two, and setting the top and bottom borders to that value, with a height of `0px`. The border-right value will be how steep you want the triangle. Keep content, position, and right as-is. So for e-ink:
```
&:before {
    content: "";
    right: 100%;
    // NO HEIGHT
    border-top: 158px solid transparent;
    border-right: 128px solid #2e3f4d;
    border-bottom: 158px solid transparent;
}
```
 Borrowed from https://css-tricks.com/snippets/css/css-triangle/.

hannahpurcell: For PreFare as another example, it looks like border-right would be approx `144px` and top/bottom would be `72px`.

hannahpurcell (commented):
Couple of tweaks. Otherwise, projection looks good

assets/css/v2/bus_eink/simulation.scss
cmaddox5: Awesome! SO much better.

hannahpurcell (approved):
Yeah, that looks great!

PR #1430, opened 2022-08-10
feat: Page load no data state

cmaddox5:
**Asana task**: [📺▶️ : 🇵🇱 | New initial loading screen](https://app.asana.com/0/1185117109217413/1202728175921073/f)

Added a new no data state that is specific for page load. This will only flash for a second as a screen is booting up. Riders will rarely see it, but it will be most helpful in Screenplay. At the moment, users will see the existing NoData page in that second, so this new page will reduce confusing when viewing a simulation.

- [ ] Tests added?


jzimbel-mbta (approved):
I hate to ask but do we need this for the v1 apps as well? Otherwise, looks great.

cmaddox5:
> I hate to ask but do we need this for the v1 apps as well? Otherwise, looks great.

Hmm, good question. I think the only thing keeping me from saying yes is our understanding that v1 will not be supported forever. But that makes more sense for widget sims than something PIOs will actually see when they view a sim. I'll ask this in Slack.

cmaddox5:
@jzimbel-mbta Had a chat with PaulH and he agreed that v1 should also get the loading state. Just pushed up a commit adding them. v1 E-Ink matches v2 E-Ink, Solari matches v2 LCD, DUPs use a new custom loading page.

hannahpurcell:
Hallo! The V1 load states looks great, but the V2 ones don't seem to be showing a load state atm (anymore). 

DUPs look great and v1 e-inks. Prefare, Bus Shelter, and V2 e-inks are currently showing "loading".

hannahpurcell (approved):
Looks great, code makes sense

PR #1432, opened 2022-08-11
fix: Add refresh rate to DUP simulations

cmaddox5:
**Asana task**: ad-hoc

Added a 20 second refresh to DUP simulations so Screenplay will continue showing updated info.

- [ ] Tests added?


hannahpurcell (approved):
Nice! Easy peasy, lookin clean

PR #1435, opened 2022-08-16
BlueBikes widget backend

jzimbel-mbta:
**Asana task**: [Create new Blue Bike pre-fare widget](https://app.asana.com/0/1185117109217413/1202791356936898/f)

This is the server code for the new "nearby BlueBikes stations" widget.

Since BlueBikes exposes their system data via two regularly-updated JSON files, I set up a genserver that caches that data, merged into a single map, and allows the widget code to grab just the stations it needs, rather than the entire set of all stations.

Here is how the data that BlueBikes exposes looks, with only the fields we care about included.
`station_information.json`: ([link](https://gbfs.bluebikes.com/gbfs/en/station_information.json))
```ts
interface InformationJson {
  data: {
    stations: StationInformation[]
  },
  last_updated: <unix timestamp integer>
}

interface StationInformation {
  station_id: string,
  name: string,
}
```

`station_status.json`: ([link](https://gbfs.bluebikes.com/gbfs/en/station_status.json))
```ts
interface StatusJson {
  data: {
    stations: StationStatus[]
  },
  last_updated: <unix timestamp integer>
}

type StationStatus = NormalStatus | OutOfServiceStatus | ValetStatus;

interface OutOfServiceStatus {
  station_id: string,
  station_status: "out_of_service"
}

interface ValetStatus {
  station_id: string,
  valet: {
    active: true
  }
}

interface NormalStatus {
  station_id: string,
  station_status: "active",
  num_bikes_available: number,
  num_docks_available: number
}
```

The genserver refreshes its data via two requests every 5 seconds, since BlueBikes specifies their TTL as 5 seconds. We will never make more requests than that.

Remaining work:

- [x] finish up WidgetInstance implementation
- [x] audio equivalence
- [x] unit tests


lib/screens_web/views/v2/audio/blue_bikes_view.ex
```diff
@@ -0,0 +1,70 @@
...
+    walk_distance_minutes: 30
+  },
+  %{
+    arrow: :e,
+    id: "61",
+    name: "Boylston St at Fairfield St",
+    status: :out_of_service,
+    walk_distance_minutes: 45
+  }
+]
```
cmaddox5, line 70: Is this here intentionally for documentation purposes?

cmaddox5 (approved):
Looks great! I feel like you got this done super quick. I did have one question, but the answer won't have any impact of approval.

lib/screens_web/views/v2/audio/blue_bikes_view.ex
jzimbel-mbta: Whoops! I pasted it in as a sample of the serialized data while working on the view functions and intended to remove it after I was done. Funny that the compiler/Credo/Dialyzer didn't complain about it at all. I guess the compiler just happily evaluates expressions like this, just like everything else, and since they have no side effects (like defining a module or a function), it's not an issue.

PR #1436, opened 2022-08-16
Revert "experiment: Switch upper and lower and remove elevator status"

cmaddox5:
Reverts mbta/screens#1422


hannahpurcell (approved):
thanks for tee-ing this up

PR #1437, opened 2022-08-16
feat: Shuttle Bus Widget Backend

cmaddox5:
**Asana task**: [Create new Shuttle Bus pre-fare widget](https://app.asana.com/0/1185117109217413/1202791356936896/f)

This adds the backend for a new Shuttle Bus widget. Most of the widget will be static text that will live in the frontend. Boarding instructions will be stored in the config along with the arrow icon to use, the minute range, and the destination of the shuttle.

- [x] Tests added?

### DIFF
--- a/lib/screens/config/v2/shuttle_bus_info.ex
+++ b/lib/screens/config/v2/shuttle_bus_info.ex
@@ -26,7 +26,7 @@ defmodule Screens.Config.V2.ShuttleBusInfo do
 
   use Screens.Config.Struct
 
-  for arrow <- ~w[n ne e se s sw w nw nil]a do
+  for arrow <- ~w[n ne e se s sw w nw]a do
     arrow_string = Atom.to_string(arrow)
     defp value_from_json("arrow", unquote(arrow_string)), do: unquote(arrow)
   end
--- a/lib/screens/config/v2/shuttle_bus_info.ex
+++ b/lib/screens/config/v2/shuttle_bus_info.ex
@@ -7,15 +7,26 @@ defmodule Screens.Config.V2.ShuttleBusInfo do
           minutes_range_to_destination: String.t(),
           destination: String.t(),
           arrow: arrow(),
+          english_boarding_instructions: String.t(),
+          spanish_boarding_instructions: String.t(),
           priority: WidgetInstance.priority()
         }
 
   @type arrow :: :n | :ne | :e | :se | :s | :sw | :w | :nw | nil
 
-  @enforce_keys [:minutes_range_to_destination, :destination, :arrow, :priority]
+  @enforce_keys [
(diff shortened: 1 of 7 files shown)


lib/screens/config/v2/shuttle_bus_info.ex
```diff
@@ -0,0 +1,42 @@
...
+    :english_boarding_instructions,
+    :spanish_boarding_instructions,
+    :priority
+  ]
+  defstruct minutes_range_to_destination: nil,
+            destination: nil,
+            arrow: nil,
+            english_boarding_instructions: nil,
+            spanish_boarding_instructions: nil,
+            priority: nil
```
jzimbel-mbta, line 30: If you want, this can reuse `@enforce_keys`:
```suggestion
  defstruct @enforce_keys
```

lib/screens/config/v2/shuttle_bus_info.ex
```diff
@@ -0,0 +1,42 @@
...
+
+  use Screens.Config.Struct
+
+  for arrow <- ~w[n ne e se s sw w nw nil]a do
```
jzimbel-mbta, line 34: `nil` will be handled correctly (`Jason` decodes JSON `null` as `nil`) so you don't need to add an explicit `value_from_json` clause for that one. Currently it would cause `{"arrow": "nil"}` to be converted to `%{arrow: nil}`.
```suggestion
  for arrow <- ~w[n ne e se s sw w nw]a do
```

lib/screens_web/views/v2/audio/shuttle_bus_info_view.ex
```diff
@@ -0,0 +1,10 @@
+defmodule ScreensWeb.V2.Audio.ShuttleBusInfoView do
+  use ScreensWeb, :view
+
+  def render("_widget.ssml", %{
+        minutes_range_to_destination: minutes_range_to_destination,
+        destination: destination
+      }) do
+    ~E|<p>Estimated <%= minutes_range_to_destination %> minute to <%= destination %></p>|
```
jzimbel-mbta, line 8: This looks fine for now since we're expecting to get specific readout text from design soon. Though I think "minute" should be "minutes", e.g. "15-20 minutes".

jzimbel-mbta (approved):
Looks good to go, just some minor suggestions. Nice work.

lib/screens_web/views/v2/audio/shuttle_bus_info_view.ex
cmaddox5: Right, this was more of a placeholder until we know exactly what we want to be read out for this widget. I knew that this line would be part of the readout, but I anticipate there being some additional text to add to it.

PR #1439, opened 2022-08-17
feat: OL Surge PreFare Slots

cmaddox5:
**Asana task**: [Add new slots to pre-fare template](https://app.asana.com/0/1185117109217413/1202809139605053/f)

Added a new slot for OL Surge widgets: `orange_line_surge_upper` and `orange_line_surge_lower`. Right now, each slot takes up half of the space under the header. If any widget in the `body_right_normal` group has a higher priority, it will not show these new widgets.

- [ ] Tests added?

### DIFF
--- a/assets/css/v2/pre_fare/body_right/surge.scss
+++ b/assets/css/v2/pre_fare/body_right/surge.scss
@@ -1,4 +1,4 @@
-// body total height: 1648px
+// body total height: 1720px
 
 .body-right-surge {
   position: relative;
@@ -10,16 +10,16 @@
   left: 0;
 
   width: 1080px;
-  height: 876px;
+  height: 830px;
   overflow: hidden;
 }
 
 .body-right-surge__lower {
   position: absolute;
-  top: 918px;
+  top: 830px;
   left: 0;
 
   width: 1080px;
-  height: 876px;
+  height: 890px;
   overflow: hidden;
 }
--- a/assets/css/v2/pre_fare/body_right/surge.scss
+++ b/assets/css/v2/pre_fare/body_right/surge.scss
@@ -1,0 +1,25 @@
+// body total height: 1648px
+
+.body-right-surge {
+  position: relative;
+}
+
+.body-right-surge__upper {
+  position: absolute;
+  top: 0;
+  left: 0;
+
+  width: 1080px;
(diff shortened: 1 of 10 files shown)


assets/css/v2/pre_fare/body_right/surge.scss
```diff
@@ -0,0 +1,25 @@
...
+  left: 0;
+
+  width: 1080px;
+  height: 876px;
+  overflow: hidden;
+}
+
+.body-right-surge__lower {
+  position: absolute;
+  top: 918px;
```
jzimbel-mbta, line 19: Should this be `876px`? I think normally we put one slight immediately next to (or below) the previous one, so I would have expected this `top:` to have the same value as the `height:` of the upper slot.

assets/css/v2/pre_fare/body_right/surge.scss
```diff
@@ -0,0 +1,25 @@
...
+  left: 0;
+
+  width: 1080px;
+  height: 876px;
```
jzimbel-mbta, line 23: I think this height and that of the upper slot should add up to the body total height. Currently we have 876 + 876 = 1752 ≠ 1648 (assuming the comment at the top of the file is accurate)

jzimbel-mbta (commented):
Just reviewing this because I need it for the BlueBikes widget 😅

assets/css/v2/pre_fare/body_right/surge.scss
cmaddox5: Thanks for bumping this. I started to add some numbers initially until I realized we didn't know dimensions yet. Now that we have them, I'll make these match the final numbers in the designs.

assets/css/v2/pre_fare/body_right/surge.scss
jzimbel-mbta: Ah that comment was not correct! The pre-fare screen body is 1720px tall.

cmaddox5: Yep. I originally copied this file from `normal.scss`, which also happened to be incorrect 😅 Got them both corrected.

jzimbel-mbta (approved):

PR #1440, opened 2022-08-17
tweak: Show partial alerts on DUP screen 1 if configured for CR/SL

cmaddox5:
**Asana task**: [DUP: Preserve CR/SL predictions](https://app.asana.com/0/1185117109217413/1202809139605042/f)

The code for this was there, but we needed to change how we determine if a stop has multiple lines. It originally was hardcoded and only Haymarkey met the criteria. Now, it just looks at the number of `sections` in the `primary` of the config. If there is more than one section, show a partial alert on screen 1 and a fullscreen alert on screen 2. 

- [ ] Tests added?

### DIFF
--- a/lib/screens/dup_screen_data.ex
+++ b/lib/screens/dup_screen_data.ex
@@ -67,9 +67,9 @@ defmodule Screens.DupScreenData do
     alerts_by_section = fetch_and_interpret_alerts(primary_departures)
     alerts = flatten_alerts(alerts_by_section)
 
-    line_count = Data.station_line_count(primary_departures)
+    mode_or_line_count = length(primary_departures.sections)
 
-    case Data.response_type(alerts, line_count, rotation_index) do
+    case Data.response_type(alerts, mode_or_line_count, rotation_index) do
       :departures ->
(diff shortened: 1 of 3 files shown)


lib/screens/dup_screen_data.ex
```diff
@@ -67,17 +67,17 @@ defmodule Screens.DupScreenData do
     alerts_by_section = fetch_and_interpret_alerts(primary_departures)
     alerts = flatten_alerts(alerts_by_section)
 
-    line_count = Data.station_line_count(primary_departures)
+    mode_or_line_count = length(primary_departures.sections)
```
hannahpurcell, line 70: could this still just be `line_count`?

hannahpurcell (approved):
Ah I love it when the code is already there and we just have to turn it on. 👏

lib/screens/dup_screen_data.ex
cmaddox5: This change was primarily motivated by the possibility of commuter rail being a section of the primary departures. The original name was `line_count` because the logic only checked for multiple subway lines, but that isn't the case anymore.

PR #1441, opened 2022-08-17
tweak: Add overrides for DUP secondary departures

cmaddox5:
**Asana task**: [DUP: Add config option to show a partial alert on the secondary departures (bus) page](https://app.asana.com/0/1185117109217413/1202809139605045/f)

This adds the ability to override the secondary departures screen on DUPs. Right now, only partial alerts can be added. This also changes the config so that we no longer need to override all screens at the same time. A value of `null` can be added for a screen override if we do not need an override for that rotation. 

- [ ] Tests added?


jzimbel-mbta (approved):
This looks ideal for unit-testing but I think it looks good! I'll add a tech-debt ticket for the tests.

PR #1443, opened 2022-08-18
feat: Audio for evergreen content

cmaddox5:
**Asana task**: [Audio equivalence for evergreen content widgets](https://app.asana.com/0/1185117109217413/1202815198320926/f)

This adds audio equivalence to evergreen content. `audio_valid_candidate?/1` will only return true if the widget config actually has a string value. For now, evergreen audio will always have the lowest priority. 

- [ ] Tests added?


PaulJKim (approved):
LGTM!

PR #1444, opened 2022-08-18
tweak: Changed minute range to include a schedule

cmaddox5:
**Asana task**: ad-hoc

Changed the config to include a schedule for the minute range we will show in the widget. 

- [ ] Tests added?


lib/screens/util.ex
```diff
@@ -135,4 +135,17 @@ defmodule Screens.Util do
...
+  @doc """
+  Returns true if given Time object falls between start_time and end_time.
+  """
+  def time_in_range?(t, start_time, stop_time) do
+    if Time.compare(start_time, stop_time) in [:lt, :eq] do
+      # The range exists within a single day starting/ending at midnight
+      Time.compare(start_time, t) in [:lt, :eq] and Time.compare(stop_time, t) == :gt
+    else
+      # The range crosses midnight, e.g. start: 5am, stop: 1am
+      Time.compare(start_time, t) in [:lt, :eq] or Time.compare(stop_time, t) == :gt
```
hannahpurcell, line 148: If time range is 11pm-1am, and t is 5pm, this equates to true -- which isn't expected for a function named `time_in_range` but we probably won't see a use case like that (except in testing) since most overnight ranges will probably be like 7am-3am or something. Could rename the function and/or add a comment with more info about the parameter restrictions? Like both need to be an AM time for it to work.

hannahpurcell (approved):
Approved with 1 need-to-fix

lib/screens/util.ex
cmaddox5: I think in your example, it would still return false. 23:00:00 would be greater than 17:00:00 and 01:00:00 would be less than 17:00:00, so neither side of the `or` would evaluate to `true`.

hannahpurcell: Oh my gosh. You're right. Thanks :doh:

PR #1446, opened 2022-08-19
CR widget backend

hannahpurcell:
**Asana task**: [Create new Commuter Rail pre-fare widget](https://app.asana.com/0/1185117109217413/1202791356936894)

Covers main backend changes, excluding audio.

Config structure:
```json
{
  "cr_departures": {
    "station": "place-rugg",
    "destination": "place-bbsta",
    "direction_to_destination": 1,
    "priority": [1],
    "travel_time_to_destination": "9-12m",
    "show_via_headsigns_message": true,
    "wayfinding_arrows": {
      "e": [1, 2, 3],
      "w": [5, 7]
    }
  }
}
```

"wayfinding_arrows" can also just be a string in cases where there is only 1 clear platform: "n"

- [ ] Tests added? Nope! Barely. Clean-up task [here](https://app.asana.com/0/1185117109217422/1202822127862361)


lib/screens/v2/candidate_generator/pre_fare.ex
```diff
@@ -58,6 +58,8 @@ defmodule Screens.V2.CandidateGenerator.PreFare do
   end
 
   @impl CandidateGenerator
+  # Error of "arity is too high"
```
jzimbel-mbta, line 61: :oh-no:

lib/screens/v2/candidate_generator/widgets/cr_departures.ex
```diff
@@ -0,0 +1,60 @@
...
+    }
+
+    destination = fetch_stop_name_fn.(cr_departures.destination)
+
+    {:ok, departures_data} = fetch_departures_fn.(opts, params)
+
+    # The Overnight and NoData widgets may not be relevant here
+    departures_instance =
+      cond do
+        departures_data == :error ->
```
jzimbel-mbta, line 38: I think this would mean we got `{:ok, :error}` back from `fetch_departures_fn.(opts, params)` above, is that right?

It also does not look like `Departure.fetch/2` can return `{:ok, :overnight}` but I might be missing something.

A `case` expression might work best here, again I'm not 100% sure of the possible return values but it looks like we're expecting something like
```ex
case fetch_departures_fn.(opts, params) do
  :error ->
    %DeparturesNoData{screen: config, show_alternatives?: true}

  {:ok, :overnight} ->
    %OvernightDepartures{}

  _ ->
    %CRDeparturesWidget{
      config: cr_departures,
      departures_data: departures_data,
      destination: destination
    }
end
```

lib/screens/v2/candidate_generator/widgets/cr_departures.ex
```diff
@@ -0,0 +1,60 @@
...
+      direction_id: cr_departures.direction_to_destination,
+      route_type: :rail,
+      stop_ids: [cr_departures.station]
+    }
+
+    destination = fetch_stop_name_fn.(cr_departures.destination)
+
+    {:ok, departures_data} = fetch_departures_fn.(opts, params)
+
+    # The Overnight and NoData widgets may not be relevant here
```
jzimbel-mbta, line 35: Oops, I just saw this code comment after writing my overly-detailed PR comment below. Feel free to disregard whatever I just said, or use whatever part of it may be useful.

lib/screens/v2/widget_instance/cr_departures.ex
```diff
@@ -0,0 +1,150 @@
+defmodule Screens.V2.WidgetInstance.CRDepartures do
+  @moduledoc false
+
+  alias Screens.Config.V2.CRDepartures
+  alias Screens.V2.Departure
+  alias Screens.V2.WidgetInstance.CRDepartures, as: CRDeparturesWidget
```
jzimbel-mbta, line 6: Could you move this alias inside of the `defimpl` block to avoid confusion?
(Also huh, I'm surprised that putting the alias out here makes it available inside the `defimpl` 🤔)

lib/screens/v2/widget_instance/cr_departures.ex
```diff
@@ -0,0 +1,150 @@
...
+    end
+
+    def audio_sort_key(_instance), do: [1]
+
+    def audio_valid_candidate?(_instance), do: true
+
+    def audio_view(_instance), do: ScreensWeb.V2.Audio.CRDeparturesView
+  end
+
+  def serialize_departure(%Departure{} = departure, wayfinding_arrows, now \\ DateTime.utc_now()) do
```
jzimbel-mbta, line 66: Normally when a widget requires a `now` value, we add it as a field on the struct ([example](https://github.com/mbta/screens/blob/master/lib/screens/v2/widget_instance/alert.ex#L19)) so that all logic inside the widget calculates time-based stuff from the same `now`. If there's time, please do that, or make a tech-debt ticket if not.

jzimbel-mbta (approved):
There's a lot of copy-pasted code, but that's fine given the circumstances / similarity with the regular departures widget. I think this looks good overall.

The only thing that worries me is the part of the config definition that I commented on. I think we won't properly parse the config with the code as it currently stands.

Please try adding some sample config to test the logic. Remember that if you start the server with `iex -S mix phx.server`, you can call any public function from any module on the command line to test out what it returns!

lib/screens/v2/widget_instance/cr_departures.ex
hannahpurcell: This pattern was taken from WidgetInstance.Departures

lib/screens/v2/widget_instance/cr_departures.ex
hannahpurcell: Ah, this issue is also in Departures

lib/screens/config/v2/cr_departures.ex
```diff
@@ -0,0 +1,45 @@
...
+          | String.t()
+
+  @type t :: %__MODULE__{
+          station: String.t(),
+          destination: String.t(),
+          direction_to_destination: 0 | 1,
+          priority: WidgetInstance.priority(),
+          travel_time_to_destination: String.t(),
+          show_via_headsigns_message: true | false,
+          wayfinding_arrows: platform_directions()
```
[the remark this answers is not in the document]
hannahpurcell: 🤔 I've tested this logic using the 2 sample configs in the PR description with no issue. What do you mean "try adding some sample config to test the logic"?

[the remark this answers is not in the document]
jzimbel-mbta: Huh ok! If it works then I guess it's fine. I did not expect keys in the `wayfinding_arrows` sub-object to properly get converted from strings `"left"` `"right"` to atoms `:left` `:right`. Normally `Jason.decode/1` maintains keys from JSON objects as strings, and we have to go through and convert them to atoms.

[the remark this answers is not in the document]
jzimbel-mbta: Oh, it's working because none of our backend code actually tries to use access those keys. The map just gets dropped into the return value of `serialize/1` unchanged, and `Jason.encode/1` handles the string keys without a problem when converting back to a JSON string to be sent back in the response.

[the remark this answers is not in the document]
hannahpurcell: I see! I'll roll with this as acceptable, then, unless you think otherwise

PR #1450, opened 2022-08-19
tweak: Add clock icon to Shuttle Bus widget

cmaddox5:
**Asana task**: ad-hoc

Adds the clock icon to the widget. Also fixes some console errors when rendering the new SVGs.

- [ ] Tests added?


hannahpurcell (dismissed):
This seems to have reverted the config update to using `minutes_range_to_destination_schedule` instead of `minutes_range_to_destination`

cmaddox5:
> This seems to have reverted the config update to using `minutes_range_to_destination_schedule` instead of `minutes_range_to_destination`

The schedule is resolved in the backend, so the front end only receives a single `minutes_range_to_destination`.

hannahpurcell:
> The schedule is resolved in the backend, so the front end only receives a single `minutes_range_to_destination`.

Got it, thanks

hannahpurcell (commented):
Can I lift my "requested changes" by just leaving a comment? (answer: no)

jzimbel-mbta (approved):

PR #1453, opened 2022-08-19
CR widget audio

hannahpurcell:
Also: shortened departures on the backend instead. 
Also: Conditionally show the "via headsign" message on the frontend visuals.

### DIFF
--- a/lib/screens/v2/widget_instance/cr_departures.ex
+++ b/lib/screens/v2/widget_instance/cr_departures.ex
@@ -25,10 +25,10 @@ defmodule Screens.V2.WidgetInstance.CRDepartures do
         }) do
       %{
         departures:
-          Enum.map(
-            departures_data,
-            &CRDeparturesWidget.serialize_departure(&1, config.wayfinding_arrows)
-          ),
+          departures_data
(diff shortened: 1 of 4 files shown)


lib/screens/v2/widget_instance/cr_departures.ex
```diff
@@ -47,10 +47,10 @@ defmodule Screens.V2.WidgetInstance.CRDepartures do
...
       %{
         departures:
-          Enum.map(
-            departures_data,
-            &CRDeparturesWidget.serialize_departure(&1, config.wayfinding_arrows)
-          ),
+          departures_data
+          |> Enum.map(&CRDeparturesWidget.serialize_departure(&1, config.wayfinding_arrows))
+          |> Enum.slice(0..2),
+        show_via_headsigns_message: config.show_via_headsigns_message,
```
cmaddox5, line 53: Instead of doing this twice, `audio_serialize` could just use `serialize`. Looks like they both return the same thing.

cmaddox5 (approved):
Looks great! One up-to-you suggestion.

PR #1454, opened 2022-08-19
Headsign pages

hannahpurcell:
Arbitrarily chose every 2 seconds


assets/src/components/v2/cr_departures.tsx
```diff
@@ -1,20 +1,78 @@
...
+          </td>
+          <td className="arrival-column"></td>
+          <td className="track-column">
+            <div className="table-header__english">Track</div>
+            <div className="table-header__spanish">Pista</div>
+          </td>
+        </tr>
+        { departures.slice(0, 3).map((departure, i) => {
+            return (
+              <tr key={i}>
```
jzimbel-mbta, line 52: Does `departure` have an ID field, or any way to identify it uniquely by its values, in a way that we can track it across data refreshes? Please use that as the `key` instead. You may need to add the departure ID to the values sent back from the server by `serialize/1`. Using array indices as `key`s causes issues/incorrect updates when the array contents change.

assets/src/components/v2/cr_departures.tsx
```diff
@@ -1,20 +1,78 @@
...
 }
 
+const DeparturesTable: React.ComponentType<any> = (props) => {
+  const {departures} = props
+
+  let [headsignPageOne, setHeadsignPageOne] = useState(true);
+
+  useInterval(() => {
+    setHeadsignPageOne(!headsignPageOne);
+  }, 2000);
```
jzimbel-mbta, line 33: [Checked with Mindy and co](https://mbta.slack.com/archives/C039ARUM48H/p1660939796076159) on this! Let's go with every 4 seconds.
```suggestion
  }, 4000);
```

jzimbel-mbta (approved):
If possible try to use a more reliable `key` value for the departures! Looks good other than that. Approved

lib/screens/v2/departure.ex
```diff
@@ -190,6 +190,14 @@ defmodule Screens.V2.Departure do
...
   def track_number(_), do: nil
 
+  def prediction_or_schedule_id(%__MODULE__{prediction: %Prediction{id: id}}) do
+    id
+  end
+
+  def prediction_or_schedule_id(%__MODULE__{schedule: %Schedule{id: id}}) do
+    id
+  end
+
```
handorff, line 200: You can just use `Departure.id` here.

handorff (approved):

PR #1474, opened 2022-08-22
fix: Alerts CandidateGenerator function for nil stop_sequences 

cmaddox5:
**Asana task**: ad-hoc

At the moment, the API is having issues with the E branch so our alerts code is failing when getting the stop_sequences for E branch stations. Adding this function clause will allow us to get local alerts even when we cannot get downstream stops from the stop_sequences.

- [ ] Tests added?

### DIFF
--- a/lib/screens/v2/candidate_generator/widgets/alerts.ex
+++ b/lib/screens/v2/candidate_generator/widgets/alerts.ex
@@ -82,6 +82,10 @@ defmodule Screens.V2.CandidateGenerator.Widgets.Alerts do
     |> Enum.to_list()
   end
 
+  defp local_and_downstream_stop_ids(nil, home_stop) do
+    [home_stop]
+  end
+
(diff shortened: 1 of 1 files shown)


hannahpurcell (approved):
Great, thanks!

The stuff going on with the API is intentional (certain route patterns have a different typicality value than usual for ✨ suspension reasons ✨ ) which makes me wonder if there are other areas in our app that we'll need to accommodate this E-branch weirdness. But this is the right solution for the error!

lib/screens/v2/candidate_generator/widgets/alerts.ex
```diff
@@ -82,6 +82,10 @@ defmodule Screens.V2.CandidateGenerator.Widgets.Alerts do
     |> Enum.to_list()
   end
 
+  defp local_and_downstream_stop_ids(nil, home_stop) do
+    [home_stop]
```
handorff, line 86: Hey this is just a test, did you get this?

PR #1481, opened 2022-08-25
Custom scrollbar on screen sims

hannahpurcell:
**Asana task**: [📺▶️ : 🇵🇱 | Scrollbar in simulations](https://app.asana.com/0/1185117109217413/1202736912335100)

At first, using `margin-left: auto` on the simulation__full-page and `margin-right: auto` on simulation__flex-zone looked to work beautifully, but that margin-right didn't translate to Screenplay through the iframe. This was the result:

Screen Shot 2022-08-24 at 2 24 38 PM

So had to get a little fancier with nested containers:  the outer container handles centering; the inner container handles the scroll.  This was applied to all "wide" screen types:  Prefare, Bus Shelter, V2 E-inks, DUPs. The design only shows scrollbars for screens with flex zones, but even the e-inks and DUPs could render weird if the user zooms in even slightly.

Screen Shot 2022-08-25 at 10 45 55 AM
Screen Shot 2022-08-25 at 10 45 11 AM

### DIFF
--- a/assets/css/dup/dup_screen_page.scss
+++ b/assets/css/dup/dup_screen_page.scss
@@ -10,6 +10,8 @@
 }
 
 .simulation-screen-scrolling-container {
+  scrollbar-color: #607180 #2E3F4D;
+
   &::-webkit-scrollbar {
     width: 12px;
   }
--- a/assets/css/dup/dup_screen_page.scss
+++ b/assets/css/dup/dup_screen_page.scss
@@ -3,9 +3,32 @@
   grid-template-columns: 1fr 1fr 1fr;
 }
 
-.simulation-page {
+.simulation-screen-centering-container {
+  display: flex;
+  justify-content: center;
   background: #3d5366;
+}
+
+.simulation-screen-scrolling-container {
+  &::-webkit-scrollbar {
+    width: 12px;
+  }
+  &::-webkit-scrollbar-thumb {
+    background-color: #607180;
+    border-radius: 12px;
+    border: 2px solid #2E3F4D;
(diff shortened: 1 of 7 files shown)


assets/css/dup/dup_screen_page.scss
```diff
@@ -3,9 +3,32 @@
...
 
-.simulation-page {
+.simulation-screen-centering-container {
+  display: flex;
+  justify-content: center;
   background: #3d5366;
+}
+
+.simulation-screen-scrolling-container {
+  &::-webkit-scrollbar {
```
cmaddox5, line 13: I think we need to be careful styling browser specific CSS because it won't necessarily work in all browsers. For example, `webkit` is only guaranteed to work for Chrome and Safari. I use Firefox as my browser and none of these styles are showing up.

cmaddox5: I will say that the scrollbar looks great when using Chrome! And `::webkit-scrollbar` is supported by all browsers except Firefox. After some quick googling, there doesn't seem to be a way to support this on FF, so we may just leave it alone for that one browser. Will leave that up to you and design.

hannahpurcell: Turns out that Paul H interpreted from the mocks that "there should be a scrollbar" and not "there should be a custom scrollbar". I didn't notice that on the Asana ticket until after I already customized the scrollbar 😅  So I checked with him in Slack, and he says this is fine

cmaddox5 (approved):
Looks good!

PR #1483, opened 2022-08-26
Include the GLX in trunk

hannahpurcell:
**Asana task**: adhoc

For now, will the whole norther part of the GL is all in-line, PIOs can create alerts that are non-branch-specific, and we can show the endpoints for that alert. For example, shuttling on the D and E starting at Govt Center and northward? We can and show endpoints for that, as if Science Park, Lechmere, and Union Sq are all a part of the GL "trunk" (even though they kinda aren't). When Medford branch pops up, then we'll need to remove these stops.

- [ ] Tests added?

### DIFF
--- a/lib/screens/stops/stop.ex
+++ b/lib/screens/stops/stop.ex
@@ -183,6 +183,10 @@ defmodule Screens.Stops.Stop do
   ]
 
   @green_line_trunk_stops [
+    # These 3 eventually will NOT be trunk stops, but are until Medford opens
+    {"place-unsqu", {"Union Square", "Union Sq"}},
+    {"place-lech", {"Lechmere", "Lechmere"}},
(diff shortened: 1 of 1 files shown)


lib/screens/stops/stop.ex
```diff
@@ -298,6 +302,8 @@ defmodule Screens.Stops.Stop do
   """
   def get_stop_sequence(informed_entities, route_id) do
     stop_sequences = Map.get(@route_stop_sequences, route_id)
+    |> IO.inspect(label: "stop sequences")
+    IO.inspect(informed_entities, label: "informed entities")
```
hannahpurcell, line 306:
```suggestion
```

cmaddox5 (approved):

PR #1484, opened 2022-08-30
tweak: Add clock to eInk no data state.

cmaddox5:
**Asana task**: [[E-Ink] Add a clock to full screen fallback image](https://app.asana.com/0/1185117109217413/1202593548281747/f)

This should be all places that were missing a clock. `OvernightDepartures` components in v1 and v2 and the v1 `FullScreenTakeover` component already had it. 

- [ ] Tests added?


hannahpurcell (approved):
🎉

PR #1487, opened 2022-08-31
feat: Add DeparturesNoData component to v2 eink

cmaddox5:
**Asana task**: [[E-ink] Add DeparturesNoData component to eInk front-end apps](https://app.asana.com/0/1185117109217413/1202131702864644/f)

Based on the [these designs](https://share.goabstract.com/5961689f-1714-4a37-bcfc-bee55d876aa7?collectionLayerId=42bf6e24-c328-4146-95ec-e1d579e765d1&mode=design). I needed to add to the GlEink template to accommodate a top body takeover.

- [ ] Tests added?

### DIFF
--- a/assets/css/v2/eink/departures_no_data.scss
+++ b/assets/css/v2/eink/departures_no_data.scss
@@ -1,0 +1,58 @@
+.departures-no-data-container {
+  width: 1200px;
+  height: 1312px;
+}
+
+.departures-no-data__main-content {
+  margin-bottom: 53px;
+  margin-left: 208px;
+}
+
+.departures-no-data__main-content__no-connection-icon-container {
+  width: 236px;
+  height: 236px;
+  margin-bottom: 48px;
+  margin-top: 111px;
+}
+
+.departures-no-data__main-content__heading {
+  width: 770px;
+  height: 448px;
+  color: rgb(0, 0, 0);
+  font-size: 104px;
+  font-family: Inter, sans-serif;
+  font-weight: bold;
+  letter-spacing: 0px;
+  line-height: 112px;
+}
+
+.departures-no-data__hairline {
+  content: "";
+  background: #171f26;
+  width: 784px;
+  height: 1px;
+  margin-left: 208px;
+  margin-top: 53px;
+}
+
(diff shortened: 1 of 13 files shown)


assets/css/v2/eink/departures_no_data.scss
```diff
@@ -0,0 +1,58 @@
...
+  font-size: 104px;
+  font-family: Inter, sans-serif;
+  font-weight: bold;
+  letter-spacing: 0px;
+  line-height: 112px;
+}
+
+.departures-no-data__hairline {
+  content: "";
+  background: #171f26;
```
jzimbel-mbta, line 31: Hmm, I thought for e-ink we only used black/white/grayscale. Could you check with Mindy that this is definitely the color we want?

assets/src/components/v2/eink/departures_no_data.tsx
```diff
@@ -0,0 +1,45 @@
+import React, { ComponentType } from "react";
+import NoConnection from "Components/v2/bundled_svg/no_connection";
+
+interface Props {
+  show_alternatives: boolean;
+  stop_id: string;
+}
+
+const coolBlack = "#171F26";
```
jzimbel-mbta, line 9: Same Q about allowable e-ink colors here

jzimbel-mbta (approved):
Approved assuming the non-grayscale colors are ok!

assets/src/components/v2/eink/departures_no_data.tsx
cmaddox5: Went back and double-checked the designs and this needs to be solid black `#000000`. Got that changed.

assets/css/v2/eink/departures_no_data.scss
cmaddox5: I'll double-check with design before merging. E-Ink NoData state also uses this same color for the divider, so I'm thinking it's ok. But they might also both be incorrect.

cmaddox5: Great callout on this. Checked with Mindy and she agreed that these should be solid black. Went ahead and did this for `NoData` as well.

PR #1489, opened 2022-08-31
Cr departures: Enum.sort_by to look at DateTime

hannahpurcell:
Why didn't we update this sort_by earlier? Was that a tech debt task? I don't see how this would adversely impact the departures on other screens, but also I don't know why we didn't see this bug earlier!


cmaddox5 (approved):
Great find! I know some months back, there was a tech debt task to go through the project and fix Enum functions using DateTime objects, but I guess this one was missed.

hannahpurcell:
Thanks! Yeah, figured out why this wasn't caught earlier -- this function is only run if we set "include_schedules" to true, and it's usually false by default. We really only use predictions AND schedules for CR.

PR #1490, opened 2022-09-01
tweak: CR Widget updates

cmaddox5:
**Asana task**: [Implement updates to Commuter Rail widget](https://app.asana.com/0/1185117109217413/1202872007662065/f)

Lots of design updates to frontend and a couple of small backend updates. There will be two new config updates for the overnight image: one for Sunday-Thursday, one for Friday and Saturday. Another idea I had for this was to ditch the config values and just hardcode the image name and append `weekday` or `weekend` to it. The config route is more traditional, but if the other route is preferable, let me know.

Screen Shot 2022-09-01 at 8 07 27 AM

- [ ] Tests added?

PR #1491, opened 2022-09-01
tweak: only show arrow for CR at Forest Hills

cmaddox5:
**https://github.com/mbta/screens/pull/1490 must be merged before this!**

**Asana task**: [CR Widget: No track number at Forest Hills](https://app.asana.com/0/1185117109217413/1202901200468405/f)

This will also require a config change for Forest Hills screens. `"wayfinding_arrows": {
    "e": [
      1,
      2,
      3,
      4,
      5
    ]
  }` will need to be changed to `"wayfinding_arrows": "e"`. With that change and `track_number` always being `nil`, only the arrow will be displayed.

Screen Shot 2022-09-01 at 8 33 47 AM

- [ ] Tests added?


lib/screens/v2/widget_instance/cr_departures.ex
```diff
@@ -194,4 +195,7 @@ defmodule Screens.V2.WidgetInstance.CRDepartures do
         []
     end
   end
+
+  defp get_track_number(_, "place-forhl"), do: nil
```
hannahpurcell, line 199: On the fence about hardcoding vs config. We can leave it for now, and If we get more stations like this in the future, we can re-assess. Let's go ahead and add a comment, though.

assets/src/components/v2/cr_departures.tsx
```diff
@@ -168,7 +168,7 @@ const CRDepartures: React.ComponentType<CRDeparturesProps> = (props) => {
             </div>
             <div className="departures-card__footer-ride-free">
               <div className="departures-card__footer-english ride-free">
-                Show your CharlieCard or CharlieTicket to ride at no charge
+                Show your Charlie Card or Charlie Ticket to ride free of charge
```
hannahpurcell, line 171: Needs to be reverted.

hannahpurcell (approved):

PR #1492, opened 2022-09-01
fix: Shuttle Bus time schedules

cmaddox5:
**Asana task**: [Shuttle Bus Widget: Our travel time estimates for peak / non-peak are not showing up at the scheduled time](https://app.asana.com/0/1185117109217413/1202901200468413/f)

Slight timezone shifting issue. We need to find the time for the service day, so the day we look at is correct. The time we look at should be local time though. Just need to have two separate shifts for date and time.

- [ ] Tests added?

### DIFF
--- a/lib/screens/v2/widget_instance/shuttle_bus_info.ex
+++ b/lib/screens/v2/widget_instance/shuttle_bus_info.ex
@@ -75,7 +75,9 @@ defmodule Screens.V2.WidgetInstance.ShuttleBusInfo do
   def audio_view(_instance), do: ScreensWeb.V2.Audio.ShuttleBusInfoView
 
   defp get_minute_range(schedule, now) do
+    # Shift to find current service day. i.e. 1AM Monday is actually still in the Sunday service day.
     {:ok, service_day_now} = DateTime.shift_zone(now, "America/Los_Angeles")
+    # Shift the time to local time.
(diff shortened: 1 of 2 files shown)


lib/screens/v2/widget_instance/shuttle_bus_info.ex
```diff
@@ -75,7 +75,10 @@ defmodule Screens.V2.WidgetInstance.ShuttleBusInfo do
   def audio_view(_instance), do: ScreensWeb.V2.Audio.ShuttleBusInfoView
 
   defp get_minute_range(schedule, now) do
-    {:ok, now} = DateTime.shift_zone(now, "America/Los_Angeles")
+    # Shift to find current service day. i.e. 1AM Monday is actually still in the Sunday service day.
+    {:ok, service_day_now} = DateTime.shift_zone(now, "America/Los_Angeles")
```
jzimbel-mbta, line 79: Does LA/California observe DST the same way as Massachusetts? :thinking:

I'm wondering if `DateTime.add(now, some_number, :hour)` might be more appropriate / correct under DST nonsense.

Would you mind writing a few tests to make sure this function works as expected under normal circumstances as well as around a DST shift?

jzimbel-mbta (approved):

jzimbel-mbta:
Created a ticket! https://app.asana.com/0/1185117109217422/1202933006890213/f

PR #1498, opened 2022-09-07
Property test for reconstructed alerts

hannahpurcell:
**Asana task**: [Create integration tests for reconstructed alert widget](https://app.asana.com/0/1185117109217413/1202502026744249)

This test uses StreamData value generators to randomly pair stop ID and alert (from a tiny subset of 3 alerts) and see if that combo serializes properly. If we find alerts down the road that break screens, that alert should be added to this test.

- [ ] Tests added?


lib/screens/stops/stop.ex
```diff
@@ -339,4 +354,13 @@ defmodule Screens.Stops.Stop do
     |> Enum.uniq()
     |> Enum.into(%{})
   end
+
+  def get_all_routes_stop_sequence() do
+    @route_stop_sequences
+  end
+
+  def get_all_subway_stops() do
```
cmaddox5, line 362: Are duplicates something that should be taken out of this list? For example, each GL trunk stop is present in at least one branch, so those will appear more than once in the result of this function.

lib/screens/util.ex
```diff
@@ -148,4 +148,18 @@ defmodule Screens.Util do
       Time.compare(start_time, t) in [:lt, :eq] or Time.compare(stop_time, t) == :gt
     end
   end
+
+  @doc """
+  Takes route id, returns route type. 
+  Not the favorite way of getting route type, but useful for property testing.
+  """
+  @spec route_type_from_id(String.t()) :: atom()
+  def route_type_from_id("Green-" <> _), do: :light_rail
```
cmaddox5, line 157: I believe `Green` without a branch letter is also valid. So you may want to remove the `-` here to catch those as well.

```suggestion
  def route_type_from_id("Green" <> _), do: :light_rail
```

cmaddox5 (approved):
This is awesome! I have a couple of comments, but the meat of this is good to go. 

Side question: do you have a recommended plan in mind for adding to this? This seems like the sort of thing that will grow over time. Will it be more of adding alerts that break screens in production, or is there a more proactive approach we can take?

lib/screens/stops/stop.ex
hannahpurcell: Oh lol, I added this function but then ultimately didn't use it. Gotta get rid of it!

hannahpurcell:
> or is there a more proactive approach we can take?

Paul H and I spitballed what we could do to be more proactive -- things like testing known alerts ahead of time, etc. But that may end up being a lot of work to test-ahead and add to the alerts list, and unplanned alerts will still fall through the cracks. So unfortunately, I can't think of anything other than a reactive approach.

PR #1505, opened 2022-09-12
tweak: CR Departures prediction vs. schedule visuals

cmaddox5:
**Asana task**: [CR Widget: train arrivals are not quite right ](https://app.asana.com/0/1185117109217413/1202901200468415/f)

Changed the visuals for CR departures depending on if a valuable prediction is available and if there are any delays. See task and designs for logic.

Screen Shot 2022-09-12 at 11 10 35 AM

- [ ] Tests added?


PaulJKim (approved):
Updated logic seems to match up with what was discussed in Slack. LGTM!

PR #1511, opened 2022-09-15
fix: CR prediction departure time when time is a string.

cmaddox5:
**Asana task**: ad-hoc

When a prediction is valuable (exists and has an assigned vehicle) and it is more than 60 minutes out, we show the clock time instead of a timestamp. Existing logic assumed all predictions are timestamps. Adding a conditional type check helps us determine what to show.

- [ ] Tests added?


PaulJKim (approved):
🚢 🇮🇹

PR #1514, opened 2022-09-22
doc: Implementation plan for screens-by-alert data source

jzimbel-mbta:
**Asana task**: [[Screenplay] Filter alerts by visibility on screens](https://app.asana.com/0/1185117109217413/1202932249366378/f)

[View the tech spec here.](https://github.com/mbta/screens/blob/jz/screens-by-alert-tech-spec/docs/tech_specs/screens_by_alert/tech_spec.md)
[View the initial proposal here.](https://github.com/mbta/screens/blob/jz/screens-by-alert-tech-spec/docs/tech_specs/screens_by_alert/proposal.md)


jzimbel-mbta:
Notes from meeting with @paulswartz on this:
- the general approach looks good
- normally would not recommend self-refreshes/"pretend requests", but given the product needs of Screenplay it makes sense
- recommends keeping the cached values in memcached or redis (memcached is probably enough for our needs) to avoid the issues highlighted in the doc re: multiple server instances keeping separate state in memory
- recommends implementing TTL as "last_updated" timestamps. Server will periodically read the cache data and do self-refreshes for any values that are past a given age
  - it's ok for some duplicated work to happen here. We hopefully do not have too many broken/malfunctioning screens at any given moment.

cmaddox5 (approved):
This looks great! The amount of detail you provided is awesome and will be so helpful come time to implement!

docs/tech_specs/screens_by_alert/tech_spec.md
```diff
@@ -0,0 +1,187 @@
...
+@type timestamped_screen_id :: {screen_id, timestamp}
+
+@type cache :: %{
+  "screens_by_alert." <> alert_id => list(timestamped_screen_id),
+  # metadata to make the "self-refresh" mechanism possible
+  "screens_last_updated." <> screen_id => timestamp
+}
+```
+
+TTL for `screens_by_alert.*` cache keys will be a number just over the "self-refresh" interval, so that in the case of a screen client going down we don't inaccurately report it as showing no alerts for a short period before the "self-refresh" mechanism kicks in.
```
hannahpurcell, line 30: `TTL for 'screens_by_alert.*' cache keys will be a number just over the "self-refresh" interval`: could you explain this idea in different words? I can't get these terms to make sense in my mind. In the typespec for the cache, it seems that the "cache key" for `screens_by_alert.*` is literally that string plus the alert id. How is TTL attached to that?

docs/tech_specs/screens_by_alert/tech_spec.md
```diff
@@ -0,0 +1,187 @@
...
+
+TTL for `screens_last_updated.*` keys will be something large, like 1 day. These values should only expire when a screen is deleted from the config and
+the self-refresh logic stops caring about it.
```
hannahpurcell, line 33: TTL: I first assumed that this was the drop-dead expiration of a value, but really it seems like this is the "time until we run some logic to determine whether the value should expire". Is that right?

docs/tech_specs/screens_by_alert/tech_spec.md
jzimbel-mbta: Thanks for the question, I could have been clearer on this!

In this case, **TTL will be handled internally by memcached** and isn't explicitly part of the data structure.

Every time you set or update a key's value in memcached, you can also tell it what that key/value pair's TTL is. See, for example, the `set` [function](https://hexdocs.pm/memcachex/Memcache.html#set/4) made available by the `memcachex` library—it accepts a `:ttl` option to specify TTL. Once that specified amount of time elapses, the key+value will automatically expire from the cache.

Our case is a little tricky though, because we also need to implement some separate TTLs that we handle in our own code—memcached's TTL feature only applies to whole key/value pairs, but we need to also apply TTLs to individual values under each alert ID key.

Assuming we use 60 seconds as the "self-refresh" interval:
```ex
# An example key/value pair in the cache.
# When we set/update this key, we set its TTL to 70 seconds.
# If this key's value isn't updated again within 70 seconds, it will automatically be removed by memcached.
"screens_by_alert.123" => [
  {"PRE-101", 100},
  # ^ One of the values in the list stored under this key.
  # The first tuple element is a screen ID, the second is a timestamp produced by `System.monotonic_time()`.
  # Every time we update this list, we remove any element whose timestamp is from more than 70 seconds ago.
  {"EIG-123", 200},
  {"BUS-456", 300}
]
```

docs/tech_specs/screens_by_alert/tech_spec.md
jzimbel-mbta: In our case it's kind of both unfortunately 😬

In the text you highlighted for this comment, it's the former ("the drop-dead expiration of a value"). We'll specify the TTL when updating these keys via memcached commands, and memcached will carry out their expirations accordingly.

For other values in the cache, specifically each screen ID in the list under an alert key, _we_ need to implement the TTL ourselves because memcached is totally agnostic of the internal structure of data you store under a key. So in that case, "we run some logic to determine whether the value should expire" is correct.

PaulJKim: A minor thing, but does the TTL need to be very large? If a screen is deleted from the config, then it seems like the self-refresh GenServer would realize that in a minute or so at which point it would stop attempting self-refreshes. It seems like the `screen_last_updated` key could be dropped immediately or soon after that.

cmaddox5: I think it's possible to cover that base while keeping the TTL large. Because the TTL is a "check if we need to self-refresh" timer, we could do a check against `State.screens()` to check that the screen still exists. If it's still there, do a self-refresh if we are past whatever the "TTL" is. 

(An out-of-scope observation, we might also be able to use this to clean up `places_and_screens.json`)

docs/tech_specs/screens_by_alert/tech_spec.md
hannahpurcell: I understand the first 3 paragraphs. ✅ 

Half of the 4th paragraph makes sense -- it pairs with your comment to my other question. We have a TTL we provide to memcached, and that time limit will immediately expire a key-value after it has had the opportunity to run the self-refresh code. The definition of "self-refresh" is kinda buried in this doc, but in essence: for each screen ID, we'll look at the cache to see if the screen has recently updated the cache itself (indicated by a very recent timestamp) and if the screen has NOT updated itself (indicating a screen outage), we trigger a run of the widget framework for that screen to force an update of the cache with realtime alert status. 

So, after this self-refresh has had a chance to occur, we'll see one of two things: either the screen affirmed that it still has an alert, or it claims it doesn't have that alert anymore. If the screen still has the alert, the key-value stored in memcached will have been updated (restarting the TTL clock and the self-refresh clock). If the screen no longer has that alert, the key-value timestamp (within memcached) will be unchanged, and then the TTL will expire that key-value.

How is this translating so far?

docs/tech_specs/screens_by_alert/tech_spec.md
hannahpurcell: > Because the TTL is a "check if we need to self-refresh" timer

🤔 Hold up... I think the "large TTL" referenced here in the doc is one thing (the timeline by which memecached will expire un-updated screens) and the "check if we need to self-refresh" is a separate timer that we manage in a GenServer. Right? So Paul K's suggestion makes sense to me -- the `screen_last_updated` TTL can be a couple minutes, since the screen probably would have refreshed a few times by then.

jzimbel-mbta: Yeah, I think you're right! I left a longer gap mostly because I don't yet have a good sense of how long the self-refresh will take**, on average, and because the GenServer will already ignore these keys anyway. The TTL is there just to have them stop taking up space in the cache after a while.

**TBH it will probably be decently fast—the prod server responds to screen data requests in about 100-200ms, and I'm sure the actual computation time is significantly smaller than that. Multiply that by how many screen clients we expect to be down at a given moment, on average—like 10 max? And that's roughly 1-2 seconds of computation time.

docs/tech_specs/screens_by_alert/tech_spec.md
jzimbel-mbta: > If the screen still has the alert, the key-value stored in memcached will have been updated (restarting the TTL clock and the self-refresh clock).

Correct

> If the screen no longer has that alert, the key-value timestamp (within memcached) will be unchanged, and then the TTL will expire that key-value.

This is also correct, but I just want to clarify two things:
1. If the screen no longer has that alert, then the cache key for that alert won't get updated. However, other screens might be showing it and thus updating the cache key, so it won't necessarily expire just because this one update no longer included it.
2. The self-refresh clock for this screen will also be updated in this case, just like in the first case where the screen still has the alert. Every screen data update that gets processed (whether from an external request, or from the self-refresh mechanism) will cause the `screens_last_updated.<id>` key for that screen to get updated with a new timestamp value.

docs/tech_specs/screens_by_alert/tech_spec.md
PaulJKim: hmm so there's little to no risk of having old data in the cache then right? Initially I was following a train of thought that led me to wonder if it's plausible that future alerts-related logic in Screenplay would still think that a screen exists even though the config for it was deleted. 

I think it should be fine though if I'm understanding the spec correctly since the deleted screen_id would naturally be removed from the alert_id's list by virtue of the timestamps expiring. Does that check out @jzimbel-mbta ? 🤔 

Ohh and maybe this concern is more something that would be addressed by eventually not having to manually update the places and screens list anymore

docs/tech_specs/screens_by_alert/tech_spec.md
hannahpurcell: Thanks! Here's an updated translation, picking up halfway through my last translation, with new questions:

> If the screen no longer has that alert...

... then the best that this screen can do is say “I have these alerts: []”… Can we somehow compare that list of alerts with the cache to more quickly break the relationship between the alert and screen?

If not, then the alert will still seem to be on that screen while other stuff happens…

If other screens still have the alert, they will one-by-one affirm that they should still be in the screens array for that alert in the cache. Plus, they will also be updating the key-value timestamp for that alert in memcached. If this is the case… how does the screen-without-the-alert remove itself from its place in the cached array?

On the other hand, if other screens one by one indicate they, too, no longer have the alert, the key-value will get staler and staler until it’s clear that the alert has been removed from all screens and should be expired. Then the TTL in memcached will kick in to expire that key-value pair.

jzimbel-mbta: > Can we somehow compare that list of alerts with the cache to more quickly break the relationship between the alert and screen?

Nope, not without having a lot more cache operations per update! That would essentially require us to turn it back into keeping a full representation of a bipartite graph in the cache, with the associated increase in potential race conditions/excessive retries of cache updates. The way we avoid this extra work is by using TTLs and never comparing a previous list of reported alerts to the new list of reported alerts.

> how does the screen-without-the-alert remove itself from its place in the cached array?
Whenever we make any update to the list under an alert key, we first get the existing list and remove any screen IDs that have expired (based on the timestamps they're paired with).

Last paragraph is correct. If all screens stop showing an alert (probably because it's been cleared in alerts-ui) then it and its associated list of screen IDs eventually expires from the cache entirely. Screenplay won't need to wait for the cached value to expire though, if the alert was cleared, because it also checks alert data in the V3 API. It will know that the alert is gone, and simply stop asking for info about it, before the cached data expires.

docs/tech_specs/screens_by_alert/tech_spec.md
```diff
@@ -80,6 +64,37 @@ System.convert_time_unit(time2 - time1, :native, :second)
...
+### Timers and TTLs
+
+In addition to the actual structure of the data being stored in the cache, we also need to carefully choose various timings and cache TTLs to properly keep that data up to date.
+
+TTLs in our case are also somewhat tricky, because some of them will be carried out internally by memcached, and others will need to be carried out by us in Elixir code.
+
+**Timers**
+| timer | interval (sec) | Why? |
+| - | - | - |
+| "self-refresh" job-runner GenServer | 70 | Job should give the slowest screen clients (e-ink, 30 sec refresh rate) time to retry after 1 missing request. |
```
hannahpurcell, line 78: So this runner executes every 70s (or whatever interval) non-conditionally? Also, is 30s the slowest refresh rate? Can the interval be 35s, then? Does it benefit us to self-refresh more or less frequently?

docs/tech_specs/screens_by_alert/tech_spec.md
```diff
@@ -80,6 +64,37 @@ System.convert_time_unit(time2 - time1, :native, :second)
...
+
+**TTLs**
+| value | Handled by | TTL (sec) | Why? |
+| - | - | - | - |
+| `screens_by_alert.*` cache items | memcached | 75 seconds | A key should expire shortly after a self-refresh runs without updating that key. |
+| screen IDs listed under a `screens_by_alert.*` key | application logic | 75 seconds | A screen ID should expire from the list shortly after a self-refresh runs without reaffirming the ID's place in the list. |
```
hannahpurcell, line 84: > A screen ID should expire from the list shortly after a self-refresh runs without reaffirming the ID's place in the list.

Is this mechanism described in this doc? (Does it need to be?)

To help wrap my head around this, here's my scenario:  let's say an alert is bumped off a screen by a higher priority alert. So instead of having alerts A and B on Screen 1, it now has alerts A and C. The screen declares to the cache it has different alerts than it did before. Alert A still has Screen 1 in its list, Alert C now has Screen 1 in its list, and B still has it in its list but with an old timestamp.

in clunky pseudo:
%{
  "screens_by_alert.A” => [{Screen1, current time}, {ScreenY, current time}]
  "screens_by_alert.B” => [{Screen1, time from 30s ago}, {ScreenX, current time}]
  "screens_by_alert.C” => [{Screen1, current time}]
  “screens_last_updated.1” => current time
}

Kinda like that. Each alert is still getting reported by at least 1 screen, so we never hit the TTL for "screens_by_alert.*"

QUESTION:  What mechanism removes Screen1 from AlertB’s screen list?

jzimbel-mbta: Every time the list under an alert ID is updated, we also _remove_ any list elements whose timestamps have expired. In this instance, ScreenX is still showing alertB, so when its client next fetches data, the resulting cache update will put a new `{ScreenX, now}` tuple in the list under `screens_by_alert.B`, _and_ remove both the previous ScreenX entry and the Screen1 entry, since they're now expired.

In fact, we could also do this cleanup when Screenplay asks for cached data. Any time we `get` an alert ID cache key, it's an opportunity to do a subsequent `set` that clears expired screen IDs from the list.

**In the case of clients that refresh more frequently, like pre-fare's 20s rate, the list may occasionally contain duplicate screen IDs with non-expired timestamps. That's ok—it's simple to remove duplicates from the list in code.

- - -

Other ways the data reaches "eventual accuracy":

If the only screen(s) still showing an alert have broken clients for some reason, then the self-refresh job will cover them in keeping the list up to date.

If an alert is no longer being shown by any screens at all, but still exists in the API data, then the `screens_by_alert.<id>` cache key will eventually expire due to lack of updates.

If an alert is no longer being shown by any screens at all, and has been cleared from the API data, then Screenplay will know about that earlier since it also checks the API regularly. The `screens_by_alert.<id>` cache key will eventually expire due to lack of updates.

PR #1515, opened 2022-09-23
feat: Log sim/real_screen requests

cmaddox5:
**Asana task**: [Log screens simulation requests](https://app.asana.com/0/1185117109217413/1202766648515759/f)

Added a `source` to screen data request logs. This will mean that all screens with either `is_real_screen=true` or `source=screenplay` will log a data request. The new source will only be used in Screenplay sims for now. URLs do not need to change for this change to work. Existing URLs with `is_real_screen=true` will continue to produce logs, but will have an added `source=real_screen` for data request log lines. 

- [ ] Tests added?

### DIFF
--- a/lib/screens_web/controllers/screen_api_controller.ex
+++ b/lib/screens_web/controllers/screen_api_controller.ex
@@ -26,10 +26,11 @@ defmodule ScreensWeb.ScreenApiController do
     end
   end
 
-  def show(conn, %{"id" => screen_id, "last_refresh" => last_refresh}) do
+  def show(conn, %{"id" => screen_id, "last_refresh" => last_refresh} = params) do
     is_screen = ScreensWeb.UserAgent.is_screen_conn?(conn, screen_id)
 
-    _ = Screens.LogScreenData.log_data_request(screen_id, last_refresh, is_screen)
+    _ =
+      Screens.LogScreenData.log_data_request(screen_id, last_refresh, is_screen, params["source"])
 
     if nonexistent_screen?(screen_id) do
       Screens.LogScreenData.log_api_response(:nonexistent, screen_id, last_refresh, is_screen)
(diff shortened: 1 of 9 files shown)


lib/screens_web/controllers/screen_api_controller.ex
```diff
@@ -26,10 +26,11 @@ defmodule ScreensWeb.ScreenApiController do
     end
   end
 
-  def show(conn, %{"id" => screen_id, "last_refresh" => last_refresh}) do
+  def show(conn, %{"id" => screen_id, "last_refresh" => last_refresh} = params) do
```
PaulJKim, line 29: Minor thing, but any particular reason to hold onto `params` here instead it pattern matching out `source` like the other two fields here?

cmaddox5: Mostly to avoid making a second function for cases where there is no source at all. URLs set for live screens and pulling up a screen in the browser will not have this `source` a majority of the time. Accessing the param using `params["source"]` allows me to conditionally add the source to the logs if it was supplied. If not, it looks for `is_real_screen=true`. If that's also not there, it skips logging.

PaulJKim: Ah okay. So I see that there's a Screenplay PR to add the source param. And you mentioned that real, rider-facing screens typically won't have the source param at all. Is that basically how we'll filter requests in Splunk?

cmaddox5: Correct. We want to give Ana a way to filter out simulation logs in her Splunk queries while also giving us a way to see data requests from Screenplay users. And because we don't want to have to change all of the live URLs, the current URLs with `is_real_screen=true` will be translated to `source=real_screen`.

PaulJKim: Make sense!

PaulJKim (approved):

PR #1522, opened 2022-09-28
fix: Prevent crash when line map API fails

cmaddox5:
**Asana task**: [🐞MatchError: no match of right hand side value](https://app.asana.com/0/1185117109217413/1202561076967107/f)

Bug in the Sentry link pointed to an API error when getting data for the line map. Changed that function to use a `with/else` so we can allow line map to fail without making the screen go into a no-data state.

- [ ] Tests added?


PaulJKim (approved):
Nice! Thanks José Valim.

PR #1526, opened 2022-10-04
feat: Always use mbta.com/alerts as Alert widget CTA url

jzimbel-mbta:
**Asana task**: [Always use mbta.com/alerts as e-Ink alert widget URL](https://app.asana.com/0/1185117109217413/1202072393748976/f)

- [x] Tests added?

### DIFF
--- a/assets/src/components/v2/alert.tsx
+++ b/assets/src/components/v2/alert.tsx
@@ -11,7 +11,6 @@ interface Props {
   icon: AlertIcon;
   header: string;
   body: string;
-  url: string;
 }
 
 interface BaseAlertProps {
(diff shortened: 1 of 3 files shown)


hannahpurcell (dismissed):

hannahpurcell:
This approach seems to affect both E-ink and Bus Shelter alerts, but the ask was only for E-ink

jzimbel-mbta:
> This approach seems to affect both E-ink and Bus Shelter alerts, but the ask was only for E-ink

Ooo good catch. Will fix.

hannahpurcell (approved):
great! 👍 like the comment, too

PR #1528, opened 2022-10-07
feat: ScreensByAlert behavior and interface

cmaddox5:
**Asana task**: [(deployment envs) Write logic to interface with screens-by-alert caching layer](https://app.asana.com/0/1185117109217413/1203086969471675/f)

I kinda expanded the scope on this one, so sorry about that. Got a little carried away. 

This adds the needed modules for interfacing with either the GenServer (local dev) or Memcached (prod). It is done in such a way that only the interface is needed in the Supervision tree. The interface itself will figure out which cache module to use. All functions work but return dummy values for now. A dependency has also been added for the `memcachex` package. 

- [x] Tests added?


lib/screens/application.ex
```diff
@@ -37,4 +38,34 @@ defmodule Screens.Application do
...
+
+  ## Examples
+
+      iex> Screens.Application.config(ScreensByAlert.Memcache, :connection_opts)
+      [
+        namespace: "localhost:4000",
+        hostname: "localhost"
+      ]
+  """
+  def config(root_key, sub_key) do
```
jzimbel-mbta, line 53: Not a huuuuge fan of this custom convenience function, as it hides the fact that it calls `Application.fetch_env!` (which should only be called at runtime), while sometimes we want to call `Application.compile_env`, for example when we're storing some config in a module attribute.

[Application.compile_env](https://hexdocs.pm/elixir/Application.html#compile_env/3) also lets you provide a "path" to search for under a key, so this custom function shouldn't be necessary in that case.

cmaddox5: Ahh, I didn't think about that. Looked nifty so I gave it a shot. Ha. Changed it to just use the Elixir `Application` module.

config/releases.exs
```diff
@@ -64,6 +64,12 @@ config :sentry,
 config :ueberauth, Ueberauth.Strategy.Cognito, client_secret: cognito_client_secret
 config :screens, ScreensWeb.AuthManager, secret_key: screens_auth_secret
 
+config :screens, ScreensByAlert.Memcache,
+  connection_opts: [
+    namespace: System.get_env("HOST"),
+    hostname: System.get_env("MEMCACHED_HOST")
+  ]
+
```
PaulJKim, line 72: Will this work with the `@cache_module` module attribute? I thought `releases.exs` is a runtime config file but I believe module attributes are evaluated/expanded at compile time 🤔

lib/screens/screens_by_alert/behaviour.ex
```diff
@@ -0,0 +1,21 @@
...
+  - `get_screens_last_updated(screen_id)` returns the timestamp that represents the last time a given screen_id was updated.
+
+  """
+  @type screen_id :: String.t()
+  @type alert_id :: String.t()
+  @type timestamp :: integer()
+  @type timestamped_screen_id :: {screen_id, timestamp}
+
+  @callback start_link(Keyword.t()) :: {:ok, pid()}
+  @callback put_data(screen_id(), list(alert_id())) :: :ok
```
PaulJKim, line 18: Do we need to define a behavior for `put`-ing the `screen_last_updated` data as well, or is that handled elsewhere in the overall solution?

PaulJKim: Hmm so when the list is empty, we would let the cache expire naturally instead of updating the entry as an empty list ourselves? I think that's what the tech spec indicates so if that's right, then yep I think that makes sense! In that case, would it make sense to make the type spec for `put_data/2` a little more generic?

config/releases.exs
cmaddox5: `@cache_module` doesn't need this particular piece of the config. It only needs what is in `prod.exs` (which I just realized was missing 😬). `connection_opts` is used when the Memcached instance is booting up in the application's supervision tree, on [this line](https://github.com/mbta/screens/blob/298213606796724633bc393c5ff5beb173d6e852/lib/screens/screens_by_alert/memcache.ex#L12) specifically.

PaulJKim: Ah yeah woops, I must've misread something. Makes sense!

lib/screens/screens_by_alert/behaviour.ex
cmaddox5: Right, they would either expire naturally or the self-refresh GenServer would refresh/expire them for us. I don't think that we would want to add a screen to `screens_by_alert` unless it was actually showing alerts. 

As for the type spec, the way that it is in the tech spec is unfortunately the best we can do given when the call would occur. We would call `put_data` once per screen refresh, so in that instance we only know the list of alerts for a given screen. We will then have to essentially flip that from `screen_id => list(alert_id)` to `alert_id => list(screen_id)` in the cache module implementation.

PaulJKim: Hmm yeah I see. Otherwise we'd have to do an additional read _first_ for each alert to get what it currently has as it's list and then append the current screen's id to each of those lists, and then re-`put` them into memcached, which seems pretty inefficient.

So it seems then that we'll be flipping the mapping on read? Sorry, I'm sure this was already discussed at some point but it might've flown over my head or I just missed it.

cmaddox5: I think the first thing you said is what Jon proposed. In the spec, it says `New cache data from one screen will be distributed across several cache keys, since our updates come in as lists of 0 or more alerts keyed on a screen ID, but the cache is keyed on alert ID), updating TTL metadata at the same time.`. So I think the flip will happen in `put_data` and not on read. I agree that this normally would be super inefficient. But worst case, our screens only show 4 alerts at one time, and that worst case rarely happens. So I don't _think_ it will be too bad.

PaulJKim: Ohhh hmm gotcha. I suppose even if it's a little inefficient, it's still better than flipping the map on read now that I think about it. If we did that, I think we'd have to iterate through the entire cache while checking for the alert_id(s) on every request.

I guess my main concern then would just be readability since it might be a little obfuscating to have the function definition read as "put this list of alert ids for this screen id" when in reality, it's doing the reverse. But that's not a big deal I think as long as it's well documented

cmaddox5: That's a good point. I'll add some comments to the behaviour.

PaulJKim (approved):
LGTM!

PR #1536, opened 2022-10-17
feat: Add DUP rotation index to data request logs.

cmaddox5:
**Asana task**: [Specify in logs which DUP screen page we're loading](https://app.asana.com/0/1164574158255689/1203081402939552/f)

Added DUP rotation index to DUP data request logs. The term `page_number` was used in the ticket so I went with that instead of `rotation_index`. Happy to stay consistent if anyone feels strongly about it.

- [ ] Tests added?


hannahpurcell (approved):
👍

PR #1538, opened 2022-10-18
feat: disable_sentry query param

cmaddox5:
**Asana task**: [[Screens] New parameter `disable_sentry` for the 4 first-gen EIGs](https://app.asana.com/0/1164574158255689/1202958313810860/f)

Added option for `disable_sentry` param to be added to URL in order to prevent Sentry from starting in React. If a DSN is present AND it's a real screen AND the param is absent or false, Sentry will run like normal. If `disable_sentry` is true, Sentry will not initialize.

- [ ] Tests added?

### DIFF
--- a/assets/src/util/sentry.tsx
+++ b/assets/src/util/sentry.tsx
@@ -24,13 +24,6 @@ const initSentry = (appString: string) => {
     environmentName: env,
     disableSentry,
   } = getDataset();
-
-  console.log(disableSentry);
-
-  if (!disableSentry && isRealScreen()) {
-    console.log("HERE");
-  }
-
   // Note: passing an empty string as the DSN sets up a "no-op SDK" that captures errors and lets you call its methods,
   // but does not actually log anything to the Sentry service.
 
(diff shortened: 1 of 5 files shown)


assets/src/util/sentry.tsx
```diff
@@ -19,23 +19,34 @@ const error = (message: string) => log(message, "error");
...
+  const {
+    sentry: sentryDsn,
+    environmentName: env,
+    disableSentry,
+  } = getDataset();
+
+  console.log(disableSentry);
+
+  if (!disableSentry && isRealScreen()) {
+    console.log("HERE");
```
hannahpurcell, line 31: 2 things: remove console.logs and update the comment for this function to mention sentry

hannahpurcell (approved):
2 cosmetic fixes needed, otherwise good

assets/src/util/sentry.tsx
cmaddox5: 😬

PR #1541, opened 2022-10-25
feat: Add cache update to framework

cmaddox5:
**Asana task**: [Update cached alert => screens data on every v2 screen data request](https://app.asana.com/0/1185117109217413/1203111452890186/f)

Cache will now be updated automatically while a screen is open. Right now, I just have the list of widgets we should cache in a module attribute. I picked this mostly because there are not many right now, and adding to the list is super easy. Feel free to suggest another way to hold this info.

- [ ] Tests added?

### DIFF
--- a/lib/screens/v2/alert_widget_instance.ex
+++ b/lib/screens/v2/alert_widget_instance.ex
@@ -1,0 +1,7 @@
+defprotocol Screens.V2.AlertWidgetInstance do
+  @type alert_id :: String.t()
+
+  @doc "Gets the ID of the alert that this widget shows."
+  @spec alert_id(t) :: alert_id()
+  def alert_id(widget)
+end
--- a/lib/screens/v2/screen_data.ex
+++ b/lib/screens/v2/screen_data.ex
@@ -472,4 +472,21 @@ defmodule Screens.V2.ScreenData do
       Enum.find_value(children, &get_containing_slot(&1, target_slot_ids))
     end
   end
+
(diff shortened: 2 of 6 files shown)


jzimbel-mbta (approved):
Looks good to go as-is, just some minor suggestions.

As for filtering to alert widgets and getting their alert IDs, I wonder if it would be worth it to define a new protocol for this? Or would that be total overkill :thinking:

Something like
```ex
defprotocol Screens.V2.AlertWidgetInstance do
  @doc "Gets the ID of the alert that this widget shows."
  def alert_id(widget)
end
```

and then we could check if a widget is an alert widget with `not is_nil(AlertWidget.impl_for(widget_struct))`.

cmaddox5:
> Looks good to go as-is, just some minor suggestions.
> 
> As for filtering to alert widgets and getting their alert IDs, I wonder if it would be worth it to define a new protocol for this? Or would that be total overkill 🤔
> 
> Something like
> 
> ```elixir
> defprotocol Screens.V2.AlertWidgetInstance do
>   @doc "Gets the ID of the alert that this widget shows."
>   def alert_id(widget)
> end
> ```
> 
> and then we could check if a widget is an alert widget with `not is_nil(AlertWidgetInstance.impl_for(widget_struct))`.

I think it may only seem overkill because there are only 2 potential widgets we can cache right now. But I think it's a bit cleaner to use the protocol in the long term if we end up adding a bunch more. I took that idea and ran with it and it works and looks great.

jzimbel-mbta (approved):
✅

PR #1542, opened 2022-10-25
feat: Screens-by-alert self-refresh job runner

jzimbel-mbta:
**Asana task**: [Implement screens-by-alert "self-refresh" mechanism](https://app.asana.com/0/1185117109217413/1203086969471679/f)

The job uses `screens_ttl_seconds` as its run interval.

On each run, it does the following:
1. Get a list of all v2 screen IDs in config that are not hidden from Screenplay,
2. Find which of those have not been updated in the last `screens_ttl_seconds` seconds,
3. Take up to the 20 oldest of those,
4. And run the widget logic on each, concurrently, to update the cache with new data.

I needed to make some changes to the existing screens-by-alert code to allow requesting info on multiple alert IDs, screen IDs, so that we don't overload the cache-interfacing process with messages on each job run, or make separate cache GETs for each ID.

I also added a new optional `opts` param to `V2.ScreenData.by_screen_id`, allowing us to skip the serialization step and save some computation when we don't care about the end result.

- [x] Tests added?
  - Ok actually I just updated existing tests, not entirely sure how to unit test this GenServer since it doesn't really take any inputs after being started

### DIFF
--- a/lib/screens/screens_by_alert/behaviour.ex
+++ b/lib/screens/screens_by_alert/behaviour.ex
@@ -12,7 +12,6 @@ defmodule Screens.ScreensByAlert.Behaviour do
   @type screen_id :: String.t()
   @type alert_id :: String.t()
   @type timestamp :: integer()
-  @type timestamped_screen_id :: {screen_id, timestamp}
 
   @doc """
   Starts the process that interfaces with the screens-by-alert cache.
--- a/lib/screens/screens_by_alert/behaviour.ex
+++ b/lib/screens/screens_by_alert/behaviour.ex
@@ -5,8 +5,8 @@ defmodule Screens.ScreensByAlert.Behaviour do
   - `start_link(opts)` is called to start the backend by the supervisor.
   - `put_data(screen_id, list(alert_id))` inserts the list of alert_ids visible on a given screen_id
     to the cache.
-  - `get_screens_by_alert(alert_id)` returns all screen_ids that are currently displaying the given alert_id.
-  - `get_screens_last_updated(screen_id)` returns the timestamp that represents the last time a given screen_id was updated.
+  - `get_screens_by_alert(alert_ids)` returns a map associating each requested alert_id with the screen IDs that are currently displaying it.
+  - `get_screens_last_updated(screen_ids)` returns a map associating each requested screen_id with its last-updated Unix timestamp.
 
   """
   @type screen_id :: String.t()
@@ -14,6 +14,9 @@ defmodule Screens.ScreensByAlert.Behaviour do
   @type timestamp :: integer()
   @type timestamped_screen_id :: {screen_id, timestamp}
 
+  @doc """
+  Starts the process that interfaces with the screens-by-alert cache.
+  """
   @callback start_link(Keyword.t()) :: {:ok, pid()}
 
   @doc """
@@ -23,6 +26,19 @@ defmodule Screens.ScreensByAlert.Behaviour do
   """
   @callback put_data(screen_id(), list(alert_id())) :: :ok
 
-  @callback get_screens_by_alert(alert_id()) :: list(timestamped_screen_id())
-  @callback get_screens_last_updated(screen_id()) :: timestamp() | nil
+  @doc """
+  Given a list of 0 or more alert IDs, returns a mapping from each requested alert ID
+  to the list of screens currently showing it.
+
+  If a cache item is missing for some alert, its screens-list value will default to [].
+  """
+  @callback get_screens_by_alert(list(alert_id())) :: %{alert_id() => list(screen_id())}
+
+  @doc """
+  Given a list of 0 or more screen IDs, returns a mapping from each requested screen ID
+  to its last-updated Unix timestamp.
+
+  If a cache item is missing for some screen, its timestamp value will default to 0.
+  """
+  @callback get_screens_last_updated(list(screen_id())) :: %{screen_id() => timestamp()}
 end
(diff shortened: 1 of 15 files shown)


lib/screens/screens_by_alert/behaviour.ex
```diff
@@ -5,15 +5,18 @@ defmodule Screens.ScreensByAlert.Behaviour do
...
-  - `get_screens_by_alert(alert_id)` returns all screen_ids that are currently displaying the given alert_id.
-  - `get_screens_last_updated(screen_id)` returns the timestamp that represents the last time a given screen_id was updated.
+  - `get_screens_by_alert(alert_ids)` returns a map associating each requested alert_id with the screen IDs that are currently displaying it.
+  - `get_screens_last_updated(screen_ids)` returns a map associating each requested screen_id with its last-updated Unix timestamp.
 
   """
   @type screen_id :: String.t()
   @type alert_id :: String.t()
   @type timestamp :: integer()
   @type timestamped_screen_id :: {screen_id, timestamp}
```
cmaddox5, line 15: Can this type be removed now that the return is a map?

jzimbel-mbta: Oh yeah, good point. The timestamps are an implementation detail at this layer. I'll clean that up.

lib/screens/screens_by_alert/self_refresh_runner.ex
```diff
@@ -0,0 +1,102 @@
...
+    #
+    # Doing the work in a separate, unlinked task process protects this GenServer
+    # process from going down if an exception is raised while running
+    # ScreenData.by_screen_id/1 for some screen.
+    for screen_id <- screen_ids_to_refresh do
+      _ =
+        Task.Supervisor.start_child(TaskSupervisor, fn ->
+          @screen_data_fn.(screen_id, skip_serialize: true)
+        end)
+    end
```
PaulJKim, line 91: Is a list comprehension being used mainly for readability purposes? If we want to perform an operation with each item in an enumerable but we also know that we don't care about the result, then that typically is a suitable use case for `Enum.each`, I think. Usually if we use a list comprehension, it's because we want the result back.

lib/screens/screens_by_alert/self_refresh_runner.ex
```diff
@@ -0,0 +1,102 @@
...
+    schedule_run()
+
+    now = System.system_time(:second)
+
+    {screen_ids_to_refresh, overflow} =
+      Screens.Config.State.v2_screens_visible_to_screenplay()
+      # get a mapping from each ID to its last updated time
+      |> ScreensByAlert.get_screens_last_updated()
+      # filter to outdated IDs
+      |> Enum.filter(fn {_screen_id, timestamp} -> now - timestamp > @screens_ttl_seconds end)
```
PaulJKim, line 64: I think you could make `Enum.filter` and `Enum.map` below into `Stream.filter` and `Stream.map`? Would save us two iterations through the map, but probably barely makes a difference performance wise 🤷

lib/screens/screens_by_alert/self_refresh_runner.ex
```diff
@@ -0,0 +1,102 @@
...
+
+    {:noreply, state}
+  end
+
+  defp schedule_run do
+    Process.send_after(self(), :run, @job_run_interval_ms)
+  end
+
+  @doc "A fake screen-data-fetching function to be used during tests, to avoid making requests."
+  def fake_screen_data_fn(_screen_id, _opts), do: nil
```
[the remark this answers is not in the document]
jzimbel-mbta: Yeah it might be... I was a little lazy with this 🫠

I'll try and find a better place to put this / a different way to prevent the real function from being called.

lib/screens/screens_by_alert/self_refresh_runner.ex
```diff
@@ -0,0 +1,102 @@
...
+  # (Not a real module--just a name assigned to the Task.Supervisor process that supervises each simulated data request run)
+  alias Screens.ScreensByAlert.SelfRefreshRunner.TaskSupervisor
+
+  require Logger
+  use GenServer
+
+  @type state :: %__MODULE__{}
+
+  # The server is not stateful.
+  defstruct []
```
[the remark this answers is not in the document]
jzimbel-mbta: I think I was following the informal best practice of having GenServers always use a struct as state, but yeah, the state could just be `nil`. I guess the idea was that if we do need to start keeping track of some state values in the future, it would be easier to just add some new keys to the existing struct.

I'll change the state to `nil` for now.

lib/screens/screens_by_alert/self_refresh_runner.ex
jzimbel-mbta: Tbh I'm not super clear on the best practices here. I have seen `for` used elsewhere for side-effecting code even though it produces a value, but `Enum.each` does seem better. I'll switch to `Enum.each`.

lib/screens/screens_by_alert/self_refresh_runner.ex
jzimbel-mbta: When I've informally tested it out before, `Stream` has only conferred performance gains when we might stop short of using all of the values that the stream spits out, or when the thing we're iterating over is huge. In this case, it's a finite and not-too-big enumerable, and we want to use all of the elements produced, so I'm sticking with `Enum` functions.

PaulJKim: Sounds good to me!

PaulJKim (approved):
LGTM!

PR #1547, opened 2022-11-01
feat: Fill in ScreensByAlert.Memcache implementation

jzimbel-mbta:
**Asana task**: [Write logic to interface with screens-by-alert memcached cache in deployment envs](https://app.asana.com/0/1185117109217413/1203221411292686/f)

This adds the logic to read/write the memcached cache.

- [ ] Tests added? **No, can't currently test this code in local dev since it depends on a real memcached instance. I'll deploy to dev-green and test there.**

*Though I _did_ thoroughly test the changes I contributed to memcachex.


lib/screens/screens_by_alert/memcache.ex
```diff
@@ -1,29 +1,176 @@
...
+    module_config = Application.fetch_env!(:screens, Screens.ScreensByAlert.Memcache)
+    connection_opts = Keyword.fetch!(module_config, :connection_opts)
 
-    Memcache.start_link(config[:connection_opts])
+    Memcache.start_link(connection_opts, name: @server)
   end
 
   @impl true
-  def put_data(_screen_id, _alert_ids) do
+  def put_data(screen_id, alert_ids) when is_binary(screen_id) and is_list(alert_ids) do
```
hannahpurcell, line 54: Does there need to be a definition without the guard to avoid uncaught errors, or is that handled?

lib/screens/screens_by_alert/memcache.ex
```diff
@@ -1,29 +1,176 @@
...
+        fn timestamped_screen_ids ->
+          unexpired_ids =
+            Enum.reject(timestamped_screen_ids, fn {_id, timestamp} ->
+              timestamp + @screens_ttl < now
+            end)
+
+          [timestamped_screen_id | unexpired_ids]
+        end,
+        retry: true,
+        default: [timestamped_screen_id],
```
hannahpurcell, line 137: When is the default used as a fallback by memcache?

hannahpurcell (commented):
Ok awesome! Tracking! Couple questions is all I have, just to check my understanding

lib/screens/screens_by_alert/memcache.ex
jzimbel-mbta: The default is used when the key does not already exist in the cache. This is basically how `Memcache.cas` works:
1. Check if the key that we want to update exists
2a. If it does, get its value, run the updater function on it to produce its new value, and set that value.
2b. If it does not, create the key using the default value instead.

lib/screens/screens_by_alert/memcache.ex
jzimbel-mbta: The guards can be thought of sort of like type annotations on the arguments.

Roughly equivalent TypeScript code:
```ts
function putData(screenID: string, alertIDs: string[]): void {...}
```

In TS, we wouldn't expect to need to handle other input types in `putData`—if another function called `putData("screen1", "alert1")`, your editor would underline it and point out the mistake.

hannahpurcell (approved):
Thanks for your answers! They were perfect and helped me learn a lot.

PR #1560, opened 2022-11-16
Added Medford branch stations to stop.ex

hannahpurcell:
We don't seem to need to add anything for these stops to the config.exs file, because none of these new stops have DUPs or Pre-Fare screens.

~Stops were added to stop.ex. Untested, but needs the stops to be assigned to trips in dev-green.~ Has been tested at this point.

To be deployed later, before the opening.

- [ ] Tests added?


PaulJKim (approved):
Seems reasonable to me! Approved but restating what's in the description: shouldn't be deployed until just before the start of revenue service

PR #1568, opened 2022-12-08
Temporary override for Heath St Sign

hannahpurcell:
Ad-hoc: When the GLX is deployed in the new GTFS later today, there is one screen that will start using the new destination name (Medford/Tufts) before we want that to be visible. So here's a hardcoded fix that we'll revert Sunday evening (along with deploying other changes)

### DIFF
--- a/assets/src/components/v2/normal_header.tsx
+++ b/assets/src/components/v2/normal_header.tsx
@@ -43,7 +43,8 @@ const NormalHeaderTitle = forwardRef(
       modifiers.push("with-icon");
     }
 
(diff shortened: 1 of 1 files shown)


cmaddox5 (commented):
This looks good for v2. Do we also need a change for v1? I know we have some v1 gl_eink screens in the config but can't remember if they are actually for live screens.

hannahpurcell:
There is precisely one screen we have to worry about on the whole E line, and it's V2, so we're good.

cmaddox5 (approved):

PR #1576, opened 2022-12-14
polish: Show boundary_downstream/upstream alerts V2 bus

cmaddox5:
Notion [task](https://www.notion.so/mbta-downtown-crossing/Alerts-don-t-appear-when-the-home-stop-is-on-the-boundary-of-a-disruption-ecf8d22084c64bf08f7c69bf731507f7)

For V2 bus, we always want to show `boundary_downstream` alerts and want to show `boundary_upstream` alerts for all effects except `:suspension`. 

- [ ] Tests added?

### DIFF
--- a/lib/screens/v2/widget_instance/alert.ex
+++ b/lib/screens/v2/widget_instance/alert.ex
@@ -189,6 +189,13 @@ defmodule Screens.V2.WidgetInstance.Alert do
     if takeover_alert?(t), do: :full_body_alert, else: :alert
   end
 
+  # Refer to https://mbta.sharepoint.com/:w:/s/CTDpublic/Eb_yKasCL9xKu_D9Xd1O5d0BG1GoHfplwoI-fpnru42M9w for how PIOs create alerts.
+  # For bus alerts, we assume that alerts are created using this guide.
+
+  # For bus suspensions, the endpoints of the alert are not directly affected by the suspension.
+  # This means that `boundary_upstream` and `boundary_downstream` stops are present in the `informed_entities` of the alerts,
+  # But the alert itself is not relevant for riders at the `boundary_upstream` stop.
+  # Because of this, we will not show a suspension alert at on a screen if the current stop is located on the `boundary_upstream`.
   def valid_candidate?(
         %__MODULE__{screen: %Screen{app_id: screen_type}, alert: %Alert{effect: :suspension}} = t
       )
@@ -198,6 +205,7 @@ defmodule Screens.V2.WidgetInstance.Alert do
       active?(t)
   end
 
+  # For all other bus alert effects, all stops in the `informed_entities` are directly affected by the alert and would be useful for riders to see.
   def valid_candidate?(%__MODULE__{screen: %Screen{app_id: screen_type}} = t)
       when screen_type in [:bus_shelter_v2, :bus_eink_v2] do
     priority(t) != :no_render and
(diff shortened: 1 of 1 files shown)


hannahpurcell:
QA-ing the performance, first. I think there's a bug here? With this suspension, shouldn't there be a flex zone alert? 

Screenshot 2022-12-20 at 2 07 30 PM
Screenshot 2022-12-20 at 4 03 06 PM

cmaddox5:
> QA-ing the performance, first. I think there's a bug here? With this suspension, shouldn't there be a flex zone alert?
> 
> Screenshot 2022-12-20 at 2 07 30 PM Screenshot 2022-12-20 at 4 03 06 PM

No, because the home stop is outside the list of affected stops. Bus alerts at minimum have to affect the home stop for it to show on screens. This alert also wouldn't show if it were the first stop in the stop range because it would then be `boundary_upstream`, which is not a location we care about for suspensions. 

The reasoning behind these decisions involves assumptions based on this [guide](https://mbta.sharepoint.com/:w:/s/CTDpublic/Eb_yKasCL9xKu_D9Xd1O5d0BG1GoHfplwoI-fpnru42M9w?rtime=sIKlANDi2kg). Mindy also added some graphics to the comments of this task that are super helpful. If things are still fuzzy, let me know and I'd be happy to walk through it.

hannahpurcell:
> The reasoning behind these decisions involves assumptions based on this [guide](https://mbta.sharepoint.com/:w:/s/CTDpublic/Eb_yKasCL9xKu_D9Xd1O5d0BG1GoHfplwoI-fpnru42M9w?rtime=sIKlANDi2kg). Mindy also added some graphics to the comments of this task that are super helpful. If things are still fuzzy, let me know and I'd be happy to walk through it.

Ohhh hmm. The guide helped fill in some blanks for me, because the notion diagrams really confused me. So, when a PIO makes a detour alert, only the skipped stops are included in the selection; when they make a suspension, the "new endpoints" for the route are selected also, in addition to the skipped stops.

I think that's how subway works, also, but I guess I didn't expect that for bus.

cmaddox5:
> > The reasoning behind these decisions involves assumptions based on this [guide](https://mbta.sharepoint.com/:w:/s/CTDpublic/Eb_yKasCL9xKu_D9Xd1O5d0BG1GoHfplwoI-fpnru42M9w?rtime=sIKlANDi2kg). Mindy also added some graphics to the comments of this task that are super helpful. If things are still fuzzy, let me know and I'd be happy to walk through it.
> 
> Ohhh hmm. The guide helped fill in some blanks for me, because the notion diagrams really confused me. So, when a PIO makes a detour alert, only the skipped stops are included in the selection; when they make a suspension, the "new endpoints" for the route are selected also, in addition to the skipped stops.
> 
> I think that's how subway works, also, but I guess I didn't expect that for bus.

That's correct. So like with subway, the stops on either side of the range of the alert aren't actually affected by the suspension directly. With that being the case, the last stop of the range of the alert (`boundary_upstream`) doesn't care at all about the suspension. None of the affected stops are reachable from it so it's irrelevant info if you are continuing downstream. 

>This alert also wouldn't show if it were the first stop in the stop range because it would then be boundary_upstream, which is not a location we care about for suspensions.

I said this earlier and that is actually incorrect. The first stop in the range is actually considered `boundary_downstream`, so it would show an alert if included. Broadway @ Cary would be the `boundary_upstream` stop of the alert, and that stop would not show an alert.

For detour, we talked about how it isn't ideal that the "endpoints" are not included in the alert because that means there is no info about the detour at the stop before the detour starts. That's out of scope here though and will be discussed in the future.

hannahpurcell (approved):
Mind adding a comment? Otherwise, fab!

PR #1578, opened 2022-12-16
polish: Prevent caching for hidden screens

cmaddox5:
Notion [task](https://www.notion.so/mbta-downtown-crossing/Bus-alerts-always-show-0-places-even-when-screen-count-is-more-than-0-d860bb7e91fd4756b994ab100d07d312)

For screens refreshing on their own, we were not checking if it is hidden from screenplay before caching. With this check, counts will be more accurate on the Alerts page in screenplay.

- [ ] Tests added?

### DIFF
--- a/lib/screens/v2/screen_data.ex
+++ b/lib/screens/v2/screen_data.ex
@@ -42,7 +42,7 @@ defmodule Screens.V2.ScreenData do
     layout_and_widgets =
       config
       |> fetch_data()
-      |> tap(&cache_visible_alert_widgets(&1, screen_id))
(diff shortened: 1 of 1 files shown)


PaulJKim (approved):
Logic makes sense to me! Mainly just to refresh my memory, but why do we hide certain screens on Screenplay again?

cmaddox5:
@PaulJKim We mostly hide test screens and screens in places that PIOs have no control over/need to view info on. With those hidden, they can make sure that the screens they see on the page are screens that we manage across the system.

PR #1579, opened 2022-12-19
polish: New state for bus_eink_v2

cmaddox5:
Notion [task](https://www.notion.so/mbta-downtown-crossing/If-all-routes-at-that-stop-are-suspended-or-detoured-the-departures-widget-should-show-an-empty-sta-fe41daacba2e44d4849b3213c51daafb)

For `bus_eink_v2`, we now show a no service state for multi route stops when multiple alerts cause the stop to show no departures. This is specific to bus E-Ink at the moment. 

Screenshot 2022-12-19 at 11 17 15 AM

- [x] Tests added?


jzimbel-mbta (approved):
Nice, looks good to me!

PR #1583, opened 2022-12-20
fix: Pair takeover alerts with bottom screen filler on GL e-ink

jzimbel-mbta:
Notion [task](https://www.notion.so/mbta-downtown-crossing/E-Ink-V2-Polish-Punchlist-12bc6f522850471ca01770802a009d97?p=99cf9f899b064ff096d385f2ffad6973&pm=s)

GL e-ink takeover alerts (which replace main content: the departures list and the line map) were not causing the bottom screen to show the "filler" content instead of the usual flex zone.

This was because the GL e-ink template has an extra layout variant intended for the DeparturesNoData widget to be displayed alongside the flex zone. The framework was choosing this variant over the one with the bottom screen takeover/filler content, because the filler widget has a lower priority than any flex zone widget.

As a "surgical" fix, I separated the two cases by having DeparturesNoData use a different slot name. We may want to revisit this in the future to find a more elegant approach.

## Before
(At Heath Street)
image

## After
image

- [x] Tests added?

### DIFF
--- a/lib/screens/v2/candidate_generator/gl_eink.ex
+++ b/lib/screens/v2/candidate_generator/gl_eink.ex
@@ -37,11 +37,17 @@ defmodule Screens.V2.CandidateGenerator.GlEink do
               Builder.with_paging({:flex_zone, %{one_medium: [:medium]}}, 2),
               :footer
             ],
+            # This layout variant is necessary for the DeparturesNoData widget
+            # to take up both the usual main_content slot, and the left_sidebar
+            # slot to its left, while still allowing the normal flex zone
+            # to appear on the bottom screen.
             top_takeover: [
-              :full_body_top_screen,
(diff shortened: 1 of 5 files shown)


lib/screens/v2/candidate_generator/gl_eink.ex
```diff
@@ -37,11 +37,17 @@ defmodule Screens.V2.CandidateGenerator.GlEink do
               Builder.with_paging({:flex_zone, %{one_medium: [:medium]}}, 2),
               :footer
             ],
+            # This layout variant is necessary for the DeparturesNoData widget
```
cmaddox5, line 40: Thanks for commenting! Would've been nice if the original dev did it...

cmaddox5 (approved):
Looks good!

PR #1590, opened 2023-01-05
feat: DUP V2 app

cmaddox5:
**Asana task**: [[Widgetize DUPs] Set up new V2 app](https://app.asana.com/0/1185117109217413/1203279317437006/f)

This PR adds a new app for the V2 version of DUPs.

The vision for the config shape is very basic. Rotations 0 and 1 always show "primary" departures and rotation 2 always shows "secondary" (or primary if no secondary is specfied). Those plus evergreen content are the only configs values I see us needing for now.

The APP layout is built similarly to how we tackle dual screens on Pre-Fare. Each DUP is has a "screen" with 3 pages, and each page has its own layout. Because rotation 0 and 1 show the same info, those layouts are the same. This made things tricky from a candidate placement perspective. What I found out is as long as a header `WidgetInstance` is created for each rotation (in other words, create a copy of one instance 3 times), only one "main content" candidate needs to be created for each layout name. So 3 `:header`s, one `:main_content_primary`, and one `:main_content_secondary` will properly map to 3 full pages of content. 

On the frontend, the main hurdle was how to show each rotation individually. That was solved in a similar way to Pre-Fare. There is an optional query param `rotation_index` that can be added to the URL to show only the content for a single rotation (or all three if no param is present). With this solution, we should only need one packaged app for each screen. Displayed pages can be controlled by the param.

I am still trying to fully understand how Outfront uses our packages so my only open question is will this work for them?

- [ ] Tests added?


lib/screens/v2/candidate_generator/dup.ex
```diff
@@ -0,0 +1,80 @@
...
+              :main_content_primary,
+              :inline_alert
+            ],
+            body_takeover: [:full_body]
+          }},
+         {:rotation_two,
+          %{
+            body_normal: [
+              :header,
+              :main_content_secondary,
```
hannahpurcell, line 41: I'm a bit fuzzy on the way the dup template works. I'd guess that 0 and 2 should be `main_content_primary` if they are supposed to hold the same main content, and 1 would be `main_content_secondary`, but I might be missing the mark here?

lib/screens/v2/screen_data/parameters.ex
```diff
@@ -38,7 +41,8 @@ defmodule Screens.V2.ScreenData.Parameters do
     gl_eink_v2: 0,
     solari_v2: 0,
     solari_large_v2: 0,
-    pre_fare_v2: 0
+    pre_fare_v2: 0,
+    dup_v2: 0
```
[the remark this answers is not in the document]
cmaddox5: This list is used to set a default value for configs that don't define `interval_offset_seconds`. The value itself is used to offset audio by a defined number of seconds. When we first implemented audio, we got feedback saying that it was sometimes hard to hear the audio because the screen on the other side of the street was drowning it out. So with this, one screen will play audio and the other will play audio `now + interval_offset_seconds` later.

lib/screens/v2/candidate_generator/dup.ex
cmaddox5: V1 DUPs currently have 0 and 1 showing the same list of departures, not 0 and 2. This is at least true for departures. I do think I can see an issue with this current layout when thinking about alerts. While 0 and 1 share logic for departures, they do not share logic for alerts that are not `:inside`. For `:boundary` alerts, 0 shows a partial and 1 shows a fullscreen. 

I _think_ we might be able to handle this by giving each rotation their own `full_body_rotation_x` in the `body_takeover` part of the layout. Then in the `CandidateGenerator` logic for `Alert`s, we can flag fullscreen vs inline on the candidate.

cmaddox5: We may actually be able to get rid of the `:inline_alert` for rotation 2. I see anything in the code that says that rotation shows an inline alert for live alerts, only overrides. And since overrides are being skipped, it may never get used.

cmaddox5: I decided to try this out. Each rotation now has its own alert slot if it will need one (I couldn't find any instance of v1 DUP showing a partial alert on rotation one, so I removed it from the layout). Each rotation also now has a `full_screen_takeover` slot. This will likely always be used for full screen alerts, but evergreen could also use it.

jzimbel-mbta: For the reason why we did the primary departures on 0 and 1, it was just because it was a little easier to group them together like that in code. The end result on the screens is the same, since they cycle through the 3 pages repeatedly:

`A- A-B-A- A-B-A- A-B ...`
vs, if primary departures were on 0 and 2:
`   A-B-A- A-B-A- A-B-A ...`

lib/screens/v2/screen_data/parameters.ex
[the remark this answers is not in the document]
jzimbel-mbta: I wonder if we could get rid of this map and just return 0 in the [fallback clause](https://github.com/mbta/screens/blob/0c50bc4a689ac6eb78991a4adab38208c1131747/lib/screens/v2/screen_data/parameters.ex#L84-L86) of `get_audio_interval_offset_seconds`? Currently it looks up the value in this map, but that value is always 0.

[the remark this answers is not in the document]
cmaddox5: That's a good idea. Makes the function returns more clear.

lib/screens/v2/candidate_generator/dup.ex
```diff
@@ -0,0 +1,78 @@
...
+  alias Screens.V2.WidgetInstance.{NormalHeader, Placeholder}
+
+  @behaviour CandidateGenerator
+
+  @impl CandidateGenerator
+  def screen_template do
+    {:screen,
+     %{
+       screen_normal: [
+         {:rotation_zero,
```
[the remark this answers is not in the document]
cmaddox5: The main reason I used words instead of digits is it felt cleaner to me in most situations. I think the digit is fine in an atom, but in filenames and frontend variables it felt less so for some reason. I don't feel super strongly about that though so feel free to push back.

assets/css/v2/dup/body/normal.scss
```diff
@@ -0,0 +1,19 @@
+.body-normal {
+  position: relative;
+}
+
+.body-normal__header {
+  position: absolute;
+  top: 0px;
+  left: 0px;
+  width: 1920px;
+  height: 248px;
```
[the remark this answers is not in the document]
cmaddox5: Good idea. I went ahead and made that util class. Only applied it to DUPs for now though.

lib/screens/v2/candidate_generator/dup.ex
```diff
@@ -0,0 +1,78 @@
...
+              :main_content_primary,
+              :inline_alert_zero
+            ],
+            screen_takeover_zero: [:full_screen_zero]
+          }},
+         {:rotation_one,
+          %{
+            body_normal_one: [
+              :header,
+              :main_content_primary
```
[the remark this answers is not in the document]
cmaddox5: At the time, I did the share slot names mostly because I thought it was convenient. I'm with you now, I don't think we should rely on this working forever. Every slot name should have a unique name now.

lib/screens/v2/candidate_generator/dup.ex
```diff
@@ -0,0 +1,79 @@
...
+          %{
+            body_normal_two: [
+              :header_two,
+              :main_content_secondary_two,
+              :inline_alert_two
+            ],
+            screen_takeover_two: [:full_screen_two]
+          }}
+       ]
+     }}
```
jzimbel-mbta, line 46: Need a few tweaks to this, based on some discussion with Paul this morning.

The main change is this:
- make inline (or "partial") alerts optional, and give them their own dedicated slot rather than drawing directly over the main content slot

IMG_1990

The rest are just suggested name changes to more closely align with the naming scheme of other existing templates.

IMG_1989

```suggestion
    {:screen,
     %{
       screen_normal: [
         {:rotation_zero,
          %{
            rotation_normal_zero: [
              :header_zero,
              {:body_zero,
               %{
                 body_normal_zero: [
                   :main_content_zero
                 ],
                 body_split_zero: [
                   :main_content_reduced_zero,
                   :bottom_pane_zero
                 ]
               }}
            ],
            rotation_takeover_zero: [:full_rotation_zero]
          }},
         {:rotation_one,
          %{
            rotation_normal_one: [
              :header_one,
              :main_content_one
            ],
            rotation_takeover_one: [:full_rotation_one]
          }},
         {:rotation_two,
          %{
            rotation_normal_two: [
              :header_two,
              {:body_two,
               %{
                 body_normal_two: [
                   :main_content_two
                 ],
                 body_split_two: [
                   :main_content_reduced_two,
                   :bottom_pane_two
                 ]
               }}
            ],
            rotation_takeover_two: [:full_rotation_two]
          }}
       ]
     }}
```

jzimbel-mbta: (And, unfortunately, corresponding changes to the frontend layout components as well. Sorry!!)

cmaddox5: This makes sense. Made all the necessary changes to the frontend too.

jzimbel-mbta (approved):
Looks good! Just one styling issue for the "bottom pane" height, and one optional suggestion. Approved assuming you address those as needed.

PR #1592, opened 2023-01-06
tweak: DUP headsign replacement

cmaddox5:
**Asana task**: [Add new DUP headsign abbreviation for "Middleborough/Lakeville"](https://app.asana.com/0/1185117109217413/1203658337706587/f)

Instead of adding a new case for every headsign with a `/`, I added a `String.replace` call that looks for a `/` without spaces around it. Needed to use regex to make it happen, but this should keep us from needing a task for each new headsign.

- [ ] Tests added?

### DIFF
--- a/lib/screens/dup_screen_data/request.ex
+++ b/lib/screens/dup_screen_data/request.ex
@@ -213,7 +213,7 @@ defmodule Screens.DupScreenData.Request do
       # Find any headsign with a slash that does NOT have a space on each side and add spaces
       # i.e. String.replace("Middleborough/ Lakeville", ~r/([^[:space:]])([\/])(.)/, "\\1 / \\3") => "Middleborough / Lakeville"
       # i.e. String.replace("Middleborough/Lakeville", ~r/([^[:space:]])([\/])(.)/, "\\1 / \\3") => "Middleborough /  Lakeville"
-      |> String.replace(~r/([^[:space:]])([\/])(.)/, "\\1 / \\3")
+      |> String.replace(~r/(.)([\/])(.)/, "\\1 / \\3")
 
     %{departure_map | destination: replaced_destination}
   end
--- a/lib/screens/dup_screen_data/request.ex
+++ b/lib/screens/dup_screen_data/request.ex
@@ -213,7 +213,7 @@ defmodule Screens.DupScreenData.Request do
       # Find any headsign with a slash that does NOT have a space on each side and add spaces
       # i.e. String.replace("Middleborough/ Lakeville", ~r/([^[:space:]])([\/])(.)/, "\\1 / \\3") => "Middleborough / Lakeville"
       # i.e. String.replace("Middleborough/Lakeville", ~r/([^[:space:]])([\/])(.)/, "\\1 / \\3") => "Middleborough /  Lakeville"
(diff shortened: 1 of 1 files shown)


lib/screens/dup_screen_data/request.ex
```diff
@@ -208,7 +207,13 @@ defmodule Screens.DupScreenData.Request do
   end
 
   defp replace_long_headsigns(%{destination: destination} = departure_map) do
-    replaced_destination = Map.get(@headsign_replacements, destination, destination)
+    replaced_destination =
+      @headsign_replacements
+      |> Map.get(destination, destination)
+      # Find any headsign with a slash that does NOT have spaces around it and add spaces
+      # i.e. String.replace("Middleborough/Lakeville", ~r/([^[:space:]])([\/])([^[:space:]])/, "\\1 / \\3") => "Middleborough / Lakeville"
+      |> String.replace(~r/([^[:space:]])([\/])([^[:space:]])/, "\\1 / \\3")
```
PaulJKim, line 215: Don't know how often this edge case would appear or how important it is that we handle it, but if the slash has a space only on one side of it, for example `test/ string", the regex fails to catch that and add a space to the one side that doesn't have the space.

cmaddox5: I'm not sure either so I went ahead and tweaked the regex to cover it. The most common case (a slash with no spaces on either side) is still covered. Now a slash only has a space on the right side, it will still add a space on both sides. This will cause an extra space on the right side, but the front end splits the string on whitespace so any extra whitespace is trimmed.

PaulJKim: Should we just add a space on both sides of the slash? I think this doesn't cover the converse case where the slash only has a space on the left. Or would the frontend handle that case alright as is?

cmaddox5: Yeah that's a good point. Always adding a space on each side doesn't cause extra spaces. The frontend string splitting always trims them.

jzimbel-mbta: The regex could be slightly adjusted to capture just the slash and 0 or 1 space on either side of it, and then you could simply replace that single capture with `" / "`:
```ex
~r/[[:space:]]?\/[[:space:]]?/
```

```ex
iex> tests = ["a/b", "a /b", "a/ b", "a / b"]
iex> Enum.map(tests, &String.replace(&1, ~r/([[:space:]]?\/[[:space:]]?)/, " / "))
["a / b", "a / b", "a / b", "a / b"]
```

Per regex101.com:
image

cmaddox5: Exactly what we need! Thanks Jon!

jzimbel-mbta: You can also use `\s` instead of `[[:space:]]`! I'd never checked until now, but from a few searches it appears they represent an identical set of whitespace characters.

PaulJKim (approved):
LGTM!

lib/screens/dup_screen_data/request.ex
cmaddox5: Yep, I learned that today too. I can't decide if the character classes make things more or less readable though. The word is nice but the extra characters muddy up the regex a bit.

PR #1594, opened 2023-01-06
feat: Client Exception Catcher

cmaddox5:
**Asana task**: ad-hoc

This new component can be used to catch all exceptions thrown in component tree and log them using the API. It uses existing query params to determine if a log is needed in this way: only real screens that have Sentry disable will use this. After logging, we throw the exception again and let the screen error like it would anyway. This is very much a flexible solution so feel free to chime in with any recommendations and questions!

- [ ] Tests added?


assets/src/components/v2/exception_catcher.tsx
```diff
@@ -0,0 +1,50 @@
+import React, { ErrorInfo, ReactElement } from "react";
+import { RouteComponentProps, withRouter } from "react-router-dom";
+import { getDataset } from "Util/dataset";
+import { isRealScreen } from "Util/util";
+interface ExceptionCatcherState {
+  hasError: boolean;
+  errorMessage?: Error;
+}
```
jzimbel-mbta, line 8: Is this accurate? It looks like while the component declares this as its state, it never actually sets such state up, or uses it.

jzimbel-mbta (approved):
Nice work, just 2 comments that should be quick to address.

assets/src/components/v2/exception_catcher.tsx
cmaddox5: A previous solution I tried used this and I just forgot to delete it after I went with this solution. Removed now.

PR #1595, opened 2023-01-06
Revert regex for long headsigns, add new DUP case for Middleborough

hannahpurcell:
Adhoc

### DIFF
--- a/lib/screens/dup_screen_data/request.ex
+++ b/lib/screens/dup_screen_data/request.ex
@@ -37,6 +37,7 @@ defmodule Screens.DupScreenData.Request do
     "Needham Heights" => "Needham Hts",
     "Houghs Neck via McGrath & Germantown" => "Houghs Neck via McGth & Gtwn",
     "Houghs Neck via Germantown" => "Houghs Neck via Germntwn",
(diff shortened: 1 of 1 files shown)


lib/screens/dup_screen_data/request.ex
```diff
@@ -37,6 +37,7 @@ defmodule Screens.DupScreenData.Request do
     "Needham Heights" => "Needham Hts",
     "Houghs Neck via McGrath & Germantown" => "Houghs Neck via McGth & Gtwn",
     "Houghs Neck via Germantown" => "Houghs Neck via Germntwn",
+    "Middleborough/Lakeville" => "Middleborough / Lakeville",
```
hannahpurcell, line 40: Is there a possibility of any other format?

cmaddox5: I searched around in the API and that's the only format I found for that headsign.

cmaddox5 (approved):

lib/screens/dup_screen_data/request.ex
hannahpurcell: oh sweet! thanks for the extra mile check

PR #1598, opened 2023-01-09
doc: Update README and add widget architecture doc

jzimbel-mbta:
**Asana task**: [Document technical architecture of Screens app's "widget framework"](https://app.asana.com/0/1185117109217413/1203530439461080/f)

NOTE FOR REVIEWERS: You can view the rendered markdown by clicking this button on the top right of a `.md` file diff.
Screen Shot 2023-01-09 at 12 41 56 PM

This is my attempt to more fully document the internal "widget framework" that we've developed for the app. I tried to follow the advice from [here](https://matklad.github.io/2021/02/06/ARCHITECTURE.md.html), somewhat.

I also zhuzhed up the main README with some sample screenshots! 😃


docs/architecture/widget_framework.md
```diff
@@ -0,0 +1,90 @@
...
+
+During the serialization step, all pages except the current one are discarded. This means the React client is only aware of the content of the current page at any given moment.
+
+### Why backend paging?
+
+Because audio readouts are not space constrained in the same way as the visual UI, and because they happen less frequently, we want to read out all info from a screen, including currently hidden pages. Since audio synthesis happens on the backend, this means the backend needs to be aware of, and in control of, paging.
+
+## Feature: Audio-only widgets
+
+Because the semantics of speech don't allow for the same "scanability" as a visual UI, we occasiona5lly need to provide a summary of content before it's read out, so that the rider can decide whether it's relevant to them.
```
panentheos, line 86: Just a small typo
```suggestion
Because the semantics of speech don't allow for the same "scanability" as a visual UI, we occasionally need to provide a summary of content before it's read out, so that the rider can decide whether it's relevant to them.
```

lib/screens/v2/widget_instance.ex
```diff
@@ -9,30 +9,91 @@ defprotocol Screens.V2.WidgetInstance do
...
   def audio_serialize(instance)
 
+  @doc """
+  Returns a priority value used to sort the widget among others being rendered to SSML.
+  """
   @spec audio_sort_key(t) :: priority()
   def audio_sort_key(instance)
 
+  @doc """
+  Indicates whether the widget is valid for inclusion in the audio readout.
```
jzimbel-mbta, line 89:
```suggestion
  Indicates whether the widget is valid for inclusion in the audio readout.

  Note that (except for audio-only widgets) a widget must be included in the visual
  layout as a prerequisite for consideration in the audio readout.

  That is, if a widget does not make it into the visual layout--due to having a low
  priority, for example--then it will not make it into the audio readout even if
  this function returns `true`.
```

README.md
```diff
@@ -1,5 +1,22 @@
 # Screens
 
+This application serves real-time information to all of our browser-based screens posted at stops and stations around the system.
+
+If you're looking for the application that controls the LED countdown clocks and the in-station PA system, please see [mbta/realtime_signs](https://github.com/mbta/realtime_signs).
```
PaulJKim, line 5: Probably would be more accurate to reference [Signs UI](https://github.com/mbta/signs_ui) here (or both)

docs/architecture/widget_framework.md
```diff
@@ -0,0 +1,90 @@
...
+
+## Request data flow / how the framework logic works
+
+Screens team members can see this as a [flowchart in Miro](https://miro.com/app/board/o9J_lDRaax4=/).
+
+1. Client requests `/v2/api/screen/{id}`. Subsequent steps take place on the server.
+1. We get the config stored under `id` in the screen configuration file.
+1. We look up the "app ID" stored in the config. We get the appropriate candidate generator module for that app ID.
+1. We get the template for this screen type by calling `candidate_generator.screen_template()`.
+1. We get a list of candidate widgets for the screen by calling `candidate_generator.candidate_instances(config)`.
```
hannahpurcell, line 37: Somewhere separate, it could be helpful to acknowledge that there are often more candidates than space on screens, and we weigh which ones should appear based on priority. Examples would also be helpful: alert widget candidates are highly prioritized and so almost always make it onto screens, while general PSAs are low-priority and will get bumped if the screen template runs out of room.

docs/architecture/widget_framework.md
```diff
@@ -0,0 +1,90 @@
...
+
+Some screen types have paged regions, where the widgets that appear in the region can rotate over the course of several data refreshes.
+
+Paging logic lives on the backend. The current page is selected based on the screen type's data refresh rate (e.g. 20 sec for pre-fare duo) and the current clock time.
+
+During the serialization step, all pages except the current one are discarded. This means the React client is only aware of the content of the current page at any given moment.
+
+### Why backend paging?
+
+Because audio readouts are not space constrained in the same way as the visual UI, and because they happen less frequently, we want to read out all info from a screen, including currently hidden pages. Since audio synthesis happens on the backend, this means the backend needs to be aware of, and in control of, paging.
```
hannahpurcell, line 82: Not quite following here. I'm not sure why the audio readout situation requires backend paging. The backend is aware of all pages. It passes this info to the audio synthesis code, which then is also aware of all pages. In the same way, can't all pages be passed to the frontend, and then frontend paging is possible? (I'm sure there are other reasons why frontend paging is not ideal, though.)

docs/architecture/widget_framework.md
```diff
@@ -0,0 +1,90 @@
+# Widget framework
+
+The widget framework comprises common code that provides a "backbone structure" to our business logic.
+
+Each screen's content is composed entirely of widgets, and when a screen requests new data, no matter which type it is, the same backend logic is used to look up the screen's configuration, fetch the appropriate real-time data, transform that data into "candidate" widgets, and then assemble some of those candidates together to be sent back to the client and rendered.
+
+## What's a widget?
```
cmaddox5, line 7: ~~I know this doc primarily focuses on visuals, but we do have widgets that are audio only. Would that be worth mentioning here?~~

hannahpurcell (approved):
Ah this is so awesome to have! At the moment, I think this reads well and is fairly comprehensive, though I know we'll keep thinking of things down the road that we could add. I'm trying to think of ways that we can keep this document living, well-maintained? We're constantly thinking of new ways to adjust the framework, and this is a great reference to keep it all documented post-implementation.

docs/architecture/widget_framework.md
```diff
@@ -0,0 +1,90 @@
...
+
+A widget is a self-contained piece of content. It's essentially a special type of Model (as in MVC) that can be dropped into template slots alongside any number of other widgets to form a full page of content.
+
+Some examples of widgets include the departures list, subway status, alert cards, and image PSAs. However, even basic screen content like the header and footer are widgets as well.
+
+On the backend, each widget implements the [`WidgetInstance` protocol](/lib/screens/v2/widget_instance.ex). The framework calls these protocol functions to find out each widget's priority and where it would like to appear on the screen. The framework then determines how best to position all available widgets within these constraints.
+
+**Widgets do not directly communicate with one another, or even know about the existence of other widgets.** Widgets only communicate with the framework via the aforementioned protocol functions. Once created, a widget has all of the data it needs to render itself.
```
cmaddox5, line 15: Our `content_summary` widget breaks this rule 😬 It is a very special case though.

docs/architecture/widget_framework.md
```diff
@@ -0,0 +1,90 @@
...
+
+Once a widget is created, it has all of the data it needs. This means all functions on a widget are pure/free of side effects, and easy to test.
+
+See the [`WidgetInstance` protocol](/lib/screens/v2/widget_instance.ex)'s inline documentation for details on each implementation function.
+
+## Request data flow / how the framework logic works
+
+Screens team members can see this as a [flowchart in Miro](https://miro.com/app/board/o9J_lDRaax4=/).
+
+1. Client requests `/v2/api/screen/{id}`. Subsequent steps take place on the server.
```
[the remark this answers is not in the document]
cmaddox5: It might be nice if a dev wants to try to "follow along", not so much as a replacement to these steps. You plain English version is much easier to follow than the code by itself.

docs/architecture/widget_framework.md
```diff
@@ -0,0 +1,90 @@
...
+
+- `region_id`: the region's name. For example: `:body`, `:flex_zone`, `:screen`.
+- `layout_id`: a name for one possible set of slots that can fill this region. For example: `:body_normal`, `:body_takeover`; `:one_large`, `:two_medium`.
+- `list(template)`: the set of slots, or even further nested regions, that fill this region under the given layout variation.
+
+Layout variations provide a mechanism for high-priority widgets, like urgent alerts, to take over large portions of a screen. They also allow paged regions to hold widgets of different sizes per page.
+
+## Feature: Paging
+
+Some screen types have paged regions, where the widgets that appear in the region can rotate over the course of several data refreshes.
```
cmaddox5, line 74: A special case here is the `elevator_status`. Could be confusing to someone new to the framework due to that widget appearing to page the same way the flex zone does.

docs/architecture/widget_framework.md
cmaddox5: I see the section for this at the bottom here, so disregard.

cmaddox5 (approved):
This is fantastic! Folks new to the framework should be able to get familiar with it very quickly now. I did have a few questions/comments, mostly around some of our "one-offs".

hannahpurcell (commented):
The recent architecture discussion with Paul H made me realize that I have some gaps in my own understanding about where we like putting different kinds of logic; the question of heavy logic in candidate generator vs widgetinstance? No clue! Is this a good place to weigh some of our philosophies around what different files should do? Like, API requests should go in the candidate_generator, etc

ARCHITECTURE.md
```diff
@@ -0,0 +1,55 @@
...
+
+## Codemap
+
+Server:
+| Module or directory | Description |
+| - | - |
+| [`ScreenData`](/lib/screens/v2/screen_data.ex) | Common logic to answer all screen data requests. We try to keep this code agnostic of widget implementation details and free of business logic. Code changes infrequently. If a new feature is needed on _all_ screen types, the change might happen here. |
+| [`config` directory](/lib/screens/config/) | Contains modules that define the config schema for each screen type. Most modules `use` a macro from [`config/struct.ex`](/lib/screens/config/struct.ex) to simplify defining their data structures. |
+| [Candidate generators](/lib/screens/v2/candidate_generator/) | These modules define "step 1" for producing the data response for each screen type. Each one adopts the [`CandidateGenerator`](/lib/screens/v2/candidate_generator.ex) behaviour, and contains code that defines the screen's visual template as well as functions to fetch data and transform it into widgets to populate that template. |
+| [`Template`](/lib/screens/v2/template.ex) | Contains types and functions related to screen templates.  |
```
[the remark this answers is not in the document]
jzimbel-mbta: Explained this in the new glossary section!

docs/architecture/widget_framework.md
jzimbel-mbta: Yeah this one is always fuzzy for me! Every time we revisit paging, we come to the conclusion that pages should be selected on the backend, but I always forget the exact reasoning.
I updated this section with a different reason (stale data if paging happens client-side), which I think is one of the main ones we come back to in discussions about paging.

docs/architecture/widget_framework.md
jzimbel-mbta: Added a section about persistent widgets / client-side widget-internal paging!

hannahpurcell (approved):
👍

PR #1599, opened 2023-01-10
tweak: Filter upcoming alerts for Alerts widget

cmaddox5:
**Asana task**: [[GL E-Ink V2] Filter out upcoming alerts](https://app.asana.com/0/1185117109217413/1203637230901491/f)

This PR changes the Alerts widget logic a bit so that we don't get upcoming alerts from the API. Upcoming alerts should not show on any screen, so there is no need for the CandidateGenerator to create a candidate for them. This means that a lot of logic in the `WidgetInstance.Alert` could be removed. We no longer need any conditional checking if the alert is active. It always will be. 

- [ ] Tests added?

### DIFF
--- a/lib/screens/v2/candidate_generator/widgets/alerts.ex
+++ b/lib/screens/v2/candidate_generator/widgets/alerts.ex
@@ -21,7 +21,7 @@ defmodule Screens.V2.CandidateGenerator.Widgets.Alerts do
         now \\ DateTime.utc_now(),
         fetch_simplified_routes_at_stop_fn \\ &Route.fetch_simplified_routes_at_stop/2,
         fetch_stop_sequences_through_stop_fn \\ &RoutePattern.fetch_stop_sequences_through_stop/1,
-        fetch_alerts_by_stop_and_route_fn \\ &Alert.fetch_by_stop_and_route/2
+        fetch_alerts_by_stop_and_route_fn \\ &Alert.fetch_by_stop_and_route/3
       )
       when app in @alert_supporting_screen_types do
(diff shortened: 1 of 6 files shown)


lib/screens/v2/candidate_generator/widgets/alerts.ex
```diff
@@ -21,15 +21,17 @@ defmodule Screens.V2.CandidateGenerator.Widgets.Alerts do
...
       when app in @alert_supporting_screen_types do
     with {:ok, routes_at_stop} <- fetch_simplified_routes_at_stop_fn.(stop_id, now),
          {:ok, stop_sequences} <- fetch_stop_sequences_through_stop_fn.(stop_id),
          reachable_stop_ids = local_and_downstream_stop_ids(stop_sequences, stop_id),
          route_ids_at_stop = Enum.map(routes_at_stop, & &1.route_id),
          {:ok, alerts} <-
-           fetch_alerts_by_stop_and_route_fn.(reachable_stop_ids, route_ids_at_stop) do
+           fetch_alerts_by_stop_and_route_fn.(reachable_stop_ids, route_ids_at_stop,
+             datetime: "NOW"
+           ) do
```
hannahpurcell, line 34: turns out, this query is buggy in the api right now 😞 https://mbta.slack.com/archives/GP5QJM98A/p1674145757579809?thread_ts=1674145440.912979&cid=GP5QJM98A

cmaddox5:
Gonna close this because of the bug Hannah mentioned. See https://github.com/mbta/screens/pull/1599 for a more reliable change.

PR #1600, opened 2023-01-10
feat: Automate screen refreshes

cmaddox5:
**Asana task**: [Automate refreshing of screens after a deploy](https://app.asana.com/0/1185117109217413/1203557095958472/f)

~~This PR adds an endpoint and a new job to the `Deploy to ECS` action for scheduling screen refreshes after a deploy. The `needs: call-workflow` line requires the first job to succeed before executing. When the job executes, it makes an API call to a new endpoint: `/api/admin/refresh_all`. This endpoint does the same thing as the existing `refresh` endpoint, except it gets _all_ screen IDs from the current config and schedules a refresh for all of them.~~

Edit: Bit of a pivot due to some concerns. The action no longer communicates with the server. Instead, it simply creates/updates a file (content of the file is irrelevant. We just need the time the file was modified) and uploads that file to S3. The server then gets the `Last-Modified` time of the file and considers that time when determining if clients need to be refreshed. 

- [ ] Tests added?

### DIFF
--- a/lib/screens/util.ex
+++ b/lib/screens/util.ex
@@ -1,6 +1,8 @@
 defmodule Screens.Util do
   @moduledoc false
 
+  alias Screens.Config.State
+
   def format_time(t) do
     t |> DateTime.truncate(:second) |> DateTime.to_iso8601()
   end
@@ -150,7 +152,7 @@ defmodule Screens.Util do
   end
 
   @doc """
-  Takes route id, returns route type. 
+  Takes route id, returns route type.
   Not the favorite way of getting route type, but useful for property testing.
   """
   @spec route_type_from_id(String.t()) :: atom()
@@ -162,4 +164,14 @@ defmodule Screens.Util do
   def route_type_from_id("CR-" <> _), do: :commuter_rail
   def route_type_from_id("Boat-" <> _), do: :ferry
   def route_type_from_id(_), do: :bus
+
+  defp outdated?(screen_id, client_refresh_timestamp) do
+    {:ok, client_refresh_time, _} = DateTime.from_iso8601(client_refresh_timestamp)
+    refresh_if_loaded_before_time = State.refresh_if_loaded_before(screen_id)
+
+    case refresh_if_loaded_before_time do
+      nil -> false
+      _ -> DateTime.compare(client_refresh_time, refresh_if_loaded_before_time) == :lt
+    end
+  end
 end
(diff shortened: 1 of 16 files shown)


lib/screens/util.ex
```diff
@@ -162,4 +164,34 @@ defmodule Screens.Util do
   def route_type_from_id("CR-" <> _), do: :commuter_rail
   def route_type_from_id("Boat-" <> _), do: :ferry
   def route_type_from_id(_), do: :bus
+
+  def outdated?(screen_id, client_refresh_timestamp) do
```
jzimbel-mbta, line 168: To save on requests, what do you think about having the screens config cache GenServer incorporate the `LAST_DEPLOY` value into config? That is, it would fetch the config JSON as usual, but also fetch `LAST_DEPLOY`. Then, it would iterate through each screen in the config and set its `refresh_if_loaded_before` to the `LAST_DEPLOY` time if that's later than the stored timestamp.

This way, we would only need to check `LAST_DEPLOY` every 15 seconds when the cached config gets updated, rather than checking it once per screen data request.

cmaddox5: That's a great idea. I added a new struct attribute to `Screens.Config.State` that holds the timestamp and moved all of the `DateTime` comparison logic to `:refresh_if_loaded_before` in the GenServer.

lib/screens/util/last_deploy/s3_fetch.ex
```diff
@@ -0,0 +1,32 @@
+defmodule Screens.Util.LastDeploy.S3Fetch do
+  @moduledoc false
+
+  @bucket "mbta-screens"
+
+  def get_last_deploy_time do
+    env = Application.get_env(:screens, :environment_name, "screens-prod")
```
hannahpurcell, line 7: If environment name is not found, it defaults to prod. I noticed elsewhere, we vary in whether we default to dev or prod. Why did you choose prod here?

hannahpurcell (approved):
Ok nice! Some of these files I've never touched before (config_cache) so I learned some things from this

lib/screens/util/last_deploy/s3_fetch.ex
cmaddox5: I can't speak for the other places, but here I don't really have a good reason for defaulting to prod. In a couple of the other places, it is dealing with where assets are stored in S3. That seems fine because that will allow us to see whatever assets would show on live screens by default. The same argument could be made here: if you can't figure out what ENV you're in, just look at the time prod last deployed and use that.

PR #1601, opened 2023-01-17
Only store alerts that are valid for Screenplay in memcached

hannahpurcell:
Notion [task](https://www.notion.so/mbta-downtown-crossing/Screenplay-Alerts-Polish-d3f44e3eec4346a19f31f24f75fa1ae6?p=369ad44cc004411b8d9a173ae80b07ff&pm=s).

We want Screenplay to track:
- Which alerts are actually currently visible on screens
- Which alerts are valid candidates on screens (though they may not have been picked because there's no room on the screen)

Basically, DON'T cache any alerts that are invalid candidates.

This is an attempt to save not-visible-but-yes-valid candidates to the cache as alerts with an empty screen list. Is that too wonky?

- [ ] Tests added?

### DIFF
--- a/lib/screens/screens_by_alert/memcache.ex
+++ b/lib/screens/screens_by_alert/memcache.ex
@@ -51,14 +51,14 @@ defmodule Screens.ScreensByAlert.Memcache do
   end
 
   @impl true
-  def put_data(screen_id, alert_ids) when is_binary(screen_id) and is_list(alert_ids) do
+  def put_data(screen_id, alert_ids, store_screen_id) when is_binary(screen_id) and is_list(alert_ids) do
     now = System.system_time(:second)
 
     # To avoid bottlenecks and unnecessarily blocking the caller, run in a separate task process
     _ =
       Task.Supervisor.start_child(TaskSupervisor, fn ->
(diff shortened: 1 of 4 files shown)


lib/screens/screens_by_alert/memcache.ex
```diff
@@ -143,14 +141,14 @@ defmodule Screens.ScreensByAlert.Memcache do
...
-  defp timestamped_screens_updater_fn({_id, now} = new_timestamped_screen_id) do
+  defp timestamped_screens_updater_fn({_id, now} = new_timestamped_screen_id, store_screen_id) do
     fn timestamped_screen_ids ->
       unexpired_ids =
         Enum.reject(timestamped_screen_ids, fn {_id, timestamp} ->
           timestamp + @screens_ttl < now
         end)
 
-      [new_timestamped_screen_id | unexpired_ids]
+      if store_screen_id, do: [new_timestamped_screen_id | unexpired_ids], else: [unexpired_ids]
```
hannahpurcell, line 151:
```suggestion
      if store_screen_id, do: [new_timestamped_screen_id | unexpired_ids], else: unexpired_ids
```

PR #1604, opened 2023-01-19
 tweak: Filter upcoming alerts for Alerts widget

cmaddox5:
**Asana task**: [[GL E-Ink V2] Filter out upcoming alerts](https://app.asana.com/0/1185117109217413/1203637230901491/f)

This PR changes the Alerts widget logic a bit so that we don't get upcoming alerts from the API. Upcoming alerts should not show on any screen, so there is no need for the CandidateGenerator to create a candidate for them. This means that a lot of logic in the `WidgetInstance.Alert` could be removed. We no longer need any conditional checking if the alert is active. It always will be. 

This is version 2 of this task. [Part 1](https://github.com/mbta/screens/pull/1599) used a API filter but that filter is buggy at the moment (see [thread](https://mbta.slack.com/archives/GP5QJM98A/p1674145757579809?thread_ts=1674145440.912979&cid=GP5QJM98A)). To avoid leaving off alerts, I went this route instead. 

- [x] Tests added?


jzimbel-mbta (approved):
Aw, too bad the API has that bug right now.

PR #1608, opened 2023-01-25
feat: DUP v2 Departures

cmaddox5:
**Asana task**: [[Widgetize DUPs] Departures Widget](https://app.asana.com/0/1185117109217413/1203279317437010/f)

This PR adds components and logic changes needed for DUPs to display the Departures widget. In general, the departures widget instance code did not need to change outside of a few specific cases for DUPs. Existing apps will remain unchanged. I was _almost_ able to use the existing frontend code, too, but there are just enough differences that making separate components made more sense. 

If you need a working config, let me know on Slack. 

v1:
v1

v2:
v2

- [ ] Tests added?


hannahpurcell (commented):
Couple of code comments, couple of UX comments.

It seems that the v2 version is always behind the predictions of the v1 version. Sometimes a couple minutes behind. Why is that? Is it avoidable? Is it a big deal?

Broadway. Top is V2, bottom is V1
Screenshot 2023-01-31 at 2 43 43 PM

Back Bay. Top is V2, bottom is V1
Screenshot 2023-01-31 at 2 48 59 PM

The v2 version often jumps from 1 minute to BRD without hitting ARR.

Not always, but sometimes the long-headsign paging gets out of sync. The rows should page in sync.

Jan-31-2023 14-51-14

cmaddox5:
> It seems that the v2 version is always behind the predictions of the v1 version. Sometimes a couple minutes behind. Why is that? Is it avoidable? Is it a big deal?

Great catch! Took all morning, but I was finally able to track down why this was happening. It seems to be a bug in the v2 departures across all screens. When we are grabbing a time to show on the screen, we default to `departure_time`. v1 defaults to `arrival_time`. After a discussion after standup, we should be using `arrival_time`. This should be more in line with what riders expect the time to mean on screens. This should also fix the issue with the rows skipping `ARR`. 

> Not always, but sometimes the long-headsign paging gets out of sync. The rows should page in sync.

I was able to fix this by lifting the `useCurrentPage` hook call all the way up so that every departures component uses the same value.

hannahpurcell:
> When we are grabbing a time to show on the screen, we default to `departure_time`. v1 defaults to `arrival_time`.

I squealed "omg!!" What a tricky bug to find! And one that explains why CR departures were so challenging to deal with! Wow, this is a big deal

hannahpurcell (commented):
Excellent! The updates look good to me.

I had forgotten this when I started reviewing, but realized it along the way: this large PR could have been broken into smaller chunks. Possibly like this: backend work + tests, frontend work without styling, then styling.

The code coverage bot suggests there are still some untested lines that were added with this PR. However, I think this bot is way less helpful than it could be, so I'm trying to look for a better code coverage bot that could actually tell us which lines need coverage _without_ needing to download that coverage report. I'd prefer that coverage in this PR, but if you don't get to it before vacay, I'll approve and we can merge anyway.

cmaddox5:
> Excellent! The updates look good to me.
> 
> I had forgotten this when I started reviewing, but realized it along the way: this large PR could have been broken into smaller chunks. Possibly like this: backend work + tests, frontend work without styling, then styling.

I agree. When I decided that the frontend needed its own components, I should have paused and pushed up a PR for the backend. In my mind, I wasn't changing much because a lot of code was able to be reused so the diff really surprised me.

> The code coverage bot suggests there are still some untested lines that were added with this PR. However, I think this bot is way less helpful than it could be, so I'm trying to look for a better code coverage bot that could actually tell us which lines need coverage _without_ needing to download that coverage report. I'd prefer that coverage in this PR, but if you don't get to it before vacay, I'll approve and we can merge anyway.

Waiting on another report, but I think I was able to turn a lot of the code I added to covered. Are you wanting to get code that was unchanged covered, too, or just focus on what was added?

hannahpurcell:
> Are you wanting to get code that was unchanged covered, too, or just focus on what was added?

Just focus on what was added, thanks! Will approve shortly

hannahpurcell (approved):
Thanks for the added test coverage 👍

PR #1614, opened 2023-01-27
Add type annotations to v2 app header components

jzimbel-mbta:
**Asana task**: ad hoc

I did some cleanup while working on the DUP header task, and thought it would be easier to review in a separate PR.

This adds type annotations to the base `NormalHeader` component, and updates related code that uses it.
TypeScript still complains about the `ref` prop on some internal components—I wasn't able to resolve this without making a mess of the code, so I'm leaving it alone.

One small "bug" the new type annotations revealed: the base component does not support `"x"` as an icon name, but the e-ink header component permitted it. I double checked, and this icon isn't used anywhere, so I removed it from the permitted values.

- [ ] Tests added?

### DIFF
--- a/assets/src/components/v2/dup/normal_header.tsx
+++ b/assets/src/components/v2/dup/normal_header.tsx
@@ -9,9 +9,9 @@ const NormalHeader = ({ icon, text, time }) => {
       icon={icon}
       text={text}
       time={time}
-      showUpdated={false}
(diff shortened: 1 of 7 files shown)


PaulJKim (approved):
LGTM!

PaulJKim:
Oh, but does look like some unit tests are broken in this branch

PR #1615, opened 2023-01-27
feat: Add pre-fare layout for two side-by-side full-screen takeovers

jzimbel-mbta:
**Asana task**: ad hoc

We didn't end up needing this layout option for this weekend's disruption, but I already added the code and we might need it in the future! ...Maybe!

- [x] Tests added?


cmaddox5 (approved):
🚀

PR #1616, opened 2023-01-31
RFC for DUP alert backend

cmaddox5:
**Asana task**: [[DUP v2] Plan backend for DUP alerts](https://app.asana.com/0/1185117109217413/1203830054341498/f)

This RFC proposes the plan for implementation of the alerts backend for V2 DUP. The dev that works on the implementation will still need to lean heavily on the [DUP alert spec](https://www.notion.so/mbta-downtown-crossing/DUP-Alert-Widget-Specification-a82acff850ed4f2eb98a04e5f3e0fe52), but this should be a good foundation.


docs/tech_specs/dup_v2_alert_widget_backend.md
```diff
@@ -0,0 +1,141 @@
...
+        opts
+        |> Alert.fetch()
+        |> Enum.filter(&relevant?/1)
+        |> choose_alert()
+end
+
+alerts = first_section_alerts ++ second_section_alerts
+```
+
+If screen is configured with only one section, `alerts` will contain up to one item. If there are two sections configured, `alerts` will contain up to two items.
```
handorff, line 72: Why the different behavior for screens with one vs two sections? In general, we'll still only show one alert _per screen_, even if there are multiple sections.

One known exception: We do have some special cases where we handle combinations of multiple shuttle/suspension alerts extending west starting at Kenmore, which would require being able to see all of those alerts at once.

cmaddox5: That behavior is how v1 currently deals with an alert being active in each section [here](https://github.com/mbta/screens/blob/master/lib/screens/dup_screen_data/data.ex#L87). If there are two sections in primary departures and each one has a relevant alert, a fullscreen alert takes the place of all 3 rotations (as of this comment, you can see that in place [here](https://screens-dev-green.mbtace.com/screen/DUP-Haymarket)). Outside of Kenmore (which we know is its own can of worms), this will only be in play for Haymarket. It is the only other screen that is configured with 2 subway sections, which is the only mode we fetch alerts for. 

In my opinion, this does feel a little aggressive. For example, if there are two delays up to 30 minutes at Haymarket, one for OL one for GL), all 3 rotations show a fullscreen alert and no departures are displayed. So I would understand wanting to move away from this logic/tweaking it so this only happens when all predictions are suspended.

docs/tech_specs/dup_v2_alert_widget_backend.md
```diff
@@ -0,0 +1,141 @@
...
+
+`serialize_takeover_alert/1` returns the following map:
+
+```
+%{
+    alert_text: %FreeTextLine{icon: :warning, text: ...},
+    color: :red | :orange | :green | :blue,
+    remedy: %FreeTextLine{icon: :shuttle | nil, text: ...}
+}
+```
```
jzimbel-mbta, line 125: I think this will also need to include values to render the alert header. A nested map might make sense, something like
```ex
%{
  ...
  header: %{
    icon: :logo,
    text: String.t(),
    time: String.t(),
    pattern: :hatched | :x | :chevron,
    color: :red | :orange | :green | :blue | :silver | :purple | :yellow
  }
}
```
(based on what I saw in the existing DUP header component and CSS)

jzimbel-mbta (commented):
Looks good! Taking note of these nice markdown features I didn't know about... :writing_hand::nerd_face:

docs/tech_specs/dup_v2_alert_widget_backend.md
```diff
@@ -0,0 +1,141 @@
...
+
+# Guide-level explanation
+
+[guide-level-explanation]: #guide-level-explanation
+
+The DUP alerts widget needs to resolve a scenario specific to DUP screens: the screen "rotates" between three separate pages. When an applicable[^1] alert exists, each rotation has specific logic that determines which "type" of alert widget to display: `partial`[^2] or `takeover`[^3].
+
+### CandidateGenerator
+
+`CandidateGenerator.Dup` is responsible for deciding what alert should appear on a screen. It fetches all active alerts affecting the screen's `stop_ids` listed in each section of the `primary_departures` config, filters out all alerts that do not directly affect the current stop (no downstream alerts), and selects the alert with the highest priority[^4]. If an alert (or alerts if configured with two sections) meets all criteria, the `CandidateGenerator` will then create three `WidgetInstance`s for the alert(s) (one for each rotation).
```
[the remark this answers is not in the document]
cmaddox5: Yep! I thought footnotes would work well here so I looked it up and they exist!

docs/tech_specs/dup_v2_alert_widget_backend.md
handorff: It sounds like this is a divergence from v1, then, but (at least for now) in v2 we'll just show one alert per screen.

I can imagine deciding later that we might want to add some more cases -- for example we might check whether the current alerts eliminate all service at a station and do something different -- but none of that is decided yet, and that would come after we ship a first pass at the widgetized DUP app.

docs/tech_specs/dup_v2_alert_widget_backend.md
```diff
@@ -0,0 +1,141 @@
...
+    |> choose_alert()
+
+second_section_alerts = []
+if length(sections) == 2 do
+    second_section_alerts =
+        opts
+        |> Alert.fetch()
+        |> Enum.filter(&relevant?/1)
+        |> choose_alert()
+end
```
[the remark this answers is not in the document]
cmaddox5: Ahh, good point. Wasn't thinking about that when I wrote this.

docs/tech_specs/dup_v2_alert_widget_backend.md
```diff
@@ -0,0 +1,141 @@
...
+
+If screen is configured with only one section, `alerts` will contain up to one item. If there are two sections configured, `alerts` will contain up to two items.
+
+With this list, we create three `WidgetInstance` objects for each rotation:
+
+```
+[
+  %DupAlert{screen: config, rotation_index: 0, alerts: alerts},
+  %DupAlert{screen: config, rotation_index: 1, alerts: alerts},
+  %DupAlert{screen: config, rotation_index: 2, alerts: alerts}
```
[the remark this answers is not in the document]
cmaddox5: That's a good idea. No need to use two different types for this.

docs/tech_specs/dup_v2_alert_widget_backend.md
cmaddox5: Didn't even think about that. Added that and a little blurb about what each value is. Let me know what you think.

docs/tech_specs/dup_v2_alert_widget_backend.md
cmaddox5: When you say this is a divergence, do you mean the logic in the doc or the suggestion in my comment? In v2, I mentioned that we still will only mention one alert even if two are present in the `alerts` attribute. The second alert is only used to determine if we need a fullscreen alert on all rotations.

docs/tech_specs/dup_v2_alert_widget_backend.md
```diff
@@ -0,0 +1,151 @@
+- Feature Name: `dup_v2_alert_widget_backend`
+- Start Date: 2023-01-30
+- RFC PR: [#1616](https://github.com/mbta/screens/pull/1616)
```
jzimbel-mbta, line 3: Another GH markdown feature: If you just type #1616, it will recognize it as a PR/issue reference number for this repo. So this can just be
```suggestion
- RFC PR: #1616
```

docs/tech_specs/dup_v2_alert_widget_backend.md
```diff
@@ -0,0 +1,151 @@
...
+
+```
+%{
+    alert_text: %FreeTextLine{icon: :warning, text: ...},
+    color: :red | :orange | :green | :blue,
+    remedy: %FreeTextLine{icon: :shuttle | nil, text: ...},
+    header: %{
+        icon: :logo,
+        text: String.t(),
+        time: String.t(),
```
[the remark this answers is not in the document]
jzimbel-mbta: Oops, my bad. I keep forgetting that the pattern graphic replaces the time. Thanks for catching that!

docs/tech_specs/dup_v2_alert_widget_backend.md
```diff
@@ -0,0 +1,151 @@
...
+
+`CandidateGenerator.Dup` is responsible for fetching relevant alerts from the API. The API query will contain an `opts` array for each section:
+
+```
+opts = [
+      stop_ids: <stop_ids from config>,
+      route_ids: <route_ids from config>,
+      route_types: ~w[light_rail subway]a
+    ]
+```
```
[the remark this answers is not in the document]
cmaddox5: Good question. At the time of writing this, I was thinking about `primary_departures` in the existing config. The main reason being that DUP v1 screens use those values. That's not a very good reason though. For flexibility, we would probably want this widget to have it's own config. Then, if for any reason we want to include alerts for stops/routes that are not in the `primary_departures`, we could easily do that. Thoughts?

docs/tech_specs/dup_v2_alert_widget_backend.md
cmaddox5: Hmm. For some reason, that doesn't seem to work in this file. I've seen it work many times in comments, but it doesn't seem to have that in md files themselves.

docs/tech_specs/dup_v2_alert_widget_backend.md
cmaddox5: I think this result is actually a bug. When I pull up Haymarket with these alerts, I'm getting an Elixir error. We do not have a function set up to handle a `:boundary` alert on a multi-line screen. That scenario was only set up for `:inside`. Based on that, I'm guessing we do _not_ want this. 

The scenario I mentioned above where there is an alert in each section on a screen, we see the following:

Screenshot 2023-02-02 at 12 57 01 PM

I think in most cases, this could be ok. Mostly if predictions are suspended because of the alerts. However, this image was captured with two delay alerts. While the text is true, I'm not sure a takeover on all rotations is the appropriate response to that scenario. Regardless, this scenario is the reason for the paragraph that was originally commented on in this thread.

docs/tech_specs/dup_v2_alert_widget_backend.md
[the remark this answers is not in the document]
cmaddox5: Decided to go with a new `alerts` config for this.

hannahpurcell (approved):

PR #1617, opened 2023-02-01
feat: DUP Evergreen

cmaddox5:
**Asana task**: [[DUP v2] Evergreen widget](https://app.asana.com/0/1185117109217413/1203830054341506/f)

This PR adds all functions needed to allow configured evergreen content to show on DUPs. Each slot on each rotation is set up with the correct dimensions for the content to display properly.


assets/css/dup_v2.scss
```diff
@@ -12,6 +12,7 @@
 
 @import "v2/placeholder";
 @import "v2/dup/normal_header";
+@import "v2/dup/evergreen/image";
```
hannahpurcell, line 15: Only `image.scss` is imported here, not `video.scss`. 

Since the two scss files are essentially identical, can they be consolidated into just `evergreen.scss`?

hannahpurcell (approved):
One optional consolidation suggestion, or a one-line fix

assets/css/dup_v2.scss
cmaddox5: Good idea. I don't see `image` and `video` ever needing different dimensions so separating them is unnecessary.

PR #1618, opened 2023-02-02
feat: Report elevator closure alerts to screens-by-alert

jzimbel-mbta:
**Asana task**: [Add elevator closure alerts to Screenplay](https://app.asana.com/0/1185117109217413/1203785095165730/f)

The Elevator Status widget now reports the IDs of all alerts that make it onto its limited number of pages.

Since Elevator Status has a kind of mini-version of the overall page serialization logic, where some content gets "cut" under certain conditions, we're forced to serialize the widget a second time to get an accurate picture of which alerts it's showing.

---

Due to recent changes that came from Screenplay, I needed to adjust the V3Api module to standardize how it builds the API's url. It now properly forms the url even when the base URL doesn't have a trailing `/`.

- [x] Tests added?

### DIFF
--- a/lib/screens/v3_api.ex
+++ b/lib/screens/v3_api.ex
@@ -69,11 +69,18 @@ defmodule Screens.V3Api do
   end
 
   defp build_url(route, params) when map_size(params) == 0 do
-    base_url() <> route
+    base_url()
+    |> URI.parse()
+    |> URI.merge(route)
+    |> URI.to_string()
   end
 
   defp build_url(route, params) do
-    "#{base_url()}#{route}?#{URI.encode_query(params)}"
+    base_url()
+    |> URI.parse()
+    |> URI.merge(route)
+    |> URI.to_string()
+    |> then(&"#{&1}?#{URI.encode_query(params)}")
   end
 
   defp base_url do
--- a/lib/screens/v2/alert_widget_instance.ex
+++ b/lib/screens/v2/alert_widget_instance.ex
@@ -1,7 +1,7 @@
 defprotocol Screens.V2.AlertWidgetInstance do
   @type alert_id :: String.t()
 
-  @doc "Gets the ID of the alert that this widget shows."
-  @spec alert_id(t) :: alert_id()
-  def alert_id(widget)
+  @doc "Gets the ID(s) of the alert(s) that this widget shows."
+  @spec alert_ids(t) :: list(alert_id())
+  def alert_ids(widget)
 end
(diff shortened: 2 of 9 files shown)


lib/screens/v3_api.ex
```diff
@@ -69,11 +69,18 @@ defmodule Screens.V3Api do
...
+    |> URI.to_string()
   end
 
   defp build_url(route, params) do
-    "#{base_url()}#{route}?#{URI.encode_query(params)}"
+    base_url()
+    |> URI.parse()
+    |> URI.merge(route)
+    |> URI.to_string()
+    |> then(&"#{&1}?#{URI.encode_query(params)}")
```
hannahpurcell, line 83: why these updates?

hannahpurcell (commented):
Oh this is so crisp! Simple, succinct. One question about a seemingly unrelated change to the api url

lib/screens/v3_api.ex
jzimbel-mbta: From the PR description:
> Due to recent changes that came from Screenplay, I needed to adjust the V3Api module to standardize how it builds the API's url. It now properly forms the url even when the base URL doesn't have a trailing `/`.

I started getting an `:nxdomain` error in all the V3 API requests from Screens when I started it up after setting new env vars for the [values that Screenplay now requires](https://github.com/mbta/screenplay/pull/206).

I guess both apps check the same env var for the V3 API base URL, and one (Screens) [expects the base URL to end with a `/`](https://github.com/mbta/screens/blob/master/lib/screens/v3_api.ex#L76) while the other (Screenplay) [does not](https://github.com/mbta/screenplay/blob/master/lib/screenplay/v3_api.ex#L69). By using functions from the `URI` module, we can handle both cases in an ISO-standard way instead of doing raw string concatenation and having the app break when a trailing slash is present/missing.

jzimbel-mbta: Actually! I just noticed a bug in Screenplay code.
- If we make a V3 API request without any query params, it does not put a slash between the base URL and the resource. [link to line](https://github.com/mbta/screenplay/blob/master/lib/screenplay/v3_api.ex#L65)
- If we make a V3 API request with query params, it _does_ put a slash between the base URL and the resource. [link to line](https://github.com/mbta/screenplay/blob/master/lib/screenplay/v3_api.ex#L69)

We should probably use the `URI` functions over there as well to avoid these kinds of bugs with constructing URLs.

hannahpurcell (approved):

PR #1619, opened 2023-02-03
fix: Make admin textareas less bouncy

jzimbel-mbta:
**Asana task**: ad hoc

Make the textareas less eager to show you what they've got.

- [ ] Tests added?


cmaddox5 (approved):
Thank you!

PR #1621, opened 2023-02-07
[tweak DUPs] Consolidate duplicate React components

hannahpurcell:
**Asana task**: [[DUP v2] Consolidate React components](https://app.asana.com/0/1185117109217413/1203887218424289)

One reusable wrapper splits the rotation index from the widget propnames and passes that index as a separate prop instead. I left the value as a string, since we seem to primarily use the string version of the index elsewhere. But can shift to using a number if there's a good reason for it.

Many file changes boil down to this:
- New wrapper lives in the dup v2 components folder
- Copied the "_one" version of each file type and made it generic (no mention of rotation)
- Fixed imports in the dup.tsx file


assets/src/apps/v2/dup.tsx
```diff
@@ -16,34 +16,27 @@ import Placeholder from "Components/v2/placeholder";
...
-  body_normal_two: NormalBodyTwo,
-  body_split_zero: SplitBodyZero,
-  body_split_one: SplitBodyOne,
-  body_split_two: SplitBodyTwo,
+  rotation_normal_zero: splitRotationFromPropNames(RotationNormal, "zero"),
+  rotation_normal_one: splitRotationFromPropNames(RotationNormal, "one"),
+  rotation_normal_two: splitRotationFromPropNames(RotationNormal, "two"),
+  rotation_takeover_zero: splitRotationFromPropNames(RotationTakeover, "zero"),
+  rotation_takeover_one: splitRotationFromPropNames(RotationTakeover, "zero"),
+  rotation_takeover_two: splitRotationFromPropNames(RotationTakeover, "zero"),
```
panentheos, line 33: This breaks the pattern of the others, should these be "zero", "one", "two" as well?

assets/src/components/v2/dup/dup_rotation_wrapper.tsx
```diff
@@ -0,0 +1,16 @@
+import React, { ReactNode } from "react";
+
+export const splitRotationFromPropNames = (WrappedComponent: React.ElementType, rotation: string) => {
+  return class extends React.Component {
+    render() {
+
+      const modifiedProps = Object.keys(this.props).reduce((newProps: any, currentKey: string) => {
```
panentheos, line 7: We could use `_.mapKeys` to do this transformation, which would probably be more compact.

assets/src/components/v2/dup/dup_rotation_wrapper.tsx
```diff
@@ -0,0 +1,16 @@
...
+        const newKey = currentKey.replace("_one", "").replace("_two", "").replace("_zero", "")
```
panentheos, line 8: This should work, but since we only ever expect one of them at a time, what about ```replace(`_${rotation}`, "")``` as a more explicit option?

assets/src/components/v2/dup/dup_rotation_wrapper.tsx
```diff
@@ -0,0 +1,16 @@
+import React, { ReactNode } from "react";
+
+export const splitRotationFromPropNames = (WrappedComponent: React.ElementType, rotation: string) => {
+  return class extends React.Component {
```
panentheos, line 4: This wrapper could be a function component, assuming that's preferred.

assets/src/apps/v2/dup.tsx
hannahpurcell: Copy pasta error! 🙇

assets/src/components/v2/dup/dup_rotation_wrapper.tsx
hannahpurcell: oooo nifty indeed

panentheos (approved):
Looks great! I did not do any additional testing of this while reviewing.

assets/src/components/v2/dup/dup_rotation_wrapper.tsx
```diff
@@ -0,0 +1,11 @@
+import _ from "lodash";
+import React from "react";
+
+export const splitRotationFromPropNames = (WrappedComponent: React.ElementType, rotation: string) => {
+  return (props: any) => {
+      
+    const modifiedProps = _.mapKeys(props, (_value, key) => key.replace(`_${rotation}`, ""))
```
jzimbel-mbta, line 7: Super minor suggestion: a regex pattern could be a bit safer as you can specify that the rotation needs to appear at the end of the key string:
```suggestion
    const rotationSuffixPattern = new RegExp(`_${rotation}$`);
    const modifiedProps = _.mapKeys(props, (_value, key) => key.replace(rotationSuffixPattern, ""))
```

For example:
```js
const rot = "one";
// The '$' at the end of the pattern anchors it to the end of the string
const re = new RegExp(`_${rot}$`);

console.log( "foo_one".replace(re, "") );
> "foo"

// Does not get replaced since it's not at the end of the string
console.log( "foo_one_x".replace(re, "") );
> "foo_one_x"
```

jzimbel-mbta (approved):
Good to go! Just a small/optional suggestion.

assets/src/components/v2/dup/dup_rotation_wrapper.tsx
hannahpurcell: It doesn't hurt to be more precise! 👌

PR #1622, opened 2023-02-10
DUP V2 alerts frontend stage 1: file organization

hannahpurcell:
**Asana task**: [title](url)

The code changes for this work were extensive, so I'm splitting into separate PRs. This one is all about file organization: I copied files from dup-v1 folders to dup-v2 locations. One exception: the code for FreeTextLine was put in a generic V2 spot because this is actually reused throughout our apps, but we just kept the code in the dup folder for some reason.

Also, the scss files are no longer 100% identical: tweaked some elements because the dup design has changed slightly (or to fix that bug in partial alerts where the `g` in Orange Line and the `y` in delay got cut off)

Also, added updated SVG for alert header.


cmaddox5 (approved):
Looks good! New location makes a lot more sense.

PR #1625, opened 2023-02-14
DUP V2 alerts frontend, phase 2: Partial & Takeover components

hannahpurcell:
**Asana task**: [[DUP v2] Partial + Takeover Alerts Frontend](https://app.asana.com/0/1185117109217413/1203830054341502)

Screenshot 2023-02-14 at 8 59 30 AM

To visually test, need to:
- In lib/screens/v2/candidate_generator/dup.ex, need to comment out departures and header instances. Add `%Placeholder{slot_names: [:full_rotation_one], color: :blue}` to the placeholder instances array.
- Make dummy data in lib/screens/v2/widget_instance/placeholder.ex. I added:
```
    def serialize(%Placeholder{color: :blue}), do: %{
      text: FreeTextLine.to_json(%FreeTextLine{icon: :warning, text: ["No", %{format: :bold, text: "Ferry service"}, "due to dock repairs"]}),
      remedy: FreeTextLine.to_json(%FreeTextLine{icon: :shuttle, text: [%{format: :bold, text: "Resumes Oct 12"}]}),
      header: %{
        text: "Long Wharf South",
        color: :teal # | :orange | :green | :blue
      }
    }
    def serialize(%Placeholder{color: :red}), do: %{
      text: FreeTextLine.to_json(%FreeTextLine{icon: :warning, text: ["No", %{format: :bold, text: "Blue Line" }, "trains"]}),
      color: :blue
    }
    def widget_type(%Placeholder{color: :blue}), do: :takeover_alert
    def widget_type(%Placeholder{color: :red}), do: :partial_alert
```

And then this alias is needed: `alias Screens.Config.V2.FreeTextLine`

### DIFF
--- a/assets/src/components/v2/dup/takeover_alert.tsx
+++ b/assets/src/components/v2/dup/takeover_alert.tsx
@@ -1,0 +1,90 @@
+import React from "react";
+
+import { Icon } from "../normal_header";
+import FreeText, { FreeTextType } from "./dup_free_text";
+import NormalHeader from "./normal_header";
+
+const LinkArrow = ({ width, color }: any) => {
+  const height = 40;
+  const stroke = 8;
+  const headWidth = 40;
+
+  const d = [
+    "M",
+    stroke / 2,
+    height / 2,
+    "L",
+    width - headWidth,
+    height / 2,
+    "L",
+    width - headWidth,
+    stroke / 2,
+    "L",
+    width - stroke / 2,
+    height / 2,
+    "L",
+    width - headWidth,
+    height - stroke / 2,
+    "L",
+    width - headWidth,
+    height / 2,
+    "Z",
+  ].join(" ");
+
+  return (
+    <svg
+      xmlns="http://www.w3.org/2000/svg"
+      viewBox={`0 0 ${width} ${height}`}
+      width={`${width}px`}
+      height={`${height}px`}
+      version="1.1"
+    >
+      <path
+        stroke={color}
+        strokeWidth={stroke}
+        strokeLinecap="round"
+        strokeLinejoin="round"
+        fill={color}
+        d={d}
+      />
+    </svg>
+  );
+};
+
+interface TakeoverAlertProps {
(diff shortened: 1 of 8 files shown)


assets/src/components/v2/dup/takeover_alert.tsx
```diff
@@ -0,0 +1,90 @@
+import React from "react";
+
+import { Icon } from "../normal_header";
+import FreeText, { FreeTextType } from "./dup_free_text";
+import NormalHeader from "./normal_header";
+
+const LinkArrow = ({ width, color }: any) => {
```
cmaddox5, line 7: I know only DUPs will use this particular SVG, but is it worth adding to the `bundled_svgs` folder? I feel it will be easier to find in the future if it is in its own folder. Totally up to you though.

assets/src/components/v2/normal_header.tsx
```diff
@@ -115,15 +115,23 @@ const NormalHeaderVersion: ComponentType<NormalHeaderVersionProps> = ({ version
   return <div className="normal-header-version">{version}</div>;
 };
 
+const NormalHeaderAccent = () => (
+  <div className="normal-header__accent-pattern-container">
+    <img className="normal-header__accent-pattern-image" src={imagePath(`dup-accent-pattern.svg`)} />
```
cmaddox5, line 120: To avoid referencing a specific app in a component shared by all, you could change `accentPattern?` to a string and pass in the `src` value. That will allow us to easily integrate different patterns into other apps if we ever need to.

cmaddox5 (approved):
Looks good! Just a couple of non-blocking suggestions.

assets/src/components/v2/dup/takeover_alert.tsx
hannahpurcell: Currently, since this is only used in one niche situation, I think keeping it here is fine. If and when it gets used elsewhere, we can move it for ease of use

PR #1631, opened 2023-02-24
Updated DUP simulation for v2

hannahpurcell:
**Asana task**: [[DUP v2] Make v2 DUP screens show up correctly in Screenplay](https://app.asana.com/0/1185117109217413/1203885464554557)

Adjusted the SimulationPage component to watch for an `alternateView` option, which adjusts the type of the widget from `screen_normal` to `simulation_screen_normal`. Might come in handy with other screens in the future.

This is the main work of the task, and there will be a small Screenplay task to follow.

- [ ] Tests added?

### DIFF
--- a/assets/src/components/v2/simulation_screen_container.tsx
+++ b/assets/src/components/v2/simulation_screen_container.tsx
@@ -11,7 +11,7 @@ import {
 
 interface SimulationScreenLayoutProps {
   apiResponse: ApiResponse;
-  opts: {[key: string]: any}
+  opts: { [key: string]: any };
 }
 
 const SimulationScreenLayout: ComponentType<SimulationScreenLayoutProps> = ({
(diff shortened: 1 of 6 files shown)


assets/src/components/v2/simulation_screen_container.tsx
```diff
@@ -11,21 +11,27 @@ import {
...
+  // Currently only applies to DUPs
+  const alternatePage = {...fullPage, type: "simulation_screen_normal"}   
 
   return (
     <div className="simulation-screen-centering-container">
       <div className="simulation-screen-scrolling-container">
         {apiResponse && (
           <div className="simulation__full-page">
-            <Widget data={fullPage ?? data} />
+            { opts.alternateView ? <Widget data={alternatePage} /> : <Widget data={fullPage ?? data} /> }
```
cmaddox5, line 34: It's not terribly difficult to follow now, but this might be a little more clear if the value of `data` is evaluated in an if/else block above the return.

cmaddox5 (approved):
Looks good! Just one suggestion.

PR #1637, opened 2023-03-07
feat: DUP v2 handle unreliable predictions - Backend

cmaddox5:
**Asana task**: [[DUP v2] Handle unreliable predictions (devops checkboxes)](https://app.asana.com/0/1185117109217413/1204081228664130/f)

Spec [link](https://www.notion.so/mbta-downtown-crossing/Departures-Widget-Specification-20da46cd70a44192a568e49ea47e09ac?pvs=4#cb82633f5d314bcfa8974b721d249bd1)

**Backend only**

This PR adds logic for handling the disabled modes checkboxes in the admin tool. This required a departure (heh) from the way we normally render a no data state for the departures widget.

All other V2 screen types display a single section in its departures widget, and that one section only ever contains one mode. DUPs can have multiple sections and each section can be any mode (but never more than one). This means we do not always want to show a full `DeparturesNoData` widget when one mode is disabled. For example, when subway predictions are disabled, we still want to show CR predictions at Back Bay.

To accomplish this, `Departures` needs to handle a `no_data_section`. This section will render similarly to our headway sections. It will contain a single row with a pill for the route and the text "Updates unavailable".  This will make it very easy to show the most accurate data for each section.

Important note: I did change the value for CR in the devops checkbox. It was set to `:commuter_rail`. While absolutely true, we use `:rail` much more often in other parts of the code. This change makes it easier for the devops config to match up with existing helper functions in other parts of the code.

- [X] Tests added?

### DIFF
--- a/lib/screens/v2/widget_instance/departures_no_data.ex
+++ b/lib/screens/v2/widget_instance/departures_no_data.ex
@@ -4,13 +4,13 @@ defmodule Screens.V2.WidgetInstance.DeparturesNoData do
   alias Screens.Config.Screen
   alias Screens.Config.V2.Alerts
 
-  defstruct screen: nil, show_alternatives?: nil, slot_name: nil, routes: []
+  defstruct screen: nil, show_alternatives?: nil, slot_name: nil, route_types: []
 
   @type t :: %__MODULE__{
           screen: Screens.Config.Screen.t(),
           show_alternatives?: boolean(),
           slot_name: atom(),
-          routes: list(atom())
+          route_types: list(atom())
         }
 
   def priority(_instance), do: [2]
@@ -19,7 +19,7 @@ defmodule Screens.V2.WidgetInstance.DeparturesNoData do
     %{
       show_alternatives: instance.show_alternatives?,
       stop_id: stop_id(instance),
-      routes: instance.routes
+      route_types: instance.route_types
     }
   end
 
--- a/lib/screens/v2/widget_instance/departures_no_data.ex
+++ b/lib/screens/v2/widget_instance/departures_no_data.ex
@@ -4,12 +4,12 @@ defmodule Screens.V2.WidgetInstance.DeparturesNoData do
   alias Screens.Config.Screen
   alias Screens.Config.V2.Alerts
 
-  defstruct screen: nil, show_alternatives?: nil, slot_names: []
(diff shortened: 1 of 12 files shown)


lib/screens/v2/widget_instance/departures_no_data.ex
```diff
@@ -4,11 +4,12 @@ defmodule Screens.V2.WidgetInstance.DeparturesNoData do
   alias Screens.Config.Screen
   alias Screens.Config.V2.Alerts
 
-  defstruct screen: nil, show_alternatives?: nil
+  defstruct screen: nil, show_alternatives?: nil, slot_names: []
```
hannahpurcell, line 7: Why is this pluralized? In `departures.ex`, this widget is created with: 

```
   Enum.map(slot_ids, fn slot_id ->
      if Enum.all?(sections, &(&1.type == :no_data_section)) do
        %DeparturesNoData{screen: config, slot_names: [slot_id]}
      else ...
```

so `slot_id` should always be just one id.

hannahpurcell (approved):
Fantastic, superb. One thought, but still approved

hannahpurcell:
Oh I forgot! I'm asking a product question about this [here](https://mbta.slack.com/archives/C039ARUM48H/p1678370566401749?thread_ts=1678215323.790149&cid=C039ARUM48H). If you're working on other stuff right now, please hold off on merging

lib/screens/v2/widget_instance/departures_no_data.ex
cmaddox5: I did that mostly to leave room for an app to give slot options for the widget to be valid in. While that has made sense for previous widgets (evergreen is the main one that needs this), I'm not sure that's actually valuable for this one. So I think I'll change it to a single slot.

cmaddox5:
> Oh I forgot! I'm asking a product question about this [here](https://mbta.slack.com/archives/C039ARUM48H/p1678370566401749?thread_ts=1678215323.790149&cid=C039ARUM48H). If you're working on other stuff right now, please hold off on merging

Decision from design is to include route types in the message. Only distinction that needs to be made is `:light_rail` will say "Green Line" instead of "Light Rail". All other types will be unchanged outside of formatting.

@hannahpurcell This is all we will need for the backend piece. I will change the no data component to use these new values in frontend branch.

hannahpurcell (approved):
thanks!

PR #1638, opened 2023-03-07
feat: DUP v2 handle unreliable predictions - Frontend

cmaddox5:
**Asana task**: [[DUP v2] Handle unreliable predictions (devops checkboxes)](https://app.asana.com/0/1185117109217413/1204081228664130/f)

Spec [link](https://www.notion.so/mbta-downtown-crossing/Departures-Widget-Specification-20da46cd70a44192a568e49ea47e09ac?pvs=4#cb82633f5d314bcfa8974b721d249bd1)

**Frontend only**

This PR adds the frontend component needed to render the new `no_data_section` in the `Departures` widget. This section is always just a route's pill, the text "Updates unavailable", and the `NoConnection` icon. If all sections are on a rotation are unavailable, the `NoData` component will render instead. 

Screenshot 2023-03-10 at 9 50 55 AM


assets/src/components/v2/dup/no_data.tsx
```diff
@@ -8,15 +8,19 @@ import NormalHeader from "./normal_header";
...
+}
+
+const NoData: ComponentType<Props> = ({ include_header }) => {
   let stationName = useOutfrontStation() || "Transit information";
   stationName = REPLACEMENTS[stationName] || stationName;
 
   return (
     <div className="no-data__container">
-      <NormalHeader text={stationName} />
+      {include_header && <NormalHeader text={stationName} />}
```
hannahpurcell, line 23: These changes set up the ability to provide `include_header` to no_data, but all the instances of it in `dup.tsx` are set to true, so were these changes needed?

assets/src/components/v2/dup/dup_free_text.tsx
```diff
@@ -3,7 +3,8 @@ import _ from "lodash";
 
 import { classWithModifier, classWithModifiers, imagePath } from "Util/util";
 
-const pillIcons = ["red", "blue", "orange", "green", "silver"];
+const textPills = ["red", "blue", "orange", "green", "silver"];
+const iconPills = ["cr", "bus"];
```
hannahpurcell, line 7: Ohhhhhh phoo... I just noticed that this file is an almost-duplicate of another v2 `free_text.tsx` file. Would you be interested in consolidating? If not, can make a clean-up task. I shouldn't have introduced this `dup_free_text.tsx` file. My bad

assets/src/components/v2/dup/no_data.tsx
cmaddox5: It is there to prevent the `departures_no_data` widget from rendering a duplicate header. The slot for `departures_no_data` is `:main_content_x` which means that a header candidate will still be given to the frontend to render. In the error state you added in `dup.tsx`, that specific no data state cannot render a header without hardcoding one in the component. So having this conditional allows each no data state to render what it needs to.

assets/src/components/v2/dup/dup_free_text.tsx
cmaddox5: I don't mind doing it. I think everything v2 is using the same one now. I tested some of the other apps (`ReconstructedTakeover` uses `FreeText`) and it looks good.

assets/src/components/v2/dup/no_data.tsx
hannahpurcell: Ohhh I see now. 💯

assets/src/components/v2/dup/dup_free_text.tsx
cmaddox5: Correct. We do technically get alerts for predictions, but it would only render if `alert_is_inline?/1` is `true` and that is hardcoded to `false` for now. I think it was left in and hardcoded to `false` so bus shelter alerts could be revisited, but that didn't happen as soon as we thought.

hannahpurcell (approved):
Looks great! Sorry, didn't mean to hold out on approving.

PR #1639, opened 2023-03-07
Empty states for DUPs

hannahpurcell:
**Asana task**: [[DUP v2] No data and disabled states](https://app.asana.com/0/1185117109217413/1204081228664132)

- Added 2 frontend states: loading & disabled/failure states.
- Carried over some functionality from DUP v1: the "debug by printing console logs to the screen" feature and the maintenance code feature (though we aren't using either right now).
- When adding LinkArrow to one of the components, I noticed the LinkArrow in takeover_alert needed fixing.

- [ ] Tests added?

### DIFF
--- a/assets/css/dup_v2.scss
+++ b/assets/css/dup_v2.scss
@@ -28,8 +28,8 @@
 
 @import "v2/multi_screen_page";
 @import "v2/dup/simulation";
-@import "v2/dup/debug.scss";
-@import "v2/dup/empty_state.scss";
+@import "v2/dup/debug";
+@import "v2/dup/empty_state";
 
 body {
   margin: 0px;
--- a/assets/css/dup_v2.scss
+++ b/assets/css/dup_v2.scss
@@ -28,6 +28,8 @@
 
 @import "v2/multi_screen_page";
 @import "v2/dup/simulation";
+@import "v2/dup/debug.scss";
+@import "v2/dup/empty_state.scss";
 
 body {
   margin: 0px;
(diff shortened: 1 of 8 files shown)


assets/css/dup_v2.scss
```diff
@@ -28,6 +28,8 @@
 
 @import "v2/multi_screen_page";
 @import "v2/dup/simulation";
+@import "v2/dup/debug.scss";
+@import "v2/dup/empty_state.scss";
```
cmaddox5, line 32: Drop file type for consistency.

```suggestion
@import "v2/dup/debug";
@import "v2/dup/empty_state";
```

assets/src/components/v2/dup/page_load_no_data.tsx
```diff
@@ -0,0 +1,39 @@
...
+import React, { ComponentType } from "react";
+import LinkArrow from "../bundled_svg/link_arrow";
+import Loading from "../bundled_svg/loading";
+import NormalHeader from "./normal_header";
+
+// Fix station name tags without rider-facing names
+const REPLACEMENTS = {
+  WTC: "World Trade Center",
+  Malden: "Malden Center",
+} as {[key:string]: string};
```
cmaddox5, line 11: Not sure the best place for it, but would it be worth pulling this bit into something common so this and `no_data` can use the same list? Or you could just leave it in this and export it to the other (or vice versa).

cmaddox5 (approved):
Looks great! Just a suggestion and a question.

PR #1641, opened 2023-03-10
Bidirectional departure sections (for DUP)

hannahpurcell:
**Asana task**: [[DUP v2] Add ability to configure bidirectional departures](https://app.asana.com/0/1185117109217413/1204083894737787)

`"bidirectional": true,` can be an attribute on a primary_departure section. The RL is easiest to test with, because you'll frequently see an Ashmont train back-to-back with a Braintree train, or two Alewife trains back-to-back. But the bidirectional feature only works for 2-row sections, not 4-row. So you'll need to make a fake configuration. (Just copy the secondary bus departures from the Broadway screen to the bottom of the primary_departures screen.)

- [x] Tests added?

### DIFF
--- a/lib/screens/v2/candidate_generator/dup/departures.ex
+++ b/lib/screens/v2/candidate_generator/dup/departures.ex
@@ -11,6 +11,7 @@ defmodule Screens.V2.CandidateGenerator.Dup.Departures do
   alias Screens.Stops.Stop
   alias Screens.Util
   alias Screens.V2.CandidateGenerator.Widgets
+  alias Screens.V2.Departure
   alias Screens.V2.WidgetInstance.Departures, as: DeparturesWidget
   alias Screens.V2.WidgetInstance.DeparturesNoData
 
@@ -138,7 +139,8 @@ defmodule Screens.V2.CandidateGenerator.Dup.Departures do
                               query: %Query{
                                 params: %Params{stop_ids: stop_ids, route_ids: route_ids} = params
                               },
-                              headway: headway
+                              headway: headway,
+                              bidirectional: is_bidirectional
                             } = section ->
       route =
         get_primary_route_for_section(stop_ids, route_ids, create_station_with_routes_map_fn)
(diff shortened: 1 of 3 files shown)


lib/screens/v2/candidate_generator/dup/departures.ex
```diff
@@ -149,8 +151,15 @@ defmodule Screens.V2.CandidateGenerator.Dup.Departures do
       else
         section_departures =
           case fetch_section_departures_fn.(section) do
-            {:ok, section_departures} -> section_departures
-            _ -> []
+            {:ok, section_departures} ->
```
PaulJKim, line 154: General question about DUPs: Can they have multiple 4-row sections? Or is always either two 2-row sections or one 4-row section?

Asking because I'm wondering if there's a way to do this logic without the addition of another config value 🤔

hannahpurcell: Yeah, a single DUP departures page can only have two 2-row sections or one 4-row section. (`primary_departures` creates 2 pages, and `secondary_departures` creates 1.)

It does need to be a config value because we will want to turn bidirectional departures "on" conditionally for some screens and not others. At least... that's what I interpreted the [task](https://app.asana.com/0/1185117109217413/1204083894737787) to mean, but now I'm wondering if that was a misinterpretation! I'll ask Paul H

hannahpurcell: Confirmed with Paul that we want to conditionally set some 2-row sections to bidirectional and others to time-based order.

PaulJKim: Ohh I see. I'm curious what the cases are where we'd want to set a 2-row section to time-based ordering! I could see a solution where we just _always_ take the first two predictions for the first two distinct destinations in the list of predictions and default to time-ordering when there's just one destination. That's essentially what we do for countdown clocks but I'm sure there's some DUP-specific nuances that I'm overlooking

hannahpurcell: Sounds like a question for the [task](https://app.asana.com/0/inbox/1199335143657941/1204083894737787/1204148904312449)!

PaulJKim: Took a look at the task thread and seems like there could be some cases where we want time ordering, or that there is at least some uncertainty. Config value seems to be the safer/more flexible solution in that case!

PaulJKim (approved):
LGTM!

PR #1646, opened 2023-03-15
fix: Use a sane timeout for Task.async_stream calls

jzimbel-mbta:
**Asana task**: [[tech debt] Use sane timeouts for Task calls](https://app.asana.com/0/1185117109217422/1204176045657639/f)

- [ ] Tests added?


cmaddox5:
Would it make more sense to set the timeout on each screen to be <= the refresh interval for that screen type? Some screens, pre-fare for example, refresh more often than every 30 seconds. Setting those timeouts to 30 seconds would cause requests to last longer than it would show on the screen anyway.

cmaddox5 (approved):

PR #1647, opened 2023-03-15
refactor: Define new SingleAlertWidget protocol and refactor BaseAlert to use it

jzimbel-mbta:
**Asana task**: ad hoc

This adds a new protocol for widgets that model a single alert.
The BaseAlert functions now access widget properties exclusively through these functions, so that it has a clearly defined set of requirements for its input values.

The `SingleAlertWidget` and `BaseAlert` protocol & module pair can be thought of as analogous to elixir's [`Enumerable` and `Enum`](https://github.com/elixir-lang/elixir/blob/main/lib/elixir/lib/enum.ex)—the protocol defines the contract for what types of values the module can work with.

- [ ] Tests added?

### DIFF
--- a/lib/screens/v2/alerts_widget.ex
+++ b/lib/screens/v2/alerts_widget.ex
@@ -1,0 +1,19 @@
+defprotocol Screens.V2.AlertsWidget do
+  @moduledoc """
+  Protocol for a widget that models zero or more alerts.
+  """
+
+  # https://hexdocs.pm/elixir/Protocol.html#module-fallback-to-any
+  @fallback_to_any true
+
+  @type alert_id :: String.t()
+
+  @doc "Gets the ID(s) of the alert(s) that this widget shows."
+  @spec alert_ids(t) :: list(alert_id())
+  def alert_ids(widget)
+end
+
+# https://hexdocs.pm/elixir/Protocol.html#module-fallback-to-any
+defimpl Screens.V2.AlertsWidget, for: Any do
+  def alert_ids(_), do: []
+end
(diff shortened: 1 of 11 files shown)


lib/screens/v2/screen_data.ex
```diff
@@ -484,10 +484,7 @@ defmodule Screens.V2.ScreenData do
     alert_ids =
       instance_map
       |> Map.values()
-      |> Enum.filter(fn widget_struct ->
-        not is_nil(AlertWidgetInstance.impl_for(widget_struct))
-      end)
-      |> Enum.flat_map(&AlertWidgetInstance.alert_ids/1)
+      |> Enum.flat_map(&AlertsWidget.alert_ids/1)
```
jzimbel-mbta, line 487: (Since `AlertsWidget.alert_ids/1` now returns `[]` for types that don't implement the protocol, we no longer need to first filter out non-alert widgets.)

hannahpurcell:
Assigning to Christian for now since I'm a bit behind on special cases work

lib/screens/v2/alerts_widget.ex
```diff
@@ -0,0 +1,19 @@
+defprotocol Screens.V2.AlertsWidget do
+  @moduledoc """
+  Protocol for a widget that models zero or more alerts.
```
hannahpurcell, line 3: In the moduledoc for these two different AlertsWidgets, could you add the widgets that qualify so far? Like, "for example: ". (We don't need to keep that list updated over time.)

PR #1650, opened 2023-03-16
feat: DUP v2 Overnight - Frontend

cmaddox5:
**Asana task**: [[DUP v2] Implement overnight mode for bus](https://app.asana.com/0/1185117109217413/1203876538101497/f)

This PR adds the DUP `OvernightDepartures` component, "AM" to overnight rows in the `NormalDepartures` widget, and an `OvernightSection` component. There are two variations of the `OvernightDepartures` component: one when there are still predictions on one of the rotations and one for if all rotations are overnight.

Screenshot 2023-03-16 at 8 58 05 AM

Screenshot 2023-03-16 at 8 58 01 AM

Screenshot 2023-03-16 at 8 59 37 AM

Screenshot 2023-03-16 at 9 17 59 AM

- [ ] Tests added?


assets/css/v2/dup/departures/time.scss
```diff
@@ -38,8 +38,19 @@
...
+  font-size: 138px;
+  line-height: 128px;
+}
+
+.departure-time__ampm {
+  margin-left: 20px;
+  font-weight: 400;
+  font-size: 96px;
+  line-height: 108px;
+  letter-spacing: 2px;
```
hannahpurcell, line 55: Where did this CSS come from? I don't see this in the figma

cmaddox5: https://www.figma.com/file/xCNFkMDWpq7y7cIH4V5dln/DUP-screens?node-id=520-10597&t=JNmkxOCFTeH9rxOe-4

assets/src/components/v2/dup/departures/overnight_section.tsx
```diff
@@ -0,0 +1,15 @@
...
+  return (
+    <div className="departures-section overnight-section">
+      <div className="overnight-section__row">
+        <FreeText lines={text} />
+      </div>
+    </div>
+  );
+};
+
+export default OvernightSection;
```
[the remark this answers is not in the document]
cmaddox5: https://www.figma.com/file/xCNFkMDWpq7y7cIH4V5dln/DUP-screens?node-id=520-9595&t=JNmkxOCFTeH9rxOe-4

assets/css/v2/dup/departures/time.scss
hannahpurcell: I feel dumb! I panned around and never found this section 😅

assets/css/v2/dup/departures/overnight.scss
```diff
@@ -0,0 +1,44 @@
...
+        margin: 0 40px 0 0;
+      }
+    }
+
+    .overnight-departures__text {
+      width: 1182px;
+      font-family: Inter, sans-serif;
+      font-weight: 600;
+      font-size: 138px;
+      line-height: 148px;
```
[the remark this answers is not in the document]
cmaddox5: This is actually an `overnight-section`, so it uses `FreeText` instead of that file. The `line-height` was not correct for `overnight-section`s so I got that fixed. Also gave `departure-section`s a static height of `416px` and reduced the text box width. 

With the change to section height, the divider was not always rendering in the right place under some section types. A more reliable way to position it is to flip `&:not(:last-child):after` to `&:not(:first-child):before`. So that's why that change in in the most recent commit.

hannahpurcell (approved):
Alrighty, I gotta stop scope-creepin with my review! Thanks for this work. Looks good

PR #1655, opened 2023-03-21
polish: Add header types DUP v2

cmaddox5:
**Asana task**: ad-hoc

WTC and Tufts overflow into the time when in the 12 o'clock hours. To prevent this, the header config can now be the string to display on the screen instead of just a `stop_id`. This gives us the flexibility needed to abbreviate words to make headers fit.

- [X] Tests added?

### DIFF
--- a/lib/screens/config/v2/dup.ex
+++ b/lib/screens/config/v2/dup.ex
@@ -33,14 +33,10 @@ defmodule Screens.Config.V2.Dup do
   end
 
   defp value_to_json(:header, %CurrentStopId{} = header) do
-    header
-    |> CurrentStopId.to_json()
-    |> Map.put(:type, :current_stop_id)
+    CurrentStopId.to_json(header)
   end
 
   defp value_to_json(:header, %CurrentStopName{} = header) do
-    header
-    |> CurrentStopName.to_json()
-    |> Map.put(:type, :current_stop_name)
+    CurrentStopName.to_json(header)
   end
 end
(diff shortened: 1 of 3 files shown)


PaulJKim (commented):
Might be a product question, but it _sort of_ seems like `current_stop_name` should behave like an override based on the description. As in, `if name is provided, use that. else fetch name using stop_id`. Just wanted to check if you think that could potentially make the config a bit easier to manage instead of having the two types. Thoughts?

cmaddox5:
> Might be a product question, but it _sort of_ seems like `current_stop_name` should behave like an override based on the description. As in, `if name is provided, use that. else fetch name using stop_id`. Just wanted to check if you think that could potentially make the config a bit easier to manage instead of having the two types. Thoughts?

That's a good point. I pulled over the logic from bus shelter, but it could be simplified to look at the name of the key instead of looking at a type first. I'll get that change together.

lib/screens/config/v2/dup.ex
```diff
@@ -21,7 +21,22 @@ defmodule Screens.Config.V2.Dup do
...
       evergreen_content: {:list, EvergreenContentItem}
     ]
+
+  defp value_from_json("header", %{"stop_id" => _} = header) do
+    CurrentStopId.from_json(header)
+  end
+
+  defp value_from_json("header", %{"stop_name" => _} = header) do
+    CurrentStopName.from_json(header)
+  end
```
PaulJKim, line 33: Just sanity checking myself here: would this mean that we prefer `stop_id` over `stop_name` based on the ordering of the pattern matching, or are we assuming that there will always only be either a `stop_id` _or_ `stop_name`?

PaulJKim (approved):
LGTM!

PR #1656, opened 2023-03-22
feat: DUP v2 Rail Overnight w/ active alert

cmaddox5:
**Asana task**: [[DUP v2] Implement overnight mode for rail](https://app.asana.com/0/1185117109217413/1203876538101495/f)

When an alert is present for rail, we won't always have predictions. If this occurs after the last schedule, we cannot guarantee that we are actually in overnight mode. To help with this, we are adding an additional check for subway and light rail: If an alert is active, only check for overnight mode if there are no more vehicles on the route. This is not a perfect solution, but this gets us closer to showing overnight exactly when we should. 

- [x] Tests added?


lib/screens/v2/candidate_generator/dup/departures.ex
```diff
@@ -308,17 +382,159 @@ defmodule Screens.V2.CandidateGenerator.Dup.Departures do
...
+         [%{type: type} | _] = routes,
+         alert,
+         now,
+         fetch_schedules_fn,
+         fetch_vehicles_fn
+       )
+       when type in [:subway, :light_rail] do
+    informed_route = get_section_route_from_alert(params.stop_ids, alert)
+    # If there are no vehicles operating on the route, assume we are overnight.
+    if fetch_vehicles_fn.(informed_route.id, nil) == [] do
```
hannahpurcell, line 460: Have you tested this out with an alert yet? For me, this fails because the output of `informed_route` is a string, and the dot access (`.id`) requires an atom. Alternatively, you could make it `informed_route[id]`. I think that'll work

cmaddox5: That `.id` should not have been there. I must have added that without thinking after doing my testing. The string returned is always a `route_id` (or `nil` which is why I added `not is_nil(informed_route)` to the conditional). That can then be used in the vehicle fetch.

hannahpurcell (approved):
Okaaayy! Short and sweet. Looks good

PR #1657, opened 2023-03-23
Cm/update ofm packaging dup v2

cmaddox5:
**Asana task**: [[DUP v2] Update and document OFM packaging process](https://app.asana.com/0/1185117109217413/1204083894737779/f)

Updated README with how to package DUP v2 for OFM. I followed the steps a couple of times for multiple screen IDs and everything looks good.


jzimbel-mbta (approved):
Apologies for the very slow review! Changes look good and I successfully tested the package locally. Good to go!

assets/src/components/v2/dup/version.tsx
```diff
@@ -0,0 +1 @@
+export const DUP_VERSION = "21.4.6.1";
```
hannahpurcell, line 1: I think we should bump the DUP version

assets/src/components/v2/dup/rotation_index.tsx
```diff
@@ -0,0 +1 @@
+export const ROTATION_INDEX = 1;
```
hannahpurcell, line 1: for dup v1, the rotation index was set in the file to 0, and the readme indicates we're still working with digits 0-2. Should this be 0?

hannahpurcell (commented):
Started looking this over just so I knew what was going on, but noticed a couple things I had questions / suggestions about

assets/src/components/v2/dup/rotation_index.tsx
cmaddox5: The committed index is not the index that is used in the packaging. In the `for` loop part executed to create the packaged zips, we export the index needed for each loop and that index is saved to this file. This file is mostly there because so that way we don't have to worry about it after the packaging is finished. One thing that might be useful is committing index 2 to the file. That will always be the last index saved so we would never need to worry about a diff when we run the script.

hannahpurcell: > This file is mostly there because so that way we don't have to worry about it after the packaging is finished.

Could you explain this differently? Not quite following. But cool, glad it's not a necessary value 👍

assets/src/components/v2/dup/version.tsx
cmaddox5: I agree, but the version is only used at the time of packaging. It is probably a good idea to create a PR when we package up the app that changes this string (that is what the last change to the v1 version file did but it wasn't done super consistently). That way, we can have a more reliable change log. What do you think?

assets/src/components/v2/dup/rotation_index.tsx
cmaddox5: A better solution might actually be just adding this file to `.gitignore`. If the file doesn't exist, the script will create it so no dev needs to worry about it existing before they can package the app.

assets/src/components/v2/dup/version.tsx
hannahpurcell: Ohhhh my bad. I forgot that's part of the packaging process. Cool, thanks!

assets/src/components/v2/dup/rotation_index.tsx
hannahpurcell: Thanks for the explain! Why do we need to save rotation_index in a file at all? Is it used later?

cmaddox5: The main reason is because the code for the packaged app needs access to the variable. After compilation, the file is no longer necessary. To help with that, I added it to `.gitignore` in [this PR](https://github.com/mbta/screens/pull/1719). With it ignored, we can just forget it ever existed and let the script worry about it.

PR #1658, opened 2023-03-23
DUP Special Alert cases

hannahpurcell:
**Asana task**: [[DUP v2] Handle alerts special cases](https://app.asana.com/0/1185117109217413/1203876547342482)

Screenshot 2023-03-23 at 1 14 05 PM
Screenshot 2023-03-23 at 1 51 00 PM

- [ ] Tests added?


lib/screens/v2/candidate_generator/dup/alerts.ex
```diff
@@ -127,14 +142,195 @@ defmodule Screens.V2.CandidateGenerator.Dup.Alerts do
...
+       [
+         %DupSpecialCaseAlert{
+           alert_ids: alert_ids,
+           serialize_map: %{
+             text: %Screens.Config.V2.FreeTextLine{
+               icon: :warning,
+               text: get_kenmore_special_text(branches, :partial_alert)
+             },
+             color: :green
+           },
```
cmaddox5, line 190: My gut tells me this is the wrong place for this bit. I get these are special cases, but putting `serialize` logic in the `CandidateGenerator` breaks the mold a bit on how we prepare `WidgetInstance`s. Is it possible to allow `DupSpecialCaseAlert.serialize` to handle this?

lib/screens/v2/candidate_generator/dup/alerts.ex
```diff
@@ -127,14 +142,195 @@ defmodule Screens.V2.CandidateGenerator.Dup.Alerts do
...
+       end}
+    else
+      {:normal, []}
+    end
+  end
 
-    {:normal, alerts}
+  @spec check_entities_for_stops(Alert.t(), list(%{branch: String.t(), stop: String.t()})) ::
+          atom()
+  def check_entities_for_stops(%{informed_entities: informed_entities}, stop_matchers) do
```
cmaddox5, line 268: The verb `check` implies a `boolean` return to me. Something like `get_branches_from_entities` seems a little closer to what the function is doing. What do you think?

lib/screens/v2/candidate_generator/dup/alerts.ex
```diff
@@ -127,14 +142,195 @@ defmodule Screens.V2.CandidateGenerator.Dup.Alerts do
...
+      Enum.any?(informed_entities, fn e ->
+        stop.stop === e.stop
+      end)
+    end)
+    |> Map.get(:branch)
   end
 
-  for {effect, key} <- Enum.with_index([:shuttle, :suspension, :station_closure, :delay]) do
+  @spec get_kenmore_special_text(list(atom()), atom()) :: list(FreeText.t())
+  def get_kenmore_special_text(["b", "c"], :partial_alert),
```
cmaddox5, line 278: Similar to my comment on `serialize_map`, I think the better home for these functions is in the `WidgetInstance`.

cmaddox5 (commented):
Yeesh, Kenmore and WTC just gotta make things complicated. Looks good overall. Couple of questions/suggestions.

assets/src/components/v2/free_text.tsx
```diff
@@ -165,12 +165,12 @@ const FreeTextLine = ({
   icon,
   text,
 }: {
-  icon?: string;
+  icon: string;
```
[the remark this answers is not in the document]
hannahpurcell: Oooo glad you caught that. I was able to fix the issue with `reconstructed_takeover.tsx` by changing line 50 to `  if (!icon) { `, but not sure how to reproduce the inline alert for departures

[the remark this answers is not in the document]
cmaddox5: Was there an issue with leaving it as an optional param? I admit, falsy values and equality in JS give me trouble so I might be missing something. But leaving it as optional and not rendering `Icon` if it is falsy should allow for backwards compatibility.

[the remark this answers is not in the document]
hannahpurcell: I'm aiming to have the container always render, but not necessarily anything inside the container. So I think the change I made will always render the container, but conditionally fill it. Not sure I would normally do that, but the FreeText file supports so many different cases, this seemed an ok compromise.

Screenshot 2023-03-27 at 5 01 41 PM

lib/screens/v2/candidate_generator/dup/alerts.ex
hannahpurcell: I'll try `get_branch_if_entity_matches_stop`? The matcher is important for this function. I'll add a comment, too!

assets/src/components/v2/free_text.tsx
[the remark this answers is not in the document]
cmaddox5: Gotcha, that makes sense.

lib/screens/v2/candidate_generator/dup/alerts.ex
hannahpurcell: I'm waffling back and forth and not sure how I feel. I went in this direction after reading @jzimbel-mbta's comment in his code `# Use the DupSpecialCaseAlert widget to directly specify the serialized data.` And it kinda makes sense that we want to break the mold in this case, because we go through all these conditionals in the `CandidateGenerator` to determine yes, we are working with a special case here, that to then pass this info to the `WidgetInstance` and then have to go through pattern matching to again say yes! this is a special case, seems a little redundant.

But if it's much cleaner to put the serialization in the WidgetInstance, then I understand breaking the mold might not be worth it

cmaddox5: That's a good point. We've had edge cases before, but this one is kinda the leader of edge cases. So makes sense that it needs to stray from convention.

jzimbel-mbta: Answered in slack, but in short: **Yes**, I think it's reasonable to deviate from the approach I had prescribed without really looking into how complicated some of these special cases are!

hannahpurcell: I'm not worried about deviating from the originally prescribed solution, but trying to figure out which organization is actually better

hannahpurcell: For now, I'll reorganize and then get your weigh-in

PR #1660, opened 2023-03-25
Add environment to header for V2 screens in dev and dev-green

PaulJKim:
**Asana task**: [Add env back to top of all v2 screens](https://app.asana.com/0/1185117109217432/1203860696691652/f)

This adds the environment name for V2 screens in `dev` and `dev-green`

Examples:
Screenshot 2023-03-24 at 8 13 33 PM
Screenshot 2023-03-24 at 8 14 23 PM
Screenshot 2023-03-24 at 8 14 35 PM
Screenshot 2023-03-24 at 8 14 42 PM

I noticed that when the title is long and overflows to the next line, part of the title will start overlapping with the environment name. I thought of putting the environment name somewhere else like underneath the time, but I figured it wouldn't be as noticeable. I think the text overlaps in a similar way for v1 screens currently.

### DIFF
--- a/assets/src/components/v2/normal_header.tsx
+++ b/assets/src/components/v2/normal_header.tsx
@@ -66,8 +66,15 @@ const NormalHeaderTitle: ComponentType<NormalHeaderTitleProps> = forwardRef(
     }
 
     const abbreviatedText = fullName ? text : abbreviateText(text);
+    const environmentName = getDatasetValue("environmentName");
 
     return (
+      <div>
+        <div className="normal-header__environment">
+          {["screens-dev", "screens-dev-green"].includes(environmentName)
+            ? environmentName
+            : ""}
+        </div>
       <div className="normal-header-title">
         {showTo && <div className="normal-header-to__text">TO</div>}
         {icon && <NormalHeaderIcon icon={icon} />}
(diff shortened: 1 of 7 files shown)


assets/src/components/v2/normal_header.tsx
```diff
@@ -66,8 +66,15 @@ const NormalHeaderTitle: ComponentType<NormalHeaderTitleProps> = forwardRef(
     }
 
     const abbreviatedText = fullName ? text : abbreviateText(text);
+    const environmentName = getDatasetValue("environmentName");
```
cmaddox5, line 69: The import is missing here. Also, the return for `getDatasetValue` is `string | undefined` so the linter might complain. You can get around that with:

```suggestion
    const environmentName = getDatasetValue("environmentName") || "";
```

cmaddox5 (changes requested):
Overall, looks great! Just need an import and I made a few suggestions.

assets/src/components/v2/normal_header.tsx
PaulJKim: Thank you! I was wondering about a good way to get around that linter error

cmaddox5 (approved):
Looks great!

PR #1667, opened 2023-03-30
polish: DUP v2 admin header config

cmaddox5:
**Notion task**: [Long station name “Tufts Medical Ctr” is abbreviated correctly](https://www.notion.so/mbta-downtown-crossing/Long-station-name-Tufts-Medical-Ctr-is-abbreviated-correctly-dbe9b4057bcb4320afd97bdece726907?pvs=4)

Header config was missing from admin tool.

I also fixed the DUP v2 header config so it can use `stop_name`. `CandidateGenerator.Dup.Alerts` was assuming the header config was always given a `stop_id` (which was true when that task was started). Now that it can be a `stop_name`, it needs to check which is being first.

- [ ] Tests added?


hannahpurcell (approved):
👍

PR #1668, opened 2023-03-30
Add retry logic to routes fetch

cmaddox5:
**Asana task**: ad-hoc

~~Our `Agent` seems to be saving an empty list at times. When this happens, we still use the `if-modified-since` header which causes the empty list to persist until the response does change. Conditionally adding the header should allow the `Agent` to get back to a good state it somehow saves bad data.~~

Went a different direction. The original logic would still always fail once because we never try again when we get an empty list. New logic adds retries to the fetch. It only retries for empty lists (`v3_api` already has retry logic for failures) up to 3 times. If it's still an issue, we log and the function will still return an empty list. It will not save that empty list to the Agent so in the next refresh it will try fetching again. 

- [ ] Tests added?

### DIFF
--- a/lib/screens/stops/stop.ex
+++ b/lib/screens/stops/stop.ex
@@ -8,8 +8,6 @@ defmodule Screens.Stops.Stop do
   So there's inconsistent use of this local data.
   """
 
-  use Retry.Annotation
-
   require Logger
 
   alias Screens.Routes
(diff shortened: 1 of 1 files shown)


lib/screens/stops/stop.ex
```diff
@@ -255,6 +258,9 @@ defmodule Screens.Stops.Stop do
            [],
            true
          ) do
+      {:ok, %{"data" => []}, _} ->
+        :bad_response
```
hannahpurcell, line 262: The function that uses this function looks for :error atom, not :bad_response, so it will be unable to match. Also, would be great to have the error logged in `create_station_with_routes_map`. And finally, we should probably let the empty list fail more gracefully down the road -- so if we log the error in `create_station_with_routes_map` but still pass an empty list, we should handle that empty list better in departures.ex.

hannahpurcell (approved):
Thanks for re-assigning to me. Yeah, this looks good to me. Let's take it for a spin

PR #1673, opened 2023-03-31
fix: Headway headsign logic

cmaddox5:
**Notion task**: [Multi-line headways include the correct destination](https://www.notion.so/mbta-downtown-crossing/Multi-line-headways-include-the-correct-destination-48251b33f716474db75396ab7257613c?pvs=4)

Something I did not think about when implementing headway mode was how to headsign matchers are there to determine the direction of the affected range of an alert. This means that headway needs the opposite direction's headsign (sounds so obvious now so not sure how I didn't think about this).

This PR adds an additional element to the headsign matcher tuples. I would rather solve this in the logic, but that seems risky because we are looking at `platform_id`s. Directional alerts may cause an ID we expect not to be present (or vice-versa). Because we know that the headsign matcher works for alerts, if we also hardcode the headsign for headway we will know for sure that the logic is correct.

Biggest concern: this affects anything that implements the `SingleAlertWidget` protocol. The upside is it can just ignore the fourth element if it does not have headway mode (which is every existing alert since headway is handled in departures).

- [ ] Tests added?

### DIFF
--- a/config/config.exs
+++ b/config/config.exs
@@ -81,121 +81,121 @@ config :screens,
 
 config :screens,
   # Maps alert informed entity contents to the appropriate headsign to show for that alert.
-  # List elements must be of the shape {informed_stop_ids, not_informed_stop_ids, headsign}.
+  # List elements must be of the shape {informed_stop_ids, not_informed_stop_ids, headsign, headway}.
   # Each set of stop IDs can be either a single string or a list of strings.
   dup_alert_headsign_matchers: %{
     # Kenmore
     "place-kencl" => [
-      {"70149", ~w[70153 70211 70187], "Boston College"},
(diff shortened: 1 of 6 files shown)


hannahpurcell (approved):
Makes sense. I'm not crazy about needing to include unneeded values for the prefare_alert_headsign_matchers, but I see why it's needed at the moment. Maybe this will get tidied when we refactor alerts a bit (and especially when dup v1 is closed out)

cmaddox5:
> 

Definitely agree with you on the new values in `prefare_alert_headsign_matchers`, so I thought about this again and came up with something else. Headsign matchers now use maps instead of tuples. This means we can add headway headsigns to just the configs that need it. So that value is gone entirely from prefare and DUP scenarios that do not display headway (mostly Kenmore stuff).

PR #1679, opened 2023-04-04
Cm/fix multi section alert layout

cmaddox5:
**Notion task**: [Station closure, inside, 1/2](https://www.notion.so/mbta-downtown-crossing/Station-closure-inside-1-2-8a2c0d7a2faa471b8878939ac43c152d?pvs=4)

On multi-section screens where the second section is not subway/light rail, we were showing a fullscreen alert instead of a partial alert. Added a new struct attribute that holds the total number of section configured in `primary_departures` in the screen config. I added the config access to the `candidate_generator` because adding it to the `widget_instance` just didn't feel right. Happy to hear any opposing opinions on that.

- [ ] Tests added?

### DIFF
--- a/lib/screens/v2/widget_instance/dup_alert.ex
+++ b/lib/screens/v2/widget_instance/dup_alert.ex
@@ -15,6 +15,7 @@ defmodule Screens.V2.WidgetInstance.DupAlert do
     :alert,
     :stop_sequences,
     :subway_routes_at_stop,
+    :primary_section_count,
     :rotation_index,
     :stop_name
   ]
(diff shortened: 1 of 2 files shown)


lib/screens/v2/widget_instance/dup_alert.ex
```diff
@@ -139,26 +140,26 @@ defmodule Screens.V2.WidgetInstance.DupAlert do
   # Table is adapted from the one in the product spec: https://www.notion.so/mbta-downtown-crossing/DUP-Alert-Widget-Specification-a82acff850ed4f2eb98a04e5f3e0fe52
   # Compile-time code converts each table row to `do_alert_layout` function clauses.
```
hannahpurcell, line 141: Maybe add a lil comment noting that TotalSections refers to sections in primary_departures (so folks don't think it's total number of sections for both primary & secondary)

hannahpurcell (approved):

PR #1680, opened 2023-04-04
polish: Make DUP 3rd rotation layout depend on whether secondary departures are configured

jzimbel-mbta:
**Asana task**: [Page 3/3 layout should match that of page 1/3 when there are no secondary departures configured](https://www.notion.so/mbta-downtown-crossing/Station-closure-inside-1-1-f46b843e4eb842ba9043f3623a420dc6?pvs=4)

[Relevant slack thread](https://mbta.slack.com/archives/C02SXM166N8/p1680186465521599)

I moved the special case for delay alerts, as well as this, into one function to try and avoid confusion.

I'm very open to suggestions on how to organize this logic better, though—does the table approach no longer make sense since the layout for page 2/3 is always full-screen, and the layout for page 3/3 is handled specially?

- [ ] Tests added?

### DIFF
--- a/lib/screens/v2/widget_instance/dup_alert.ex
+++ b/lib/screens/v2/widget_instance/dup_alert.ex
@@ -93,16 +93,14 @@ defmodule Screens.V2.WidgetInstance.DupAlert do
 
     parameters = {effect, location, affected_line_count, total_line_count, rotation_index}
 
-    delay_severity = t.alert.severity
+    special_case = alert_layout_special_case(t, parameters)
 
-    if effect == :delay and delay_severity < 5 do
-      # We don't show delays < 20 minutes
-      log_layout_mismatch(t, parameters, delay_severity)
-      :no_render
+    if special_case do
+      special_case
     else
       case do_alert_layout(parameters) do
         :no_render ->
-          log_layout_mismatch(t, parameters, delay_severity)
+          log_layout_mismatch(t, parameters, :not_applicable)
           :no_render
 
         layout ->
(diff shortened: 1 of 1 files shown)


lib/screens/v2/widget_instance/dup_alert.ex
```diff
@@ -94,16 +94,14 @@ defmodule Screens.V2.WidgetInstance.DupAlert do
 
     parameters = {effect, location, affected_line_count, t.primary_section_count, rotation_index}
 
-    delay_severity = t.alert.severity
+    special_case = alert_layout_special_case(t, parameters)
```
cmaddox5, line 97: I think one thing that makes this a little confusing is that rotation 2 is considered a special case. I don't really think of it as special in the same sense that delays are. It just uses a different set of rules because it (potentially) holds secondary info. (Hopefully that makes sense) Is it possible to handle primary and secondary rotation layouts in different functions? So you could pattern match on `t.rotation_index` and pull all of the logic starting at line 226 into the function for rotation index 2. With that separation, the table still holds up because it is used exclusively by the primary rotations. What do you think?

jzimbel-mbta: Thanks for nudging me in the right direction! Hopefully the updated approach makes more sense. It separates delay alerts out right at the top now, and the table only gives the mapping for rotation `zero`.

jzimbel-mbta: The changes are substantial enough that it will probably be easier to review the diff in the split view now.

lib/screens/v2/widget_instance/dup_alert.ex
```diff
@@ -81,145 +81,159 @@ defmodule Screens.V2.WidgetInstance.DupAlert do
...
-    Logger.info(
+  defp log_layout_mismatch(t, delay_severity) do
+    log_fields =
+      t
+      |> get_layout_parameters()
+      |> Map.merge(%{
+        delay_severity: delay_severity,
+        rotation_index: t.rotation_index,
+        alert_id: t.alert.id,
+        stop_id: t.screen.app_params.alerts.stop_id
```
jzimbel-mbta, line 228: One note on this change: my previous code was wrong. It didn't access the field correctly, and was based on an older version of the config schema that allowed multiple stop IDs.

cmaddox5 (approved):
Yep, I like this a lot better! The updated comments help a ton.

PR #1682, opened 2023-04-05
polish: Overnight row no schedule tomorrow

cmaddox5:
**Notion task**: [Moon icon is displayed in place of scheduled time when there is no scheduled departure the next morning](https://www.notion.so/mbta-downtown-crossing/Moon-icon-is-displayed-in-place-of-scheduled-time-when-there-is-no-scheduled-departure-the-next-morn-89a60129ea5c428aa22a0dc0586e295c?pvs=4)

If a route is in overnight mode and has no scheduled trips the next service day, we still want to show the row with a moon icon instead of a time. This one takes some hardcoding to test, so reach out if you have any trouble testing locally.

Screenshot 2023-04-05 at 3 49 29 PM

- [ ] Tests added?


lib/screens/v2/candidate_generator/dup/departures.ex
```diff
@@ -501,6 +487,52 @@ defmodule Screens.V2.CandidateGenerator.Dup.Departures do
 
   defp get_overnight_schedules_for_section(_, _, _, _, _, _, _), do: []
 
+  defp get_overnight_departure_for_route(
+         nil,
+         _first_schedule_tomorrow,
```
PaulJKim, line 492: Does this fully replicate the first condition in the original original logic? What if `last_schedule_today` is not nil but `first_schedule_tomorrow` is?

cmaddox5: It replicates half of it. We know for sure that `last_schedule_today == nil` means something went wrong so don't show overnight just in case. If `first_schedule_tomorrow == nil`, That may be because there are no schedules for that route/direction the following day. In that case, we show a row for the route/direction with a moon icon instead of a time (line 514 is where this is handled).

PaulJKim (approved):
LGTM!

hannahpurcell:
Having a couple issues with this branch -- might be because my testing methods are outdated. (Just making a comment because didn't want to ping you on slack while you're vacationing!)

PR #1688, opened 2023-04-06
Polish: Yellow DUP alerts

hannahpurcell:
[Polish](https://www.notion.so/mbta-downtown-crossing/4b67fc8f4ef04868901d85a412cdd48f?v=8c5d6c5153b443a58cedcf0d8b9838d7&p=e39ce62f5a69401bb88dfd633b262aba&pm=s)

When an alert affects multiple station lines and yellow is used, the svg and fonts are dark.

### DIFF
--- a/assets/css/v2/dup/normal_header.scss
+++ b/assets/css/v2/dup/normal_header.scss
@@ -21,6 +21,9 @@
   &--silver {
     background-color: $line-color-silver;
   }
+  &--yellow {
+    background-color: $alert-yellow;
+  }
 }
 
 .normal-header-icon {
@@ -87,7 +90,6 @@
 }
 
 .normal-header--yellow {
-  background-color: $alert-yellow;
   .normal-header-title__text {
     color: #171F26;
   }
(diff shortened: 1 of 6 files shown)


assets/css/v2/dup/normal_header.scss
```diff
@@ -85,3 +85,10 @@
   color: white;
   position: absolute;
 }
+
+.normal-header--yellow {
+  background-color: $alert-yellow;
+  .normal-header-title__text {
```
cmaddox5, line 91: I know this `title__text` bit probably needs to be down here so it isn't overridden above, but should `background_color` be moved up so all `background-color`s live in the same place?

cmaddox5 (approved):
👍 Just one non-blocking suggestion.

assets/css/v2/dup/normal_header.scss
hannahpurcell: Yeah, that's probably good for the future!

PR #1689, opened 2023-04-06
Kenmore case allows alert with multiple branches

hannahpurcell:
[Polish](https://www.notion.so/mbta-downtown-crossing/4b67fc8f4ef04868901d85a412cdd48f?v=8c5d6c5153b443a58cedcf0d8b9838d7&p=1c9efef552b3449195de4dd64b9909b9&pm=s)

Kenmore case allows either multiple alerts that affect single branches or one alert that affects multiple branches.

Before:
Screenshot 2023-04-06 at 10 16 35 AM
After:
Screenshot 2023-04-06 at 10 16 14 AM

### DIFF
--- a/lib/screens/v2/candidate_generator/dup/alerts.ex
+++ b/lib/screens/v2/candidate_generator/dup/alerts.ex
@@ -263,12 +263,12 @@ defmodule Screens.V2.CandidateGenerator.Dup.Alerts do
           atom()
   def get_branches_if_entity_matches_stop(%{informed_entities: informed_entities}, stop_matchers) do
     stop_matchers
-    |> Enum.map(fn stop ->
-      Enum.find_value(informed_entities, fn e ->
-        if stop.stop === e.stop, do: Map.get(stop, :branch)
+    |> Enum.filter(fn stop ->
+      Enum.any?(informed_entities, fn e ->
+        stop.stop === e.stop
       end)
     end)
-    |> Enum.reject(&is_nil(&1))
+    |> Enum.map(&Map.get(&1, :branch))
   end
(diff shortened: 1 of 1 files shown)


lib/screens/v2/candidate_generator/dup/alerts.ex
```diff
@@ -254,21 +254,21 @@ defmodule Screens.V2.CandidateGenerator.Dup.Alerts do
...
-  # Given an alert, see if any of its informed entities match a list of stops-of-interest (called stop_matchers here).
-  # If it has an informed entity on the list, return its branch.
-  @spec get_branch_if_entity_matches_stop(
+  # Given an alert, see if its informed entities match a list of stops-of-interest (called stop_matchers here).
+  # If a stop matcher is found, the branch is added the the returned list.
+  @spec get_branches_if_entity_matches_stop(
           Alert.t(),
           list(%{branch: String.t(), stop: String.t()})
         ) ::
           atom()
```
cmaddox5, line 263: Need to update this now that you return a list. 

```suggestion
          [atom()]
```

Also, should this be `String.t()` instead of `atom()`? `branch` is a string in the `stop_matchers` you give to the function above.

cmaddox5 (changes requested):

lib/screens/v2/candidate_generator/dup/alerts.ex
hannahpurcell: ah yes! I believe this is why Dialyzer was complaining. 👍

cmaddox5 (approved):
🚀

PR #1691, opened 2023-04-07
Revert "polish: Overnight row no schedule tomorrow"

hannahpurcell:
Reverts mbta/screens#1682. It was resulting in a [Sentry error ](https://mbtace.sentry.io/issues/4068805421/?environment=screens-dev&project=6061747&query=is%3Aunresolved&referrer=issue-stream&stream_index=0)and paused design QA.


PaulJKim:
Oof my bad for not reviewing more thoroughly. I had a hunch that some of cases covered by the original logic might not be anymore but didn't look closely enough at the updated logic

PR #1692, opened 2023-04-10
polish fix: overnight no schedule tomorrow

cmaddox5:
**Notion task**: [Moon icon is displayed in place of scheduled time when there is no scheduled departure the next morning](https://www.notion.so/mbta-downtown-crossing/Moon-icon-is-displayed-in-place-of-scheduled-time-when-there-is-no-scheduled-departure-the-next-morn-89a60129ea5c428aa22a0dc0586e295c?pvs=4)

The code change in the first PR gave an error when we were not in overnight but we did not have a schedule for tomorrow. To tackle this, I added another function clause for when `first_schedule_tomorrow` is `nil`. If `now` is after `last_schedule_today.departure_time`, and there are no schedules tomorrow, we know we can safely show overnight mode. 

- [x] Tests added?


hannahpurcell:
One thing I noticed when trying to debug Friday: the params in the default function def for `get_overnight_schedules_for_section` does not match the params for the later instances. That made it difficult to understand what function did what

lib/screens/v2/candidate_generator/dup/departures.ex
```diff
@@ -516,6 +513,28 @@ defmodule Screens.V2.CandidateGenerator.Dup.Departures do
     nil
   end
 
+  # If now is after today's last schedule and there are no schedules tomorrow,
+  # we still want a departure row without a time (will show a moon icon)
+  defp get_overnight_departure_for_route(
+         first_schedule_today,
+         last_schedule_today,
```
jzimbel-mbta, line 520: I think this would be a separate issue to fix, but do we need to handle the case when there are no schedules _today_?

The remaining scenarios I can think of that (afaik) aren't already covered are
- No schedules today, no schedules tomorrow
- No schedules today, yes schedules tomorrow

jzimbel-mbta (approved):
Good to go!

lib/screens/v2/candidate_generator/dup/departures.ex
cmaddox5: I honestly hadn't thought of that. I had always considered no schedules today meaning there was some kind of issue, but this task is specifically for routes that don't run on certain days. Because predictions take priority, I think it's safe to assume a route not running today is in overnight mode. Throughout the day, predictions would show as long as there are any, and we would always have an overnight row in the queue when we start running out of predictions. Does that sound right?

cmaddox5: Had a chat with PaulH. If a route has no schedules today, don't assume it's in overnight mode. This does mean that a route will not show overnight mode at all on the service day it is not running, but that's ok. We will let it have the opportunity to show again when it reaches the next service day.

PR #1693, opened 2023-04-10
polish: Overnight for early morning

cmaddox5:
**Notion task**: [Outstanding issues with overnight mode](https://www.notion.so/mbta-downtown-crossing/Outstanding-issues-with-overnight-mode-040f5c0849ee446d9135409871b50674?pvs=4)

When we change service day at 3am, we were incorrectly switching to a no data state instead of continuing to display overnight mode. To correct this, an extra conditional has been added to check that `now` is before the first schedule of the day. 

- [ ] Tests added?

### DIFF
--- a/lib/screens/v2/candidate_generator/dup/departures.ex
+++ b/lib/screens/v2/candidate_generator/dup/departures.ex
@@ -541,7 +541,8 @@ defmodule Screens.V2.CandidateGenerator.Dup.Departures do
         %Departure{schedule: first_schedule_tomorrow}
 
       # After 3am but before the first scheduled trip of the day.
-      DateTime.compare(now, first_schedule_today.departure_time) == :lt ->
+      not is_nil(first_schedule_today) and
+          DateTime.compare(now, first_schedule_today.departure_time) == :lt ->
         %Departure{schedule: first_schedule_today}
 
       true ->
--- a/lib/screens/v2/candidate_generator/dup/departures.ex
+++ b/lib/screens/v2/candidate_generator/dup/departures.ex
@@ -458,29 +458,14 @@ defmodule Screens.V2.CandidateGenerator.Dup.Departures do
           &(&1.route.id == route_id and &1.direction_id == direction_id)
         )
 
-      cond do
-        is_nil(last_schedule_today) or is_nil(first_schedule_tomorrow) ->
-          nil
-
-        DateTime.compare(now, first_schedule_tomorrow.departure_time) == :gt ->
-          Logger.warn(
-            "[get_overnight_schedules_for_section] now is after first_schedule_tomorrow."
-          )
-
-          nil
-
-        # Before 3am and between the `departure_time` for today's last schedule and tomorrow's first schedule
-        DateTime.compare(now, last_schedule_today.departure_time) == :gt and
-            DateTime.compare(now, first_schedule_tomorrow.departure_time) == :lt ->
-          %Departure{schedule: first_schedule_tomorrow}
-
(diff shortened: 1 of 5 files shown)


lib/screens/v2/candidate_generator/dup/departures.ex
```diff
@@ -501,37 +501,76 @@ defmodule Screens.V2.CandidateGenerator.Dup.Departures do
...
+
+        nil
+
+      # Before 3am and between the `departure_time` for today's last schedule and tomorrow's first schedule
+      DateTime.compare(now, last_schedule_today.departure_time) == :gt and
+          DateTime.compare(now, first_schedule_tomorrow.departure_time) == :lt ->
+        %Departure{schedule: first_schedule_tomorrow}
+
+      # After 3am but before the first scheduled trip of the day.
+      DateTime.compare(now, first_schedule_today.departure_time) == :lt ->
```
PaulJKim, line 544: Is it possible for `first_schedule_today` to be `nil`? Or is it safe to assume that if `last_schedule_today` is `nil`, then `first_schedule_today` will be nil as well?

lib/screens/v2/candidate_generator/dup/departures.ex
```diff
@@ -501,37 +501,76 @@ defmodule Screens.V2.CandidateGenerator.Dup.Departures do
...
+        )
+
+        nil
+
+      # Before 3am and between the `departure_time` for today's last schedule and tomorrow's first schedule
+      DateTime.compare(now, last_schedule_today.departure_time) == :gt and
+          DateTime.compare(now, first_schedule_tomorrow.departure_time) == :lt ->
+        %Departure{schedule: first_schedule_tomorrow}
+
+      # After 3am but before the first scheduled trip of the day.
```
PaulJKim, line 543: For these two cases, how are we certain that it's before 3am if we're only checking `now` against the `last_schedule_today` and the `first_schedule_tomorrow`?

lib/screens/v2/candidate_generator/dup/departures.ex
```diff
@@ -501,37 +501,76 @@ defmodule Screens.V2.CandidateGenerator.Dup.Departures do
...
-         tomorrow,
+         now,
          route_ids_serving_section
        ) do
     today =
-      case fetch_schedules_fn.(fetch_params, nil) do
+      case fetch_schedules_fn.(fetch_params, now) do
         {:ok, schedules} when schedules != [] ->
           # We want the last schedules of the current day.
           # Need to reverse the list of fetched schedules so that List.first/1 looks at the correct time of day.
```
PaulJKim, line 562: Since we're not reversing the list here anymore, I think we could remove this documentation

lib/screens/v2/candidate_generator/dup/departures.ex
PaulJKim: Hmm okay I think I'm following. Essentially we're inferring if it's currently before or after 3am based on how `now` relates to `last_schedule_today` and `first_schedule_tomorrow`? Or I guess mainly just `last_schedule_today` to determine that, and then we compare against `first_schedule_tomorrow` to know if we should show overnight mode?

cmaddox5: Yep, that's right!

lib/screens/v2/candidate_generator/dup/departures.ex
cmaddox5: I _think_ it's safe to assume, but you know what assuming does. Added a `nil` check just in case.

PR #1695, opened 2023-04-11
polish: departure row dividers

cmaddox5:
**Asana task**: [Typical partial alerts are rendered correctly](https://www.notion.so/mbta-downtown-crossing/Typical-partial-alerts-are-rendered-correctly-38f9c0f09f564cb9af5085d435775e0a?pvs=4)

Changed up dividers so they do not show next to a partial alert.

- [ ] Tests added?


hannahpurcell (approved):
👍

PR #1696, opened 2023-04-11
polish: dup_alert priority

cmaddox5:
**Notion task**: [End of disrupted subway service is detected correctly](https://www.notion.so/mbta-downtown-crossing/End-of-disrupted-subway-service-is-detected-correctly-c86eb01642774ae9a93b43e2249bf34a?pvs=4)

When we are overnight but there is an active alert to show on screen, we want widget priority to be fullscreen alert > overnight > partial alert.

- [ ] Tests added?


hannahpurcell (approved):
👍

PR #1697, opened 2023-04-11
polish: ashmont braintree headway message

cmaddox5:
**Asana task**: [Multi-line headways accommodate the longest destination](https://www.notion.so/mbta-downtown-crossing/Multi-line-headways-accommodate-the-longest-destination-7de73f5f45064fbbb609d760fb65af66?pvs=4)

Headway messages are too long when headsign is Ashmont/Braintree. Fixed this by condensing `minutes` to `m` for just that headsign. Also fixed some alignment issues for headway pills.

- [ ] Tests added?

### DIFF
--- a/lib/screens/v2/widget_instance/departures.ex
+++ b/lib/screens/v2/widget_instance/departures.ex
@@ -100,18 +100,24 @@ defmodule Screens.V2.WidgetInstance.Departures do
 
     text =
       if is_only_section do
+        time_range =
+          if headsign == "Ashmont/Braintree" do
+            [%{format: :bold, text: "#{lo}-#{hi}m"}]
+          else
+            [%{format: :bold, text: "#{lo}-#{hi}"}, "minutes"]
+          end
+
         %FreeTextLine{
           icon: "subway-negative-black",
-          text: [
-            %{
-              color: pill_color,
-              text: "#{String.upcase(route)} LINE"
-            },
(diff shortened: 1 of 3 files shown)


lib/screens/v2/widget_instance/departures.ex
```diff
@@ -100,18 +107,24 @@ defmodule Screens.V2.WidgetInstance.Departures do
 
     text =
       if is_only_section do
+        time_range =
+          if headsign == "Ashmont/Braintree" do
```
PaulJKim, line 111: Is it worth making this conditional on the length of the headsign instead of hardcoding it for Ashmont/Braintree?

cmaddox5: Hmm, I'm not sure. A part of me thinks we want this to be case by case. I think design didn't want to abbreviate this headsign so we went this direction, but we may want to treat future headsigns differently. Another hesitation is the unreliability of looking at length when not using a monospace font. I will check with design though so I'm not making any incorrect assumptions.

cmaddox5: Got an answer back from Paul. The one-off is enough for now since it's the only headsign we anticipate needing to hardcode for.

PaulJKim: Sounds good!

PR #1699, opened 2023-04-12
fix: headway and no-data sections take up 1 row when sharing with another section

jzimbel-mbta:
**Asana task**: [headway_section and no_data_section should only take up one line](https://www.notion.so/mbta-downtown-crossing/headway_section-and-no_data_section-should-only-take-up-one-line-a1d8471b90f14df8bc1a30b28d6b2386?pvs=4)

headway-section
Headway section, on departures configured with two sections

no-data-section
No-data section, on departures configured with two sections

---

headway-section-single-section-departures
Headway section, on departures configured with one section

no-data-section-single-section-departures
No-data section, on departures configured with one section

- [ ] Tests added?


hannahpurcell (changes requested):
The merge seems to have affected the fix

Screenshot 2023-04-18 at 4 06 35 PM

jzimbel-mbta:
😩

jzimbel-mbta:
Should be good now! I gave the selectors increased specificity so I could specify height calculation for those particular cases.

My previous change just gave all sections variable height based on the total height of their contents, but based on changes on master since then, it looks like we need to fix the height at 416px for normal sections that share space with a second section.

assets/css/v2/dup/departures/no_data_section.scss
hannahpurcell, line 1: Ooo there's a general "comment on file" button now!

There's one extremely unlikely but also easy-to-resolve case where there is an alert AND there's a 
"data unavailable" situation.

At the bottom of this file:
```
.body-split__main-content-reduced .no-data-section {
  height: 208px;
}
```
Before:
Screenshot 2023-04-19 at 8 23 46 AM

After:
Screenshot 2023-04-19 at 8 29 07 AM
Screenshot 2023-04-19 at 8 29 35 AM

hannahpurcell (approved):
One improvement. Otherwise, should be golden!

PR #1700, opened 2023-04-12
polish: destination headsign paging fix

cmaddox5:
**Notion task**: [No headsigns overflow 1024px / overlap with the following morning’s schedule time](https://www.notion.so/mbta-downtown-crossing/No-headsigns-overflow-1024px-overlap-with-the-following-morning-s-schedule-time-3344c4da7bf744d3b19f71688a3a6495?pvs=4)

An incorrect CSS class was causing text wrapping on some headsigns. Fixing this class fixed the paging issues. 

I also went ahead and change v2 DUPs to use local Inter files.

- [ ] Tests added?

### DIFF
--- a/assets/css/dup_v2.scss
+++ b/assets/css/dup_v2.scss
@@ -1,5 +1,4 @@
-@import url("https://rsms.me/inter/inter.css");
-
+@import "fonts/inter_font_face";
 @import "colors";
 @import "v2/common/widget";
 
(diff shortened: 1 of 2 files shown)


hannahpurcell (approved):
👍 Lol, this solution is still funny

cmaddox5:
> 👍 Lol, this solution is still funny

Glad you think so 😬

PR #1702, opened 2023-04-13
polish: DUP v2 all screen URL

cmaddox5:
**Notion task**: ad-hoc

`responseMapper` was missing in the `MultiScreenPage` props, which was causing some layout issues. Adding it renders the page correctly.

- [ ] Tests added?

### DIFF
--- a/assets/src/apps/v2/dup.tsx
+++ b/assets/src/apps/v2/dup.tsx
@@ -122,7 +122,10 @@ const App = (): JSX.Element => {
     <Router>
       <Switch>
         <Route exact path="/v2/screen/dup_v2">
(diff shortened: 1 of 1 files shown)


hannahpurcell:
What are the layout differences? I can't spot the diff between this and what's in dev-green right now

cmaddox5:
> What are the layout differences? I can't spot the diff between this and what's in dev-green right now

As each screen is loading, the loading state is not taking up the space it should. The background color doesn't take up the whole space and the arrow and link in the bottom right are missing. The screens resolve themselves once the page fully loads, but until then things are a little out of place.

PR #1703, opened 2023-04-13
polish: DUP v2 route pills

cmaddox5:
**Notion task**: [Text within all pills is vertically centered](https://www.notion.so/mbta-downtown-crossing/Text-within-all-pills-is-vertically-centered-3f7b8a4e5b544389bec502ea55f3758c?pvs=4)

Our imported Helvetica font was causing the text in our route pills to be offcenter. To combat this, we are importing a font file for `font-weight: 700` (which is the only `font-weight` used by these pills). This center the text as we expect without the need of any `transform: translateY` adjustments. 

Along with the new file, I combed through all route pill CSS to make sure everything was the same. The biggest offender was `overnight-section` (mostly because there were two CSS class blocks for some reason).

- [ ] Tests added?


hannahpurcell (approved):
👍 🦜 👍

PR #1706, opened 2023-04-18
tech debt: Made a test for each DUP special case

hannahpurcell:
Belated old work. Finally made the Dup special cases tests. One test for each case. They are faux integration tests because they run the candidate_generator code through the serialization.

- [x] Tests added?


cmaddox5 (approved):
Makes sense to me!

PR #1709, opened 2023-04-19
tweak: Use line pills and adjust bolding in alert issue line

jzimbel-mbta:
**Asana task**: ad hoc-ish

These changes ensure that the issue text never takes up more than 2 lines.

image

- [ ] Tests added?

### DIFF
--- a/assets/css/v2/dup/free_text.scss
+++ b/assets/css/v2/dup/free_text.scss
@@ -270,6 +270,7 @@
     border-radius: 72px;
     font-size: 96px;
     line-height: 117px;
+    vertical-align: bottom;
   }
 }
 
--- a/assets/css/v2/dup/free_text.scss
+++ b/assets/css/v2/dup/free_text.scss
@@ -269,7 +269,7 @@
     width: 248px;
     border-radius: 72px;
     font-size: 96px;
-    line-height: 117px;
+    line-height: 128px;
     vertical-align: bottom;
   }
 }
--- a/lib/screens/v2/widget_instance/dup_alert/serialize.ex
+++ b/lib/screens/v2/widget_instance/dup_alert/serialize.ex
@@ -42,13 +42,21 @@ defmodule Screens.V2.WidgetInstance.DupAlert.Serialize do
     }
   end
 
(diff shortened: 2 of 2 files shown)


assets/css/v2/dup/free_text.scss
```diff
@@ -270,6 +270,7 @@
     border-radius: 72px;
     font-size: 96px;
     line-height: 117px;
```
hannahpurcell, line 272: I am pro- `line-height: 128px`, as you mentioned over slack. I think it looks right. Do you still think it's off?

Screenshot 2023-04-20 at 8 31 23 AM

lib/screens/v2/widget_instance/dup_alert/serialize.ex
hannahpurcell, line 1: The change to the issue text also affects the partial alert

Screenshot 2023-04-20 at 8 33 45 AM

assets/css/v2/dup/free_text.scss
jzimbel-mbta: Looks good to me!

lib/screens/v2/widget_instance/dup_alert/serialize.ex
jzimbel-mbta: Agh. Thanks for catching. So much for the single function to produce the line text for both partial and full-screen alerts.

lib/screens/v2/widget_instance/dup_alert/serialize.ex
```diff
@@ -62,62 +70,74 @@ defmodule Screens.V2.WidgetInstance.DupAlert.Serialize do
     end
   end
 
-  defp get_line_text_builder(t) do
-    case DupAlert.get_affected_lines(t) do
-      [line_color] ->
-        fn _ -> [bold("#{line_color} Line")] end
+  defp line_to_pill_atom(line_string) do
```
hannahpurcell, line 73: I think some of these functions would be good as utils, but instead of figuring that out here, I'd say that can go in the "alert refactor" task, since it'll be easier to find a place for this util while also figuring out where to put all the other common code, etc.

hannahpurcell (approved):
Ok! I think we're golden now, thanks 👍

PR #1710, opened 2023-04-20
First round of flagging common alert code

hannahpurcell:


lib/screens/alerts/alert.ex
```diff
@@ -289,6 +292,7 @@ defmodule Screens.Alerts.Alert do
     end
   end
 
+  # This was only used for the AlertPriorityController. What was that? Is it still used?
```
cmaddox5, line 295: I've actually never heard of that 🤔 It still has a route but it isn't being used in the frontend. It's also v1 only. Might be able to remove it when/if we cleanup v1.

lib/screens/alerts/alert.ex
```diff
@@ -1,3 +1,6 @@
+# This file has primarily v1 stuff, but there are a couple of utils we use in v2.
```
cmaddox5, line 1: Would it be useful to create a module `Screens.Alerts.V1.Alert` (or something similar) that has only functions that are used by v1? The benefit is that if/when we remove v1 references from code, we will have more confidence that what we are removing is not used by v2.

lib/screens/v2/widget_instance/alert.ex
```diff
@@ -136,6 +139,7 @@ defmodule Screens.V2.WidgetInstance.Alert do
     defp serialize_icon(unquote(e)), do: unquote(icon)
   end
 
+  # possibly useful for other alerts
```
cmaddox5, line 142: I could see this potentially being useful for other widgets, too. A very standard `Util` function.

lib/screens/alerts/alert.ex
jzimbel-mbta: I'm one of the contributors to this file in the blame, but...
image

jzimbel-mbta: I'd say it's safe to remove!

lib/screens/v2/widget_instance/common/base_alert.ex
```diff
@@ -248,9 +248,15 @@ defmodule Screens.V2.WidgetInstance.Common.BaseAlert do
     |> MapSet.new(& &1.route_id)
   end
 
+  # Include a comment to define "informed routes": the routes affected by an alert that also exist at the current stop
+  #
+  # This is the updated version of "informed_routes" from alert.ex
+  # The other should be deleted
+  # Note: this does NOT reference downstream stops. Make sure that's fine and we don't need it
```
jzimbel-mbta, line 255: +1 to checking on this, I vaguely remember the team deciding to stop showing any alerts besides those happening now and affecting the home stop in some way.

If that's true then we can get rid of a decent chunk of logic in this function as well as `location/1`. (`boundary_upstream` and `boundary_downstream` would no longer be applicable.)

lib/screens/v2/single_alert_widget.ex
```diff
@@ -1,3 +1,6 @@
+# Rename "ContextualAlert"? "EnrichedAlert"?
+# Just "Alert" - would that work? We'd need to rename the other module, then
```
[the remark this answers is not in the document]
jzimbel-mbta: We could also go with Paul's suggested approach: do away with the protocol, and just define a new struct with the appropriate fields
(I think: `alert: Screens.Alerts.Alert.t()`, `location: SomeNewStruct???.t()`)

This would follow the same lines as the `Screens.V2.Departure` module/struct, which merges together a Prediction struct and a Schedule struct and provides functions that operate on that combined data.

The main question for that approach is, **can we define one common data structure for the "location context" of all of our alert-y widgets?**

lib/screens/v2/widget_instance/elevator_status.ex
```diff
@@ -260,6 +291,7 @@ defmodule Screens.V2.WidgetInstance.ElevatorStatus do
     end)
   end
 
+  # Seems familiar... keep an eye out for duplicate,
```
jzimbel-mbta, line 294: I think I cut some corners here because elevator_closure alerts have a simpler set of possible informed entity shapes. I believe it's always `{facility: "facility id", stop: "platform stop id or parent station id"}`.

A general function would need to handle more cases.

lib/screens/v2/widget_instance/elevator_status.ex
```diff
@@ -188,6 +199,24 @@ defmodule Screens.V2.WidgetInstance.ElevatorStatus do
...
+    # active = Alert.happening_now?(alert, now)
+
+    # case option do 
+    #   :upcoming -> not active and Enum.any?(entities, fn e -> e.stop == parent_station_id end)
+    #   :here -> active and Enum.any?(entities, fn e -> e.stop == parent_station_id end)
+    #   :connecting -> more complicated logic OR run BaseAlert.location === :upstream / :downstream
+    #   :elsewhere -> 
+    #     get the parent stations for the entities
+    #     active and Enum.any?(parent_stations, fn station != parent_station_id end)
+  # end
```
jzimbel-mbta, line 218: We might be able to do something like this, although it would need to support multiple options unioned together, so that we can express "is this upcoming _and_ here?" for example.

We would need to be careful that we're not just duplicating `BaseAlert.location` for parts of this.

Overall, "relevance" seems much more per-screen-type and even per-widget-type, though. I don't think a high-level concept like this would fit well in a module that contains simpler utility functions. It would be better to use those simple functions to build different versions of this (which is kind of what we're already doing) for all of our different definitions of "relevant".

lib/screens/alerts/alert.ex
hannahpurcell: I'm looking back at my comments and seeing they were a little disorienting -- like I left the question "what is this", but elsewhere noted "used here". Lol, that's unhelpful. I'll be trying to clean all this up. Christian is correct, this is needed for v1, so I'll label it as such, and we can remove it when v1 is retired!

lib/screens/v2/single_alert_widget.ex
```diff
@@ -18,6 +21,7 @@ defprotocol Screens.V2.SingleAlertWidget do
   @doc """
   Gets the Screens.Alerts.Alert struct that this widget contains.
   """
+  # What to change this to?
```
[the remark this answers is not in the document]
hannahpurcell: I think my comment was "overcome by events" and should have been taken out. 🤔  I was trying to think through whether a function like this would be useful for both a single alert widget and a multi-alert widget (elevatorstatus). Currently, the `alert()` function is only used in BaseAlert in the `location()` function and the `informed_entities` function. I think those functions are best served by still only taking 1 alert and not a collection of alerts, so I'd leave this function as-is.

lib/screens/v2/widget_instance/alert.ex
```diff
@@ -358,6 +365,11 @@ defmodule Screens.V2.WidgetInstance.Alert do
 
   # For GL, we want to list all affected branches for the alert and not just the branch serving the home stop.
   # This allows us to show a pill for each branch in the informed_entities of the alert (or GL pill if all branches are affected).
+  
+  # This is a special exception to how informed_routes are generated for other screens.
+  # I'm not sure that these should be two clauses of the same function, because they aim to return different kinds of values
+  # One is informed routes at the home stop (seen in BaseAlert also)
+  # The other is informed routes for the alert
```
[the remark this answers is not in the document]
hannahpurcell: I'm not following in your explanation of the bug in the GL clause. Could you describe differently? It seems to be outputting what we'd expect -- two branches if there are only two branches, "Green" if more than two.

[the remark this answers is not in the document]
hannahpurcell: I think I will generalize the `get_affected_routes` function from `reconstructed_alert.ex`, move it to replace this function in BaseAlert, and will write it in a way that it will be more clear what it is doing.

lib/screens/v2/widget_instance/elevator_status.ex
hannahpurcell: I think the function I stubbed out here handles the unioning of options, unless I'm misunderstanding your comment?

True, it's not useful to rehash BaseAlert.location, but the location logic of this function is super simple. Here, not here, and downstream (but BaseAlert.location is used for downstream). BaseAlert is way more complicated.

As for the overall statement about relevance, agreed! 👍

PR #1711, opened 2023-04-20
Alerts Refactor Investigation: CandidateGenerator

cmaddox5:
I don't see a ton of value in a major refactor of the CandidateGenerators. Each one uses logic that is very specific to the widget it is creating candidates for. They do have some shared workflows (fetch, filter, create instance), but not much outside of that. Even the flagged items are mostly ways to allow other files to use functions that aren't specific to a single CandidateGenerator, not any actually refactoring of duplicated code.

### DIFF
--- a/lib/screens/v2/candidate_generator/widgets/elevator_closures.ex
+++ b/lib/screens/v2/candidate_generator/widgets/elevator_closures.ex
@@ -45,6 +45,7 @@ defmodule Screens.V2.CandidateGenerator.Widgets.ElevatorClosures do
     end
   end
 
+  # Should live in Screens.Alerts.Alert along with other alert fetches.
   def fetch_elevator_closures do
     case Screens.V3Api.get_json("alerts", %{
            "filter[activity]" => "USING_WHEELCHAIR",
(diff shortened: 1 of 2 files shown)


jzimbel-mbta (commented):
Yeah, not too much to factor out in these modules.

jzimbel-mbta:
Doesn't "relevant" have different meanings for different alert-y widgets and different screen types, though? That is, it's kind of subjective. I think we'd be better off having the utility functions answer basic questions, and build up the widget-specific logic from those.

hannahpurcell:
@jzimbel-mbta Is your comment about the third line of my message? My note about `relevant?` is that different function names are used for the same stage of logic. I think it would be good for all these CGs to use the same function name, whether it's filter_alerts, relevant, or relevant_alert -- and if we want to force the naming convention, I wonder if a protocol is helpful?

jzimbel-mbta:
@hannahpurcell Ah my bad! I mistook it as "make one `relevant?` function in an alert util functions module".

Yeah, it could make sense to enforce the naming scheme. Maybe just in a behaviour (aka "bundle of functions with specific names") rather than a protocol, since protocols need to be associated with a particular struct def and we don't really have any kind of candidate generator struct. Nor should we, since that module is all about converting API data into widget data.

[One example of a behaviour + implementations](https://github.com/mbta/screens/blob/master/lib/screens/blue_bikes/client.ex)

PR #1712, opened 2023-04-20
polish: no-data-section CSS

cmaddox5:
**Notion task**: [headway_section and no_data_section should only take up one line](https://www.notion.so/mbta-downtown-crossing/headway_section-and-no_data_section-should-only-take-up-one-line-a1d8471b90f14df8bc1a30b28d6b2386?pvs=4)

CSS selectors were not selecting the correct class, so styles were not being applied. I also changed some of the styles so the row matches design.

Screenshot 2023-04-20 at 10 52 49 AM

- [ ] Tests added?

### DIFF
--- a/assets/css/v2/dup/departures/no_data_section.scss
+++ b/assets/css/v2/dup/departures/no_data_section.scss
@@ -16,10 +16,6 @@
 
   .no-data-section__row {
     height: 208px;
-
-    .free-text__line {
-      width: 1400px;
-    }
   }
 }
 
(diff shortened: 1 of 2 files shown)


assets/css/v2/dup/departures/no_data_section.scss
```diff
@@ -2,20 +2,24 @@
...
     width: 144px;
   }
 
-  &__row {
+  .no-data-section__row {
     height: 208px;
+
+    .free-text__line {
+      width: 1400px;
+    }
```
hannahpurcell, line 22: conflicts with the css in the free text file

hannahpurcell (approved):
Jumping in to review because I'd like these changes in my branch. There's one thing I'll have to change, but I can do that in my branch

PR #1716, opened 2023-04-21
Making FreeText styles more concise for DUPs

hannahpurcell:
Adhoc.

Ok, tried to make styles for FreeText more robust for DUPs, but not sure whether it is clearly cleaner. Thoughts?

Adjacent changes:
- Removed styles from other files that refer to DUP free text
- Added line breaks to free text in cases where we want it, instead of doing it with css (more brittle)
- Added border to partial alert because someone noticed it was missing (turns out it was a later change, but oh well, it's added now.)
- Made "seek alternate route" unbold (another late change, but I didn't realize that until I updated it)

### DIFF
--- a/assets/css/v2/dup/alert/partial_alert.scss
+++ b/assets/css/v2/dup/alert/partial_alert.scss
@@ -1,10 +1,5 @@
-.partial-alerts {
-  position: absolute;
-  bottom: 0;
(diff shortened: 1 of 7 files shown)


cmaddox5 (approved):
Lots of nice cleanup. Looks great!

PR #1717, opened 2023-04-24
fix: bidirection with one departure bug

cmaddox5:
**Asana task**: [FunctionClauseError: no function clause matching in Screens.V2.Departure.route_id/1](https://app.asana.com/0/1185117109217422/1204461103486671/f)

See comment in code for description of problem. Only happens on bidirectional sections (was able to deduce this by seeing only Back Bay and Tufts in the Sentry error).

- [X] Tests added?


hannahpurcell (approved):
💡

jzimbel-mbta (approved):
Nice work!

PR #1718, opened 2023-04-25
Bug: styling for full headway on dups

hannahpurcell:
- Full screen headway fixed
- Partial alert centered
- Prevent 'y' from getting cut off. It messes with line-height elsewhere, so I adjusted a margin, notified design.


cmaddox5 (approved):
Looks like there's a formatting issue, but the changes look good. Now let's hope all 100 scenarios are covered 🙃

PR #1720, opened 2023-04-25
fix: headway-section only-child styling

cmaddox5:
**Asana task**: ad-hoc

When the change was made to conditional render sections that have something to show, it broke headway `only-child` CSS selectors. When a `headway-section` is displayed with a blank section on the same page, incorrect styles were being applied. I added a new prop to `headway_section`s so we are more confident which style of headway we are showing.

- [ ] Tests added?


hannahpurcell (approved):
👍  that should do the trick

PR #1727, opened 2023-05-01
feat: Subway Status 2.0 - Backend

cmaddox5:
**Asana task**: [[Subway Status 2.0] Update backend logic](https://app.asana.com/0/1185117109217413/1203775859825083/f)

Diff is very large, but most of the changes are from all the tests I added. All logic was added following the flowchart in this [Miro board](https://miro.com/app/board/uXjVO2fwvkU=/). No config changes necessary. Front-end is broken now, so debugging needs to be done using backend logging or with the Network tab in the browser.

- [ ] Tests added?


lib/screens/v2/widget_instance/subway_status.ex
```diff
@@ -332,112 +449,115 @@ defmodule Screens.V2.WidgetInstance.SubwayStatus do
...
-          type: :single,
-          route: serialize_route_pill("Green"),
-          status: "#{alert_count} alerts",
-          location: %{full: "mbta.com/alerts/subway", abbrev: "mbta.com/alerts/subway"}
-        }
-    end
+  defp serialize_alert_summary(alert_count, route_pill) do
+    %{
+      route_pill: route_pill,
+      status: "#{alert_count} alerts",
```
jzimbel-mbta, line 517: Noticed this while working on my stuff: it looks like the wording for this case is now "<count> current alerts"
image

lib/screens/v2/widget_instance/subway_status.ex
```diff
@@ -336,112 +415,166 @@ defmodule Screens.V2.WidgetInstance.SubwayStatus do
...
-          route: serialize_route_pill("Green"),
-          status: "#{alert_count} alerts",
-          location: %{full: "mbta.com/alerts/subway", abbrev: "mbta.com/alerts/subway"}
-        }
-    end
+  defp serialize_alert_summary(alert_count, route_pill) do
+    %{
+      route_pill: route_pill,
+      status: "#{alert_count} current alerts",
+      location: %{full: "mbta.com/alerts/subway", abbrev: "mbta.com/alerts/subway"}
```
jzimbel-mbta, line 484: mbta.com/alerts now redirects to /alerts/subway, and also the designs use /alerts, so I believe this should be just the plain string `"mbta.com/alerts"`, no full/abbrev map or longer url needed.

lib/screens/v2/widget_instance/subway_status.ex
```diff
@@ -121,6 +232,17 @@ defmodule Screens.V2.WidgetInstance.SubwayStatus do
     end
   end
 
+  defp serialize_gl_pill_with_branches(route_ids) do
+    branches =
+      route_ids
+      |> Enum.filter(&String.contains?(&1, "Green-"))
+      |> Enum.map(fn "Green-" <> branch ->
+        branch |> String.downcase() |> String.to_existing_atom()
+      end)
```
jzimbel-mbta, line 241: Could `route_ids` ever have duplicates? Wondering if this needs a `|> Enum.uniq()`.

jzimbel-mbta (commented):
Looks great! I think this is about as readable as such a complicated "flowchart in code" can be.
Just some minor questions/comments/suggestions, should be quick to address.

lib/screens/v2/widget_instance/subway_status.ex
cmaddox5: Not at this point. We use `alert_routes/1` to get the `route_ids` and there is an `Enum.uniq()` call there.

lib/screens/v2/widget_instance/subway_status.ex
```diff
@@ -336,112 +415,166 @@ defmodule Screens.V2.WidgetInstance.SubwayStatus do
...
+      serialize_single_alert_row_for_route(grouped_alerts, "Green")
+    else
+      trunk_alerts =
+        Enum.filter(green_line_alerts, fn alert ->
+          alert_affects_gl_trunk_or_whole_line?(alert)
+        end)
 
-    case trunk_alerts do
-      [] ->
+      branch_alerts = green_line_alerts -- trunk_alerts
```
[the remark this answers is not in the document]
cmaddox5: TIL! This is much better!

lib/screens/v2/widget_instance/subway_status.ex
```diff
@@ -62,54 +100,127 @@ defmodule Screens.V2.WidgetInstance.SubwayStatus do
...
+    %{
+      type: :contracted,
+      alerts: alert_rows
+    }
+  end
+
+  # Only executed when route displays one status.
+  def serialize_single_alert_row_for_route(grouped_alerts, route_id) do
     alerts = Map.get(grouped_alerts, route_id)
+    total_alerts = grouped_alerts |> Enum.flat_map(&elem(&1, 1)) |> Enum.uniq() |> length()
```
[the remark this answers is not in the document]
cmaddox5: Good catch. GL unfortunately requires us to use `uniq_by` so that we don't count multiple branch alerts twice. This will allow us to count a Gov't Ctr closure as 2 alerts instead of 5.

jzimbel-mbta (approved):
🐬

PR #1728, opened 2023-05-01
Hotfix for letter platform codes

PaulJKim:
**Asana task**: [[fix] Use `platform_code` field on stop data to get CR track number](https://app.asana.com/0/1185117109217432/1204204347671395/f)

Description
Hotfix for bug when platform code is a letter. Track "number" is fine to always store as a string since we don't do any arithmetic with it.

Note: The change in logic here will incur an update to the config. In the config, the track numbers are stored as integers in the `wayfinding_arrows` field. What's the best way to go about making that config update?

### DIFF
--- a/lib/screens/departures/departure.ex
+++ b/lib/screens/departures/departure.ex
@@ -39,7 +39,7 @@ defmodule Screens.Departures.Departure do
           time: DateTime.t(),
           crowding_level: crowding_level,
           inline_badges: list(map()),
-          track_number: pos_integer() | nil
+          track_number: String.t() | nil
         }
 
   @type query_params :: %{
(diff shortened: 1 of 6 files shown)


jzimbel-mbta:
Re: the issue with existing config, could you add this clause to `Screens.Config.V2.CRDepartures`? (Add before the more permissive clause that's already there.)
```ex
defp value_from_json("wayfinding_arrows", wayfinding_arrows) when is_map(wayfinding_arrows) do
  Map.new(wayfinding_arrows, fn {direction, track_numbers} ->
    {direction, Enum.map(track_numbers, &to_string/1)}
  end)
end
```

It should "patch" the config by converting any integer-valued track numbers to strings.

jzimbel-mbta (changes requested):
Looks good, just need the little change to the config schema mentioned above!

jzimbel-mbta:
Oh and in that same config schema file, also update the type spec for `platform_directions`:
```ex
%{
  left: list(String.t()),
  right: list(String.t())
}
```

jzimbel-mbta (approved):

PR #1729, opened 2023-05-02
refactor: Alerts CandidateGenerator behaviour

cmaddox5:
**Asana task**: [Explore using a Behaviour to enforce CG naming conventions](https://app.asana.com/0/0/1204505661047502/f)

Two big things in this PR:
1. Refactored `ElevatorClosures` so it uses a fetch function in `Screens.Alerts.Alert` instead of handling all fetch logic on its own. Was the only module that took that approach. This makes the module testable (tech debt task [here](https://app.asana.com/0/1185117109217422/1204506798439485/f)).
2. Refactored `ElevatorClosures`, `Alert`, `Dup.Alert`, and `ReconstructedAlert` so they implement a new behaviour `AlertsCandidateGeneratorBehaviour`. This behaviour will help standardize function names and parameters.

- [ ] Tests added?

### DIFF
--- a/lib/screens/alerts/parser.ex
+++ b/lib/screens/alerts/parser.ex
@@ -1,6 +1,17 @@
 defmodule Screens.Alerts.Parser do
   @moduledoc false
 
+  def parse_result(%{"data" => data, "included" => included}) when is_list(data) do
+    facility_data = parse_facility_data(included)
+
+    alerts =
+      data
+      |> Enum.map(&parse_alert/1)
+      |> Enum.reject(&is_nil/1)
+
+    {alerts, facility_data}
+  end
+
   def parse_result(%{"data" => data}) when is_list(data) do
     data
     |> Enum.map(&parse_alert/1)
@@ -170,5 +181,15 @@ defmodule Screens.Alerts.Parser do
 
   defp parse_cause(cause) do
     Map.get(@causes, cause, :unknown)
+  end
+
+  defp parse_facility_data(nil), do: %{}
+
+  defp parse_facility_data(facilities) do
+    facilities
+    |> Enum.map(fn %{"attributes" => %{"short_name" => short_name}, "id" => id} ->
+      {id, short_name}
+    end)
+    |> Enum.into(%{})
   end
 end
--- a/lib/screens/v2/alerts_candidate_generator_behaviour.ex
+++ b/lib/screens/v2/alerts_candidate_generator_behaviour.ex
@@ -1,0 +1,28 @@
+defmodule Screens.V2.AlertsCandidateGeneratorBehaviour do
+  @moduledoc """
+  Behavior for fetching and filtering alerts in a CandidateGenerator.
+
+  - `fetch(opts, fetch_alerts_fn)` is called fetch alerts.
+  - `relevant_alerts(alerts, config, opts)` filters the list of alerts.
+  """
+
+  # Parameters
+  @type opts :: keyword()
+  @type fetch_alerts_fn :: fun()
+  @type config :: Screens.Config.Screen.t()
+
+  # Returns
+  @type alerts :: list()
+  @type facilities :: map()
+  @type fetch_result :: alerts() | {alerts(), facilities()}
+
+  @doc """
+  Fetches a list of alerts using the function provided in parameters.
+  """
+  @callback fetch(opts(), fetch_alerts_fn()) :: {:ok, fetch_result()}
+
+  @doc """
+  Filters alerts so only relevant alerts are considered when creating a WidgetInstance struct.
+  """
+  @callback relevant_alerts(alerts(), config(), opts()) :: alerts()
+end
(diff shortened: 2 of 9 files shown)


lib/screens/v2/alerts_candidate_generator_behaviour.ex
hannahpurcell, line 1: I like really `relevant_alerts` as a behavior function, but not sure about `fetch`. I guess what we gain is forcing typing on that fetched result. Is that worth making things a little more complicated? Maybe!

lib/screens/alerts/parser.ex
```diff
@@ -1,6 +1,17 @@
...
 
+  def parse_result(%{"data" => data, "included" => included}) when is_list(data) do
+    facility_data = parse_facility_data(included)
+
+    alerts =
+      data
+      |> Enum.map(&parse_alert/1)
+      |> Enum.reject(&is_nil/1)
+
+    {alerts, facility_data}
```
hannahpurcell, line 12: What do you think about using this parse function to put the facility data in with the informed entities? It seems like the `facility_data` map is basically used as a lookup table later for the facility ids in informed entities. It would be great to have that data put together here, so then we don't need to pass the facility lookup table all the way down to the widget instance. (But that means quite a bit of refactoring of the widget instance.) If this at all sounds like an interesting idea, please do get @jzimbel-mbta's weigh-in, too

lib/screens/v2/alerts_candidate_generator_behaviour.ex
cmaddox5: That was the idea. Just want to make sure every module is on the same page as to what their fetch function needs to do. The one-off return for `elevator_closures` makes it a little less clear than I would like. @jzimbel-mbta What do you think?

lib/screens/alerts/parser.ex
cmaddox5: I liked that idea so I tried it out. Was a little tricky, but now we don't need the facility map anywhere outside of `parser.ex`.

hannahpurcell (commented):
The elevator closures refactor seems to be working great!

lib/screens/v2/candidate_generator/dup/alerts.ex
```diff
@@ -53,18 +55,25 @@ defmodule Screens.V2.CandidateGenerator.Dup.Alerts do
...
+         {:ok, alerts} <- fetch([route_ids: subway_route_ids_at_stop], fetch_alerts_fn),
          {:ok, stop_sequences} <-
            fetch_parent_station_sequences_fn.(stop_id, subway_route_ids_at_stop) do
       alerts
-      |> Enum.filter(&relevant_alert?(&1, config, stop_sequences, subway_routes_at_stop, now))
+      |> relevant_alerts(config,
+        stop_sequences: stop_sequences,
+        subway_routes_at_stop: subway_routes_at_stop,
+        now: now
+      )
```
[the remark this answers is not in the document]
cmaddox5: Not sure I follow. Do you mean just getting rid of `stop_sequences:` from the parameter list? That won't work because it is part of the `opt` list. The only parameters that are required are `alerts` and `config` because those are the only common parameters in all widgets. Any after that are just any values needed by that particular widget and need to be given as a keyword list.

lib/screens/v2/candidate_generator/widgets/alerts.ex
```diff
@@ -21,17 +23,24 @@ defmodule Screens.V2.CandidateGenerator.Widgets.Alerts do
...
     with {:ok, routes_at_stop} <- fetch_simplified_routes_at_stop_fn.(stop_id, now),
          {:ok, stop_sequences} <- fetch_stop_sequences_through_stop_fn.(stop_id),
          reachable_stop_ids = local_and_downstream_stop_ids(stop_sequences, stop_id),
          route_ids_at_stop = Enum.map(routes_at_stop, & &1.route_id),
          {:ok, alerts} <-
-           fetch_alerts_by_stop_and_route_fn.(reachable_stop_ids, route_ids_at_stop) do
+           fetch(
+             [reachable_stop_ids: reachable_stop_ids, route_ids_at_stop: route_ids_at_stop],
+             fetch_alerts_by_stop_and_route_fn
+           ) do
```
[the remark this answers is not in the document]
cmaddox5: Good catch. I think I just tunneled on the actual variable name 😅 Fixed now.

cmaddox5:
After discussing in Slack, we are not going the `behaviour` route here. Would add an unnecessary layer to the `CandidateGenerator` part of the framework. I will instead manually align all `CandidateGenerators` so they all fetch alerts from `Screens.Alerts.Alert` and filter alerts using a local function called `relevant_alerts`. That will go in a separate PR.

lib/screens/v2/candidate_generator/dup/alerts.ex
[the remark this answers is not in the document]
jzimbel-mbta: To elaborate, this is a sometimes-confusing feature / syntactic sugar in elixir:
```ex
relevant_alerts(alerts, config, stop_sequences: stop_sequences, subway_routes_at_stop: subway_routes_at_stop, now: now)
```
is exactly equivalent to
```ex
relevant_alerts(alerts, config, [stop_sequences: stop_sequences, subway_routes_at_stop: subway_routes_at_stop, now: now])
```
That is, all of the "named" arguments are actually part of one keyword-list argument, usually named `opts`.

[relevant docs](https://elixir-lang.org/getting-started/keywords-and-maps.html#keyword-lists), see the "Furthermore, when a keyword list is the last argument of a function..." part

PR #1730, opened 2023-05-04
refactor: RoutePatterns

cmaddox5:
**Asana task**: [Refactor a couple RoutePattern fetches that appear in Alert CGs](https://app.asana.com/0/0/1204505661047503/f)

There actually isn't much to refactor here. The functions seem similar but the returns are very different. I tried to improve the comments so devs can understand the difference. Also cleaned up some code I saw while I was in there. 

- [ ] Tests added?


jzimbel-mbta (approved):
Love some good documentation

PR #1731, opened 2023-05-04
Make Polly read 'minutes', not 'meters'

PaulJKim:
**Asana task**: [🐞[Screens] Polly announces "meters" instead of "minutes"](https://app.asana.com/0/1185117109217432/1203713399498235/f)

Description
Adds essentially an override for the status readout for the subway status widget.


cmaddox5 (approved):
This bug is so funny. The 2.0 update for this widget should help avoid this issue, but in the meantime this looks great.

PR #1732, opened 2023-05-04
fix: Use defined order for header sizes array

jzimbel-mbta:
**Asana task**: ad hoc

`Object.keys` doesn't guarantee any consistent ordering in the array it returns, so we may have been randomly giving `useTextResizer` its `sizes` array in reverse order here!

- [ ] Tests added?

### DIFF
--- a/assets/src/components/v2/normal_header.tsx
+++ b/assets/src/components/v2/normal_header.tsx
@@ -170,7 +170,7 @@ const NormalHeader: ComponentType<Props> = ({
   accentPattern,
 }) => {
   const { ref: headerRef, size: headerSize } = useTextResizer({
-    sizes: Object.keys(TitleSize),
+    sizes: [TitleSize.small, TitleSize.large],
     maxHeight: maxHeight,
     resetDependencies: [text],
   });


PaulJKim:
Hmm it seems to me like documentation for `Object.keys()` says that it _does_ return the keys in their defined order!

jzimbel-mbta:
Huh, you're right! Just to be extra safe, [I tested it on an `enum`](https://www.typescriptlang.org/play?#code/PTAEHUFMBsGMHsC2lQBd5oBYoCoE8AHSAZVgCcBLA1UABWgEM8BzM+AVwDsATAGiwoBnUENANQAd0gAjQRVSQAUCEmYKsTKGYUAbpGF4OY0BoadYKdJMoL+gzAzIoz3UNEiPOofEVKVqAHSKymAAmkYI7NCuqGqcANag8ABmIjQUXrFOKBJMggBcISGgoAC0oACCbvCwDKgU8JkY7p7ehCTkVDQS2E6gnPCxGcwmZqDSTgzxxWWVoASMFmgYkAAeRJTInN3ymj4d-jSCeNsMq-wuoPaOltigAKoASgAywhK7SbGQZIIz5VWCFzSeCrZagNYbChbHaxUDcCjJZLfSDbExIAgUdxkUBIursJzCFJtXydajBFHsRDeeTuYgUABeKAA3ooSoJEAxoNBQABeUAAInZnOg-N4rLcjmYKD5-MYZClosUAF9gghOIJ4O4AtB4MwABQAeWkACtILBUAF4pA8II9TgaSQGZAAJTOgDcwRUOGaHjIXkQ8D6DGB7BoXwlnGY7AYUv4sGg6kSwfgehEXn5AFFVgxEAt9Pyktj+eAHKgAOTCABykAk-KCKgNXzI70EkH4UqOqEcClc0jwoCcAZ0wyw+hQCEQ0OEl3DEkD0REBg42IWTFYHB4QRKQA) since under the hood TS might create the corresponding JS object in a different order, but it looks like we're all good. False alarm!

PR #1735, opened 2023-05-08
refactor: Alerts with location feature branch

hannahpurcell:
First thing was removing some old stuff from alert.ex. It has nothing to do with the later refactor, so separate PR to reduce review load.


lib/screens/alerts/alert.ex
```diff
@@ -240,6 +230,8 @@ defmodule Screens.Alerts.Alert do
     end
   end
 
+  # Todo: These private format_query_params are the same in Departures and Route
```
hannahpurcell, line 233:
```suggestion
  # credo:disable-for-next-line
  # Todo: These private format_query_params are the same in Departures and Route
```

cmaddox5 (approved):

PR #1741, opened 2023-05-12
Suppress alert 495152 on Porter & Charles/MGH pre-fare screens

jzimbel-mbta:
**Asana task**: ad hoc

[Slack thread](https://mbta.slack.com/archives/C039ARUM48H/p1683832377453069?thread_ts=1683553807.621649&cid=C039ARUM48H)

- [ ] Tests added?

### DIFF
--- a/lib/screens/v2/widget_instance/reconstructed_alert.ex
+++ b/lib/screens/v2/widget_instance/reconstructed_alert.ex
@@ -686,8 +686,11 @@ defmodule Screens.V2.WidgetInstance.ReconstructedAlert do
   def temporarily_override_alert(%__MODULE__{} = t) do
     # Prevent Porter and Charles/MGH pre-fare screens from incorrectly communicating
     # a RL alert that affects both the Ashmont and Braintree branches.
-    t.alert.id != "495152" and
-      t.screen.app_params.reconstructed_alert_widget.stop_id not in ["place-portr", "place-chmnl"]
+    not (t.alert.id == "495152" and
+           t.screen.app_params.reconstructed_alert_widget.stop_id in [
+             "place-portr",
+             "place-chmnl"
+           ])
   end
 
   defimpl Screens.V2.WidgetInstance do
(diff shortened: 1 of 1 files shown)


lib/screens/v2/widget_instance/reconstructed_alert.ex
```diff
@@ -684,10 +684,10 @@ defmodule Screens.V2.WidgetInstance.ReconstructedAlert do
   def alert_ids(%__MODULE__{} = t), do: [t.alert.id]
 
   def temporarily_override_alert(%__MODULE__{} = t) do
-    # All screens that would show either of the following alerts
-    # will show static Evergreen content instead
-    # https://app.asana.com/0/1185117109217413/1203840359759366/f
-    t.alert.id not in ["482408", "482406"]
+    # Prevent Porter and Charles/MGH pre-fare screens from incorrectly communicating
+    # a RL alert that affects both the Ashmont and Braintree branches.
+    t.alert.id != "495152" and
```
cmaddox5, line 689: Should this be an `or` instead? We still want this alert to show at Ashmont.

hannahpurcell: Alert is valid if alert ID is not 495152 AND the screen is not at Porter and Charles MGH. I think this is right

cmaddox5: `495152 != 495152 and "place-asmnl" not in ["place-portr", "place-chmnl"]` would return `false`, though, and we want this alert to return `true` at Ashmont. I have major end-of-week brain right now though so it's taking me longer than normal to wrap my head around this.

jzimbel-mbta: Oof, yeah you're right. The name of the function threw me off, I thought it needed to return `true` to suppress the alert and `false` otherwise. Thank you, good catch.

jzimbel-mbta: Another reason to give boolean-returning / predicate functions names that sound like questions. This one as it works now would be named `not_temporary_override?`

hannahpurcell: OH ohhhh got it, thanks!!

cmaddox5: O yikes. I was actually right on accident. I thought this was `valid_candidate?` 😅 .

jzimbel-mbta: Ok, the logic is updated. It's now equivalent to
> If we're showing the bad alert at Porter or Charles:
>   return false
> else:
>   return true

hannahpurcell (approved):

PR #1742, opened 2023-05-15
Alerts refactor, step 2: Add location context to all WidgetInstances

hannahpurcell:
Replaces https://github.com/mbta/screens/pull/1737. Added fetch_location_context to all alert CGs, refactored WidgetInstances to use that context and interact with BaseAlert. 

I marked some ideas for future work with TODO, which I know breaks credo. My question there is, should that work be done now, or can it wait for the future? If it can wait, I'll have credo ignore the line. (I don't think it's a problem to keep TODOs in code as a reminder for future work, but I still want to keep the credo check so it helps catch TODOs we did NOT want to intentionally leave there.)

Tests are in a separate PR: https://github.com/mbta/screens/pull/1743

Needs:
- [x] testing of Mattapan outages on different screens
- [x] decide which todos we should keep and which should be ditched.
- [x] check on [that bug](https://github.com/mbta/screens/pull/1710#discussion_r1175724871) Jon noticed in the OG PR (@jzimbel-mbta please advise)

### DIFF
--- a/lib/screens/stops/stop.ex
+++ b/lib/screens/stops/stop.ex
@@ -437,29 +437,35 @@ defmodule Screens.Stops.Stop do
           DateTime.t()
         ) :: {:ok, LocationContext.t()}
   def fetch_location_context(app, stop_id, now) do
-    alert_route_types = get_route_type_filter(app, stop_id)
-    {:ok, routes_at_stop} = Route.fetch_routes_by_stop(stop_id, now, alert_route_types)
-    route_ids_at_stop = Enum.map(routes_at_stop, & &1.route_id)
-
-    {:ok, stop_sequences} =
-      cond do
-        app in [BusEink, BusShelter, GlEink] ->
-          RoutePattern.fetch_stop_sequences_through_stop(stop_id)
-      
-        app in [PreFare, Dup] ->
-          RoutePattern.fetch_parent_station_sequences_through_stop(stop_id, route_ids_at_stop)
-      end
-
-    {:ok,
-     %LocationContext{
-       home_stop: stop_id,
-       stop_sequences: stop_sequences,
-       upstream_stops: upstream_stop_id_set(stop_id, stop_sequences),
-       downstream_stops: downstream_stop_id_set(stop_id, stop_sequences),
-       routes: routes_at_stop,
-       route_ids_at_stop: route_ids_at_stop,
-       alert_route_types: alert_route_types
-     }}
+    with alert_route_types <- get_route_type_filter(app, stop_id),
+         {:ok, routes_at_stop} <- Route.fetch_routes_by_stop(stop_id, now, alert_route_types),
+         route_ids_at_stop <- Enum.map(routes_at_stop, & &1.route_id),
+         {:ok, stop_sequences} <-
+            (cond do
+              app in [BusEink, BusShelter, GlEink] ->
+                RoutePattern.fetch_stop_sequences_through_stop(stop_id)
+            
+              app in [PreFare, Dup] ->
+                RoutePattern.fetch_parent_station_sequences_through_stop(stop_id, route_ids_at_stop)
+            end)
+          do
+      {:ok,
+        %LocationContext{
+          home_stop: stop_id,
+          stop_sequences: stop_sequences,
+          upstream_stops: upstream_stop_id_set(stop_id, stop_sequences),
+          downstream_stops: downstream_stop_id_set(stop_id, stop_sequences),
+          routes: routes_at_stop,
+          route_ids_at_stop: route_ids_at_stop,
+          alert_route_types: alert_route_types
+        }}
+    else
+      :error ->
+        Logger.error(
+          "[fetch_location_context fetch error] Failed to get location context for an alert: stop_id=#{stop_id}"
+        )
+        :error
+    end
   end
 
   # Returns the route types we care about for the alerts of this screen type / place
--- a/lib/screens/stops/stop.ex
+++ b/lib/screens/stops/stop.ex
@@ -439,31 +439,30 @@ defmodule Screens.Stops.Stop do
   def fetch_location_context(app, stop_id, now) do
     with alert_route_types <- get_route_type_filter(app, stop_id),
          {:ok, routes_at_stop} <- Route.fetch_routes_by_stop(stop_id, now, alert_route_types),
-         route_ids_at_stop <- Enum.map(routes_at_stop, & &1.route_id),
+         route_ids <- Route.route_ids(routes_at_stop),
          {:ok, stop_sequences} <-
-            (cond do
+           (cond do
               app in [BusEink, BusShelter, GlEink] ->
                 RoutePattern.fetch_stop_sequences_through_stop(stop_id)
-            
+
               app in [PreFare, Dup] ->
-                RoutePattern.fetch_parent_station_sequences_through_stop(stop_id, route_ids_at_stop)
-            end)
-          do
+                RoutePattern.fetch_parent_station_sequences_through_stop(stop_id, route_ids)
+            end) do
       {:ok,
-        %LocationContext{
-          home_stop: stop_id,
-          stop_sequences: stop_sequences,
-          upstream_stops: upstream_stop_id_set(stop_id, stop_sequences),
-          downstream_stops: downstream_stop_id_set(stop_id, stop_sequences),
-          routes: routes_at_stop,
-          route_ids_at_stop: route_ids_at_stop,
-          alert_route_types: alert_route_types
-        }}
+       %LocationContext{
+         home_stop: stop_id,
+         stop_sequences: stop_sequences,
+         upstream_stops: upstream_stop_id_set(stop_id, stop_sequences),
+         downstream_stops: downstream_stop_id_set(stop_id, stop_sequences),
+         routes: routes_at_stop,
+         alert_route_types: alert_route_types
+       }}
     else
       :error ->
         Logger.error(
           "[fetch_location_context fetch error] Failed to get location context for an alert: stop_id=#{stop_id}"
         )
+
         :error
     end
   end
--- a/lib/screens/stops/stop.ex
+++ b/lib/screens/stops/stop.ex
@@ -463,6 +463,8 @@ defmodule Screens.Stops.Stop do
           list(atom())
   def get_route_type_filter(app, _) when app in [BusEink, BusShelter], do: [:bus]
   def get_route_type_filter(GlEink, _), do: [:light_rail]
+  # Ashmont should not show Mattapan alerts for PreFare or DUP
+  def get_route_type_filter(_, "place-asmnl"), do: [:subway]
   def get_route_type_filter(PreFare, _), do: [:light_rail, :subway]
   # WTC is a special bus-only case
   def get_route_type_filter(Dup, "place-wtcst"), do: [:bus]
--- a/lib/screens/stops/stop.ex
+++ b/lib/screens/stops/stop.ex
@@ -10,8 +10,14 @@ defmodule Screens.Stops.Stop do
 
   require Logger
 
+  alias Screens.Config.V2.{BusEink, BusShelter, GlEink, PreFare, Dup}
+  alias Screens.LocationContext
+  alias Screens.RoutePatterns.RoutePattern
   alias Screens.Routes
+  alias Screens.Routes.Route
   alias Screens.Stops.StationsWithRoutesAgent
+  alias Screens.Stops.Stop
+  alias Screens.Util
   alias Screens.V3Api
 
   defstruct id: nil,
@@ -390,6 +396,10 @@ defmodule Screens.Stops.Stop do
     |> Enum.uniq()
   end
 
+  def get_all_routes_stop_sequence do
+    @route_stop_sequences
+  end
+
   defp sequence_match?(stop_sequence, informed_entities) do
     ie_stops =
       informed_entities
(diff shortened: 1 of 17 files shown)


lib/screens/v2/candidate_generator/widgets/alerts.ex
hannahpurcell, line 1: @jzimbel-mbta

lib/screens/v2/widget_instance/alert.ex
hannahpurcell, line 1: @jzimbel-mbta

lib/screens/v2/location_context.ex
hannahpurcell, line 1: @cmaddox5

lib/screens/v2/candidate_generator/widgets/reconstructed_alert.ex
hannahpurcell, line 1: @cmaddox5

lib/screens/stops/stop.ex
hannahpurcell, line 1: @cmaddox5

hannahpurcell (commented):
My attempt at distributing the review. If this is too burdensome, do folks have other ideas for lightening the load?

lib/screens/v2/candidate_generator/widgets/reconstructed_alert.ex
```diff
@@ -23,32 +22,24 @@ defmodule Screens.V2.CandidateGenerator.Widgets.ReconstructedAlert do
...
+      |> Enum.filter(&relevant?(&1, config, location_context, now))
       |> Enum.map(fn alert ->
         %ReconstructedAlert{
           screen: config,
           alert: alert,
           now: now,
-          stop_sequences: stop_sequences,
-          routes_at_stop: routes_at_stop,
+          location_context: location_context,
+          # TODO: these two items look like location stuff? Should they be?
```
cmaddox5, line 40: For `informed_stations_string`, I think no. It is getting info based on the `informed_entities` so it makes sense to keep that separate. `is_terminal_station` might be a good thing to have in the `location_context` though. The logic for the function should be the same everywhere so moving that to `fetch_location_context` seems safe.

lib/screens/v2/widget_instance/elevator_status.ex
```diff
@@ -18,6 +18,8 @@ defmodule Screens.V2.WidgetInstance.ElevatorStatus do
 
     defstruct station: nil
 
+    # TODO: just curious, do the separate `alert_ids` definitions lead to double-counting
```
cmaddox5, line 21: I was very curious as well so I looked into this. An `alert_id` is counted twice if it has a `DetailPage` and is also on a `ListPage`, but the duplicates are filtered on line 553.

lib/screens/v2/location_context.ex
```diff
@@ -0,0 +1,25 @@
+defmodule Screens.LocationContext do
+  @moduledoc false
+
+  alias Screens.Routes.Route
+  alias Screens.RouteType
+  alias Screens.Stops.Stop
+
+  defstruct home_stop: "",
```
cmaddox5, line 8: If a struct is created without one of these attributes and the default is used, will all logic still work as expected? If not, may be helpful to move some of the attributes into a `@enforce_keys` list.

cmaddox5 (commented):
Wow, great work. I think I got to all of my assigned files. Just a few comments.

lib/screens/v2/location_context.ex
hannahpurcell: Hmm, I see. I suppose I might as well enforce all keys for now, and we can un-enforce any that pose a problem in the future. 

Using the default `home_stop` of `""` is fine for GL e-ink, but it breaks the Elevator widget. Using the default `[]` for `stop_sequences` doesn't break any screen, but it does mess up how the alert displays on DUPs. Does that mean we need to enforce `home_stop`, but don't necessarily need to enforce `stop_sequences`, or must we enforce both?

cmaddox5: I can see an argument for either way, but I would lean towards just enforcing attributes that need to be there for the screen to work. If the default throws off the visuals, the dev working with the struct should be able to catch that on their own.

lib/screens/v2/candidate_generator/widgets/reconstructed_alert.ex
hannahpurcell: On second thought, `is_terminal_station` is only used by Pre-fare alerts right now, so maybe I won't add it to LocationContext unless it becomes necessary for more screens later?

cmaddox5: That's fair. I don't see that level of context being important for any other alert widget.

cmaddox5 (approved):
Changes to the files I was tagged on look good! Great work!

jzimbel-mbta:
The files you tagged me on lost their comments since they became outdated :(

Listing them here so I can find the ones assigned to me:
- lib/screens/v2/widget_instance/subway_status.ex]
- lib/screens/v2/widget_instance/common/base_alert.ex
- lib/screens/v2/widget_instance/dup_alert.ex
- lib/screens/v2/candidate_generator/dup/alerts.ex
- lib/screens/v2/candidate_generator/widgets/alerts.ex
- lib/screens/v2/widget_instance/alert.ex

Looks like there's no way to link to a file in a PR, in a way that it always points to the most recent version of that file, so these will have to just be plaintext. Time to submit a PR for github's PR feature...

jzimbel-mbta:
> check on https://github.com/mbta/screens/pull/1710#discussion_r1175724871 Jon noticed in the OG PR

Discussed in slack—it actually wasn't a bug, the function was working as intended but not very clearly documented! So no code change needed, just maybe a clarification in the comment above the function def.

lib/screens/v2/widget_instance/common/base_alert.ex
```diff
@@ -236,23 +200,68 @@ defmodule Screens.V2.WidgetInstance.Common.BaseAlert do
...
-    t
-    |> SAW.routes_at_stop()
+  def active_routes_at_stop(%{location_context: %{routes: routes}}) do
+    routes
     |> Enum.filter(& &1.active?)
     |> MapSet.new(& &1.route_id)
   end
 
-  @spec informed_routes_at_home_stop(t()) :: MapSet.t(Route.id())
+  def informed_subway_routes(%{screen: %Screen{app_id: app_id}} = t) do
```
jzimbel-mbta, line 215: Needs a `@spec`, a `@doc` wouldn't hurt either

lib/screens/v2/widget_instance/common/base_alert.ex
```diff
@@ -236,23 +200,68 @@ defmodule Screens.V2.WidgetInstance.Common.BaseAlert do
...
+    # If the alert impacts CR or other lines, weed that out
+    |> Enum.filter(fn e ->
+      Enum.member?(["Red", "Orange", "Green", "Blue"] ++ @green_line_branches, e)
+    end)
+    |> Enum.uniq()
+    |> consolidate_gl(app_id)
+  end
+
+  # GL E-ink consolidates the GL branches to Green Line if there are > 2 branches
+  # We wdant to list all affected branches for the alert and not just the one serving the home stop
```
jzimbel-mbta, line 228:
```suggestion
  # We want to list all affected branches for the alert and not just the one serving the home stop
```

Is this comment also true for the pre-fare clause? Seems like it is.

lib/screens/v2/widget_instance/common/base_alert.ex
```diff
@@ -262,27 +271,42 @@ defmodule Screens.V2.WidgetInstance.Common.BaseAlert do
...
         %{stop: ^home_stop, route: nil}, _uninformed ->
           {:halt, empty_set}
 
+        # If entity is home stop AND a route, then just that route is marked informed
         %{stop: ^home_stop, route: route}, uninformed ->
           {:cont, MapSet.delete(uninformed, route)}
 
+        # Removed the case with downstream alerts because:
+        #  - we don't want to consider downstream alerts for a takeover (bus, prefare, dup)
+        #  - we don't want ever show downstream alerts on bus shelter / bus e-ink
```
jzimbel-mbta, line 298:
```suggestion
        #  - we don't want to ever show downstream alerts on bus shelter / bus e-ink
```

jzimbel-mbta (commented):
NOICE. I had mostly very minor comments. One about maybe merging `routes` and `route_ids_at_stop` into one field, if that makes sense.

lib/screens/v2/widget_instance/common/base_alert.ex
```diff
@@ -64,12 +45,13 @@ defmodule Screens.V2.WidgetInstance.Common.BaseAlert do
   - the widget does not have a map of headsign matchers (`SingleAlertWidget.headsign_matchers(t)` returns nil)
   """
   @spec get_headsign_from_informed_entities(t()) :: headsign | nil
-  def get_headsign_from_informed_entities(t) do
-    with headsign_matchers when is_map(headsign_matchers) <- SAW.headsign_matchers(t) do
+  def get_headsign_from_informed_entities(%{screen: %Screen{app_id: app_id}} = t)
+      when app_id in [:dup_v2, :pre_fare_v2] do
+    with headsign_matchers when is_map(headsign_matchers) <- headsign_matchers(t) do
```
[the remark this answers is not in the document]
hannahpurcell: Oh I see! The `when app_id in []` seems a redundant then. Should I remove that bit, then, or change the spec? Either seems fine to me

lib/screens/v2/widget_instance/common/base_alert.ex
hannahpurcell: Thanks! Shouldn't be

lib/screens/v2/widget_instance/common/base_alert.ex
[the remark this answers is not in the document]
jzimbel-mbta: I think your change is an improvement! We shouldn't be calling this function for screen/widget types that don't do headsign matching, so raising an exception in those cases is the right thing to do.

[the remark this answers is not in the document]
jzimbel-mbta: So I'd recommend changing the spec and keeping the guard.

lib/screens/v2/widget_instance/common/base_alert.ex
```diff
@@ -236,23 +200,68 @@ defmodule Screens.V2.WidgetInstance.Common.BaseAlert do
...
   def informs_all_active_routes_at_home_stop?(t) do
-    MapSet.subset?(active_routes_at_stop(t), informed_routes_at_home_stop(t))
+    MapSet.subset?(active_routes_at_stop(t), MapSet.new(informed_routes_at_home_stop(t)))
   end
 
   @spec active_routes_at_stop(t()) :: MapSet.t(route_id())
-  def active_routes_at_stop(t) do
-    t
-    |> SAW.routes_at_stop()
+  def active_routes_at_stop(%{location_context: %{routes: routes}}) do
```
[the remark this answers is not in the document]
jzimbel-mbta: Hmm, ok... though if the reason for not using a helper function is performance, I wouldn't worry about it. It's a minuscule amount of work to do that mapping.

I would disagree if the argument is about ergonomics as well, because BaseAlert already has other helper functions that do similar basic "accessing" of values, e.g. [`informed_entities`](https://github.com/mbta/screens/blob/bde871e76870f24cca7e19c0292260d1d306c748/lib/screens/v2/widget_instance/common/base_alert.ex#L82).

If you still want to keep the duplicate data fields, then ok, but I definitely prefer merging them.

lib/screens/v2/widget_instance/common/base_alert.ex
jzimbel-mbta: I'm not following this comment's meaning then—is it talking about what the `affected_routes` list should contain?

As in, in the places where this function gets called for GL e-ink, we should make sure that `affected_routes` includes all branches affected by the alert, instead of just the affected branches that also serve the home stop?

lib/screens/v2/widget_instance/common/base_alert.ex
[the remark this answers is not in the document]
hannahpurcell: Since the helper would be "get route ids from a list of routes", I think that would better live in `Routes`, which means that module would need to be imported to other modules where it isn't currently needed. 

I guess I'll do the merging and see how it feels.

lib/screens/v2/widget_instance/common/base_alert.ex
hannahpurcell: OH! You are totally right. That comment was actually a relic from when I thought the consolidation only happened for GL e-ink, and then later made the clause for Pre-Fare, and didn't integrate the two clauses better. 

This made me realize that I could organize this a little better. Like, the GL clause would work for Pre-Fare if the conditional was `if length(green_routes) > 3`, so I'm going to consolidate a bit

lib/screens/stops/stop.ex
```diff
@@ -417,7 +428,61 @@ defmodule Screens.Stops.Stop do
...
+    route_ids_at_stop = Enum.map(routes_at_stop, & &1.route_id)
+
+    {:ok, stop_sequences} =
+      if app in [BusEink, BusShelter, GlEink] do
+        RoutePattern.fetch_stop_sequences_through_stop(stop_id)
+      else
+        RoutePattern.fetch_parent_station_sequences_through_stop(stop_id, route_ids_at_stop)
+      end
+
+    {:ok,
```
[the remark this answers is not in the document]
hannahpurcell: Oh thanks for that catch! I only half-started the error handling. Lemme think this through... 

So, the api fetches that could fail with an error are getting `routes_at_stop` or `stop_sequences`. I'd imagine those could fail on a real screen if just the api goes a lil wonky, so what would we want to see? The alert not get generated as a widget, and the issue logged. So yeah, I'll add an :error case for if those function calls fail.

lib/screens/v2/widget_instance/common/base_alert.ex
```diff
@@ -1,48 +1,29 @@
...
-  Common logic for structs that implement the `Screens.V2.SingleAlertWidget` protocol.
-
-  ***
-  None of the code in this module should access struct fields directly.
-
-  In order to maintain clarity on what values these functions require,
-  use only callbacks from `Screens.V2.SingleAlertWidget` to interact with inputs.
-  If new values are required, add new callbacks to the protocol and update implementing modules accordingly.
-  ***
+  Common logic for Alerts.
```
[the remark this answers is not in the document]
hannahpurcell: Mmmm yes. Got me thinking: not all functions in this BaseAlert file require location context -- `informed_entities`, `effect`, `informed_subway_routes`, specifically. Should those live in `Screens.Alerts.Alert`? At the moment, it seems to mostly handle fetching and setting on alerts, and less getting (and these functions are getters). I'm torn:

1. Should these getters be moved to `Screens.Alerts.Alert`, and then the `BaseAlert` module can be renamed `LocalAlert` or something indicating it's functions with locations attached
2. Are things better organized as they are, where the fetch-and-set stuff is stored together, and post-processing stuff is stored together?

I like #2, but then what do I call this module, and where should it go?

lib/screens/v2/widget_instance/subway_status.ex
```diff
@@ -153,6 +153,8 @@ defmodule Screens.V2.WidgetInstance.SubwayStatus do
     %{full: direction, abbrev: direction}
   end
 
+  # TODO: get_endpoints is a common function
+  # could also be consolidated?
```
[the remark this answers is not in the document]
hannahpurcell: 👍 I took a closer look at the code for this, and making this consolidation could disrupt `subway_status` a little more than I want to at this time. Ok if I leave the TODO, add a little more detail, and disable the credo check for that todo?

lib/screens/v2/widget_instance/common/base_alert.ex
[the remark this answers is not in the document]
jzimbel-mbta: I think it would make "logical" sense to put the functions that only need an Alert, no location context, in `Screens.Alerts.Alert`.
But if that makes the code here way more verbose—`Alert.informed_entities(alert)` is a bit longer than `informed_entities(alert)`—feel free to leave them.

I do think this module could stand to be renamed either way, since it is like 99% about querying an Alert combined with a LocationContext.

> what do I call this module

`LocalAlert` seems good, or `LocalizedAlert` if it doesn't feel too long. `LocatedAlert`, `ScopedAlert`, `PositionedAlert`, `AlertXPlace`? :thinkies:

> where should it go

I think since it's 100% about the data, it could live right in `Screens.<ModuleName>` similar to `Screens.Alerts.Alert`, `Screens.RouteType`, stuff like that. The fact that you can pass certain `WidgetInstance` structs to its functions is just a convenient coincidence.

lib/screens/v2/localized_alert.ex
```diff
@@ -12,12 +12,11 @@ defmodule Screens.V2.WidgetInstance.Common.BaseAlert do
...
   alias Screens.V2.WidgetInstance.{DupAlert, ElevatorStatus, ReconstructedAlert}
 
-  @type t ::
-          AlertWidget.t()
+  @type t :: AlertWidget.t()
           | DupAlert.t()
           | ReconstructedAlert.t()
           | ElevatorStatus.t()
-          | %{alert: Alert.t(), location_context: LocationContext.t()}
+          | %{alert: Alert.t(), location_context: LocationContext.t(), screen?: Screen.t()}
```
jzimbel-mbta, line 20: Very small thing: `screen?:` as a key means it literally has a key named `:screen?` (since elixir atoms can have question marks). If you want to mark the field as optional like the `?` achieves in TypeScript, it's this much less elegant syntax:
```suggestion
          | %{optional(:screen) => Screen.t(), alert: Alert.t(), location_context: LocationContext.t()}
```
(It needs to come first because reasons)

jzimbel-mbta (approved):
Looks great! 🚀
(Approved assuming you get CI passing and address that last thing about the `screen?:` key)

lib/screens/v2/localized_alert.ex
hannahpurcell: Oh my gosh 😅 hahahhaa I actually laughed aloud at this. My "codeswitching" failed miserably. So glad you noticed

hannahpurcell:
> Looks great! 🚀 (Approved assuming you get CI passing and address that last thing about the `screen?:` key)

Thanks!! Merging as long as CI except for tests pass, because test fixes are in a separate PR

PR #1745, opened 2023-05-17
feat: E-Ink Subway Status

cmaddox5:
**Asana task**: [[Subway Status 2.0] E-Ink frontend](https://app.asana.com/0/1185117109217413/1203775859825087/f)

Added the Subway Status component for E-Ink. Was able to leave the backend alone. Only difference is this component only shows `ContractedAlert`s. If there is a scenario where an `ExtendedAlert` is visible, show it as `contracted` and display the header.

- [ ] Tests added?


assets/css/v2/lcd_common_styles/subway_status.scss
```diff
@@ -1,170 +1,172 @@
 .subway-status {
```
cmaddox5, line 1: This file replaced the old `subway_status` styles. It used to be called `subway_status_v2`. No changes were actually made.

assets/src/apps/v2/bus_shelter.tsx
```diff
@@ -34,7 +34,7 @@ import Placeholder from "Components/v2/placeholder";
 import LinkFooter from "Components/v2/bus_shelter/link_footer";
 import NormalHeader from "Components/v2/lcd/normal_header";
 import NormalDepartures from "Components/v2/departures/normal_departures";
-import SubwayStatus from "Components/v2/lcd/subway_status_v2";
+import SubwayStatus from "Components/v2/subway_status/lcd/subway_status";
```
hannahpurcell, line 37: This organization change seems odd, but tbh our v2 components are all kinda disorganized. So I'mma ask about it in slack, so how we'd like things organized.

assets/static/images/pills/eink/gl-c.svg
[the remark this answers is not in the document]
cmaddox5: My bad. Should be good now.

assets/src/apps/v2/bus_shelter.tsx
cmaddox5: The main reason for the new folder is so the common file has a sensible place to live. I couldn't think of an existing folder that would make sense for it, so I had it share a folder with the two components. Very much open to change on it.

hannahpurcell: I guess I like things being organized by widgets, so I like there being a `subway_status` folder with a `common` file inside. At the moment, we don't plan on putting anymore styles in the lcd and e_ink folders, so maybe the css files could just be called `lcd_subway_status` and `eink_subway_status`

assets/src/components/v2/subway_status/eink/subway_status.tsx
```diff
@@ -0,0 +1,312 @@
...
+      <ContractedStatus
+        alerts={alerts}
+        showInlineBranches={showInlineBranches}
+      />
+      <div className="subway-status_status_rule" />
+    </div>
+  );
+};
+
+const getRoutePillObject = (
```
hannahpurcell, line 69: I'm using this Miro as my frame of reference, and it seems like the routePill should never be more complicated than just the line name. Why is there so much logic here and in `SubwayStatusRoutePill` to calculate the big route pill?

cmaddox5: If you look at [this Figma file](https://www.figma.com/file/TjbJshbl7fk63GDebLokmu/Widgets-%E2%80%94-Subway-status?type=design&node-id=345-26902&t=wXqe84q2C22dPaKT-4), there is an update to GL pill that isn't reflected in the Miro board. If there is only one GL alert and it is not a trunk alert/affects the whole line, we show a GL + branch icon combo similar to LCD. If the alert affects the trunk/whole line or there are more than one alerts, we show the GL pill and inline the branch icons next to the alert.

assets/src/components/v2/subway_status/eink/subway_status.tsx
```diff
@@ -0,0 +1,312 @@
...
+    color === LineColor.Green &&
+    ((isContracted(section) && section.alerts.length > 1) ||
+      isExtended(section));
+
+  return (
+    <div className="subway-status_row">
+      <div className="subway-status_route-pill-container">
+        <SubwayStatusRoutePill routePill={routePill} />
+      </div>
+      <ContractedStatus
```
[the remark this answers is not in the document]
cmaddox5: That's fair. Changed the name of that component and the `ContractedAlert` component.

assets/src/apps/v2/bus_shelter.tsx
cmaddox5: You mean move the CSS files out of their eink/lcd folders and into a `subway_status` folder? That's definitely doable, but is a bit of a departure from the way the rest of our CSS is organized (not to say the current way is the best way).

hannahpurcell (commented):
Ok, this component is complex! But so is the design logic. So, I think the core logic is good. But I did have a hard time orienting myself in `subway_status.tsx`. Couple things that may help: Why is `BasicAlert` named basic? Could a comment be added as a general summary for `shouldShowHeader` and `getRoutePillObject`? Any other comments that could help?

Separate tiny styling thing: the first letter for the alert without a pill should be centered over the circle pill below it. So the S in Suspension should be aligned with the circle pill E, I think.

Screenshot 2023-05-25 at 11 33 32 AM

assets/src/apps/v2/bus_shelter.tsx
hannahpurcell: I think both the CSS and the TSX files could be called `lcd_subway_status.extension` and put right in the subway_status folder. But if that doesn't gel with our organization system, that's fine! It was a nitpick anyway

cmaddox5:
> Ok, this component is complex! But so is the design logic. So, I think the core logic is good. But I did have a hard time orienting myself in `subway_status.tsx`. Couple things that may help: Why is `BasicAlert` named basic? Could a comment be added as a general summary for `shouldShowHeader` and `getRoutePillObject`? Any other comments that could help?
> 
> Separate tiny styling thing: the first letter for the alert without a pill should be centered over the circle pill below it. So the S in Suspension should be aligned with the circle pill E, I think.
> 
> Screenshot 2023-05-25 at 11 33 32 AM

The reason for the name `BasicAlert` is the name `Alert` is already taken by the model. This was also the name Jon used in LCD so I borrowed from there. 

Comments were definitely needed so I added some to help with some of the functions/components.

As for the styling, I'm checking with Betsy real quick to see what the alignment should be. The [Figma](https://www.figma.com/file/TjbJshbl7fk63GDebLokmu?node-id=345:24836#454801149) shows what I have, but it's unclear if that is actually the requirement.

assets/src/apps/v2/bus_shelter.tsx
cmaddox5: I don't mind that for the components so I changed those names and moved the files. I think I'm gonna leave the CSS files how they are now to avoid a one-off for this component. I like the widget organization over app like how the components are organized, but it would be a larger task to get all CSS into that.

hannahpurcell (approved):
Ohhh you're totally right on the styling. The alignment seemed more strikingly "off" in the context of the screen, but I see it's still a little odd in the design, too. Ok, I'm happy to approve this, and you can go forward depending on Betsy's reply.

Thanks for the comments!

cmaddox5:
> Ohhh you're totally right on the styling. The alignment seemed more strikingly "off" in the context of the screen, but I see it's still a little odd in the design, too. Ok, I'm happy to approve this, and you can go forward depending on Betsy's reply.
> 
> Thanks for the comments!

FYI, looks like that is the alignment we need to go with: https://www.figma.com/file/TjbJshbl7fk63GDebLokmu?node-id=0:140#454863577.

PR #1746, opened 2023-05-17
Added window property for js capture in iframes.

cmaddox5:
**Asana task**: [Configure Fullstory to capture iframes](https://app.asana.com/0/1185117109217413/1204351329872461/f)

Added this to Screenplay, but misread the docs saying it also needs to be in the app showing in the `iframe`. This is only set in prod on simulation pages, so live screens won't have it.

- [ ] Tests added?


hannahpurcell (approved):
Let's try it!

PR #1748, opened 2023-05-17
Added window property for js capture in iframes. Part 2

cmaddox5:
I had the conditional backwards...

- [ ] Tests added?


hannahpurcell (approved):
derp

PR #1749, opened 2023-05-17
feat: Fullstory initialization

cmaddox5:
**Asana task**: [Configure Fullstory to capture iframes](https://app.asana.com/0/1185117109217413/1204351329872461/f)

I believe I misunderstood the docs. Fullstory needs to be initialized in the iframe source domain with the new window attribute. The attribute doesn't do anything on it's own. In order to make sure it is not initialized on live screens, I only add the org ID to `conn.assigns` from the `simulation` endpoint and prevent the frontend from attempting the init on real screens. 

This [PR](https://github.com/mbta/devops/pull/951) must be merged before this one.

- [ ] Tests added?

### DIFF
--- a/assets/src/util/fullstory.tsx
+++ b/assets/src/util/fullstory.tsx
@@ -7,10 +7,13 @@ import { isRealScreen } from "./util";
  * a real production screen AND the URL does not contain the disable_sentry param.
  */
 const initFullstory = () => {
-  const fullstoryOrgId = getDatasetValue("fullstoryOrgId");
+  const screenplayFullstoryOrgId = getDatasetValue("screenplayFullstoryOrgId");
 
-  if (fullstoryOrgId && !isRealScreen()) {
-    FullStory.init({ orgId: fullstoryOrgId, recordCrossDomainIFrames: true });
+  if (screenplayFullstoryOrgId && !isRealScreen()) {
+    FullStory.init({
+      orgId: screenplayFullstoryOrgId,
+      recordCrossDomainIFrames: true,
+    });
   }
 };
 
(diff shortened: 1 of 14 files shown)


assets/src/util/fullstory.tsx
```diff
@@ -0,0 +1,20 @@
+import { getDatasetValue } from "Util/dataset";
+import * as FullStory from "@fullstory/browser";
+import { isRealScreen } from "./util";
+
+/**
+ * Initializes Sentry if the DSN is defined AND this client is running on
+ * a real production screen AND the URL does not contain the disable_sentry param.
+ */
```
hannahpurcell, line 8: Comment needs updating

assets/src/util/fullstory.tsx
```diff
@@ -0,0 +1,20 @@
...
+const initFullstory = () => {
+  const screenplayFullstoryOrgId = getDatasetValue("screenplayFullstoryOrgId");
+
+  if (screenplayFullstoryOrgId && !isRealScreen()) {
```
hannahpurcell, line 12: What is the purpose of the `!isRealScreen()` check? In all situations where `screenplayFullstoryOrgId` is set, won't it not be a real screen?

cmaddox5: Mostly to make me feel more confident that this will not initialize for real screens. It very well may be overkill because we only get to this code from the simulation endpoint.

assets/src/util/fullstory.tsx
cmaddox5: 😬

PR #1756, opened 2023-05-22
Bugfix: Turn off cr overnight

hannahpurcell:
Big oof. I forgot to turn off the CR widget overnight mode.


lib/screens/v2/widget_instance/overnight_cr_departures.ex
```diff
@@ -1,3 +1,7 @@
+# Changing the validity of this candidate to 0, so it will get tossed
```
jzimbel-mbta, line 1: "false" instead of "0", maybe?

jzimbel-mbta (approved):
Good to go!

lib/screens/v2/widget_instance/overnight_cr_departures.ex
hannahpurcell:
```suggestion
# Changing the validity of this candidate to false, so it will get tossed
```

PR #1757, opened 2023-05-22
tweak: E-Ink Subway Status backend

cmaddox5:
**Asana task**: ad-hoc

E-Ink data needs to be a tad different than LCD. Each route is able to show up to two alerts at all times. Any time an `:extended` alert would be shown, we show a header instead. This means we need to only condense alerts for a route if there are 3+ alerts instead of 2+ alerts. 

- [x] Tests added?


jzimbel-mbta (approved):
I think this is good! Updating `serialize_alert` to handle the normal service case is a nice improvement.

PR #1760, opened 2023-05-22
upgrade several packages to fix macos build

panentheos:
**Asana task**: [Upgrade packages to enable building on Mac M1](https://app.asana.com/0/1185117109217413/1204652846824095/f)

This upgrades the Elixir/Erlang runtimes as well as several other packages, in order to build the app on newer Mac hardware.
* Upgrades Elixir/Erlang. While the previous versions were technically compatible, there was a known issue that required modifying the build process to get the runtimes installed. The newer versions don't have this issue.
* Replaces `Application.get_env` with `Application.compile_env` in module bodies to address a deprecation warning. This required upgrading a couple packages as well.
* Removes the explicit `:gettext` from the compilers list, since the new version no longer requires this.
* Removes the `:race_conditions` flag from the dialyzer configuration, since newer OTP versions don't support this.
* Replaces `node-sass` with `sass`, since `node-sass` is deprecated and does not build on recent hardware. `sass` is the recommended compiler now and should be a drop-in replacement.

Note: I tested this locally, and everything seems to work, but I'd appreciate some additional testing by someone who's more familiar with the system, to make sure everything looks ok.


lib/screens_web/router.ex
```diff
@@ -13,7 +13,7 @@ defmodule ScreensWeb.Router do
   end
 
   pipeline :redirect_prod_http do
-    if Application.get_env(:screens, :redirect_http?) do
+    if Application.compile_env(:screens, :redirect_http?) do
```
jzimbel-mbta, line 16: Oh interesting, I didn't know that these `pipeline` blocks were evaluated at compile time. The compiler complained about this?

jzimbel-mbta (approved):
I think this looks good—the main thing we've had to pay extra attention to in past upgrades was the version of Phoenix and its related packages.

Phoenix versions past 1.6 remove a feature that we use (in a _slightly_ nonstandard way 😬) to produce SSML that gets piped through AWS Polly for our on-demand audio equivalence.

So as a sanity check, if you haven't already could you try loading audio for a screen? You can do so by visiting one or more of these routes:
/audio/[id]/readout.mp3?disposition=inline
/audio/[id]/debug
where id is a Solari screen ID: `MUL-104`, `MUL-106`, etc

/v2/audio/[id]/readout.mp3?inline=
/v2/audio/[id]/debug
where id is a bus shelter or pre-fare screen ID: `BUS-108`, `PRE-101`, etc.

✅ Approving now assuming no problems there.

We will also want to check that the JS/CSS changes don't cause problems on any screens—the e-ink ones run pretty outdated, minimal browsers that have given us some trouble in the past. That can happen in dev-green after this is merged, though.

lib/screens_web/router.ex
panentheos: Yes, this was the one spot in our code that it warned about this.

PR #1761, opened 2023-05-24
feat: Updated subway status audio equivalence

jzimbel-mbta:
**Asana task**: [[Subway Status 2.0] Audio implementation](https://app.asana.com/0/1185117109217413/1204531586392668/f)

⚠️ NOTE TO REVIEWER
The last commit significantly complicates the code in order to make the readout more natural sounding. If it seems like too much compared to the branch diff before that commit, let me know and I'll revert.

The "natural" readout generally has about the same playback time as the "non-natural" one—the pauses are just replaced by words.

---

I would also recommend reviewing with the split layout, because I did a from-scratch rewrite of the file. Any code shared with the original is just a coincidence.

- [ ] Tests added?

### DIFF
--- a/lib/screens_web/views/v2/audio/subway_status_view.ex
+++ b/lib/screens_web/views/v2/audio/subway_status_view.ex
@@ -14,12 +14,12 @@ defmodule ScreensWeb.V2.Audio.SubwayStatusView do
   end
 
   defp render_sections(sections_map) do
-    sections_map
-    |> put_in_order()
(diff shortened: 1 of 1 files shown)


cmaddox5 (approved):
This looks and sounds great! I don't really have any feedback on how any of this is done. It very much sounds better. 

Only concern I have is character count. Looks like you were able to make it seem more "human" by tricking Polly with punctuation. This does add more characters on screens that already have high counts. I think that is out-of-scope for this task, but something to watch out for during QA.

PR #1765, opened 2023-05-30
Cm/pre fare alerts candidate generator

cmaddox5:
**Asana task**: [title](url)

Description

- [ ] Tests added?

### DIFF
--- a/lib/screens/v2/candidate_generator/widgets/reconstructed_alert.ex
+++ b/lib/screens/v2/candidate_generator/widgets/reconstructed_alert.ex
@@ -96,7 +96,13 @@ defmodule Screens.V2.CandidateGenerator.Widgets.ReconstructedAlert do
             create_alert_instances(moderate_delays, false, common_parameters)
 
         Enum.any?(downstream_disruptions) ->
-          create_alert_instances(downstream_disruptions, true, common_parameters) ++
+          fullscreen_alerts =
+            find_closest_downstream_alerts(downstream_disruptions, stop_id, stop_sequences)
+
+          flex_zone_alerts = downstream_disruptions -- fullscreen_alerts
+
+          create_alert_instances(fullscreen_alerts, true, common_parameters) ++
+            create_alert_instances(flex_zone_alerts, false, common_parameters) ++
             create_alert_instances(moderate_delays, false, common_parameters)
 
         true ->
@@ -129,6 +135,48 @@ defmodule Screens.V2.CandidateGenerator.Widgets.ReconstructedAlert do
         is_full_screen: is_full_screen
       }
     end)
+  end
+
+  defp find_closest_downstream_alerts(alerts, stop_id, stop_sequences) do
+    # Map each alert with its distance from home.
+    Enum.map(alerts, fn %{informed_entities: ies} = alert ->
+      distance =
+        ies
+        |> Enum.filter(&String.starts_with?(&1.stop, "place-"))
+        |> Enum.map(&get_home_distance_from_ie(stop_id, &1, stop_sequences))
+        |> Enum.min()
+
+      {alert, distance}
+    end)
+    |> Enum.group_by(&elem(&1, 1), &elem(&1, 0))
+    |> Enum.into([])
+    |> Enum.sort_by(&elem(&1, 0))
+    # The first item will be all alerts with the shortest distance.
+    |> List.first()
+    |> elem(1)
+  end
+
+  # Find the closest informed station and calculate its distance from the home_stop_id.
+  # Assumes informed_entity is for a parent station.
+  defp get_home_distance_from_ie(home_stop_id, informed_entity, stop_sequences)
+
+  # GL distance logic to be done later.
+  defp get_home_distance_from_ie(_home_stop_id, %{route: "Green" <> _}, _stop_sequences), do: 99
+
+  # Loop through each stop_sequence to find the shortest distance between home and ie.
+  defp get_home_distance_from_ie(home_stop_id, %{stop: stop_id}, stop_sequences) do
+    stop_sequences
+    |> Enum.filter(&(home_stop_id in &1 and stop_id in &1))
+    |> Enum.map(&calculate_distance(&1, home_stop_id, stop_id))
+    |> Enum.min(fn -> 99 end)
+  end
+
+  # Alerts can be upstream or downstream. If it is an upstream alert, distance will be negative. Make it positive and use that as the distance.
+  defp calculate_distance(stop_sequence, home_stop_id, ie_stop_id) do
+    abs(
+      Enum.find_index(stop_sequence, &(&1 == home_stop_id)) -
+        Enum.find_index(stop_sequence, &(&1 == ie_stop_id))
+    )
   end
 
   defp relevant_alerts(alerts, config, stop_sequences, routes_at_stop, now) do
--- a/lib/screens/v2/candidate_generator/widgets/reconstructed_alert.ex
+++ b/lib/screens/v2/candidate_generator/widgets/reconstructed_alert.ex
@@ -81,16 +81,7 @@ defmodule Screens.V2.CandidateGenerator.Widgets.ReconstructedAlert do
         )
 
       moderate_delays =
-        Enum.filter(
-          relevant_alerts,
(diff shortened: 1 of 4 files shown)


lib/screens/v2/candidate_generator/widgets/reconstructed_alert.ex
```diff
@@ -26,26 +38,142 @@ defmodule Screens.V2.CandidateGenerator.Widgets.ReconstructedAlert do
...
+
+  defp build_distance_map(home_stop_id, stop_sequences) do
+    Enum.reduce(stop_sequences, %{}, fn stop_sequence, distances_by_stop ->
+      stop_sequence
+      # Index each element by its distance from home_stop_id. For example if home_stop_id is at position 2, then indices would start at -2.
+      |> Enum.with_index(-Enum.find_index(stop_sequence, &(&1 == home_stop_id)))
+      # Convert negative distances to positive, and put into a map.
+      |> Map.new(fn {stop, d} -> {stop, abs(d)} end)
+      # Merge with the distances recorded from previous stop sequences.
+      # If a stop already has a distance recorded, we use the smaller of the two. <-- *** Unsure if this is always what we'd want to do! ***
```
hannahpurcell, line 161: I can't think of a situation where this would happen. Two stations should be the same number of stops apart in every stop sequence. Do you have an example? If not, I guess I'd just say go with d1?

lib/screens/v2/candidate_generator/widgets/reconstructed_alert.ex
```diff
@@ -26,26 +38,142 @@ defmodule Screens.V2.CandidateGenerator.Widgets.ReconstructedAlert do
...
+      |> Map.merge(distances_by_stop, fn _stop, d1, d2 -> min(d1, d2) end)
+    end)
+  end
+
+  # Default to 99 if stop_id is not in distance map.
```
hannahpurcell, line 166: Could you add a spec here to help later readers know what shape input is expected?

hannahpurcell, line 166: Also, if we have to fall back to 99, didn't something go wrong? Should an error be logged, or is it nbd

lib/screens/v2/candidate_generator/widgets/reconstructed_alert.ex
```diff
@@ -26,26 +38,142 @@ defmodule Screens.V2.CandidateGenerator.Widgets.ReconstructedAlert do
...
+  defp get_distance(home_stop_distance_map, %{route: "Green" <> _, stop: stop_id})
```
hannahpurcell, line 167: Does this match on just plain "Green"?

hannahpurcell (commented):
Some questions, couple suggestions, should be good to go in a jiffy!

lib/screens/v2/candidate_generator/widgets/reconstructed_alert.ex
```diff
@@ -26,26 +38,142 @@ defmodule Screens.V2.CandidateGenerator.Widgets.ReconstructedAlert do
...
+          relevant_alerts,
+          &(LocalizedAlert.location(%{alert: &1, location_context: location_context}) in [
+              :downstream,
+              :upstream
+            ] and
+              (&1.effect != :delay or &1.severity >= 7))
+        )
+
+      moderate_delays =
+        Enum.filter(relevant_alerts, &(&1.effect == :delay and &1.severity in 5..6))
```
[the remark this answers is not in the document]
cmaddox5: I usually look at `interpret_severity/1` in `alert.ex` to see how we match `severity` to a string. That was set up a long time ago, though, so I hope it's still accurate. I can't find anything in the API docs for it 😬

lib/screens/v2/candidate_generator/widgets/reconstructed_alert.ex
cmaddox5: This could happen with delays. Because delays are just any severe delay, the home stop and the informed stop may not be connected in any stop sequence. In this case, I think it still makes sense to prioritize alerts that are reachable via the home stop.

lib/screens/v2/candidate_generator/widgets/reconstructed_alert.ex
```diff
@@ -26,26 +38,142 @@ defmodule Screens.V2.CandidateGenerator.Widgets.ReconstructedAlert do
...
-          is_terminal_station: is_terminal?(stop_id, location_context.stop_sequences)
-        }
-      end)
+      relevant_alerts = relevant_alerts(alerts, config, location_context, now)
+      is_terminal_station = is_terminal?(stop_id, location_context.stop_sequences)
+
+      immediate_disruptions =
+        Enum.filter(
+          relevant_alerts,
+          &(LocalizedAlert.location(%{alert: &1, location_context: location_context}) in [
```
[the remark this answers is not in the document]
hannahpurcell: Hmmm it might be out of scope of your work anyway! So I think just leave as is

lib/screens/v2/candidate_generator/widgets/reconstructed_alert.ex
hannahpurcell: > delays are just any severe delay

Delays are any severe delay _on the current line_. When we fetch alerts earlier in this file, we fetch alerts that affect this line. So the home stop and informed stop should both be in the stop sequence

lib/screens/v2/candidate_generator/widgets/reconstructed_alert.ex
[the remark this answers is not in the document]
hannahpurcell: It may still be nice to have a few helpers: `is_high_severity`, `is_moderate_severity`, `is_low_severity` maybe? Totally optional

lib/screens/v2/candidate_generator/widgets/reconstructed_alert.ex
cmaddox5: See [here](https://github.com/mbta/screens/pull/1765#discussion_r1219858317) for another reason I think a fallback is good to have.

lib/screens/v2/candidate_generator/widgets/reconstructed_alert.ex
hannahpurcell: What I mean is a trunk alert (Park St to Copley, and the home stop is Govt Center) will go through the section function def and go through the Kenmore branch logic and get assigned a distance of 6, but it should have a distance of 1. (Park St is 1 stop from Govt Center)

lib/screens/v2/candidate_generator/widgets/reconstructed_alert.ex
hannahpurcell: To respond to that PS here: Gov't Center _does_ have a relationship to Kenmore, even if the alert we're looking at is only on the E branch. For Govt Center, the `home_stop_distance_map` includes the route sequence for all the routes that go through it. 

The problem is when we have a home stop on the branch, like Ashmont. Then an alert on the _other_ branch does cause the default of 99 to kick in, since the Ashmont route sequence doesn't have the opposite branch stations. So

lib/screens/v2/candidate_generator/widgets/reconstructed_alert.ex
hannahpurcell: Oh sorry, I didn't understand from your comment that you made a code change. Re-reviewing now

hannahpurcell: Ok yeah, the update did not resolve this issue, but it makes the solution easier to add. The second case should be "If home stop is on the trunk, and the entity is NOT on the trunk, the use the distance to kenmore"

cmaddox5: Exactly the mental nudge I needed, thank you! The spec even says that but I didn't pick up on it when I read it.

lib/screens/v2/candidate_generator/widgets/reconstructed_alert.ex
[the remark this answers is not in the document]
hannahpurcell: Fab, thanks!

hannahpurcell (approved):
🚀

PR #1767, opened 2023-05-30
polish: e-ink Subway Status height/width fix

cmaddox5:
**Notion task**: [Multiple green line alerts are shown when present](https://www.notion.so/mbta-downtown-crossing/Multiple-green-line-alerts-are-shown-when-present-0eff9c28fa794661b9baab0dfa22fd2b?pvs=4)

Fixed height and width values. Also added a `background-color` so that simulations look normal.

Turns out I misunderstood the spec and design. The only difference between e-ink and LCD is e-Ink shows a header anytime an `:extended` alert would be shown on an LCD screen. Collapse logic is the same on both types.

- [ ] Tests added?


lib/screens/v2/widget_instance/subway_status.ex
jzimbel-mbta, line 1: Does your commit 353256dbc1864b77f7d0a9dec0fcfa8bc6ce32b9 essentially `git revert` this file to how it was before the tweaks that were made for e-ink? Have any e-ink-specific changes been retained?

jzimbel-mbta (approved):
LGTM!

lib/screens/v2/widget_instance/subway_status.ex
cmaddox5: No, I totally misunderstood the spec when I made those changes 😞 All logic is shared between LCD and e-Ink now. Only difference is `:extended` alerts are converted to `:contracted` alerts, but that is done on the frontend.

jzimbel-mbta: Ok no worries! I guess the silver lining is that the code is comparatively nice and simple now.

PR #1768, opened 2023-05-30
tweak: Add new paged layout variation for pre-fare left body region

jzimbel-mbta:
**Asana task**: [Update Pre-Fare template](https://app.asana.com/0/1185117109217413/1204649346992049/f)

- [x] Tests added?

### DIFF
--- a/assets/package-lock.json
+++ b/assets/package-lock.json
@@ -5950,7 +5950,7 @@
     "arr-diff": {
       "version": "4.0.0",
       "resolved": "https://registry.npmjs.org/arr-diff/-/arr-diff-4.0.0.tgz",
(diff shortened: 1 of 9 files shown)


assets/package-lock.json
cmaddox5, line 1: Are these changes from running a fresh `npm install`? Just want to make sure the changes are intentional.

cmaddox5 (approved):
Looks good! Just one non-blocking question.

assets/package-lock.json
jzimbel-mbta: Yeah, I needed to run `npm --prefix assets install` to pick up Brett's recent version upgrades. Before that, the page was loading but didn't seem to run any JS, so it remained blank (without any exceptions thrown).

jzimbel-mbta: And I believe you're supposed to check in changes to package-lock.json, so I did that 🤷

PR #1769, opened 2023-05-31
polish: Station Closure text

cmaddox5:
**Notion task**: [Station closure / one line / three stops](https://www.notion.so/mbta-downtown-crossing/Station-closure-one-line-three-stops-321a98f4d3944ed7ab7b7179ce89555f?pvs=4)

The server consolidates stations into a count when there are more than 3 stations in a single closure. There are some combinations of station names that are too long to show even though there are 3 or less stations closed. I tweaked the `PerEffect` logic for `Bypassing` so it consolidates stations into a count on its own.

To make this logic consistent between LCD and E-Ink, I also refactored the hook logic out into the common file.

- [ ] Tests added?


jzimbel-mbta (approved):
Looks good to me!

PR #1771, opened 2023-06-01
feat: WidgetTreeErrorBoundary component

jzimbel-mbta:
**Asana task**: [Deploy 🐞: Client can crash when screen data schema changes during a prod deploy](https://app.asana.com/0/1185117109217413/1204579243927091/f)

This adds an error boundary that protects the data request cycle from being halted by render errors. By doing so, we can keep requesting data and trying to render it.

This is mostly useful for deploys where the data sent by the server changes—for a short period, the old client receives new data until the deploy completes and the client gets told to reload itself. By having the client continue to request new data through that "rough patch", we ensure that it will eventually get the `{force_reload: true}` response and reload, picking up the new client code.

The error boundary does a couple extra things to avoid especially undesirable behavior:
1. Every time we fetch new data, the error boundary gives the "normal render" another go even if it previously caught an error. This lets us avoid having one error (e.g. in a paged widget that only appears every few renders) put the screen into its no-data state indefinitely.
2. If it continually catches the same error multiple times over several renders, it only logs that error once every 10 minutes, to avoid overloading our vendor logging quotas.

- [ ] Tests added?


cmaddox5 (approved):
Yeesh, this is a lot of work just to keep screens from crashing. Looks great though! Still impressed at how quickly you diagnosed the issue.

PR #1776, opened 2023-06-06
fix: Prevent Error Boundary on packaged DUPs

cmaddox5:
**Asana task**: ad-hoc

While working on another task, I noticed that the new error boundary does not play nicely with packaged DUPs. Because Outfront is constantly reloading these pages, I think it's ok to just remove it from DUPs. Open to alternatives though.

- [ ] Tests added?

### DIFF
--- a/assets/src/components/v2/screen_container.tsx
+++ b/assets/src/components/v2/screen_container.tsx
@@ -4,6 +4,7 @@ import React, {
   ComponentType,
   useState,
   useEffect,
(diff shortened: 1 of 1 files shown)


jzimbel-mbta (approved):
Just checking—the error only happens when the error boundary component is actually mounted, right? We don't break anything by simply importing that module (and thus its own import of the React Router stuff) in the DUP app?

PR #1777, opened 2023-06-07
tweak: v2 DUP screen IDs

cmaddox5:
**Asana task**: [Modify DUP app to use unique screen IDs](https://app.asana.com/0/1185117109217413/1204742012101948/f)

Decided to keep the old hook just so I didn't have to worry about v1 for now. None of those are live anyway. When we tear down the v1 app, we can remove the hook then.

I also created a script for replacing the v2 DUP screens in the config. Would that be useful to check in? I don't foresee needing it again after the first time we use it.

- [ ] Tests added?

### DIFF
--- a/assets/src/apps/v2/dup.tsx
+++ b/assets/src/apps/v2/dup.tsx
@@ -37,7 +37,7 @@ import {
 import PageLoadNoData from "Components/v2/dup/page_load_no_data";
 import NoData from "Components/v2/dup/no_data";
 import OvernightDepartures from "Components/v2/dup/overnight_departures";
-import useOutfrontStation from "Hooks/use_outfront_station";
+import useOutfrontPlayerName from "Hooks/use_outfront_player_name";
 
 const TYPE_TO_COMPONENT = {
   screen_normal: NormalScreen,
@@ -122,10 +122,10 @@ const responseMapper: ResponseMapper = (apiResponse) => {
 };
 
 const App = (): JSX.Element => {
-  const station = useOutfrontStation();
+  const playerName = useOutfrontPlayerName();
 
-  if (station !== null) {
-    const id = `DUP-${station.replace(/\s/g, "")}-V2`;
+  if (playerName !== null) {
+    const id = `DUP-${playerName.trim()}`;
     return (
       <MappingContext.Provider value={TYPE_TO_COMPONENT}>
         <ResponseMapperContext.Provider value={responseMapper}>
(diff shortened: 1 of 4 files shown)


assets/src/components/v2/dup/normal_header.tsx
```diff
@@ -30,6 +33,7 @@ const NormalHeader = ({
       showTo={false}
       classModifiers={color}
       accentPattern={accentPattern}
+      playerName={playerName}
```
jzimbel-mbta, line 36: Would it make sense to drop the player name into the existing `version` string, here in the DUP normal header, rather than passing it down to the more generic header component to be added to the string there?

So like
```ts
let version = DUP_VERSION;
if (code) {
  version = `${version}; Maintenance code: ${code}`;
}
if (playerName) {
  version = `${version}-${playerName}`;
}

return <DefaultNormalHeader ... version={version} ... />;
```

jzimbel-mbta (approved):
Looks good, one minor Q and one suggestion. Approved assuming you respond as you see fit!

assets/src/components/v2/dup/normal_header.tsx
jzimbel-mbta: Now that I look more closely, it might also make sense to add the player name before the maintenance code. (Though we don't actually use those at all in the new DUP app, so this would be more about future-proofing in case we start)

assets/src/apps/v2/dup.tsx
```diff
@@ -122,10 +122,10 @@ const responseMapper: ResponseMapper = (apiResponse) => {
 };
 
 const App = (): JSX.Element => {
-  const station = useOutfrontStation();
+  const playerName = useOutfrontPlayerName();
 
-  if (station !== null) {
-    const id = `DUP-${station.replace(/\s/g, "")}-V2`;
+  if (playerName !== null) {
+    const id = `DUP-${playerName.trim()}`;
```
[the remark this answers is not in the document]
cmaddox5: Based on the spreadsheet they gave us, the names will always be something like `COP-DS-007`. So `{StationID}-DS-XXX`. We don't have to worry about multi word station names like we used to.

assets/src/components/v2/dup/normal_header.tsx
cmaddox5: I think that makes sense. No reason for `playerName` to exist anywhere in the `DefaultNormalHeader` component.

PR #1781, opened 2023-06-13
LCD subway status text-fitting fixes

jzimbel-mbta:
**QA task**: [Station closure / extended / three stops](https://www.notion.so/mbta-downtown-crossing/Station-closure-extended-three-stops-4c21e366f22a45ed99b52c867d392cda?pvs=4)

This fixes a few remaining issues with automatic text-fitting on the new subway status widget:
- (After fixing a subtle bug with it) Use the `isDone` value returned by `useTextResizer` / `useSubwayStatusTextResizer` to set overflow-hiding styles appropriately, instead of trying to infer when the resizing is done.
- Add text-fitting to extended alert location text, because we need it in certain very rare scenarios.
- (Not flagged in QA, but a known bug) In contracted alerts, prevent the status+location from wrapping below the pill icon in cases where the text is still too long after all resizing attempts.

[Delays / extended / whole route](https://www.notion.so/mbta-downtown-crossing/Delays-extended-whole-route-6bb469f525174f678fdc83d5d53cab55?pvs=4) is a somewhat related item flagged during QA, but this behavior was actually following the designs. Pending a response from Paul H, I might add a tweak for that to this PR.

- [ ] Tests added?

### DIFF
--- a/assets/css/v2/lcd_common_styles/subway_status.scss
+++ b/assets/css/v2/lcd_common_styles/subway_status.scss
@@ -101,6 +101,14 @@
     }
   }
 
+  // Prevents text-container from wrapping to a second line below the pill
+  // if it's too long.
+  .subway-status_alert-sizer--hide-overflow {
+    overflow: hidden;
+    white-space: nowrap;
+    text-overflow: clip;
+  }
+
   .subway-status_alert_route-pill-container {
     display: inline-block;
     // If the row has no pill, we still keep this element as a pill-sized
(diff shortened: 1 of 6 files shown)


cmaddox5 (approved):
Logic looks good to me! One specific alert I wanted to check out is this one:

Screenshot 2023-06-13 at 10 10 33 AM

Definitely fixed the wrapping problem, but the text is very cramped. That may be intentional considering there are no other resizing steps to do here. Just wanted to flag it.

PR #1786, opened 2023-06-15
RFC: Logging screen responses to S3

hannahpurcell:
**Asana task**: [[timebox 1d] Evaluate technical options & make go/no-go decision](https://app.asana.com/0/1185117109217413/1204496091984220)

This is a condensed RFC to outline our approach to recording screen responses in S3 and break down the possible budgetary burden. This was generated after loose brainstorming with #ctd-engineering [here](https://mbta.slack.com/archives/GP5QJM98A/p1685559879489779), and then a detailed discussion with @paulswartz. 

Primary goal of making this a shortened RFC is just determining whether the budgetary burden is acceptable to @ryan-mahoney and the CTD Infra team. (I'll have Ryan look first before bugging the Infra team.) If it's not, then we'll need to iterate on alternatives.

### DIFF
--- a/docs/tech_specs/0004_store_screen_responses.md
+++ b/docs/tech_specs/0004_store_screen_responses.md
@@ -1,0 +1,73 @@
+- Feature Name: `store_screen_responses`
+- Start Date: 2023-06-12
+- RFC PR: [mbta/technology-docs#0000](https://github.com/mbta/technology-docs/pull/0000)
+- Asana task: [Log and store all screen data responses](https://app.asana.com/0/1185117109217413/1204451311903806)
+- Status: Proposed
+
+# Summary / Motivation
+[summary]: #summary
+
+Problem statement:  The Screens team wants to be able to see visual renderings and data responses of past screen content in order to investigate rider reports of incorrect messages or do our own auditing of screen content.
+
+Suggested approach: we want to store the json screen data responses in S3, as frequently as the screen content updates (every 15-30s). With the screen responses stored, we can run the json through frontend rendering and get (at least an MVP view) of past screens. We only need to store 90 days of data, at most.
+
+Note: At the moment, this will be a tool used mainly by the dev team, so we're not going to be too sophisticated with the retrieval and rendering process of the json right now. We did ideate about making a beautiful "screen time-travel view", but there are a few challenges with it that are explained in the #rationale-and-alternatives section. So we're going with an MVP user story here: an engineer needs to be able to see the data responses for historic screen content and be able to locally render the screen at that time. This method of local rendering is already doable, so data storage will be the focus of this work.
+
+# Guide-level explanation
+[guide-level-explanation]: #guide-level-explanation
+
+For e-inks, this is every 30s. For other lcd screens, it could be as frequent as 15 or 20s.
+
+Explain the proposal as if it was already implemented and you were teaching it to a new developer
+that just joined the team. That generally means:
+
+- Introducing new named concepts.
+- Explaining the feature largely in terms of examples.
+- Explaining how programmers should *think* about the feature, and how it should impact the way they
+  work on this project. It should explain the impact as concretely as possible.
+- If applicable, provide sample error messages, deprecation warnings, or migration guidance.
+- If applicable, describe the differences between teaching this to senior developers and to junior
+  developers.
+
+We will store screen data responses in S3. This data response is a json blob that gets sent to the frontend renderer. Screen data requests happen at regular intervals for each screen type (and screens of the same type have intervals offset from one another). E-inks refresh every 30s, while lcd screens refresh every 15 or 20s. During each refresh, we already run a function that caches the alerts that are being displayed on a particular screen. So, with that logic we will also now push the screen data response to S3. 
+
+There are a few attributes we want to make sure are included / added to the json blob to make future debugging more useful. Namely, we want to include alert ids and possibly trip ids for departures (since they are not necessarily needed for rendering the screen but would be helpful for later debugging.)
+
+These responses will be stored by a screen ID prefix, and then within the screen ID by datetime. 
+
+**What kind of load does this put on our S3 resources?**
+
+A few important variables: we foresee wanting to retain no more than 90 days of history. And from trying out a couple example files, the blobs should be around 5-5.5KB in size. 
+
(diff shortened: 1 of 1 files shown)


docs/tech_specs/0004_store_screen_responses.md
```diff
@@ -0,0 +1,73 @@
+- Feature Name: `store_screen_responses`
+- Start Date: 2023-06-12
+- RFC PR: [mbta/technology-docs#0000](https://github.com/mbta/technology-docs/pull/0000)
```
hannahpurcell, line 3:
```suggestion
- RFC PR: [mbta/screens#0000](https://github.com/mbta/screens/pull/1786)
```

ryan-mahoney (approved):
Very clear + well researched. Appreciate that you spoke to Paul and LAMP already, as that would have been my recommendation.

I think this makes sense. I also suspect the cost will be less than the 1% described.

I apologize if I missed this, but even though you need longer retention time than what we typically do for Splunk; you might consider applying an expiration to these objects, for example, 2 years as that will keep the cost at a steady level over the long term.

It might be worth noting in this document that there is no PII data as generally when we are thinking about data storage, it is beneficial to be explicit about this point even though you have characterized the type of data, and it is evident that it would not contain PII.

We can chat about this, but have you looked into Athena already? This seems like it would support your limited querying needs while being S3 compatible.

Great work :+1:

ryan-mahoney:
Note re: implementation, I could see getting some guidance from LAMP re: the naming of files/bucket as in the long term, LAMP may be somehow responsible for this archive, so adhering to standards they have (if they have them for something like this) could be helpful.

docs/tech_specs/0004_store_screen_responses.md
```diff
@@ -0,0 +1,73 @@
...
+- RFC PR: [mbta/screens#1786](https://github.com/mbta/screens/pull/1786)
+- Asana task: [Log and store all screen data responses](https://app.asana.com/0/1185117109217413/1204451311903806)
+- Status: Proposed
+
+# Summary / Motivation
+[summary]: #summary
+
+Problem statement:  The Screens team wants to be able to see visual renderings and data responses of past screen content in order to investigate rider reports of incorrect messages or do our own auditing of screen content.
+
+Suggested approach: we want to store the json screen data responses in S3, as frequently as the screen content updates (every 15-30s). With the screen responses stored, we can run the json through frontend rendering and get (at least an MVP view) of past screens. We only need to store 90 days of data, at most.
```
hannahpurcell, line 12:
```suggestion
Suggested approach: we want to store the json screen data responses in S3, as frequently as the screen content updates (every 15-30s). With the screen responses stored, we can run the json through frontend rendering and get (at least an MVP view) of past screens. We only need to store 90 days of data, at most. This json will only include content we display on screens and will never contain PII.
```

docs/tech_specs/0004_store_screen_responses.md
```diff
@@ -0,0 +1,73 @@
...
+- Logging to Sentry using the json source_type
+
+We stuck with the S3 solution rather than introduce a DB because:
+- It should cover our requirements
+- We already use S3 in the project stack
+- The cost shouldn't be a problem (after discussing with some engineering leads)
+
+As for using Parquet (recommended by the LAMP team) that would just help make the data more queryable, which we don't really need for our use case.
+
+As for logging to Sentry, this was kind of a late-breaking idea, and I didn't immediately find anyone in CTD who has programmatically queried Sentry in this way before, so I just stuck with the S3 option as prime choice. (If the PR reviewers have other insight, this could be an interesting option.)
```
cmaddox5, line 69: Biggest issue here is our department quota. We would potentially cause other teams to drop errors because we would generate so many. Either that or we would need to drop our own logs, which would take away our opportunity to see responses towards the end of each month.

docs/tech_specs/0004_store_screen_responses.md
```diff
@@ -0,0 +1,73 @@
...
+
+We also considered these alternatives to streamline the frontend rendering before deciding it was out of scope for this work:
+- Store jpgs of the screen content
+- Store html of the screens (tricky, because some screens have frontend paging that will not be represented in the html.)
```
cmaddox5, line 73: The frontend paging issue may actually just be the same situation as backend paging, but instead of pages being omitted in the response, they are omitted from the DOM. We would be able to see all pages in the server response, but would get a more accurate snapshot if we were to look at the HTML. (Not really advocating for this, just something I thought about.)

cmaddox5 (approved):
I think this looks like a great plan. 

An idea that I vaguely remember talking about previously is DUPs. With the switch to unique screen IDs, I see that we have 200 DUPs that share 12 different configs. That could be a pretty sizeable reduction in number of responses we need to save (at least until screens in the same station start showing different info, but that's not planned at the moment). We also have some stations with two Pre-Fare screens that show the same info most of the time. That may be a riskier and more unnecessary consolidation, but another screen type to think about.

docs/tech_specs/0004_store_screen_responses.md
hannahpurcell: OH I misspoke! I meant Splunk here!

hannahpurcell, line 69:
```suggestion
- Logging to Splunk using the json source_type

We stuck with the S3 solution rather than introduce a DB because:
- It should cover our requirements
- We already use S3 in the project stack
- The cost shouldn't be a problem (after discussing with some engineering leads)

As for using Parquet (recommended by the LAMP team) that would just help make the data more queryable, which we don't really need for our use case.

As for logging to Splunk, this was kind of a late-breaking idea, and I didn't immediately find anyone in CTD who has programmatically queried Splunk in this way before, so I just stuck with the S3 option as prime choice. (If the PR reviewers have other insight, this could be an interesting option.)
```

docs/tech_specs/0004_store_screen_responses.md
```diff
@@ -0,0 +1,73 @@
...
+- We already use S3 in the project stack
+- The cost shouldn't be a problem (after discussing with some engineering leads)
+
+As for using Parquet (recommended by the LAMP team) that would just help make the data more queryable, which we don't really need for our use case.
+
+As for logging to Splunk, this was kind of a late-breaking idea, and I didn't immediately find anyone in CTD who has programmatically queried Splunk in this way before, so I just stuck with the S3 option as prime choice. (If the PR reviewers have other insight, this could be an interesting option.)
+
+We also considered these alternatives to streamline the frontend rendering before deciding it was out of scope for this work:
+- Store jpgs of the screen content
+- Store html of the screens (tricky, because some screens have frontend paging that will not be represented in the html.)
```
hannahpurcell, line 73:
```suggestion
- Store html of the screens
```

hannahpurcell:
> we have 200 DUPs that share 12 different configs

Christian and I took a look at this because that number was off. The actual number is closer to 40 ids for 12 configs, which is a bit of a reduction, but not as enormous. So if we choose to reduce response logging down the road, we'd save a fraction of space / processing

PR #1788, opened 2023-06-16
polish: Alert count logic

cmaddox5:
**Notion tasks**: [Multiple rows show up expanded](https://www.notion.so/mbta-downtown-crossing/Multiple-rows-show-up-expanded-e54526b1088b4074bb06d12b10f8be12?pvs=4)
[Incorrect station name appears](https://www.notion.so/mbta-downtown-crossing/Incorrect-station-name-appears-20c2bc0377ff495fb16b089b768d0139?pvs=4)

Total alert count logic was off when there was an alert that affected multiple routes. The function now looks at total affected routes and total unique alert IDs and returns the max of the two counts. This allows `:extended` vs `:contracted` logic to pick the correct type.

- [x] Tests added?

### DIFF
--- a/lib/screens/v2/widget_instance/subway_status.ex
+++ b/lib/screens/v2/widget_instance/subway_status.ex
@@ -135,36 +135,42 @@ defmodule Screens.V2.WidgetInstance.SubwayStatus do
   end
 
   def serialize_one_row_for_all_routes(grouped_alerts) do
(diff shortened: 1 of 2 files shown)


hannahpurcell (approved):
Ok nice! I feel like this one must have been tricky to deduce.

PR #1792, opened 2023-06-20
feat: Pre-Fare alert serializer tests

cmaddox5:
New/updated tests needed for https://github.com/mbta/screens/pull/1785.

- [ ] Tests added?

### DIFF
--- a/test/screens/v2/widget_instance/reconstructed_alert_test.exs
+++ b/test/screens/v2/widget_instance/reconstructed_alert_test.exs
@@ -18,7 +18,7 @@ defmodule Screens.V2.WidgetInstance.ReconstructedAlertTest do
   defp setup_base(_context) do
     %{
       widget: %ReconstructedAlert{
-        alert: %Alert{id: "123"},
+        alert: %Alert{id: "123", updated_at: ~U[2023-06-09T09:00:00Z]},
         screen: %Screen{app_params: nil, vendor: nil, device_id: nil, name: nil, app_id: nil},
         location_context: %LocationContext{
           home_stop: nil,
@@ -107,6 +107,10 @@ defmodule Screens.V2.WidgetInstance.ReconstructedAlertTest do
     %{widget | is_terminal_station: is_terminal_station}
   end
 
+  defp put_is_full_screen(widget, is_full_screen) do
+    %{widget | is_full_screen: is_full_screen}
+  end
+
   defp ie(opts) do
     %{
       stop: opts[:stop],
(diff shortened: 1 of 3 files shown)


test/screens/v2/widget_instance/reconstructed_alert_test.exs
```diff
@@ -228,13 +330,20 @@ defmodule Screens.V2.WidgetInstance.ReconstructedAlertTest do
     setup @valid_alert_setup_group
 
     test "returns takeover for a closure alert at this station", %{widget: widget} do
+      widget = put_is_full_screen(widget, true)
       assert [1] == WidgetInstance.priority(widget)
       assert [:full_body] == WidgetInstance.slot_names(widget)
       assert :reconstructed_takeover == WidgetInstance.widget_type(widget)
     end
```
hannahpurcell, line 337: Ahhh hm, this test made me realize that naming is getting a little bit hairy. This test is about takeover alerts, and `is_full_screen` is set to true. When `is_full_screen` is true, that may mean it's a takeover or "the other kind", so then we run `takeover_alert` in the widgetInstance file to decide whether it's a takeover or "the other kind". One problem is that "the other kind" is sorta called "fullscreen" alert, because all of those serializer functions are called `serialize_fullscreen_alert`, which will never be run for a takeover, even though it's a "fullscreen" alert, too. The same words are getting repeated for different situations that I think will make this code hard to use in the future.

Couple related thoughts: is `takeover_alert?` still a good function to be in the WidgetInstance, or should it be in the candidateGenerator?

I like the name of `serialize_inside_flex_alert` because it's so clear what that function is for. Should the other flex serializers (outside and boundary) also be renamed to have "flex" in it?

test/screens/v2/widget_instance/reconstructed_alert_test.exs
```diff
@@ -378,23 +485,19 @@ defmodule Screens.V2.WidgetInstance.ReconstructedAlertTest do
...
-        routes: [
-          %{color: :orange, text: "OL", type: :text},
-          %{color: :red, text: "RL", type: :text}
-        ],
+        issue: "No trains",
+        location: "Shuttle buses replace Orange Line trains at Malden Center",
+        cause: nil,
         effect: :shuttle,
-        urgent: true,
-        region: :inside,
```
hannahpurcell, line 396: When merge conflicts are resolved between this and the backend branch, I think `region` will need to be added back in, because it's used by the frontend.

hannahpurcell: Similarly, `endpoints` was added, and so was `unaffected_routes` in certain cases

test/screens/v2/widget_instance/reconstructed_alert_test.exs
```diff
@@ -1205,164 +1463,764 @@ defmodule Screens.V2.WidgetInstance.ReconstructedAlertTest do
...
+        effect: :suspension,
+        urgent: false,
+        region: :outside,
+        remedy: "Seek alternate route"
+      }
+
+      assert expected == ReconstructedAlert.serialize(%{alert_widget | is_full_screen: false})
+    end
+
+    test "handles fullscreen OL downstream suspension" do
```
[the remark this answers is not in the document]
hannahpurcell: And then same with the GL tests below

hannahpurcell (approved):
Ok yeah! Only 1 consolidation question.

test/screens/v2/widget_instance/reconstructed_alert_test.exs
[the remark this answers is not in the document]
cmaddox5: Good idea. Makes it easier to follow considering setup is more complicated.

PR #1796, opened 2023-06-22
Initial type definitions for disruption diagram

jzimbel-mbta:
**Asana task**: [Write the data structure for the serialized line map](https://app.asana.com/0/1185117109217413/1204736316281207/f)

The structure focuses on describing the overall type and layout of the diagram, and leans heavily on the client to infer the specific presentation. For example, it doesn't say what symbol to use for each slot, it just gives info about the type and extent of the disruption + where the current stop is.

- [ ] Tests added?


docs/disruption_diagram_alternate_structure_examples.md
```diff
@@ -0,0 +1,159 @@
...
+```ts
+{
+  effect: "station_closure",
+  line: "green",
+  current_station_slot_index: 11,
+  closed_station_slot_indices: [2, 4],
+  slots: [
+    {type: "arrow", label_id: "place-clmnl"},
+    {
+      label: {full: "Saint Paul Street", abbrev: "St. Paul St"},
```
cmaddox5, line 117: Will these abbreviations be the same that are in `stop.ex`? Or will this widget need its own set of abbreviations?

cmaddox5: Also, what is the logic for using full vs abbreviations? I see a mix of both in some examples, but no defined logic for which to use when.

docs/disruption_diagram_alternate_structure_examples.md
```diff
@@ -0,0 +1,159 @@
...
+![image](assets/disruption_diagrams/2.png)
+
+(Correction: The first "Downtown Crossing" label should be "Kendall/MIT")
+
+```ts
+{
+  effect: "suspension",
+  line: "red",
+  current_station_slot_index: 2,
+  effect_region_slot_index_range: [3, 4],
```
cmaddox5, line 75: Would this actually be `[3, 5]`? Or is the range in this case just for the icons and the gray lines would be something the client would figure out on its own?

docs/disruption_diagram_alternate_structure_examples.md
```diff
@@ -0,0 +1,159 @@
...
+![image](assets/disruption_diagrams/1.png)
+
+```ts
+{
+  effect: "shuttle",
+  line: "green",
+  current_station_slot_index: 11,
+  effect_region_slot_index_range: [0, 8],
+  slots: [
+    {type: "terminal", label_id: "place-lake"},
```
cmaddox5, line 17: What is the advantage of passing the `stop_id` to the client for terminals? Could we just respond with the station's name instead?

cmaddox5:
I'm a big fan of the alternate approach. The first approach just didn't click for me. I think that no matter how we slice it, the frontend is gonna be a beast, so any chance we have to simplify the backend is great.

docs/disruption_diagram_alternate_structure_examples.md
jzimbel-mbta: Yes, the range is for the slots that should have "disruption" icons—either the shuttle black circles or the suspension/station closure stop signs.

Using indices of the slots rather than the edges ensures consistent behavior even when the range starts at the left end of the line or ends at the right end.

docs/disruption_diagram_alternate_structure_examples.md
jzimbel-mbta: Yeah, any value that's unique-per-station should work. Stop/station IDs are already unique so I figured we could use those.

I also want to check with design/product whether we could come up with some consistent algorithm for the end label presentation, so that we don't need to hardcode each one. If we can, then using the station's name would let us keep sending the same data from the server and just use it differently on the client.

docs/disruption_diagram_alternate_structure_examples.md
jzimbel-mbta: Yeah, that was my thinking, though I haven't confirmed with Betsy.

> I see a mix of both in some examples, but no defined logic for which to use when.
If you mean the screenshots from my markdown files with example data, I got those from the figma. It does seem to use abbreviation inconsistently, but I assumed that was just because Betsy wasn't focused on getting the labels perfectly right in every example.

According to the figma, the rule is to abbreviate middle-slot labels whenever `label_height > 8 + max(left_end_label_height, right_end_label_height)`

hannahpurcell: ++ the abbreviation system is definitely wonky atm. I don't think it will affect the data structure you've chosen, but will add complexity to the serialization / frontend work

jzimbel-mbta:
Thanks for the feedback! I'll clean up the branch, removing the less popular proposed structure, and request a final review.

lib/screens/v2/disruption_diagram/model.ex
```diff
@@ -0,0 +1,71 @@
...
+  @type serialized_response :: continuous_disruption_diagram() | discrete_disruption_diagram()
+
+  @type continuous_disruption_diagram :: %{
+          effect: :shuttle | :suspension,
+          # A 2-element list, giving indices of the effect region's *first and last disrupted stops*.
+          # For example in this scenario:
+          #     0     1     2     3     4     5     6     7     8
+          #    <= === O ========= O - - X - - X - - X - - O === =>
+          #                             |---range---|
+          # The range is [4, 6].
```
hannahpurcell, line 20: So sorry for my delayed review! The way this alert would be crafted by OIO's, and the way the alert text will read, is "Shuttling between Station 3 and Station 7". And the range or "emphasis" label on maps will also cover the same [area](https://www.figma.com/file/l4AZl8UXyKLAFI8OtultQk/Pre-Fare-%E2%80%94-Alerts-V2?type=design&node-id=565-11629&mode=dev). But I do kinda see the benefit of doing it this way, so that determining the icons for the stations are easier? What do you think?

hannahpurcell (approved):
Just 1 question, but otherwise, a long overdue approval

lib/screens/v2/disruption_diagram/model.ex
jzimbel-mbta: Yeah, it's a little weird but the value of the range is purely based on visuals, not the semantics of the different types of alerts. It's simply "which are the first and last slots that should get disruption-style symbols?"

That lets the client focus on presentation rather than having to know more details about how the different alerts should be interpreted.

Also—if the shuttle replacement/suspension starts or ends at a terminus, then there would be no way to represent that in the data if we had the range start/end at the first/last "non-disrupted" stops.

hannahpurcell: Aha! Sweet, thanks for the extra explanation. This is perfect

lib/screens/v2/disruption_diagram/model.ex
```diff
@@ -0,0 +1,71 @@
...
+          effect_region_slot_index_range: list(non_neg_integer()),
+          line: line_color(),
+          current_station_slot_index: non_neg_integer() | nil,
```
cmaddox5, line 23: A question I just thought about that I want to squeeze in. Why can this be `nil`?

lib/screens/v2/disruption_diagram/model.ex
jzimbel-mbta: Both of you flagged this and you were totally right! The range does need to include the "boundary" stops. And, my reason for taking this approach originally:

> if the shuttle replacement/suspension starts or ends at a terminal, then there would be no way to represent that in the data if we had the range start/end at the first/last "non-disrupted" stops. (We couldn't get that terminal stop to have a disruption-style symbol.)

can be worked around with a pretty straightforward check on the client: "If the start of the range is 0 use a 'disruption icon' for the starting border stop. If the end of the range is (length(diagram) - 1) use a 'disruption icon' for the ending border stop."

I'll update the comment now.

PR #1799, opened 2023-06-26
polish: branch vs trunk logic

cmaddox5:
**Notion task**: [Fix GL branch vs trunk alert](https://www.notion.so/mbta-downtown-crossing/Fix-GL-branch-vs-trunk-alert-d5a287156e734f3d960580ade8591a38?pvs=4)

Our previous branch vs trunk logic was causing some confusion when a branch alert included a stop on the trunk (i.e. GL-C shuttle starting at Kenmore). In summary, the new logic is:

A station closure is a trunk alert if any closed station is on the trunk. Shuttles and suspensions are trunk if the list of affected stops fits in more than one branch. A delay with no stop range is a trunk alert if it affects all branches. Otherwise, it uses the same logic as shuttles and suspensions.

- [ ] Tests added?

### DIFF
--- a/lib/screens/v2/widget_instance/subway_status.ex
+++ b/lib/screens/v2/widget_instance/subway_status.ex
@@ -485,8 +485,8 @@ defmodule Screens.V2.WidgetInstance.SubwayStatus do
       end)
       |> Enum.uniq()
 
-    Enum.count(gl_stop_sequences, fn stop_sequence ->
-      alert_stops != [] and Enum.all?(alert_stops, &(&1 in stop_sequence))
+    alert_stops != [] and Enum.count(gl_stop_sequences, fn stop_sequence ->
+      Enum.all?(alert_stops, &(&1 in stop_sequence))
     end) > 1
   end
 
--- a/lib/screens/v2/widget_instance/subway_status.ex
+++ b/lib/screens/v2/widget_instance/subway_status.ex
@@ -371,11 +371,10 @@ defmodule Screens.V2.WidgetInstance.SubwayStatus do
 
   defp serialize_alert(
          %Alert{effect: :station_closure, informed_entities: informed_entities},
-         route_id
+         _route_id
        ) do
     # Get closed station names from informed entities
-    stop_id_to_name = Stop.stop_id_to_name(route_id)
-    stop_names = get_stop_names_from_informed_entities(informed_entities, stop_id_to_name)
+    stop_names = get_stop_names_from_informed_entities(informed_entities)
     {status, location} = format_station_closure(stop_names)
 
     %{status: status, location: location, station_count: length(stop_names)}
@@ -417,9 +416,7 @@ defmodule Screens.V2.WidgetInstance.SubwayStatus do
         },
         route_ids
       ) do
-    stop_id_to_name = route_ids |> Enum.flat_map(&Stop.stop_id_to_name/1) |> Enum.into(%{})
-    stop_names = get_stop_names_from_informed_entities(informed_entities, stop_id_to_name)
-
+    stop_names = get_stop_names_from_informed_entities(informed_entities)
     {status, location} = format_station_closure(stop_names)
 
     %{
@@ -447,17 +444,7 @@ defmodule Screens.V2.WidgetInstance.SubwayStatus do
     )
   end
 
-  defp alert_affects_gl_trunk_or_whole_line?(%Alert{informed_entities: informed_entities}) do
-    gl_trunk_stops = Stop.gl_trunk_stops()
-
-    alert_trunk_stops =
-      informed_entities
-      |> Enum.map(fn e -> Map.get(e, :stop) end)
-      |> Enum.reject(&is_nil/1)
-      |> Enum.uniq()
-      |> Enum.into(MapSet.new())
-      |> MapSet.intersection(gl_trunk_stops)
-
+  defp alert_affects_whole_green_line?(%Alert{informed_entities: informed_entities}) do
     alert_whole_line_stops =
       informed_entities
       |> Enum.map(fn e -> Map.get(e, :route) end)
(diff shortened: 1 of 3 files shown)


lib/screens/v2/widget_instance/subway_status.ex
```diff
@@ -468,8 +455,44 @@ defmodule Screens.V2.WidgetInstance.SubwayStatus do
...
+  #
+  defp alert_affects_gl_trunk?(%Alert{informed_entities: informed_entities}, gl_stop_sequences) do
+    alert_stops =
+      informed_entities
+      |> Enum.filter(fn
+        %{stop: nil} -> false
+        ie -> String.starts_with?(ie.stop, "place-") and String.starts_with?(ie.route, "Green-")
+      end)
+      |> Enum.uniq()
+
```
jzimbel-mbta, line 487: Is `alert_stops` intended to be a list of informed entities? It looks like it should be a list of stop IDs since its elements get checked against `gl_stop_sequences`.

lib/screens/v2/widget_instance/subway_status.ex
```diff
@@ -468,8 +455,44 @@ defmodule Screens.V2.WidgetInstance.SubwayStatus do
...
+    Enum.count(gl_stop_sequences, fn stop_sequence ->
+      alert_stops != [] and Enum.all?(alert_stops, &(&1 in stop_sequence))
+    end) > 1
```
jzimbel-mbta, line 490: The result of `alert_stops != []` will be the same every time this anonymous function is called, would it make sense to do the check outside of it?
```suggestion
    alert_stops != [] and Enum.count(gl_stop_sequences, fn stop_sequence ->
      Enum.all?(alert_stops, &(&1 in stop_sequence))
    end) > 1
```

lib/screens/v2/widget_instance/subway_status.ex
```diff
@@ -483,8 +506,13 @@ defmodule Screens.V2.WidgetInstance.SubwayStatus do
     if gl_alert_count == 0 do
       serialize_single_alert_row_for_route(grouped_alerts, "Green", total_alert_count)
     else
+      gl_stop_sequences = Stop.get_gl_stop_sequences()
```
jzimbel-mbta, line 509: Since `gl_stop_sequences` is only used for membership checks (as far as I can tell), you could speed things up by converting this list of lists to a list of mapsets before passing it to `alert_affects_gl_trunk_or_whole_line?/2`. The time complexity of each `in` call within `alert_affects_gl_trunk?/2` would go from `O(n)` to `O(log n)`.
```suggestion
      gl_stop_sets = Enum.map(Stop.get_gl_stop_sequences(), &MapSet.new/1)
```

jzimbel-mbta (commented):
There's one issue that I think is a bug, but my other 2 comments are about performance.

lib/screens/v2/widget_instance/subway_status.ex
cmaddox5: Yes it would. Thanks!

lib/screens/v2/widget_instance/subway_status.ex
cmaddox5: Good call. This is much better.

lib/screens/v2/widget_instance/subway_status.ex
cmaddox5: Yikes. Good catch. Not sure how I missed that.

lib/screens/v2/widget_instance/subway_status.ex
jzimbel-mbta: Or maybe for slightly better readability and performance:
```ex
# Assuming gl_stop_sequences is now a list of sets as suggested in my other comment...

alert_stops =
  # ...current pipeline, plus this extra bit
  |> MapSet.new()

if MapSet.size(alert_stops) > 0 do
  Enum.count(gl_stop_sequences, &MapSet.subset?(alert_stops, &1)) > 1
else
  false
end
```

cmaddox5: I feel like `MapSet`s are my most underappreciated and underutilized data structure in Elixir. This is what I should have been doing from the beginning 😅

lib/screens/v2/widget_instance/subway_status.ex
```diff
@@ -468,8 +455,47 @@ defmodule Screens.V2.WidgetInstance.SubwayStatus do
...
+    informed_entities
+    |> Enum.map(fn e -> Map.get(e, :stop) end)
+    |> Enum.reject(&is_nil/1)
+    |> Enum.uniq()
+    |> Enum.any?(fn informed_stop ->
+      Enum.count(gl_stop_sets, &(informed_stop in &1)) > 1
+    end)
+  end
+
+  #
```
jzimbel-mbta, line 478: Might be helpful to have a brief description of the logic here like in the clause above, how about:
> If the affected stops are fully contained in more than one branch, the alert affects the trunk

lib/screens/v2/widget_instance/subway_status.ex
```diff
@@ -468,8 +455,47 @@ defmodule Screens.V2.WidgetInstance.SubwayStatus do
...
+  defp alert_affects_gl_trunk?(%Alert{informed_entities: informed_entities}, gl_stop_sequences) do
```
jzimbel-mbta, line 479: For naming consistency with the clause above:
```suggestion
  defp alert_affects_gl_trunk?(%Alert{informed_entities: informed_entities}, gl_stop_sets) do
```
(+ update the name where it's used ofc)

jzimbel-mbta (approved):
✅ Good to go! Just 2 quick things.

lib/screens/v2/widget_instance/subway_status.ex
cmaddox5: Guess I never finished my thought here 😬

PR #1803, opened 2023-06-28
polish: Total alert logic

cmaddox5:
**Asana task**: ad-hoc

I found another edge case 😄 If you have 1 branch alert and 1 multi-line alert that affects GL trunk, it was not being counted as 3 alerts. This was causing the non-GL route to show as `extended` because all GL alerts were counted as affecting a single route. Now, the logic looks at all unique affected routes and recognizes that trunk and branch alerts need to be counted separately. 

- [ ] Tests added?

### DIFF
--- a/lib/screens/v2/widget_instance/subway_status.ex
+++ b/lib/screens/v2/widget_instance/subway_status.ex
@@ -709,12 +709,10 @@ defmodule Screens.V2.WidgetInstance.SubwayStatus do
   defp get_total_alerts(grouped_alerts) do
     total_affected_routes =
       grouped_alerts
-      |> Map.keys()
-      |> Enum.uniq_by(fn
-        "Green-" <> _ -> "Green"
-        route_id -> route_id
-      end)
-      |> length()
+      |> Enum.flat_map(&elem(&1, 1))
+      |> Enum.uniq()
+      |> Enum.map(&get_total_affected_routes_for_alert/1)
+      |> Enum.sum()
 
     total_alerts =
       grouped_alerts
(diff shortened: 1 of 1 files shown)


lib/screens/v2/widget_instance/subway_status.ex
```diff
@@ -709,12 +709,10 @@ defmodule Screens.V2.WidgetInstance.SubwayStatus do
...
     total_affected_routes =
       grouped_alerts
-      |> Map.keys()
-      |> Enum.uniq_by(fn
-        "Green-" <> _ -> "Green"
-        route_id -> route_id
-      end)
-      |> length()
+      |> Enum.flat_map(&elem(&1, 1))
+      |> Enum.uniq()
```
jzimbel-mbta, line 713: `|> Enum.uniq_by(& &1.id)` might be a bit faster since it doesn't need to recursively check entire alert structs for equality.

Also just checking that I understand what's happening here: `grouped_alerts |> Enum.flat_map(&elem(&1, 1)) |> Enum.uniq()` just re-creates the original list of alerts from the `grouped_alerts` map, right? It merges all of the map's values into one list, and then removes duplicates.

Would it be easier to change the argument of `get_total_alerts` to be the original list of alerts so that we don't need to do that extra work? Or do we not have access to the original list in all of the places where this is called.

jzimbel-mbta (approved):
Looks good, just one thought!

lib/screens/v2/widget_instance/subway_status.ex
cmaddox5: Good points all around. With the move to using the original list of alerts, I noticed we can just call the function once in the beginning so that it isn't called so much.

PR #1807, opened 2023-06-29
[refactor] Replacing the old bundled folder (Review 2nd)

hannahpurcell:
In this PR: new svgs are added, and then these new SVGs are used everywhere that the old bundled svgs were used.

I didn't replace all the svgs from the old folder because some of them were more complex to replace, but I wanted to start phasing out this folder a bit.


assets/css/v2/eink/normal_header.scss
```diff
@@ -83,16 +84,10 @@
...
-.normal-header-updated__icon {
-  display: inline-block;
-  vertical-align: top;
-}
-
 .normal-header-updated__img {
-  position: absolute;
   height: 25px;
   width: 25px;
-  margin-right: 4px;
```
cmaddox5, line 95: A little nitpicky, but this change put more space than [designed](https://www.figma.com/file/uyXu91sE66baBZMg81XOmk/ScreenKit-2.0?type=design&node-id=0-1544&mode=dev) between the icon and the text.

assets/src/components/v2/cr_departures/cr_departures_header.tsx
```diff
@@ -1,13 +1,15 @@
 import React from "react";
 
-import Free from "Components/v2/bundled_svg/free";
-import { pillPath } from "Util/util";
+import Free from "../../../../static/images/svgr_bundled/free.svg";
+import CRPill from '../../../../static/images/svgr_bundled/pills/commuter-rail.svg'
 import RoutePill from "../departures/route_pill";
+import { getHexColor } from "Util/svg_utils";
```
cmaddox5, line 6: I see this file being added in your 4th PR. Can that be brought in here so this change works on its own?

hannahpurcell: Ope, good call!

cmaddox5 (approved):

PR #1810, opened 2023-06-29
polish: Last minute misc fixes

cmaddox5:
**Asana task**: [GL multi-branch station closures omit stops](https://www.notion.so/mbta-downtown-crossing/GL-multi-branch-station-closures-omit-stops-867ac83ee89b4feaa9c1d80386e3d1e6?pvs=4) and ad-hoc

3 big things:
1. Fixed `max-width` so that it adjusts as more branch pills are added.
2. Alert count logic was still a little off. If an alert affects GL at all, count it as 1 alert. No need to only consolidate when all branches are affected.
3. Changed a GL serializer function so that we show the full list of multi-branch station closures.

- [ ] Tests added?

### DIFF
--- a/assets/css/v2/lcd_common_styles/subway_status.scss
+++ b/assets/css/v2/lcd_common_styles/subway_status.scss
@@ -47,13 +47,16 @@
 
   .subway-status_status--contracted.subway-status_status--one-alert {
     // 114px
-    height: $status-top-padding + $pill-height + $contracted-status-with-one-alert-bottom-padding;
+    height: $status-top-padding + $pill-height +
+      $contracted-status-with-one-alert-bottom-padding;
   }
 
   .subway-status_status--contracted.subway-status_status--two-alerts:not(.subway-status_status--has-second-pill) {
     // 193px
     // Second alert still gets the height of a pill, despite not having one visible
-    height: $status-top-padding + (2 * $pill-height) + $intra-alert-margin-without-second-pill + $contracted-status-with-two-alerts-bottom-padding;
+    height: $status-top-padding + (2 * $pill-height) +
+      $intra-alert-margin-without-second-pill +
+      $contracted-status-with-two-alerts-bottom-padding;
   }
 
   // &--contracted.subway-status_status--two-alerts.subway-status_status--has-second-pill {}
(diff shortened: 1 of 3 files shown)


assets/css/v2/lcd_common_styles/subway_status.scss
```diff
@@ -134,9 +137,18 @@
...
     max-width: 830px;
     margin-left: 24px;
+
+    &--2-branches {
+      max-width: 671px;
+    }
+
+    &--3-branches {
+      max-width: 605px;
+    }
```
jzimbel-mbta, line 147: Does this address a different problem than "[GL multi-branch station closures omit stops](https://www.notion.so/mbta-downtown-crossing/GL-multi-branch-station-closures-omit-stops-867ac83ee89b4feaa9c1d80386e3d1e6?pvs=4)"? Wondering what issue this fixes.

I think I found that the  `max-width` property didn't really matter, because the element has `width: fit-content` and either gets shrunk to fit within the row, or overflows and gets clipped at the right edge of the row. So its width can be whatever in either case.

It should still wrap while `useTextResizer` is doing its thing—that is, while `isDone` is false—since [the sizer element sets a width](https://github.com/mbta/screens/blob/56a6c5e612f3f65eb266cef4fcf133bdea629c63/assets/css/v2/lcd_common_styles/subway_status.scss#L113-L116). Or at least it did the last time I messed with the resizing stuff

cmaddox5: I found that `max-width` was working, but in most cases the string never gets close to hitting it. As you add more branch pills, that available space goes down. So say you have a station closure at Saint Mary's Street and Sutherland Road: 

Screenshot 2023-06-30 at 12 08 50 PM

The text is still less than the max-width so it renders assuming everything fits. This causes the resizer to think it is ok rendering the station names as-is. If you reduce the max-width to what I have above, the resizer knows that there isn't enough space for the full station names so it tries the next method:

Screenshot 2023-06-30 at 12 11 09 PM

Does that make sense?

jzimbel-mbta: Oh weird, yeah that's not what I had expected. I thought I had it wrapping because it exceeded the widget of the containing `sizer` element, and had the appropriate properties set up there. I must not have checked properly for this case. Good catch.

PR #1813, opened 2023-07-05
Revert "[refactor] SVG better bundling feature branch"

hannahpurcell:
I didn't appropriately test this. I'm using one of these "bundled" icons in the pre-fare text resize logic now, and it revealed that the svgs are not loading quickly enough with this method. I have to re-evaluate this solution.


cmaddox5 (approved):
Interesting. I didn't think to test that either, so sorry for letting this get by. If you need any help finding a solution, let me know.

PR #1817, opened 2023-07-12
bugfix: Fix DUP playername logic. Use `getDeviceInfo`

hannahpurcell:
The original logic assumed player_name was passed with tags, but it actually should be accessed with a different function `getDeviceInfo`.


cmaddox5 (approved):

cmaddox5 (approved):
🚀

PR #1820, opened 2023-07-17
Revert "Revert "Revert #1813. Re-instate SVG better bundling""

hannahpurcell:
I'm sorry 😭 

(For future readers, we needed a late-Friday deploy to suppress planned alerts in advance of a weekend outage, but I really didn't want to deploy anything hefty on a Friday evening and have to monitor. So we reverted the only big thing that was recently merged: the svg work. Un-reverting now, for the umpteenth time. And will prioritize deploying to prod in the first half of the week.)


cmaddox5 (approved):
😂

PR #1822, opened 2023-07-18
fix: AlertBanner RL bug

cmaddox5:
**Asana task**: ad-hoc

When Ashmont and Braintree pills are added to the route pill list, they are being passed to the client as a list of lists. It should just be a singular list. I also added a `key` to the frontend to get rid of a console error. That change is on line 399. All other changes in that file are from format on save.

- [ ] Tests added?

### DIFF
--- a/assets/src/components/v2/pre_fare_single_screen_alert.tsx
+++ b/assets/src/components/v2/pre_fare_single_screen_alert.tsx
@@ -4,12 +4,12 @@ import { getHexColor, STRING_TO_SVG } from "Util/svg_utils";
 
 import { classWithModifier, classWithModifiers, formatCause } from "Util/util";
 
-import ClockIcon from "../../../static/images/svgr_bundled/clock-negative.svg"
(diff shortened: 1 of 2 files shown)


hannahpurcell:
I had to re-write the pill logic for the audio work and also fixed this bug. Since it hasn't been reviewed yet, perhaps the audio one should be reviewed first? Unsure

cmaddox5:
> I had to re-write the pill logic for the audio work and also fixed this bug. Since it hasn't been reviewed yet, perhaps the audio one should be reviewed first? Unsure

I agree. I _think_ the backend bug is fixed in your PR. I will keep this open just in case, but #1824 should definitely be reviewed first @jzimbel-mbta. @hannahpurcell, can you add the missing `key` prop on the `LinePill`s when there are more than one? That was also added here and will be needed to get rid of a console error. I can comment on the line in the PR.

jzimbel-mbta:
I think this PR can be closed—looks like the fixes have been applied in #1824.

PR #1824, opened 2023-07-24
[feature] Prefare alerts 2.0 audio

hannahpurcell:
**Asana task**: [Pre-Fare Alerts audio implementation](https://app.asana.com/0/1185117109217413/1204970356810193)

Needed to refactor serialization a bit, because the audio "frontend" needed different data format than visual frontend.

- [x] Tests added? (1 test left to fix)


assets/src/components/v2/pre_fare_single_screen_alert.tsx
```diff
@@ -266,13 +272,13 @@ const PreFareAlertBanner: React.ComponentType<{routes: any[]}> = ({routes}) => {
     banner = <div className={classWithModifiers("alert-banner", ["large--two-routes", getAlertColor(routes)])}>
       <span><span className="alert-banner__attention-text">ATTENTION,</span> riders to</span>
       {routes.map((route) => {
-        const LinePill = STRING_TO_SVG[route]
-        return <LinePill className="alert-banner__route-pill--long" color={getHexColor(getRouteColor(route))} />
+        const LinePill = STRING_TO_SVG[route.svg_name]
+        return <LinePill className="alert-banner__route-pill--long" color={getHexColor(getRouteColor(route.route_id))} />
```
cmaddox5, line 276:
```suggestion
        return <LinePill className="alert-banner__route-pill--long" key={route.svg_name} color={getHexColor(getRouteColor(route.route_id))} />
```

This is the change I mentioned in the other PR. Nothing wrong visually, but will help get rid of a console error.

hannahpurcell: Thank you!

lib/screens_web/views/v2/audio/reconstructed_alert_fullscreen_view.ex
```diff
@@ -0,0 +1,206 @@
...
+  use ScreensWeb, :view
+
+  alias Screens.Alerts.Alert
+  alias Screens.Util
+
+  def render("_widget.ssml", alert) do
+    ~E|<p><%= render_banner(alert) %><%= render_alert(alert) %></p>|
+  end
+
+  def render_banner(%{
```
jzimbel-mbta, line 11: This case is for a fullscreen alert, right? Since the `:unaffected_routes` key gets set only in that case? If so, a comment would help here. The significance of that field is not clear.

lib/screens_web/views/v2/audio/reconstructed_alert_fullscreen_view.ex
```diff
@@ -0,0 +1,206 @@
...
+        unaffected_routes: _unaffected_routes
+      }),
+      do: nil
+
+  def render_banner(%{
+        region: _region,
+        routes: routes
+      }) do
+    if routes |> hd() |> Map.has_key?(:headsign) do
```
jzimbel-mbta, line 20: Does `routes` always have at least one element? (Just want to make sure the `hd/1` call can't fail)

lib/screens_web/views/v2/audio/reconstructed_alert_fullscreen_view.ex
```diff
@@ -0,0 +1,206 @@
...
+      destinations = Enum.map(routes, fn route -> route.headsign end)
+
+      if length(destinations) < 3 do
+        ~E|Attention, riders to <%= Util.format_name_list_to_string_audio(destinations) %>. |
+      else
+        ~E|Attention, riders. |
+      end
+    else
+      ~E|Attention, <%= hd(routes).route_id %> riders. |
```
jzimbel-mbta, line 29: So is it a rule that if the first map in `routes` does not have a `:headsign` key, then we only care about its `:route_id` key instead? And don't care about the rest of the list?

In general, this code seems to make implicit assertions about the shape of the list and its elements that are not made very clear—this makes the code difficult to follow since I can't look at a typespec or doc comment to understand what the data looks like. I have to look back at the reconstructed alert widget code and walk through the logic to try and parse out the different possible shapes of these maps.

lib/screens_web/views/v2/audio/reconstructed_alert_fullscreen_view.ex
```diff
@@ -0,0 +1,206 @@
...
+        unaffected_routes: _unaffected_routes
+      }),
+      do: nil
+
+  def render_banner(%{
+        region: _region,
+        routes: routes
+      }) do
+    if routes |> hd() |> Map.has_key?(:headsign) do
+      destinations = Enum.map(routes, fn route -> route.headsign end)
```
jzimbel-mbta, line 21: So if the first map has a `:headsign` key, they all do?

lib/screens_web/views/v2/audio/reconstructed_alert_fullscreen_view.ex
```diff
@@ -0,0 +1,206 @@
...
+    ~E|There are no <%= get_line_name(route_svg_names) %> trains between <%= elem(endpoints, 0) %> and <%= elem(endpoints, 1) %><%= render_cause(cause) %>. Please seek an alternate route.|
+  end
+
+  # Boundary shuttle
+  def render_alert(%{
+        region: :boundary,
+        effect: :shuttle,
+        issue: issue,
+        routes: route_svg_names,
+        endpoints: endpoints,
```
jzimbel-mbta, line 83: Could pattern-match the two values so that they have names and you don't need to use `elem/2`: `endpoints: {left_endpoint, right_endpoint},`

Here and below.

lib/screens_web/views/v2/audio/reconstructed_alert_fullscreen_view.ex
```diff
@@ -0,0 +1,206 @@
...
+  end
+
+  defp get_line_name(routes) do
+    route_ids =
+      routes
+      |> Enum.map(fn route -> route.route_id end)
+
+    lines_with_branches =
+      route_ids
+      |> Enum.filter(&String.contains?(&1, "-"))
```
jzimbel-mbta, line 179: Maybe `String.contains?(&1, "Green-")` just to be safe?

Alternatively, you can filter and map in one go with a list comprehension:
```ex
lines_with_branches = for "Green-" <> branch_letter <- route_ids, do: branch_letter
```
(It skips any elements that don't match the pattern)

lib/screens_web/views/v2/audio/reconstructed_alert_fullscreen_view.ex
```diff
@@ -0,0 +1,206 @@
...
+      |> Enum.map(fn "Green-" <> branch_letter -> branch_letter end)
+
+    lines_without_branches =
+      route_ids
+      |> Enum.filter(&(!String.contains?(&1, "-")))
```
jzimbel-mbta, line 184: `Enum.reject` might be a little easier to follow here, `Enum.reject(&String.contains?(&1, "-"))`

lib/screens_web/views/v2/audio/reconstructed_alert_fullscreen_view.ex
```diff
@@ -0,0 +1,206 @@
...
+    |> Enum.map(fn route -> route.text end)
+    |> Util.format_name_list_to_string_audio()
+  end
+
+  defp get_line_name(routes) do
+    route_ids =
+      routes
+      |> Enum.map(fn route -> route.route_id end)
+
+    lines_with_branches =
```
jzimbel-mbta, line 177: Maybe rename to `branch_letters`? The list only contains the letters.

jzimbel-mbta (commented):
Mostly little things—I'm not as familiar with the reconstructed alert code, so I had a few questions while trying to follow the logic.

assets/src/components/v2/pre_fare_single_screen_alert.tsx
```diff
@@ -216,35 +223,34 @@ const PreFareSingleScreenAlert: React.ComponentType<PreFareSingleScreenAlertProp
...
 // use that, otherwise "yellow"
-const getAlertColor = (routes: string[]) => {
-  const colors = routes.map(r => getRouteColor(r))
+const getAlertColor = (routes: EnrichedRoute[]) => {
+  const colors = routes.map(r => getRouteColor(r.route_id))
   const uniqueColors = new Set(colors).size
-  return uniqueColors === 1 ? colors[0] : "yellow"
+
+  const answer = uniqueColors == 1 ? colors[0] : "yellow"
+  return answer
```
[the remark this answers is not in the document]
hannahpurcell: 😅 ah just a relic of using print statements to debug!

lib/screens/stops/stop.ex
```diff
@@ -460,9 +460,12 @@ defmodule Screens.Stops.Stop do
               app in [PreFare, Dup] ->
                 RoutePattern.fetch_parent_station_sequences_through_stop(stop_id, route_ids)
             end) do
+      stop_name = fetch_stop_name(stop_id)
+
       {:ok,
        %LocationContext{
          home_stop: stop_id,
+         home_stop_name: stop_name,
```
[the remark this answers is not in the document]
hannahpurcell: Currently, `home_stop_name` is fetched exclusively from the API, not from our hardcoded values. So I could add a new function that does the fetch and then runs the value through our hardcoded stop lists to get the abbrev name. 

It seems rare for us to abbreviate the stop name of a screen, though I guess we must do it for a couple headers like Tufts Medical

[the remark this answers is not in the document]
hannahpurcell: Ah, this LocationContext is app agnostic, so this is used for bus, too. We don't do abbrev for bus stops. I'd say we should skip this, then, but I could also just use the full name for both full and abbrev for bus stops.

lib/screens/util.ex
```diff
@@ -138,6 +138,19 @@ defmodule Screens.Util do
     |> Enum.join(", ")
   end
 
+  @doc """
+  Same as regular string list formatter, but for audio (extra comma for clarity, "and" instead of "&")
+  """
+  @spec format_name_list_to_string_audio([String.t()]) :: String.t()
+  def format_name_list_to_string_audio([string]), do: "#{string}"
+  def format_name_list_to_string_audio([s1, s2]), do: "#{s1} and #{s2}"
```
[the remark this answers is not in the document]
hannahpurcell:
```suggestion
  def format_name_list_to_string_audio([s1, s2]), do: "#{s1}, and, #{s2}"
```

lib/screens/v2/widget_instance/reconstructed_alert.ex
```diff
@@ -44,28 +45,40 @@ defmodule Screens.V2.WidgetInstance.ReconstructedAlert do
   # }
 
   @type takeover_serialized_response :: %{
+          optional(:other_closures) => list(String.t()),
```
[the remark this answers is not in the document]
hannahpurcell: Only 1 of 3 `serialize_takeover_alert` functions return `other_closures` in the response, so the key will not always be present

lib/screens/v2/widget_instance/reconstructed_alert.ex
```diff
@@ -44,28 +45,40 @@ defmodule Screens.V2.WidgetInstance.ReconstructedAlert do
...
+          # Unique to transfer station case
+          optional(:is_transfer_station) => boolean(),
           issue: String.t() | list(String.t()),
           cause: Alert.cause() | nil,
           # List of SVG filenames
-          routes: list(String.t()),
+          routes: list(enriched_route()),
           effect: :suspension | :shuttle | :station_closure | :delay,
           updated_at: String.t(),
+          # It seems that region was also added to the flex serialized response?
```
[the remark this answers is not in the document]
hannahpurcell: Ah sure. I was confused why `region` was only identified on the fullscreen structure because it also seems to appear in the flex responses.

lib/screens/v2/widget_instance/reconstructed_alert.ex
```diff
@@ -172,36 +183,68 @@ defmodule Screens.V2.WidgetInstance.ReconstructedAlert do
...
+        route_id: route_id,
+        svg_name: format_for_svg_name(dest),
+        headsign: dest
+      }
+    end)
+  end
+
+  # If headsign is for the trunk, use "Green" as route_id
+  defp build_pills_from_headsign(_route_id, headsign)
+       when headsign in ["North Station & North", "Copley & West"] do
```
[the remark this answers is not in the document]
hannahpurcell: I didn't mess with the headsign-generating logic, just the headsign-presentation logic. Not sure how the headsigns are generated. Looks like that happens in `get_destination`

lib/screens/v2/widget_instance/reconstructed_alert.ex
```diff
@@ -353,7 +410,8 @@ defmodule Screens.V2.WidgetInstance.ReconstructedAlert do
       effect: :suspension,
       updated_at: format_updated_at(updated_at, now),
       region: get_region_from_location(location),
-      endpoints: endpoints
+      endpoints: endpoints,
+      is_transfer_station: location === :inside
```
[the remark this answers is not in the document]
hannahpurcell: This is part of why I thought the `serialize_fullscreen_alert` should be renamed to `serialize_single_screen_alert`, to set it apart from takeovers. By the time we get to `serialize_fullscreen_alert`, we know that the alert is not a takeover. This alert clause is for a suspension, and if the location is `:inside` the alert, but somehow not a takeover, then we know there are other lines still running at that location, so it's a transfer station

lib/screens/v2/widget_instance/reconstructed_alert.ex
```diff
@@ -432,8 +491,9 @@ defmodule Screens.V2.WidgetInstance.ReconstructedAlert do
       end
 
     %{
-      issue: format_routes(affected_routes),
-      unaffected_routes: format_routes(unaffected_routes),
+      issue: nil,
```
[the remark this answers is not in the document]
hannahpurcell: This function clause is for a station closure at a transfer station. `issue` was an unused field for the frontend, and it was easier to set up the audio with the assertion that `issue` would be nil in such cases.

lib/screens_web/views/v2/audio/reconstructed_alert_fullscreen_view.ex
hannahpurcell: Ooooo nice! No it doesn't! Fixing

lib/screens_web/views/v2/audio/reconstructed_alert_fullscreen_view.ex
hannahpurcell: Yeah, that's right. Either they all get condensed to just simple RL / BL route pills, or they have direction included. [Miro](https://miro.com/app/board/uXjVM-ve3yE=/?moveToWidget=3458764557094546650&cot=14)

lib/screens_web/views/v2/audio/reconstructed_alert_fullscreen_view.ex
hannahpurcell: Oh geez, yep

lib/screens_web/views/v2/audio/reconstructed_alert_fullscreen_view.ex
hannahpurcell: Woof, yeah, sorry this was so convoluted. I don't see a great way of simplifying, but I did add comments to explain. Was the rest of the file commented well enough to follow?

jzimbel-mbta (approved):
Good to go! Thanks for your patience :snail:

PR #1830, opened 2023-08-02
[tweak] Disruption diagram maths

hannahpurcell:
In order of changes:
- increase L for when the first icon is a big diamond
- adjust all icons to be centered right on the middle of the slot (based on width of the icon) to make consistent between circle icons and all the other icons
- translate the "..." label to the left, to be most centered over the slot
- update math for EffectBackgroundComponent so that the dotted line is perfectly spaced like the design
- adjust translation of icon in AlertEmphasisComponent so that if L is increased to, say, 50, it is still centered correctly


cmaddox5 (approved):
Looks so much more polished! Thanks!

PR #1833, opened 2023-08-11
feat: Working triptych app skeleton with viewport shifter, request logging, packaging instructions

jzimbel-mbta:
**Asana task**: [Get basic triptych app working & set up packaging for OF process](https://app.asana.com/0/1185117109217413/1204882696236163/f)

This ended up being a bit of a snowball of changes—I needed to add some new frontend and backend code to support how we plan to identify individual triptych panes based on their player names.

The diff is huge, but it's almost all from straightforward refactors I did to make working on the OFM client apps less error-prone.

# Summary of changes since the first round of reviews

## Triptych player name mapping config
The server now expects us to maintain a mapping from triptych player name to screen ID in a config file.

Implementation is identical to that of the main screen config—we use a local JSON file in priv/ for local development, and per-environment files in S3 for deployment environments. The mapping is cached on the server, and can be accessed with functions in [`Screens.TriptychPlayer.State`](https://github.com/mbta/screens/blob/jz/triptych-packaging/lib/screens/triptych_player.ex).

This required a terraform change to add permission for the new file in S3. [Link to relevant PR](https://github.com/mbta/devops/pull/1148)

## Triptych API endpoint
A new [API endpoint](https://github.com/mbta/screens/blob/jz/triptych-packaging/lib/screens_web/controllers/v2/screen_api_controller.ex#L86): `/v2/screen/:player_name/triptych`.

The triptych client app uses this endpoint when running in a standalone package.

It maps the player name to a screen ID using the new config, then runs identical logic to all other screen types.

## Determine pane from MRAID tags
The triptych client now [determines which pane of the trio it is via the `Array_configuration` MRAID tag](https://github.com/mbta/screens/blob/jz/triptych-packaging/assets/src/util/outfront.tsx#L67).

---

☝️ ^ The above 3 changes ^ collectively mean that we only need to package one triptych app now! (Not 3)

---

## Improved common OFM functions
Utility functions related to OFM stuff have been collected in [Util/outfront.tsx](https://github.com/mbta/screens/blob/jz/triptych-packaging/assets/src/util/outfront.tsx) and [Hooks/outfront.tsx](https://github.com/mbta/screens/blob/jz/triptych-packaging/assets/src/hooks/outfront.tsx), and cleaned up.

The OFM hooks no longer use state or effects—just a memoized function call. This is because the values in the MRAID object are not expected to change. By using `useMemo` instead of `useState` + `useEffect`, we don't need to worry about handling an "initial" value like `null` on the first render, as well as the real value in subsequent renders. `useMemo` returns the real value synchronously on the first render.

### Fake MRAID object
The process for testing packaged DUP/triptych apps locally is now closer to how they work on real OFM screens.

You can call [a function](https://github.com/mbta/screens/blob/jz/triptych-packaging/assets/src/util/outfront.tsx#L152) to set up a fake MRAID object that matches the interface of the real one.

### Improved debug element
Improved the on-screen debug element for OFM clients. Newest lines now appear first so that relevant info doesn't scroll out of view after the first couple renders.

<details>

### left
image

### middle
image

### right
image

</details>

- [ ] Tests added?


assets/css/v2/triptych/body/normal.scss
jzimbel-mbta, line 1: The current triptych template doesn't have a "body" region, so this module/component wasn't used or needed

assets/css/v2/triptych/screen_container.scss
```diff
@@ -1,7 +1,16 @@
 .screen-container {
   position: relative;
-  width: 1080px;
+  width: 3240px;
   height: 1920px;
   margin: 0px auto;
   overflow: hidden;
 }
+
+.screen-container-blink {
```
jzimbel-mbta, line 9: Oops, actually I didn't need to add this. We'll never use a "blink" element on this screen type. Will remove

assets/src/components/v2/triptych/no_data.tsx
```diff
@@ -0,0 +1,27 @@
...
+import LcdNoData from "Components/v2/lcd/no_data";
+
+interface Props {
+  show_alternatives: boolean;
+}
+
+const NoData: ComponentType<Props> = ({
+  show_alternatives: showAlternatives,
+}) => {
+  // TODO: We likely want to show something different from the usual when we fail to fetch data for this screen type.
```
jzimbel-mbta, line 11: (Do we have designs for this? I'm assuming we'd prefer to just show some baked-in PSA image or something instead of telling riders the app, which they weren't already keeping a particularly close eye out for since it's mixed in with ads, broke. Like an employee at a store rushing over to inform you that some specific item is out of stock when you weren't even looking for it.)

priv/triptych_preview.png
jzimbel-mbta, line 1: Just a placeholder until we're able to get a screenshot of the real app, since OFM requires one in their template.json.

assets/src/components/v2/triptych/no_data.tsx
hannahpurcell: Yeahhh no designs for this. I'd guess we'd fall back to evergreen widget (which will have the psas). Hopefully we wouldn't often run into a situation where even that widget couldn't render? 😬  So I don't think we'd want this frontend component.

assets/src/hooks/v2/use_api_response.tsx
```diff
@@ -163,8 +164,8 @@ const useBaseApiResponse = ({
   let refreshRateOffsetMs = parseInt(refreshRateOffset, 10) * 1000;
   let apiPath = `/v2/api/screen/${id}${routePart}?last_refresh=${lastRefresh}${isRealScreenParam}${screenSideParam}${requestorParam}`;
 
-  if (isDup()) {
-    apiPath = `https://screens.mbta.com${apiPath}&rotation_index=${ROTATION_INDEX}`;
+  if (isOFM()) {
+    apiPath = `http://localhost:4000${apiPath}&rotation_index=${ROTATION_INDEX}&pane=${TRIPTYCH_PANE}`;
```
hannahpurcell, line 168: Won't want to commit this bit

lib/screens/log_screen_data.ex
```diff
@@ -19,7 +19,8 @@ defmodule Screens.LogScreenData do
         is_screen,
         requestor,
         screen_side \\ nil,
-        rotation_index \\ nil
+        rotation_index \\ nil,
+        triptych_pane \\ nil
```
hannahpurcell, line 23: Ah! Wonderful you remembered adding this to the logging

assets/src/hooks/v2/use_api_response.tsx
jzimbel-mbta, line 168: Whoops, accidentally committed this change from an earlier attempt at packaging. I'll reverse the localhost url back to screens.mbta.com.

Also, I think I will need to conditionally add the `rotation_index` and `pane` params instead of always adding both for either OFM app—it would make the logs kind of confusing if both were included for any DUP/triptych request.

assets/src/components/v2/triptych/README.md
```diff
@@ -0,0 +1,82 @@
...
+      <script type="text/javascript" src="polyfills.js"></script>
+      <script type="text/javascript" src="triptych_v2.js"></script>
+    </body>
+  </html>
+  ```
+
+- Set the version string in assets/src/components/v2/triptych/version.tsx to `current_year.current_month.current_day.1`.
+- In assets/webpack.config.js, change `publicPath` in the font config to have value `'fonts/'`.
+- **Only if you are packaging for local testing**
+  - replace `const playerName = useOutfrontPlayerName();` in assets/src/apps/v2/triptych.tsx with `const playerName = "BKB-TRI-001";` (or any other player name from one of the triptych screen IDs (`TRI-${playerName}`)). This data is provided by Outfront's "wrapper" app that runs on the real triptych screens, but we need to set it ourselves during testing. Think of it as a sort of frontend environment variable.
```
jzimbel-mbta, line 35: That example player name is probably not valid—I just made a guess based on what some of the DUP player names look like. We'll need to update this line once we have a real player name to work with.

hannahpurcell (commented):
A lot of the layout components + CSS were stubbed out, which I didn't expect. I knew we'll need to do them eventually, but I figured the packaging step was primarily to get the bare bones app exported, even if it was just the blank screen with slot_names. (But perhaps Paul expected the placeholders like you did. Maybe a boring screen would have been a waste of Outfront's time.) Couple notes to get us started, while Outfront looks at the app

assets/src/components/v2/triptych/no_data.tsx
jzimbel-mbta: Yeah I'm talking about when we lose network connection. I guess if that happened then OFM would hopefully look into it, though.

assets/src/util/util.tsx
```diff
@@ -51,16 +82,20 @@ export const getScreenSide = (): ScreenSide | null => {
...
-  const rotationIndex = isDup()
+export const getRotationIndex = (): RotationIndex | null => {
+  const rotationIndex = isOFM()
     ? ROTATION_INDEX.toString()
     : getDatasetValue("rotationIndex");
 
   return isRotationIndex(rotationIndex) ? rotationIndex : null;
 };
 
+export const getTriptychPane = (): TriptychPane | null => {
```
cmaddox5, line 93: So the big difference between DUPs and Triptychs is that each "pane" is a unique `playerName`. So I don't think this will be as reliable for triptychs. We may need to add the pane to the screen's config. In the CSV they gave us, we can see which `playerName` is for which pane.

jzimbel-mbta: Ahhhhh shoot. Ok. Will need to rethink this.

assets/css/v2/triptych/screen/takeover.scss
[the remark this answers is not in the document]
jzimbel-mbta: This is a reasonable point. I can update this code to match the template/layout you went with for the frontend slots.

jzimbel-mbta:
FYI: CI is currently failing because the app won't start without access to the new JSON file in S3. It should pass once my terraform change gets applied.

hannahpurcell (commented):
Not much, but couple o' items (unrelated to the upcoming merge resolution)

assets/src/components/dup/dup_screen_page.tsx
```diff
@@ -37,7 +37,7 @@ const ScreenPage = ({
 }: {
   screenContainer: React.ComponentType;
 }): JSX.Element =>
-  isDup() ? (
+  isDUP() ? (
```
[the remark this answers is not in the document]
jzimbel-mbta: Yeah, my bad, this is an artifact of my initially merging the check into a new `isOFM` function, and then needing to go back and have separate `isDUP` and `isTriptych` functions. In the process I accidentally renamed the original `isDup` to `isDUP`. 😵‍💫

I'll restore this function's original name.

lib/screens/application.ex
```diff
@@ -30,9 +31,9 @@ defmodule Screens.Application do
       # ScreensByAlert server process
       Screens.ScreensByAlert,
       # Task supervisor for ScreensByAlert self-refresh jobs
-      {Task.Supervisor, name: Screens.ScreensByAlert.SelfRefreshRunner.TaskSupervisor},
+      {Task.Supervisor, name: Screens.ScreensByAlert.SelfRefreshRunner.TaskSupervisor}
       # ScreensByAlert self-refresh job runner
-      {Screens.ScreensByAlert.SelfRefreshRunner, name: Screens.ScreensByAlert.SelfRefreshRunner}
+      # {Screens.ScreensByAlert.SelfRefreshRunner, name: Screens.ScreensByAlert.SelfRefreshRunner}
```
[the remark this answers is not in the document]
jzimbel-mbta: Oy, thank you for catching this!

assets/css/triptych_v2.scss
```diff
@@ -27,13 +27,26 @@
...
+@import "v2/triptych/viewport";
 
+// Layout/slot elements
 @import "v2/triptych/screen/full_screen";
-@import "v2/train_crowding";
 
+// No-data states
+@import "v2/lcd_common_styles/page_load_no_data";
+@import "v2/lcd_common_styles/no_data";
+@import "v2/triptych/no_data";
```
hannahpurcell, line 40: I'll try to get this reviewed / merged before the no-data branch. It's possible these may / may not be needed depending on the implementation of that, but we can leave that for the no-data pr to resolve.

assets/src/components/dup/dup_screen_page.tsx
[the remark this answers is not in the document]
hannahpurcell: I checked the latest updates, and it still looks like the changes are there, so we might be talking about different things.

Since this is a DUP v1 file, I'm wondering why the change to the function name needed to be made? We don't have any DUP v1s in the world anymore and I'm not concerned about keeping it maintained, but maybe that's a philosophically bad approach?

assets/src/apps/dup.tsx
```diff
@@ -17,7 +17,7 @@ import {
   MultiRotationPage,
   SimulationPage,
 } from "Components/dup/dup_screen_page";
-import { isDup } from "Util/util";
+import { isDup } from "Util/outfront";
```
hannahpurcell, line 20: I unresolved a previous comment thread to ask about this in better detail. (Just noting here so you don't miss it.)

hannahpurcell (approved):
Great!! I think this is ready. Two non-blocking comments

jzimbel-mbta:
GH won't let me directly reply to your comment about the edit to the old file, so—
Yeah, if I hadn't updated this reference to that function, I think the code would fail to compile or at least print some warnings to the console.

The real reason it's updated though, is simply that it's easier to do so. I used the vscode refactor function (<kbd>F2</kbd>) to rename the function as well as automatically update all references to it. It would have been harder to go back afterward and undo the renames in the old files.

assets/src/apps/dup.tsx
jzimbel-mbta: Yeah, if I hadn't updated this and other references to `isDup` in the old code, I think the code would fail to compile or at least print some warnings to the console even though we don't have any screens configured to use that app bundle anymore.

The real reason it's updated though, is simply that it's easier to do so. I used the vscode refactor function (F2) to rename the function as well as automatically update all references to it. It would have been harder to go back afterward and undo the renames in the old files.

hannahpurcell: > The real reason it's updated though, is simply that it's easier to do so. I used the vscode refactor function (F2)

Love it! That's a great reason.

PR #1837, opened 2023-08-24
tweak: PreFare simulations

cmaddox5:
**Asana task**: [[Pre-Fare Alerts] "New look" in Screenplay](https://app.asana.com/0/1185117109217413/1205259333913109/f)

Now that there are two different "flex zones" on PreFare screens, a little tweaking was needed so users can see all pages on both screens. To make things simpler, PreFare has its own set of simulation components and CSS. A single server change was needed so grouping logic did not incorrectly group pages with the same page number of different screens.

Screenshot 2023-08-24 at 3 13 26 PM

- [ ] Tests added?

### DIFF
--- a/assets/src/components/v2/pre_fare/simulation_screen_container.tsx
+++ b/assets/src/components/v2/pre_fare/simulation_screen_container.tsx
@@ -1,0 +1,117 @@
+import React, { ComponentType, useContext } from "react";
+import {
+  LastFetchContext,
+  ResponseMapperContext,
+} from "Components/v2/screen_container";
+import Widget, { WidgetData } from "Components/v2/widget";
+import {
+  ApiResponse,
+  useSimulationApiResponse,
+} from "Hooks/v2/use_api_response";
+import WidgetTreeErrorBoundary from "Components/v2/widget_tree_error_boundary";
+
+interface SimulationScreenLayoutProps {
+  apiResponse: ApiResponse;
+  opts: { [key: string]: any };
+}
+
+const SimulationScreenLayout: ComponentType<SimulationScreenLayoutProps> = ({
+  apiResponse,
+  opts,
+}) => {
+  const responseMapper = useContext(ResponseMapperContext);
+  const data = responseMapper(apiResponse);
+  const { fullPage, flexZone } = data;
+  let leftScreenPages: WidgetData[] = [];
+  let rightScreenPages: WidgetData[] = [];
+  if (flexZone) {
+    leftScreenPages = flexZone.filter(
+      (widget: WidgetData) => widget.type === "body_left_flex"
+    );
+    rightScreenPages = flexZone.filter(
+      (widget: WidgetData) => !leftScreenPages.includes(widget)
+    );
+  }
+
+  // If "alternateView" was provided as an option, we use the simulation version of screen normal
+  // Currently only applies to DUPs
+  const widgetData = opts.alternateView
+    ? { ...fullPage, type: "simulation_screen_normal" }
+    : fullPage;
+
+  return (
+    <div className="simulation-screen-centering-container">
+      <div className="simulation-screen-scrolling-container">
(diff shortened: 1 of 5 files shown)


lib/screens/v2/screen_data.ex
```diff
@@ -410,7 +410,14 @@ defmodule Screens.V2.ScreenData do
 
     instance_map
     |> Enum.group_by(
-      fn {paged_slot_id, _} -> Template.get_page(paged_slot_id) end,
+      fn
+        {paged_slot_id, instance} ->
+          if Map.has_key?(instance, :is_full_screen) and instance.is_full_screen do
```
jzimbel-mbta, line 415: Separately: it feels a little janky to distinguish paged stuff on the left screen from that in the right screen flex-zone, by a key that happens to be on the one type of widget that can currently occupy that space.

Is there any other way to distinguish them?

If I'm following the code correctly, `paged_slot_id` should look something like `{0, :paged_main_content_left}` for a widget occupying one of the paged left screen slots. Checking for that slot ID would be a bit better, since it's independent of whatever widget is going in the slot.

So maybe something like:
```ex
|> Enum.group_by(
  fn
    {{_n, :paged_main_content_left} -> :paged_main_content_left
    {paged_slot_id, _} -> Template.get_page(paged_slot_id)
  end,
  fn {paged_slot_id, instance} -> {Template.unpage(paged_slot_id), instance} end
)
```

This still isn't ideal since it's specific to pre-fares, but maybe a bit better? :shrug:

jzimbel-mbta (approved):
Looks good! Just some thoughts and suggestions.

I do wonder how we would generalize the logic that divides paged content into its separate containers, but any approach that works for the narrow case of pre-fares is good enough for now.

assets/src/components/v2/pre_fare/simulation_screen_container.tsx
```diff
@@ -0,0 +1,117 @@
...
+    rightScreenPages = flexZone.filter(
+      (widget: WidgetData) => !leftScreenPages.includes(widget)
+    );
+  }
+
+  // If "alternateView" was provided as an option, we use the simulation version of screen normal
+  // Currently only applies to DUPs
+  const widgetData = opts.alternateView
+    ? { ...fullPage, type: "simulation_screen_normal" }
+    : fullPage;
```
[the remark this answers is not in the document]
cmaddox5: Yes! Totally overlooked this.

lib/screens/v2/screen_data.ex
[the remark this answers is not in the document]
cmaddox5: O yeah, I like that much better. Thanks!

cmaddox5: Yeah this is a tricky situation and definitely the what I spent the most time figuring out. This problem only exists because we get data for both screens at the same time (or if we ever have two flexzones on one screen for other apps). Because of this, I don't see any screen besides PreFare having this issue unless we add a flexzone to another screen.

Would it be cleaner to leverage `app_id` in this? Maybe bring that in as a parameter, refactor the key grouping anon function into a named function, and use a guard. What do you think?

jzimbel-mbta: I think yeah, it might be best to just have special behavior for the pre-fare app. A separate named function leaves room for defining clauses for for other apps later, if need be, and separates the app-specific stuff from the common logic a little bit better.

jzimbel-mbta: tl;dr Yes, a named `paged_content_key` (or some better name) function that also takes app ID as an argument, assuming that's easy to access from this function, sounds good.

cmaddox5: Got that added. Mind taking one more look?

jzimbel-mbta (approved):
Double approved ✅✅

PR #1839, opened 2023-08-29
tweak: Train Crowding Widget Logging

cmaddox5:
**Asana task**: [Add logging to triptych app - pt 1](https://app.asana.com/0/1185117109217413/1205345095627639/f)

Added a log for each car so we can make some Splunk charts. The biggest change is a new config for the `Carriage` objects in the API. I wanted to make sure we tracked the `car_number` for the logs. Once merged, I will work on some Splunk charts/tables.

- [ ] Tests added?


lib/screens/v2/candidate_generator/widgets/train_crowding.ex
```diff
@@ -88,4 +96,15 @@ defmodule Screens.V2.CandidateGenerator.Widgets.TrainCrowding do
...
+  defp log_crowding_info(prediction, crowding_config) do
+    Enum.each(
+      prediction.vehicle.carriages,
+      fn %Screens.Vehicles.Carriage{} = carriage ->
+        Logger.info(
+          "[train_crowding car_crowding_info] station_id=#{crowding_config.station_id} direction_id=#{crowding_config.direction_id} car_number=#{carriage.car_number} vehicle_id=#{prediction.vehicle.id} crowding_level=#{carriage.occupancy_status} trip_id=#{prediction.trip.id} prediction_id=#{prediction.id}"
+        )
+      end
+    )
+  end
```
hannahpurcell, line 109: Instead of separate logging for each car, can we log the array of all car crowding for a vehicle?

cmaddox5: The main reason for splitting out the cars is for easier grouping and comparisons in Splunk tables/dashboards. I worry that keeping all info in the same log will require some complicated log parsing. I guess we could go with one log and, as long as we ensure sorting is consistent, look for any time that the serialized arrays are not equal. Is that what you had in mind?

hannahpurcell: Yeah, that's what I had in mind. What is the use case for grouping and comparisons on the car level?

lib/screens/v2/candidate_generator/widgets/train_crowding.ex
```diff
@@ -88,4 +107,17 @@ defmodule Screens.V2.CandidateGenerator.Widgets.TrainCrowding do
...
   end
+
+  defp log_crowding_info(prediction, crowding_config, true, screen_id, triptych_pane) do
+    crowding_levels =
+      prediction.vehicle.carriages
+      |> Enum.sort_by(& &1.car_number)
+      |> Enum.map_join(",", & &1.occupancy_status)
+
+    Logger.info(
+      "[train_crowding car_crowding_info] screen_id=#{screen_id} triptych_pane=#{triptych_pane} station_id=#{crowding_config.station_id} direction_id=#{crowding_config.direction_id} trip_id=#{prediction.trip.id} prediction_id=#{prediction.id} vehicle_id=#{prediction.vehicle.id} car_crowding_levels=#{crowding_levels}"
```
hannahpurcell, line 118: Which are the primary values that we need in splunk? `vehicle_id` is here, but not in the other log statement, so I'm wondering which values will help us most in the logs: prediction id, vehicle id, trip id?

lib/screens/v2/candidate_generator/widgets/train_crowding.ex
```diff
@@ -51,6 +56,12 @@ defmodule Screens.V2.CandidateGenerator.Widgets.TrainCrowding do
            params |> Map.to_list() |> fetch_alerts_fn.() do
       next_train_prediction = List.first(predictions)
 
+      if opts[:is_real_screen] do
```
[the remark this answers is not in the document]
cmaddox5: The motivation here is `opts` is a pretty common name for a parameter list in Elixir. The `opts` term in `by_screen_id` actually already existed for `ScreensByAlert` (see [here](https://github.com/mbta/screens/blob/1ffe934b87f8f683fd06f66529b559106ce13479/lib/screens/v2/screen_data.ex#L48). It basically allows functions to hold as many parameters as it needs and condenses them down to a single variable. Then the implementation of that function can attempt to access what it expects. You can safely access any keyword from the list, but any that don't exist are just `nil`.

lib/screens/v2/candidate_generator/widgets/train_crowding.ex
cmaddox5: I think the amount I included is overkill. We really only need `screen_id`, `triptych_pane`, `trip_id`, and `car_crowding_levels`. The rest I included can either be easily found using the `screen_id` or will not be needed (like `prediction_id`).

lib/screens_web/controllers/v2/screen_api_controller.ex
```diff
@@ -80,7 +80,14 @@ defmodule ScreensWeb.V2.ScreenApiController do
           screen_side
         )
 
-        json(conn, ScreenData.by_screen_id(screen_id))
+        json(
+          conn,
+          ScreenData.by_screen_id(screen_id,
+            is_real_screen: is_screen,
+            screen_id: screen_id,
+            triptych_pane: triptych_pane
```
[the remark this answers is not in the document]
cmaddox5: I think one of the advantages to using an `opts` list is that they can be arbitrary in length and a function doing a lookup in `opts` only needs to look for what they know they need. Similar to `params` in this action. This action is used by all screen types but references very specific values only needed by certain screen types. Because a lookup of key that doesn't exist is safe for these lists, we just need to do some `nil` checks before using them. 

I definitely see where you are coming from though. My biggest hesitation with conditionally adding it is that it will make the code here a little more complex for no drastic benefit. What do you think?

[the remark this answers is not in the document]
hannahpurcell: Ok, I can see that. Works for me!

lib/screens/v2/candidate_generator/widgets/train_crowding.ex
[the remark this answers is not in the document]
hannahpurcell: We chatted over a huddle, and found a way to re-org!

lib/screens/v2/candidate_generator/triptych.ex
```diff
@@ -20,12 +20,13 @@ defmodule Screens.V2.CandidateGenerator.Triptych do
...
         config,
-        crowding_widget_instances_fn \\ &Widgets.TrainCrowding.crowding_widget_instances/1,
+        opts,
+        crowding_widget_instances_fn \\ &Widgets.TrainCrowding.crowding_widget_instances/2,
         evergreen_content_instances_fn \\ &Widgets.Evergreen.evergreen_content_instances/1,
         local_evergreen_set_instances_fn \\ &Widgets.LocalEvergreenSet.local_evergreen_set_instances/1
       ) do
     [
-      fn -> crowding_widget_instances_fn.(config) end,
+      fn -> crowding_widget_instances_fn.(config, opts) end,
```
hannahpurcell, line 29: From our huddle conversation, I suggested this so the train_crowding widget only has the stuff it needs
```suggestion
      fn -> crowding_widget_instances_fn.(config, opts[:logging_options]) end,
```

hannahpurcell: And then the associated changes are needed in the train_crowding candidate_generator

cmaddox5: Ahh I see. I misunderstood. Will make that change now.

hannahpurcell (approved):
Great!

PR #1842, opened 2023-08-30
fix: Handle unrecognized screen IDs in `screen_name_for_id`

jzimbel-mbta:
**Asana task**: ad hoc

Little fix for an issue I noticed while poking around. This change lets the server properly log data requests from unrecognized screens, and send back 404s.

- [ ] Tests added?


hannahpurcell (approved):
👍

PR #1847, opened 2023-09-01
feat: Triptych simulations

cmaddox5:
**Asana task**: [Make triptych/crowding app show up correctly in Screenplay](https://app.asana.com/0/1185117109217413/1205301761265225/f)

Added a simulation component and styling for Triptychs. Very simple layout right now. Just show all 3 panes together in a row with no gaps. 

- [ ] Tests added?


assets/css/v2/triptych/simulation.scss
```diff
@@ -0,0 +1,77 @@
...
+  display: flex;
+  overflow-y: hidden;
+  overflow-x: auto;
+  padding-bottom: 20px;
+
+  .simulation-three-flex-panes {
+    height: 344px;
+    width: 580.5px;
+    z-index: 999;
+    display: flex;
```
jzimbel-mbta, line 35: I'd prefer sticking with the usual approach of `position: relative/absolute` + `overflow: hidden` to avoid the possibility of one overflowing widget pushing other content around.

assets/src/apps/v2/triptych.tsx
```diff
@@ -125,7 +127,7 @@ const App = (): JSX.Element => {
         <Route exact path="/v2/screen/:id/simulation">
           <MappingContext.Provider value={TYPE_TO_COMPONENT}>
             <ResponseMapperContext.Provider value={responseMapper}>
-              <SimulationScreenPage />
+              <SimulationScreenPage opts={{ alternateView: true }} />
```
jzimbel-mbta, line 130: Just making sure I understand when we want this opt, since I previously removed it for the triptych app (oops)—is the alternate view for any screen type that doesn't have any kind of "flex zone"/paged region? Seems that way since this is also used for DUPs.

jzimbel-mbta (approved):
Looks good, just would prefer to keep the CSS a bit more defensive against overflow, especially since some of that content will be images submitted by third parties.

assets/src/apps/v2/triptych.tsx
cmaddox5: The opt is to indicate that a separate component will handle the rendering of the simulation. For most of our screens, we are ok rendering the full page as a single image. For DUPs, we needed something special in order to meet design requirements. That's why DUPs use the option. 

Now that I'm thinking about it, Triptychs do not need this option right now because we are ok shipping a super simple view of the screen. I added the option for those because I anticipated a special design, but that's not in scope right now. Gonna just use the standard template for now and we can add a custom template in the future if needed.

assets/css/v2/triptych/simulation.scss
cmaddox5: I think I made the changes you are suggesting. Let me know if I misunderstood.

PR #1853, opened 2023-09-05
[tweak] Test all crowding levels (including no data)

hannahpurcell:
[Polish](https://www.notion.so/mbta-downtown-crossing/5-level-crowding-levels-map-to-correct-3-level-crowding-in-app-671ce31242494cc0a1a1100d86aacea6)

Primary goal: verify each possible output from the API is serialized correctly.

- [x] Tests added?


test/screens/v2/widget_instance/train_crowding_test.exs
```diff
@@ -5,6 +5,16 @@ defmodule Screens.V2.WidgetInstance.TrainCrowdingTest do
...
 
+  defp put_crowding_levels(widget, carriages) do
+    %{
+      widget
+      | prediction: %{
+          widget.prediction
+          | vehicle: %{widget.prediction.vehicle | carriages: carriages}
+        }
+    }
+  end
```
jzimbel-mbta, line 16: There's a cool [convenience function](https://hexdocs.pm/elixir/Kernel.html#put_in/2) (secretely a macro 😈) that makes updates to deeply-nested structures like this easier:

```suggestion
  defp put_crowding_levels(widget, carriages) do
    put_in(widget.prediction.vehicle.carriages, carriages)
  end
```

Just an FYI though; this is fine as-is. There are also some related functions/macros that work in a similar way, like `update_in`, `get_and_update_in`. They're all in the `Kernel` module.

jzimbel-mbta (approved):
LGTM!

test/screens/v2/widget_instance/train_crowding_test.exs
hannahpurcell: Oooo NICE! Good to know!

PR #1855, opened 2023-09-07
tweak(triptych): Do not render slots outside the viewport when using the 3-slot split layout

jzimbel-mbta:
**Asana task**: ad hoc

Panes besides the displayed one now only render the identifiers, just in case something has gone really wrong and we're seeing those for some reason. (Plus, they aren't expensive to render)

Screenshot 👇 notice that the `.left-pane` and `.right-pane` elements have no children since they are outside the viewport.
image

- [ ] Tests added?

### DIFF
--- a/assets/src/components/v2/triptych/triptych_three_pane.tsx
+++ b/assets/src/components/v2/triptych/triptych_three_pane.tsx
@@ -1,6 +1,7 @@
 import React from "react";
 
 import Widget, { WidgetData } from "Components/v2/widget";
+import { TriptychPane, getTriptychPane } from "Util/outfront";
 
 interface Props {
   left_pane: WidgetData;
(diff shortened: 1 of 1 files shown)


assets/src/components/v2/triptych/evergreen_content.tsx
```diff
@@ -0,0 +1,51 @@
...
+  triptychPane,
+}) => {
+  const dupReadyUrl = imagePath(assetUrl);
+  const isPlaying = useIsOnScreen();
+  const playerName = usePlayerName();
+  let identifiers = `${TRIPTYCH_VERSION} ${playerName ? playerName : ""}`;
+
+  return (
+    <>
+      {isInViewport(triptychPane) && (
```
cmaddox5, line 41: Could this viewport logic actually live above this in `triptych_three_pane`? We have access to what we need there too so could just make this a layout change instead of a widget change.

jzimbel-mbta: Oh duh, yeah I think it could go there. I would just need to make sure that the middle and left screens still draw in the right positions if the left is not rendered.

Will make this change and then reassign you.

jzimbel-mbta: So much simpler now, thanks for pointing that out.

cmaddox5 (approved):

PR #1859, opened 2023-09-11
tweak: Improve DUP overnight performance pt 1

cmaddox5:
**Asana task**: [[DUP v2] Timebox: Investigate slowness fetching data](https://app.asana.com/0/1185117109217413/1204606315677361/f)

Added a conditional that checks if there is room for overnight rows before running the logic. The logic itself is slow (Part 2 of this task is improving that overall) so we should avoid running it unless we think we need to.

- [ ] Tests added?


jzimbel-mbta (approved):
👍

PR #1860, opened 2023-09-11
fix: nil check in logging conditional

cmaddox5:
**Asana task**: ad-hoc

I forgot to check `next_train_prediction` for `nil` when logging. This is causing a good amount of errors overnight when all predictions are `nil`. 

- [ ] Tests added?

### DIFF
--- a/lib/screens/v2/candidate_generator/widgets/train_crowding.ex
+++ b/lib/screens/v2/candidate_generator/widgets/train_crowding.ex
@@ -56,7 +56,7 @@ defmodule Screens.V2.CandidateGenerator.Widgets.TrainCrowding do
            params |> Map.to_list() |> fetch_alerts_fn.() do
       next_train_prediction = List.first(predictions)
 
-      if next_train_prediction && logging_options.is_real_screen do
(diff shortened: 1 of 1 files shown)


lib/screens/v2/candidate_generator/widgets/train_crowding.ex
```diff
@@ -56,7 +56,7 @@ defmodule Screens.V2.CandidateGenerator.Widgets.TrainCrowding do
            params |> Map.to_list() |> fetch_alerts_fn.() do
       next_train_prediction = List.first(predictions)
 
-      if logging_options.is_real_screen do
+      if not is_nil(next_train_prediction) and logging_options.is_real_screen do
```
PaulJKim, line 59: nit: I personally like `if next_train_prediction && logging_options.is_real_screen do` better for the brevity, but just personal preference. Not sure if using `is_nil/1` is more idiomatic though

cmaddox5: I could've sworn I tried that earlier, but the problem is I was using `and` instead of `&&`... Thanks for the suggestion cause I definitely like this better.

PR #1861, opened 2023-09-13
fix: GL E-Ink DeparturesNoData

cmaddox5:
**Asana task**: [DevOps Checkbox does not work for all screens](https://app.asana.com/0/1185117109217413/1205451640523154/f)

Correct the slot name used in props. After this, all screens are working when devops checkboxes are checked.

**Correction: Looks like gl_eink_single was not set up to handle these checkboxes. Since we still have single live single screen GL E-Ink screens, I added the logic so it doesn't show predictions when it shouldn't.

- [ ] Tests added?


cmaddox5:
I'm really not sure why I didn't take that approach initially. I must have thought the slot names being the same would give the framework a harder time picking the right layout. Just tried it out and that doesn't seem to be the case.

hannahpurcell (approved):
👍 👍

PR #1864, opened 2023-09-14
tweak: Exclude triptychs from self-refresh job; Move unrelated logic out of Screens.Config.State

jzimbel-mbta:
**Asana task**: ad hoc

In preparation for triptychs appearing in Screenplay, this prevents the screens-by-alert self refresh job from wasting effort on triptychs since they never show alerts and are likely to load less frequently than the cache TTL.

👉 I tried to put the list of non-alert-showing screen types in a high visibility spot, since we'll need to keep it manually updated. (I couldn't think of a simple programmatic way to determine whether a given screen can show alert widgets.)
I'm very open to suggestions on this, though. Either a way to make the list even more visible, or a way to automatically determine this screen type trait.

- [ ] Tests added?

### DIFF
--- a/lib/screens/config/state.ex
+++ b/lib/screens/config/state.ex
@@ -174,18 +174,17 @@ defmodule Screens.Config.State do
     {:reply, app_params, state}
   end
 
-  def handle_call({:screen_ids, filter_fn}, _from, %__MODULE__{config: config} = state)
-      when is_function(filter_fn, 1) do
+  def handle_call({:screen_ids, nil}, _from, %__MODULE__{config: config} = state) do
+    {:reply, Map.keys(config.screens), state}
+  end
+
+  def handle_call({:screen_ids, filter_fn}, _from, %__MODULE__{config: config} = state) do
     ids =
       config.screens
       |> Enum.filter(filter_fn)
       |> Enum.map(fn {screen_id, _screen_config} -> screen_id end)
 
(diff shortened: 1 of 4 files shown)


lib/screens/config/state.ex
```diff
@@ -169,20 +174,24 @@ defmodule Screens.Config.State do
     {:reply, app_params, state}
   end
 
-  def handle_call({:mode_disabled?, mode}, _from, %__MODULE__{config: config} = state) do
-    %Devops{disabled_modes: disabled_modes} = config.devops
+  def handle_call({:screen_ids, filter_fn}, _from, %__MODULE__{config: config} = state)
+      when is_function(filter_fn, 1) do
```
PaulJKim, line 178: Nit: since this guard is already applied in the "client" function above, it seems somewhat redundant to have it here too. What do you think of matching explicitly against the default/`nil` case first, and then having this pattern defined below without the guard?

PaulJKim (approved):
One minor comment, but otherwise looks good

lib/screens/config/state.ex
jzimbel-mbta: Oops good catch! Yes, the client function should be responsible for guarding against bad input, not the server handler function.
I think I did this to separate this case from the one where `filter_fn` is `nil`, but I could just swap the two clauses and it would work fine without a guard.

PR #1865, opened 2023-09-14
feat: Basic JSON editor for triptych player mapping in admin tool

jzimbel-mbta:
**Asana task**: [Make triptych player name config editable in the admin tool](https://app.asana.com/0/1185117109217413/1205478591787279/f)

This is nearly a 100% copy of how we implemented the main screens config editor.

I did a better job of factoring out common code on the frontend. There's already a fair amount of abstraction on the backend with the `use ConfigCache` macro, and I didn't want to mess with it or add an extra layer of complexity to reduce duplication between the main screens config cache and the triptych player one.

- [ ] Tests added?


lib/screens/triptych_player/validator.ex
```diff
@@ -0,0 +1,47 @@
...
+
+    case invalid_keys do
+      [] -> :ok
+      keys -> {:error, "Mapping contains non-string keys: #{inspect(keys)}"}
+    end
+  end
+
+  defp validate_screen_id_values(mapping) do
+    triptych_screen_ids = MapSet.new(Map.values(mapping))
+    all_screen_ids = MapSet.new(Screens.Config.State.screen_ids())
```
cmaddox5, line 32: Should this be filtered down to just triptychs so we only look for the difference of the mapping and `triptych_v2` screen IDs? Likely overkill, but could add some confidence that only triptych IDs are in the mapping.

cmaddox5 (commented):
I know this would be a bigger change than it sounds, but could this config be saved as a list for the sake of sorting? Right now this function will lose any ordering you might have added because it is saved as a map. It would be much easier to follow this config if player names that belong to the same ID are next to each other.

jzimbel-mbta:
I like that idea, will see about implementing it. I can store it as a map in the genserver for fast lookups, and serialize it to a list of pairs,
```json
[
  ["BKB-DS-001", "TRI-BackBay-Orange-0-1"],
  ["BKB-DS-002", "TRI-BackBay-Orange-0-1"],
  "..."
]
```
in the JSON for readability.

lib/screens/triptych_player/validator.ex
jzimbel-mbta: Yes, definitely. I added support for passing a filter function to `Config.State.screen_ids` and then didn't use it in my own new code. 🤦

jzimbel-mbta:
I fixed the ordering (everywhere! The two JSON text editors, plus all admin tables) with the last two commits on this PR. Screen configs are consistently sorted alphanumerically by their screen IDs, and triptych player config sorted alphanumerically by player name.
Should be good to go now.

cmaddox5 (approved):
Sweet! Looks great!

PR #1867, opened 2023-09-18
tweak: Include triptych player name in data request logs

jzimbel-mbta:
**Asana task**: ad hoc

This will help us match up logs with real screens, and make sure that the placement of each player matches up with what OFM's station maps said.

- [ ] Tests added?

### DIFF
--- a/lib/screens_web/controllers/v2/screen_api_controller.ex
+++ b/lib/screens_web/controllers/v2/screen_api_controller.ex
@@ -24,6 +24,7 @@ defmodule ScreensWeb.V2.ScreenApiController do
     screen_side = params["screen_side"]
     rotation_index = params["rotation_index"]
     triptych_pane = params["pane"]
+    triptych_player_name = params["player_name"]
     ofm_app_package_version = params["version"]
 
     LogScreenData.log_data_request(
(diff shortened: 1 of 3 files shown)


lib/screens_web/controllers/v2/screen_api_controller.ex
```diff
@@ -24,6 +24,7 @@ defmodule ScreensWeb.V2.ScreenApiController do
     screen_side = params["screen_side"]
     rotation_index = params["rotation_index"]
     triptych_pane = params["pane"]
+    triptych_player_name = params["player_name"]
```
cmaddox5, line 27: If we ever want to log DUP player names, we may want to consider renaming these params on the client. Just a thought, can't do that now.

cmaddox5 (approved):

lib/screens_web/controllers/v2/screen_api_controller.ex
jzimbel-mbta: Oh hmm, we _could_... I could re-key the param over in `show_triptych/2` before passing off to `show/2`.

cmaddox5: Totally up to you. I think it will only be unclear once we add player name to params for DUPs. But I don't see us needing to do that in the near future.

jzimbel-mbta: I'll leave it be for now.

PR #1873, opened 2023-09-27
[fix] Explicitly expire child processes of a Task.Supervisor

hannahpurcell:
Child processes of Task.Supervisors do not expire, even if they are meant to be temporary. We need to explicitly end the process if it's running too long. At the moment, the timeout is set for 10s.

### DIFF
--- a/lib/screens/ol_crowding/dynamic_supervisor.ex
+++ b/lib/screens/ol_crowding/dynamic_supervisor.ex
@@ -48,6 +48,7 @@ defmodule Screens.OlCrowding.DynamicSupervisor do
       restart: :transient
     }
 
-    DynamicSupervisor.start_child(__MODULE__, spec)
+    {:ok, child_pid} = DynamicSupervisor.start_child(__MODULE__, spec)
+    _ = :timer.exit_after(10_000, child_pid, :kill)
   end
 end
--- a/lib/screens/ol_crowding/dynamic_supervisor.ex
+++ b/lib/screens/ol_crowding/dynamic_supervisor.ex
@@ -1,8 +1,10 @@
 defmodule Screens.OlCrowding.DynamicSupervisor do
   @moduledoc false
 
+  require Logger
+
   use DynamicSupervisor
-  alias Screens.OlCrowding.Logger
+  alias Screens.OlCrowding.LogCrowdingInfo
 
   def start_link(init_arg) do
     DynamicSupervisor.start_link(__MODULE__, init_arg, name: __MODULE__)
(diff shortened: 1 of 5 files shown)


cmaddox5:
Should the fallback for the `DynamicSupervisor` we discussed in Slack also be a part of this change? I don't think that fallback can be applied the same way as these, but will be a similar enough change for it to be in scope for this PR.

lib/screens/util.ex
```diff
@@ -218,4 +218,16 @@ defmodule Screens.Util do
...
   def translate_carriage_occupancy_status(_), do: nil
+
+  @doc """
+    Adds a timeout to a function. Mainly used for child processes of a Task.Supervisor
+    which don't come with a timeout by default.
+  """
+  @spec fn_with_timeout((() -> val), non_neg_integer()) :: (() -> val) when val: any()
+  def fn_with_timeout(fun, timeout) do
+    fn ->
+      _ = :timer.exit_after(timeout, :normal)
```
jzimbel-mbta, line 229: As mentioned in slack, it turns out that this _must_ use the `:kill` reason for the desired behavior to happen.
```suggestion
      _ = :timer.exit_after(timeout, :kill)
```

jzimbel-mbta (approved):
Good to go, just need that one change to the exit reason in order to force the process to stop.

hannahpurcell:
> Should the fallback for the `DynamicSupervisor` we discussed in Slack also be a part of this change?

Sure! I lost sight of that once we found the more realistic bug. I'm having trouble thinking through the erlang process driven logic. Would this just be adding a direct call to `:timer.kill_after()` right after the start_child line?

cmaddox5:
> > Should the fallback for the `DynamicSupervisor` we discussed in Slack also be a part of this change?
> 
> @cmaddox5 Sure! I lost sight of that once we found the more realistic bug. I'm having trouble thinking through the erlang process driven logic. Would this just be adding a direct call to `:timer.kill_after()` right after the start_child line?

I think it might be safer to get the `pid` from the return of `start_child` and call `kill_after` on that `pid`. So instead using the timer function before, it would be after in this particular case.

hannahpurcell:
> I think it might be safer to get the `pid` from the return of `start_child`

Ooohhh of course it returns a pid. Ok  ✍️

lib/screens/ol_crowding/dynamic_supervisor.ex
```diff
@@ -48,6 +48,7 @@ defmodule Screens.OlCrowding.DynamicSupervisor do
       restart: :transient
     }
 
-    DynamicSupervisor.start_child(__MODULE__, spec)
+    {:ok, child_pid} = DynamicSupervisor.start_child(__MODULE__, spec)
```
cmaddox5, line 51: In order to avoid crashing the app when one of these fails to start, can this be handled in a `case` instead and if the result is anything but `{:ok, child_pid}` we can just fail silently (or with a log, up to you on that)?

hannahpurcell: mmmm mmhmm 👍 👍

cmaddox5 (approved):

PR #1874, opened 2023-09-28
feat: Phoenix LiveDashboard

cmaddox5:
**Asana task**: ad-hoc

This adds the Phoenix LiveDashboard to the admin endpoint. Will only be accessible by users who are in the `screens-admin` group. You can access it via `/admin/dashboard`. On the page, there are various tabs and panels that show different pieces of info about our server. This includes a `Processes` tab that actually lists out all of the running processes. I also enabled a request logger, but you have to manually enable a cookie to see it work. 

- [ ] Tests added?


jzimbel-mbta (approved):
Nice, this should definitely come in handy.

PR #1878, opened 2023-09-29
fix: Phoenix LiveDashboard

cmaddox5:
**Asana task**: ad-hoc

There was some missing configuration on our `Endpoint` module and it was causing socket connection to fail in deployed environments. After testing these changes in dev-green, the socket connections are good to go.

- [ ] Tests added?

### DIFF
--- a/config/prod.exs
+++ b/config/prod.exs
@@ -12,7 +12,8 @@ import Config
 config :screens, ScreensWeb.Endpoint,
   url: [host: "example.com", port: 80],
   server: true,
-  cache_static_manifest: "priv/static/cache_manifest.json"
+  cache_static_manifest: "priv/static/cache_manifest.json",
+  check_origin: ["https://*.mbta.com", "https://*.mbtace.com"]
 
 # Do not print debug messages in production
 config :logger, level: :info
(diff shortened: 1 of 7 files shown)


config/prod.exs
```diff
@@ -12,7 +12,12 @@ import Config
 config :screens, ScreensWeb.Endpoint,
   url: [host: "example.com", port: 80],
   server: true,
-  cache_static_manifest: "priv/static/cache_manifest.json"
+  cache_static_manifest: "priv/static/cache_manifest.json",
+  check_origin: [
+    "https://screens.mbta.com",
+    "https://screens-dev.mbtace.com",
+    "https://screens-dev-green.mbtace.com"
+  ]
```
jzimbel-mbta, line 20: Slightly nervous about this setting. I looked it up in the docs, and it seems to only apply to websocket connections, but if it messes with CORS settings for regular requests, that will cause problems for the DUP and triptych clients. Just something to keep an eye out for when this goes to prod.

jzimbel-mbta (approved):

config/prod.exs
cmaddox5: I should have mentioned this in the description, but I did make sure to test a DUP package before posting the PR. I didn't see anything off, even with a fake mraid object.

PR #1879, opened 2023-09-29
feat: OL Crowding Heuristics Implementation

cmaddox5:
**Asana task**: [Add logging to triptych app - pt 2](https://app.asana.com/0/1185117109217413/1205353399357052/f)

A lot of changes here are refactoring so that credo doesn't complain about too many parameters or functions having too many nested conditionals. The meat of this PR is in `get_instance_using_dwell_time`. The gist is that we wait for the train to be stopped at the previous station. Once it is stopped, we get the prediction for that particular part of the trip in order to get the predicted dwell time (`departure_time - arrival_time`). When `now` is >= to that time (minus 10 seconds just to give ourselves a little time before the train is supposed to leave), we want to show the widget. The original logic of the train being in transit to the current station will always take precedent. This is only to be used as an extra effort to show the widget a little sooner. 

- [ ] Tests added?


hannahpurcell:
> Once it is stopped, we get the prediction for that particular part of the trip in order to get the predicted dwell time (departure_time - arrival_time). When now is >= to that time (minus 10 seconds just to give ourselves a little time before the train is supposed to leave), we want to show the widget.

Before reviewing, I have questions on the PR description: do we need to calculate the dwell time? It's just the departure time minus 10 seconds.

Also, we don't want to show the widget in this case. We want to log to Splunk to say that "heuristic #1 was met" and include the crowding level at that time. Then we'll make a Splunk dashboard to answer these questions:
- ...for what % of trips does that heuristic log appear BEFORE we'd normally show the widget? 
- ...what % of the time does the crowding level change between that heuristic log when the widget actually shows up?

I remember there being 2 heuristics to evaluate, and it seems like this PR only captures one. The second one is: two back-to-back readings of crowding are identical.

This might be TMI at this point, but rewriting the problem this way helped it click for me. We're trying to figure out if we we can have high confidence crowding predictions earlier than just "once the train has left a station". Can we have high confidence predictions when:
- The train is scheduled to depart a station in 10s? (first heuristic)
- Two back-to-back readings of crowding were identical? (second heuristic)

cmaddox5:
> Before reviewing, I have questions on the PR description: do we need to calculate the dwell time? It's just the departure time minus 10 seconds.

This is definitely a case of me overthinking what we needed... That does seem exactly right so unless I run into problems with that, I will make that change.

> Also, we don't want to show the widget in this case.

Guess I got too excited trying to add the widget to the screen more often 😅 . Easy to fix though.

> I remember there being 2 heuristics to evaluate, and it seems like this PR only captures one. The second one is: two back-to-back readings of crowding are identical.

This reminds me that I had a concern about this heuristic at the time and completely forgot to communicate it. Are these for two consecutive screen refreshes or some other interval to be done in the background? Both I feel pose different challenges. Two consecutive screen refreshes seems like most screens will never have a chance to catch that (although if we are just logging, not really a big issue. Just may not be a heuristic most stations use). A background interval isn't really a problem, but will mean I need to spin up a different background process to track it.

hannahpurcell:
> Are these for two consecutive screen refreshes or some other interval to be done in the background?

Ah, I don't remember. Maybe ask Paul H? We decided during that heuristics meeting and I cannot at all remember which way we decided

cmaddox5:
Added an additional heuristic for consecutive crowing classes. Heuristic logic now live in their own function `pick_heuristic`. Dwell time logic is the same as before (except we look at `departure_time` instead of calculating it). New logic is simply log when two consecutive predictions have the same crowding classes.

hannahpurcell:
> Are these for two consecutive screen refreshes or some other interval to be done in the background?

which one did you end up choosing?

cmaddox5:
> which one did you end up choosing?

2 consecutive screen refreshes. The time-based heuristic will always get priority just so we aren't forever looking for 2 consecutive classes that are the same. Each refresh that still isn't currently past that time will compare to the previous. If it's the same, log it. Otherwise, replace what is in the cache with the new prediction.

cmaddox5:
After our chat in Slack, I agree that these heuristics should be independent. That has been changed.

lib/screens/v2/candidate_generator/widgets/train_crowding.ex
```diff
@@ -63,36 +88,206 @@ defmodule Screens.V2.CandidateGenerator.Widgets.TrainCrowding do
...
+
+    {:ok, predictions} = fetch_predictions_fn.(params)
+    List.first(predictions)
+  end
+
+  defp check_time_based_heuristic(
+         cached_prediction,
+         common_params
+       ) do
+    show_widget_after_dt = DateTime.add(cached_prediction.departure_time, -10)
```
hannahpurcell, line 244: Should maybe call this var `show_widget_before_dt`

lib/screens/v2/candidate_generator/widgets/train_crowding.ex
```diff
@@ -63,36 +88,206 @@ defmodule Screens.V2.CandidateGenerator.Widgets.TrainCrowding do
...
+      trip_id: trip_id
+    }
+
+    {:ok, predictions} = fetch_predictions_fn.(params)
+    List.first(predictions)
+  end
+
+  defp check_time_based_heuristic(
+         cached_prediction,
+         common_params
```
hannahpurcell, line 242: Maybe pass in the 10s hardcoded value to the function as a parameter? Currently, we're only testing 10s, but we might in the future add more logs for earlier and earlier to see where the confidence breaks down.

lib/screens/v2/candidate_generator/widgets/train_crowding.ex
```diff
@@ -63,36 +88,206 @@ defmodule Screens.V2.CandidateGenerator.Widgets.TrainCrowding do
...
+
+    if DateTime.compare(
+         common_params.now,
+         show_widget_after_dt
+       ) in [
+         :eq,
+         :gt
+       ] do
+      log_crowding_info(
+        :dwell,
```
hannahpurcell, line 254: Maybe adjust this to match the "rebrand" of time-based heuristic. Perhaps `:some_time_before_departure`

hannahpurcell (commented):
Some clean-up thoughts, but I think the logic is sound 👍

lib/screens/ol_crowding/dynamic_supervisor.ex
```diff
@@ -50,18 +44,6 @@ defmodule Screens.OlCrowding.DynamicSupervisor do
...
-
-      {:ok, child_pid, _} ->
-        _ = :timer.exit_after(10_000, child_pid, :kill)
-
-      {:error, error} ->
-        Logger.error("crowding_dyn_supervisor_process_error #{inspect(error)}")
-
-      _ ->
-        Logger.warn("Something went wrong with starting the crowding dynamic supervisor process")
-    end
```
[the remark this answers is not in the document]
cmaddox5: Not anymore because the logger now has an internal shutdown mechanism (see [here](https://github.com/mbta/screens/blob/f91f869c60094883b8b0092c3d678fed22e211ca/lib/screens/ol_crowding/logger.ex#L17)). It felt cleaner to have the process shut itself down than have it be shut down from the outside.

lib/screens/v2/candidate_generator/widgets/train_crowding.ex
```diff
@@ -63,36 +88,206 @@ defmodule Screens.V2.CandidateGenerator.Widgets.TrainCrowding do
...
+        previous_station_prediction_current_trip =
+          fetch_previous_station_prediction(
+            train_crowding_config,
+            relevant_platform_id,
+            next_train_prediction.trip.id,
+            common_params.fetch_predictions_fn
+          )
+
+        if is_nil(previous_station_prediction_current_trip) do
+          []
```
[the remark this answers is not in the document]
cmaddox5: Yeah! Good idea.

lib/screens/v2/candidate_generator/widgets/train_crowding.ex
cmaddox5: Well `now` needs to be >= this dt. I added a comment to this var that should help. Let me know if it's still unclear.

lib/screens/v2/candidate_generator/widgets/train_crowding.ex
cmaddox5: Good catch. I went with `:time_based` because that's what I call the heuristic in code. Is that name clear or is it too general?

lib/screens/v2/candidate_generator/widgets/train_crowding.ex
```diff
@@ -63,36 +88,206 @@ defmodule Screens.V2.CandidateGenerator.Widgets.TrainCrowding do
...
+
+    previous_stop_index =
+      Enum.find_index(
+        ol_stop_sequence,
+        &(elem(&1, 0) == train_crowding_config.station_id)
+      ) - 1
+
+    {_, platform_id_tuple} = Enum.at(ol_stop_sequence, previous_stop_index)
+
+    relevant_platform_id = elem(platform_id_tuple, train_crowding_config.direction_id)
```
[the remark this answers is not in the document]
cmaddox5: I tried it out and it gets pretty confusing. The `find_index` function is a little too complicated to map out correctly in a pipe (the formatter was actually not letting me put parentheses where they actually needed to be). The other two lines get confusing because we need two consecutive `elem` calls. I did change the variable name, though.

lib/screens/v2/candidate_generator/widgets/train_crowding.ex
hannahpurcell: I do think this value should be called something else, because it suggests we're logging after departure time, but really we're logging _before_ departure time; that is, 10s before departure time. `now` needs to be later than that time, which is still before the vehicle has even departed. 

TL;DR the value represented by this variable is pre-departure, whether by 10s, 15s, or more.

lib/screens/v2/candidate_generator/widgets/train_crowding.ex
```diff
@@ -239,9 +244,13 @@ defmodule Screens.V2.CandidateGenerator.Widgets.TrainCrowding do
 
   defp check_time_based_heuristic(
          cached_prediction,
-         common_params
+         common_params,
+         previous_departure_time_cushion
        ) do
-    show_widget_after_dt = DateTime.add(cached_prediction.departure_time, -10)
+    # cached_prediction.departure_time minus previous_departure_time_cushion is when we expect crowding to be reliable.
+    # When now >= this time, show the widget.
```
hannahpurcell, line 251: Still refers to showing the widget. Perhaps instead:

```suggestion
    # When now >= this time, log it.
```

lib/screens/v2/candidate_generator/widgets/train_crowding.ex
hannahpurcell: I'm on the fence, but I suppose we can leave it for now (until we add another time based heuristic). But there is another comment referring to dwell time on line 284.

hannahpurcell (approved):
Pending two other tweaks, you should be good to go

PR #1886, opened 2023-10-10
Replace local config code with library config code

PaulJKim:
**Asana task**: [[Permanent Config] Split common code into new repo](https://app.asana.com/0/1185117109217413/1205452334782277/f)

Description

This PR replaces the config code with the same code but from the new `screens-config-lib` library that was set up. The change does not move anything related to the config state modules though, since that will remain as is.


cmaddox5 (commented):
Definitely a beefy PR, but there really isn't much to review. I think CI is the best reviewer of this and it found no issues. One thing I just remembered is that some of our READMEs need to be updated now. One that I found is [here](https://github.com/mbta/screens/blob/27d29d426a0153a3fd807f2300a23643114e8fa0/ARCHITECTURE.md?plain=1#L37). That actually may be the only place, but I might have overlooked one. Do you mind updating that paragraph to document this new way of maintaining configs?

PaulJKim:
Ty for calling that out @cmaddox5 ! I scanned through other docs in the repo and didn't find any more references, but lmk if you think of/find any others

cmaddox5 (approved):
Looks good!

PR #1888, opened 2023-10-17
Cm/crowding app wording updates

cmaddox5:
**Asana task**: [Make text changes on crowding app](https://app.asana.com/0/1185117109217413/1205701336219427/f)

Made wording changes to plain JS app and updated React app so it matches the plain JS.

- [ ] Tests added?


hannahpurcell (approved):
Ah nice, thanks for fixing for both the app versions

jzimbel-mbta (approved):

PR #1893, opened 2023-10-24
Chore: Update workflow to v2, delete beanstalk workflows

krisrjohnson21:
#### Summary of changes
This PR updates the deploy workflow to use v2 of the `deploy-ecs` workflow. This updated workflow assumes an IAM role instead of relying on long-lasting AWS keys for an IAM user

This PR also deletes two decommissioned workflows for Elastic Beanstalk dev and dev-green environments. Flagging explicitly for team approval.

Successful deploy [here](https://github.com/mbta/screens/actions/runs/6633617416)


ianwestcott (approved):

cmaddox5 (approved):
Awesome! Thanks for the help!

PR #1897, opened 2023-10-26
Fixed D branch headsign.

cmaddox5:
**Asana task**: [[frontend] Terminal station closure causes JS errors](https://www.notion.so/mbta-downtown-crossing/frontend-Terminal-station-closure-causes-JS-errors-7f00d9f3653d43d4a048c2be61328044?pvs=4)

This ended up being a server issue. When building the banner, we use the `North Station & North` icon in the banner but the headsign was only `North Station`. Because the key was missing, the svg filename was `nil`.

- [ ] Tests added?


hannahpurcell (approved):
👍

PR #1898, opened 2023-10-26
Cm/govt ctr gl affected pill fix

cmaddox5:
**Notion task**: [[alert, backend] Wrong affected route/line for station closure including Gov Ctr](https://www.notion.so/mbta-downtown-crossing/alert-backend-Wrong-affected-route-line-for-station-closure-including-Gov-Ctr-7e306485d1dc4cf7882cc0f51a2640b1?pvs=4)

When showing affected route pills, we only ever get the first in the list because that's all we have room for. Change it so if there's more than one GL branch in the list, we just respond with the GL pills and not 4 branch pills.

- [ ] Tests added?

### DIFF
--- a/lib/screens/v2/widget_instance/reconstructed_alert.ex
+++ b/lib/screens/v2/widget_instance/reconstructed_alert.ex
@@ -133,7 +133,9 @@ defmodule Screens.V2.WidgetInstance.ReconstructedAlert do
       if is_nil(route_id) do
         Alert.informed_entities(alert)
       else
-        alert |> Alert.informed_entities() |> Enum.filter(&(&1.route == route_id))
+        alert
+        |> Alert.informed_entities()
+        |> Enum.filter(&String.starts_with?(&1.route, route_id))
       end
 
     {direction_id, route_id} =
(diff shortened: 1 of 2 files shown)


lib/screens/v2/widget_instance/reconstructed_alert.ex
```diff
@@ -181,15 +181,24 @@ defmodule Screens.V2.WidgetInstance.ReconstructedAlert do
...
+      |> Enum.filter(&(&1.route_type in [0, 1] and &1.route in routes_at_stop))
+      |> Enum.group_by(fn %{route: route} -> route end)
+      |> Enum.flat_map(fn
+        {route_id, _} ->
+          headsign = get_destination(t, location, route_id)
+          build_pills_from_headsign(route_id, headsign)
+      end)
+      |> Enum.uniq()
+
+    if Enum.count(pills, & &1.route_id) > 1 do
```
PaulJKim, line 195: not 100% sure, but wouldn't you want to do a `uniq_by` by route_id first and then just do `Enum.count/1` instead here?

cmaddox5: I'm not really sure what I was thinking yesterday, but this conditional was way too aggressive (and also nonsensical considering `Enum.count/1` would accomplish the exact same thing `Enum.count(pills, & &1.route_id)` is trying to do 😬). I modified the conditional a bit so it makes sure that the list has GL in it before replacing whatever is in it with a GL pill. 

Basically in the case of a station closure at Gov't Ctr (which I believe is the only screen we have where this issue is possible right now), `pills` has 4 items here that equal `%{route_id: "Green", svg_name: "gl-{BRANCH}"}`. Normally, `uniq_by` would work, but we need to make sure that the `svg_name` is changed in the case that more than one GL item exists.

cmaddox5: Ok, made one more change. I got `get_destination` and `get_route_pills` working using `affected_routes` instead of `informed_entities`. A big underlying issue with the original logic is `informed_entities` don't list a route as `Green` if all branches are affected. Instead, it creates 4 entities for each branch. `LocalizedAlert.consolidated_informed_subway_routes` does the consolidation we need here, but it wasn't being used in this case. By using that, we can now expect this GL consolidation to happen before we create the pills. This makes it easier to know if we are creating 1-3 branch pills or 1 GL pill.

PaulJKim (approved):

PR #1905, opened 2023-10-27
Tweaked resizer logic so it maxHeight matches the card height on page.

cmaddox5:
**Asana task**: [[alert, frontend] “Use shuttle bus” + accessibility notice layout don’t match designs](https://www.notion.so/mbta-downtown-crossing/alert-frontend-Use-shuttle-bus-accessibility-notice-layout-don-t-match-designs-1c3bbf27b5884948a1492c7d833eed70?pvs=4)

This was a text resizer issue. `maxHeight` was hardcoded, but we know enough about the rest of the page to determine exactly what it needs to be.

- [ ] Tests added?


assets/src/components/v2/pre_fare_single_screen_alert.tsx
```diff
@@ -40,11 +45,19 @@ const standardLayout = (
...
   location: string | null,
+  bannerHeight: number,
   disruptionDiagram?: DisruptionDiagramData
 ) => {
+  const maxTextHeight =
+    SCREEN_HEIGHT -
+    FOOTER_HEIGHT -
+    BOTTOM_MARGIN -
+    ALERT_CARD_PADDING -
+    bannerHeight;
```
jzimbel-mbta, line 56: For some reason I feel like I would understand this better as `SCREEN_HEIGHT - (FOOTER_HEIGHT + BOTTOM_MARGIN + ALERT_CARD_PADDING + bannerHeight)`, but that might be just me

jzimbel-mbta (approved):
Nice. Good to go after addressing merge conflicts 👍

assets/src/components/v2/pre_fare_single_screen_alert.tsx
cmaddox5: I agree, that does make more sense. Bundled that change in with the merge.

PR #1909, opened 2023-10-30
feature: New widget endpoint for einks

hannahpurcell:
**Asana task**: [[Mercury] Finish widget endpoint](https://app.asana.com/0/1185117109217413/1205234924224431) 

- [ ] Tests added?


lib/screens_web/controllers/v2/screen_controller.ex
```diff
@@ -76,6 +76,16 @@ defmodule ScreensWeb.V2.ScreenController do
     |> render("index_multi.html")
   end
 
+  def widget(conn, %{"id" => app_id} = params)
+      when app_id in @app_id_strings do
+    app_id = String.to_existing_atom(app_id)
+
+    conn
+    |> assign(:app_id, app_id)
+    |> assign(:widget_data, if(params["widget"], do: Poison.encode!(params["widget"]), else: nil))
```
jzimbel-mbta, line 85: We already use the `Jason` library for JSON parsing/encoding elsewhere in the application. It's considered to be generally better performance-wise than `Poison`, and we don't need duplicate JSON libraries, so let's stick with that.
```suggestion
    |> assign(:widget_data, if(params["widget"], do: Jason.encode!(params["widget"]), else: nil))
```

jzimbel-mbta, line 85: Separately, should we even send a 200 OK response if the `"widget"` key is missing or false/nil? Consider sending a 400 Bad Request instead since this endpoint requires that key in order to do its job.

That can be done like so: (based off of [this](https://github.com/mbta/screens/blob/master/lib/screens_web/controllers/v2/screen_api_controller.ex#L169-L173))
```ex
conn
|> put_status(:bad_request)
|> text("POST /{screen_id}/widget request must contain a JSON body with `widget` key")
```

mix.exs
```diff
@@ -81,7 +81,8 @@ defmodule Screens.MixProject do
       {:memcachex, "~> 0.5.5"},
       {:telemetry_poller, "~> 0.4"},
       {:telemetry_metrics, "~> 0.4"},
-      {:screens_config, git: "https://github.com/mbta/screens-config-lib.git"}
+      {:screens_config, git: "https://github.com/mbta/screens-config-lib.git"},
+      {:poison, "~> 4.0"}
```
jzimbel-mbta, line 85: Please revert this after switching to `Jason`

mix.lock
jzimbel-mbta, line 1: Please revert this

lib/screens_web/controllers/v2/screen_controller.ex
```diff
@@ -76,6 +76,16 @@ defmodule ScreensWeb.V2.ScreenController do
     |> render("index_multi.html")
   end
 
+  def widget(conn, %{"id" => app_id} = params)
```
jzimbel-mbta, line 79: Need to move this def either down or up in the file so that it doesn't come between the clauses of `index/2` (per the compile error message in most recent CI run)

lib/screens_web/controllers/v2/screen_controller.ex
jzimbel-mbta: I checked—setting the `widget_data` assign to `nil` would cause `data-widget-data="<%= @widget_data %>"` in the view template to evaluate to `data-widget-data=""`. On the frontend, our `JSON.parse(getDatasetValue("widgetData"))` would throw an error for that:

```js
> JSON.parse("")
// => Uncaught SyntaxError: JSON.parse: unexpected end of data at line 1 column 1 of the JSON data
```

jzimbel-mbta (commented):
Just 2 quick things:
1. Undo the inclusion of the `Poison` package and use the existing `Jason` package
2. Consider sending a non-200 response when we don't receive expected data in a widget page request

lib/screens_web/controllers/v2/screen_controller.ex
hannahpurcell: OHHHhh thank you

hannahpurcell: Ahhhhh much improved! 🙏

lib/screens_web/router.ex
```diff
@@ -84,6 +84,9 @@ defmodule ScreensWeb.Router do
 
       get "/:id", ScreenController, :index
       get "/:id/simulation", ScreenController, :simulation
+      post "/:id/widget", ScreenController, :widget
+      # Get not needed for application, but for testing in the browser w/o the Mercury skin
+      get "/:id/widget", ScreenController, :widget
```
jzimbel-mbta, line 89: Oh I just thought of something, does the GET version of this endpoint work? Because the controller function expects `widget` to be an actual map, not a JSON string.

Phoenix has a convenience feature where it automatically decodes JSON in a POST request's body, which is why the controller function works properly for those. But I'm not sure if it does the same for JSON in a GET request's url params.

If the controller function receives a JSON string for `widget`, then `Jason.encode!(widget_data)` will double-encode it and make the frontend code fail.

```ex
iex> json_string = "{\"foo\": \"bar\"}"
iex> Jason.encode!(json_string)
"\"{\\\"foo\\\": \\\"bar\\\"}\""
```
😱

hannahpurcell: Hmm ok. This get isn't really supposed to work on its own, but was only needed for getting the proper csrf token needed to do other testing in postman. And when the widget endpoint handler used to have the line 
```
    |> assign(:widget_data, if(params["widget"], do: Jason.encode!(params["widget"]), else: nil))
```
then the get request "succeeded" and provided the csrf header needed to set up the post requests. 

I guess I'll just remove this since it won't be useful for anyone in the future

jzimbel-mbta (approved):
🐬

PR #1912, opened 2023-11-03
feat: Consistent description of disrupted stop ranges throughout pre-fare alerts

jzimbel-mbta:
**Notion task**: [[alert, frontend] Inconsistency? With affected stop range description](https://www.notion.so/mbta-downtown-crossing/alert-frontend-Inconsistency-With-affected-stop-range-description-9571590061054c859e4b0d0d3f95555b?pvs=4)

I ended up making a relatively narrow change to `get_endpoints` for this, instead of actually changing the widget struct's data directly.
The idea is that this is presentational in the same way as formatting a DateTime is. The logic can still work under the assumptions that all stop sequences are in direction ID 0 order, and we only flip it, as needed, when "presenting" it in the serialized response.

- [x] Tests added?

### DIFF
--- a/lib/screens/alerts/informed_entity.ex
+++ b/lib/screens/alerts/informed_entity.ex
@@ -24,4 +24,9 @@ defmodule Screens.Alerts.InformedEntity do
       ie
     )
   end
+
+  @spec parent_station?(t()) :: boolean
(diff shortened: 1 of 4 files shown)


cmaddox5 (approved):
🚀

PR #1918, opened 2023-11-13
fix: Skip CSRF protection on /widget POSTs since they really fetch a new HTML page

jzimbel-mbta:
**Asana task**: ad hoc

Phoenix normally expects a CSRF token with any POST requests to an endpoint that uses the `:browser` pipeline.
The widget endpoint is an exception, since it's basically an HTML page GET request, but we can't pass a body with a GET request.

This removes CSRF protection from the widget endpoint.

---

I also got the GET version of the endpoint working again with some fiddling! Wasn't sure how to directly test it otherwise, besides manually inspecting HTML received in Postman.

You can test the GET version with URLs like so:
```
localhost:4000/v2/widget/gl_eink_v2?widget={%22mode_cost%22:%22$2.40%22,%22mode_icon%22:%22subway-negative-black.svg%22,%22mode_text%22:%22Subway%22,%22text%22:%22For%20real-time%20predictions%20and%20fare%20purchase%20locations:%22,%22type%22:%22fare_info_footer%22,%22url%22:%22mbta.com/stops/place-hsmnl%22}
```

(Use the JS built-in `encodeURI` function to encode a JSON string for inclusion in the url.)

image

- [ ] Tests added?

### DIFF
--- a/lib/screens_web/controllers/v2/screen_controller.ex
+++ b/lib/screens_web/controllers/v2/screen_controller.ex
@@ -118,7 +118,25 @@ defmodule ScreensWeb.V2.ScreenController do
     render_not_found(conn)
   end
 
-  def widget(conn, %{"id" => app_id, "widget" => widget_data})
+  # Handles widget page GET requests with widget data as a query param.
+  # Phoenix does not automatically decode JSON received in query params.
+  def widget(conn, %{"app_id" => app_id, "widget" => json_data}) when is_binary(json_data) do
+    case Jason.decode(json_data) do
+      {:ok, widget_data} ->
+        widget(conn, %{"app_id" => app_id, "widget" => widget_data})
+
+      {:error, _} ->
+        conn
+        |> put_status(:bad_request)
+        |> text(
+          "GET /v2/widget/#{app_id} request must contain a `widget` query param containing JSON"
+        )
+    end
+  end
+
+  # Handles widget page POST requests with widget data as a JSON request body.
+  # Phoenix automatically decodes JSON received in POST body.
+  def widget(conn, %{"app_id" => app_id, "widget" => widget_data})
       when app_id in @app_id_strings do
     app_id = String.to_existing_atom(app_id)
 
@@ -128,12 +146,12 @@ defmodule ScreensWeb.V2.ScreenController do
     |> render("index_widget.html")
   end
 
-  def widget(conn, %{"id" => app_id}) do
+  def widget(conn, %{"app_id" => app_id}) do
     app_id = String.to_existing_atom(app_id)
 
     conn
     |> put_status(:bad_request)
-    |> text("POST /#{app_id}/widget request must contain a JSON body with `widget` key")
+    |> text("POST /v2/widget/#{app_id} request must contain a JSON body with `widget` key")
   end
 
   def simulation(conn, params) do
--- a/lib/screens_web/router.ex
+++ b/lib/screens_web/router.ex
@@ -9,6 +9,12 @@ defmodule ScreensWeb.Router do
     plug :protect_from_forgery
   end
 
+  pipeline :browser_no_csrf do
+    plug :accepts, ["html"]
+    plug :fetch_session
+    plug :fetch_flash
+  end
+
   pipeline :api do
     plug :accepts, ["json"]
   end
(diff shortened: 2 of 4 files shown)


lib/screens_web/router.ex
```diff
@@ -9,6 +9,12 @@ defmodule ScreensWeb.Router do
     plug :protect_from_forgery
   end
 
+  pipeline :browser_no_csrf do
+    plug :accepts, ["html"]
+    plug :fetch_session
+    plug :fetch_flash
+  end
```
jzimbel-mbta, line 16: Would it make sense to build the `:browser` pipeline on top of this one, since they share everything but CSRF? I wasn't sure how to do this elegantly...

lib/screens_web/router.ex
```diff
@@ -79,12 +85,17 @@ defmodule ScreensWeb.Router do
   end
 
   scope "/v2", ScreensWeb.V2 do
+    scope "/widget" do
+      pipe_through [:redirect_prod_http, :browser_no_csrf]
+      post "/:app_id", ScreenController, :widget
+      get "/:app_id", ScreenController, :widget
+    end
```
jzimbel-mbta, line 92: I needed to put the endpoint in a separate scope in order to apply a different set of plugs/pipelines to it.
This was the best naming I could come up with; I'd be happy to go with a different one!

jzimbel-mbta (commented):
Leaving a couple of thoughts

lib/screens_web/router.ex
hannahpurcell: Yes, that makes sense! I'd probably do:

```
  pipeline :browser do
    plug :accepts, ["html"]
    plug :fetch_session
    plug :fetch_flash
  end

  pipeline :csrf_protection do
    plug :protect_from_forgery
  end
```

and then for the appropriate scopes, you'd do
- `pipe_through [:redirect_prod_http, :browser]` or
- `pipe_through [:redirect_prod_http, :browser, :csrf_protection]`

lib/screens_web/controllers/v2/screen_controller.ex
```diff
@@ -118,7 +118,25 @@ defmodule ScreensWeb.V2.ScreenController do
     render_not_found(conn)
   end
 
-  def widget(conn, %{"id" => app_id, "widget" => widget_data})
+  # Handles widget page GET requests with widget data as a query param.
+  # Phoenix does not automatically decode JSON received in query params.
+  def widget(conn, %{"app_id" => app_id, "widget" => json_data}) when is_binary(json_data) do
```
hannahpurcell, line 123: Why do we need both a GET and a POST? Should I have used a GET to begin with?

jzimbel-mbta: Good question honestly, I'm not sure. GET _is_ more standard for requesting a webpage, and with `encodeURI` it's not too difficult to plop the JSON into the URL.

The only real concern I can think of is about URL length. I only tested the GET endpoint with one of our simplest widgets, the e-ink footer. From some stackoverflow searching, it looks like the maximum URL length is around 8000 characters, so I think we'd be fine, but it's still a slight concern.

I'd be happy to remove the POST endpoint if you think that makes sense!
Or we could do something like have them load our widget page in an iframe, and they supply the widget data strictly client-side. This would be similar to how we're able to get OFM tags using `parent.mraid`. We could do `parent.widgetData`. We're probably past the point where it's reasonable to consider new ideas, though—the current approach is more than good enough.

lib/screens_web/router.ex
jzimbel-mbta: This makes sense but I'm a little worried that it now requires extra work to do the default thing. "Secure by default" and all that.

How about this?
```ex
pipeline :browser_no_csrf do
  plug :accepts, ["html"]
  plug :fetch_session
  plug :fetch_flash
end

pipeline :browser do
  plug :browser_no_csrf
  plug :protect_from_forgery
end
```
I just checked, and you can reuse one pipeline inside another without issues!

hannahpurcell: Sure! Works for me

lib/screens_web/controllers/v2/screen_controller.ex
jzimbel-mbta: Yeah, that works.

PR #1923, opened 2023-11-21
Suppress SubwayStatus at MFA and Heath St for Surge.

cmaddox5:
**Asana task**: ad-hoc

Suppressing SubwayStatus at MFA and Heath St for Surge.

- [ ] Tests added?


jzimbel-mbta (approved):
Looks good, just need to remember to revert this afterward.

I imagine we could have a task to comb through recent commits and pick out the GL surge ones--the alternative would be to manually record these changes in notion or a slack canvas or something.

PR #1927, opened 2023-11-22
Cm/new evergreen schedule module

cmaddox5:
**Asana task**: ad-hoc

Added business logic needed to parse new schedule module. See comments for details. Tl;dr is the new module allows for recurrent schedules instead of the continuous schedule that we have currently. This makes scheduling nightly content during the surge MUCH easier.

- [x] Tests added?


jzimbel-mbta (approved):
Not sure if I can review this but.. LGTM! 😃👍

PR #1932, opened 2023-11-27
Add a set of modules for generic ETS-table-based cache, and switch all GenServer-based caches to use it

jzimbel-mbta:
**Asana task**: [Try an ETS table for screens config cache](https://app.asana.com/0/1185117109217413/1205272117453438/f)

# 👇 Use these draft PRs to assist with review!

This PR is huge. These 3 PRs split the changes into new files, changed files, and deleted files.

#1933: Implement new ETS table caches
#1934: Update existing code to use the new caches
#1935: Delete now-unused old cache implementation

<details>
<summary><h2>:warning: Weird Erlang Stuff Alert</h2></summary>

---

[`:ets.match/2`](https://www.erlang.org/doc/man/ets#match-2) and [`:ets.select/2`](https://www.erlang.org/doc/man/ets#select-2) use these weird-as-heck "match specification" data structures to grab specific parts of a table entry—if a quick explanation of how these work would help, just ask. At a high level, they're analogous to pattern matching. They're as close as Erlang could get to being able to pass a pattern as a function argument (you can't actually do that, for reasons).
`:_` is equivalent to `_` in a normal pattern; `:"$1"`, `:"$2"` etc are equivalent to binding a part of the structure to a variable.

```ex
table_name = :screens_config
screen_id = "PRE-000"

table_as_map = %{
  "EIG-000" => %{...},
  "PRE-000" => %{
    foo: :some_value_i_dont_care_about,
    bar: :value_i_want
  }
}

# This:
[[bar]] = :ets.match(table_name, {screen_id, %{foo: :_, bar: :"$1"}})

# is analogous to this:
%{^screen_id => %{foo: _, bar: bar}} = table_as_map
```

---
</details

Remaining work:
- [x] Re-implement `Screens.SignsUiConfig.State` as `Screens.SignsUiConfig.Cache`
- [x] Re-implement `Screens.TriptychPlayer.State` as `Screens.TriptychPlayer.Cache`
- [x] Replace all references to the old GenServer-based modules with the new ETS-based ones
- [x] Delete the old GenServer-based modules
- [x] Appease Dialyzer
- [x] Appease Credo
- [x] Do an informal server speed test with old vs new caching
- [x] Split PR into one with new modules and another with deleted/changed modules
- [ ] Test??

&nbsp;
- [ ] Tests added?


lib/screens/config/cache.ex
```diff
@@ -0,0 +1,180 @@
...
+
+  def last_deploy_timestamp do
+    with_table do
+      [[last_deploy_timestamp]] = :ets.match(@table, {:last_deploy_timestamp, :"$1"})
+      last_deploy_timestamp
+    end
+  end
+
+  def service_level(screen_id) do
+    with_table default: 1 do
```
cmaddox5, line 54: With this macro being called with a default of 1, is there a particular reason the `case` is needed? Is it possible for the `[]` case to match?

cmaddox5: Actually, I think I understand. Is the `[]` case there just in case the table does exist but the `screen_id` doesn't?

lib/screens/config/cache.ex
```diff
@@ -0,0 +1,180 @@
...
+      @table
+      |> :ets.select(match_spec)
+      |> Map.new()
+    end
+  end
+
+  @doc """
+  Gets the entire config struct.
+
+  👉 WARNING: This function is expensive to run and returns a large amount of data.
```
cmaddox5, line 167: Does this make performance of the admin tool worse, or is this function about as expensive as the `GenServer` implementation that this PR replaces?

lib/screens/config/cache.ex
```diff
@@ -0,0 +1,180 @@
...
+    end
+  end
+
+  @doc """
+  Returns a list of all screen IDs that satisfy the given filter.
+  The filter function will be passed a tuple of {screen_id, screen_config} and should return true if that screen ID should be included in the results.
+  """
+  def screen_ids(filter_fn) do
+    with_table default: [] do
+      filter_reducer = fn
```
cmaddox5, line 119: I don't think there's anything wrong with using `foldl` here, but the function I mentioned in another comment, `fun2ms`, might reduce the amount of code needed here. Unless there is a substantial performance impact, totally up to you which you use.

cmaddox5 (commented):
Woof, I was very rusty when it comes to macros. I think I have a good grasp now and left some comments. 

Also, I know that the randomizer picked me to review, but this would be a good PR for the rest of the devs to look at. Maybe not for a super in depth review, but the early visibility would be useful.

lib/screens/config/cache.ex
jzimbel-mbta: Yup! There are two different failure modes for most of these functions that read from the cache.
1. Table doesn't exist (shouldn't happen, but...)
2. Table exists but no match is found for the given pattern (a bit more likely to happen)

jzimbel-mbta: The `:default` value passed to `with_table`, or its fallback behavior of returning `{:error, "table #{x} doesn't exist"} handles (1), and the `[]` case handles (2).

lib/screens/config/cache.ex
```diff
@@ -0,0 +1,180 @@
...
+      end
+    end
+  end
+
+  @doc """
+  Returns a list of all screen IDs.
+  """
+  def screen_ids do
+    with_table default: [] do
+      # Not sure how to avoid matching against the struct this way.
```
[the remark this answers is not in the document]
jzimbel-mbta: Oh my bad, I fixed this just before posting for review and forgot to remove this comment.

Table entries containing screen configs now look like `{:screen, screen_id}, screen_config}` so they're easy to find.
Previously it was just `{screen_id, screen_config}` so I had to find them by checking if the key was a string, or if the value was a `%Screen{}` struct.

I'll delete this comment.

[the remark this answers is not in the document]
jzimbel-mbta: I was wary of using `fun2ms` since it seems to be geared toward erlang code specifically. It uses erlang's version of macros ("parse transforms"). I'm not sure if it would work correctly with elixir code.

lib/screens/config/cache.ex
jzimbel-mbta: I think `fun2ms` imposes the same limitations on what kinds of checks/guards you can do, since it translates `fun` to a match spec. `fun` can't call an arbitrary `filter_fn`, it can only use `is_*` guards.

From the docs:
> The fun is very restricted, it can take only a single parameter (the object to match): a sole variable or a tuple. It must use the is_ guard tests. Language constructs that have no representation in a match specification (if, case, receive, and so on) are not allowed.

jzimbel-mbta:
I can share speed stats in a bit but the informal ones I did aren't especially exciting.
Overall performance with the current app + current config was about the same.

The main advantage of this new approach is that performance of screen data requests, and other logic related to a single screen, should no longer be affected by
- the total size of the config file,
- the number of screens we have registered, or
- how many other data requests are being handled concurrently,

at all.

Also, editing the config file will no longer cause cache reads to be blocked for a bit while the cache updates its contents.

I'll try to confirm this scaling-related performance improvement by giving the server an outrageously huge config file and doing a lot of simultaneous data requests. The ETS version should handle it pretty gracefully, the GenServer one should start struggling.

cmaddox5 (approved):
Thanks for the great answers! I gave this another look and don't have any suggestions. 🚀

lib/screens/config/cache.ex
```diff
@@ -0,0 +1,176 @@
...
+  use Screens.Cache.Client, table: :screens_config
+
+  @type table_contents :: list(table_entry)
+
+  @type table_entry ::
+          {{:screen, screen_id :: String.t()}, ScreensConfig.Screen.t()}
+          | {:last_deploy_timestamp, DateTime.t()}
+          | {:devops, ScreensConfig.Devops.t()}
+
+  def ok?, do: table_exists?()
```
hannahpurcell, line 18: 👀 where is this used?

PR #1940, opened 2023-12-07
tweak/fix: Remove `ordered: false` from Task.async_stream calls to avoid process leak

jzimbel-mbta:
**Asana task**: ad hoc

elixir-lang/elixir#11760

We're on elixir 1.13 in our deployment environments, and it seems to still suffer from this bug.

Removing `ordered: false` shouldn't make a difference here--the number of tasks spawned is small and the stream is immediately passed to `Enum.flat_map`, which needs to get the full list of resulting widget instance structs before it can return anything anyway.

- [ ] Tests added?


cmaddox5 (approved):
🤞

PR #1941, opened 2023-12-08
fix: PreFare Polish - Inside shuttle

hannahpurcell:
[Notion](https://www.notion.so/mbta-downtown-crossing/fd75b83b5904405fbc71ffbcdd0cec47?v=d4971a59c7eb45ae9841e522c53ad368&p=c8510e7dd7774113b21dfe280f8b8dbb&pm=s)

- "You Are Here" diamond becomes red octagon if affected
- Gap between map and issue text is reduced to exactly 32px for all layouts where map is below issue text
- “All shuttles buses are accessible” is regular font weight for two-screen takeover, not for single-pane

Edit: Actually, this PR covered many issues for other other notion tasks. Including links and screenshots.

[Fixed emphasis margin and "you are here" icon](https://www.notion.so/mbta-downtown-crossing/fd75b83b5904405fbc71ffbcdd0cec47?v=d4971a59c7eb45ae9841e522c53ad368&p=1a5456f07a5845269d2ed7cb1d40ad0b&pm=s)
Screenshot 2023-12-11 at 1 49 34 PM


assets/src/components/v2/disruption_diagram/disruption_diagram.tsx
```diff
@@ -379,7 +378,7 @@ const MiddleSlotComponent: ComponentType<MiddleSlotComponentProps> = ({
   let icon;
   if (slot.show_symbol) {
     if (isCurrentStop) {
-      if (isAffected && effect in ["station_closure", "suspension"]) {
+      if (isAffected) {
```
jzimbel-mbta, line 381: Oh good thing this was changed, cuz that previous condition didn't work the way we wanted it to—`x in obj` in JS checks if `x` is in the keys of `obj`. So `x in ["station_closure", "suspension"]` is only true if `x` is 0, 1, "length", "map", etc. Fields of the array object, not their values. JS is weird!

To get the behavior we were looking for, it would have been `["station_closure", "suspension"].includes(effect)`.

assets/src/components/v2/disruption_diagram/disruption_diagram.tsx
```diff
@@ -764,8 +758,9 @@ const DisruptionDiagram: ComponentType<DisruptionDiagramData> = (props) => {
           <g
             id="alert-emphasis"
             transform={`translate(0, ${
-              EMPHASIS_HEIGHT / 2 +
-              (EMPHASIS_HEIGHT - MAX_ICON_HEIGHT) * scaleFactor
+              40 // Half the height of the emphasis icon
+              + 24 * scaleFactor // Half the height of the largest icon, "you are here" octagon
+              + 8 * scaleFactor // Emphasis padding
```
jzimbel-mbta, line 763: Any reason for switching to number literals here? It looks like the `EMPHASIS_HEIGHT` and `MAX_ICON_HEIGHT` constants are still defined.

Oh, it looks like the new constant values are slightly different from this now. EMPHASIS_HEIGHT (80) - MAX_ICON_HEIGHT (64) = 16, not 24. So is that why, maybe?

(The added comments do help a lot, either way)

assets/src/components/v2/pre_fare_single_screen_alert.tsx
```diff
@@ -162,14 +162,14 @@ const MultiLineLayout: React.ComponentType<MultiLineLayoutProps> = ({
   );
 };
 
-interface FalloutLayoutProps {
+interface FallbackLayoutProps {
```
jzimbel-mbta, line 165: ☢️😁

jzimbel-mbta (approved):
Truly awed and humbled by the math that's going on here. I can't say I completely follow all of the steps involved in scaling/positioning everything, but I don't see any obvious issues!

🌈 :shipit:

assets/src/components/v2/disruption_diagram/disruption_diagram.tsx
hannahpurcell: Double thumbs-up. Yeah, I think I originally was the one to add it, so I'm glad I got to fix it and realize the mistake. 👍 👍

assets/src/components/v2/disruption_diagram/disruption_diagram.tsx
hannahpurcell: Good point... I started realizing that the things I thought were constant actually varied in different situations 😓  So the max icon height for a diamond _endpoint_ is 64, while the max icon height for a diamond in the _middle slot_ is 48, so I started diverging from the constants..... While the comments are helpful, that is still a bit confusing if someone were to look at this down the road. So I'll try to clean it up a bit

PR #1946, opened 2023-12-19
fix: Pre-fare polish -- Text only resizes for bypassed stations

hannahpurcell:
[Notion](https://www.notion.so/mbta-downtown-crossing/fd75b83b5904405fbc71ffbcdd0cec47?v=d4971a59c7eb45ae9841e522c53ad368&p=13aa9bc00369477387eea57175a34294&pm=s)

During implementation, it was assumed that all alerts would need to have font resized, but it seems that really it's only explicitly designed for station closures. All other fonts are static.

Also, it turns out the max height for the text is also [defined in the design](https://www.figma.com/file/l4AZl8UXyKLAFI8OtultQk/Pre-Fare-%E2%80%94-Alerts-V2?type=design&node-id=558-14241&mode=dev), and we don't need to do all those funky calculations with bannerHeight, etc.

### DIFF
--- a/assets/src/components/v2/pre_fare_single_screen_alert.tsx
+++ b/assets/src/components/v2/pre_fare_single_screen_alert.tsx
@@ -13,17 +13,6 @@ import ISAIcon from "../../../static/images/svgr_bundled/isa.svg";
 import WalkingIcon from "../../../static/images/svgr_bundled/nearby.svg";
 import ShuttleBusIcon from "../../../static/images/svgr_bundled/bus.svg";
 
-const SCREEN_HEIGHT = 1720,
-  // The footer has effect type and updated time
-  FOOTER_HEIGHT = 84,
-  // Color margin at the very edge of the alert
-  BOTTOM_MARGIN = 32
-
-const getBannerPadding = (effect: string) => {
-  if (effect === "shuttle") return 80
-  else return 120
-}
-
 interface PreFareSingleScreenAlertProps {
   issue: string;
   location: string;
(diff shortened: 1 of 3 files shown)


assets/src/components/v2/pre_fare_single_screen_alert.tsx
```diff
@@ -42,43 +37,44 @@ interface StandardLayoutProps {
...
+      <div ref={contentBlockRef}>
+        <StandardIssueSection
+          issue={issue}
+          location={location}
+          contentTextSize={effect === "station_closure" ? contentTextSize : "large"}
+        />
+        <RemedySection
+          effect={effect}
+          remedy={remedy}
+          contentTextSize="large"
```
cmaddox5, line 75: Since we know what size each remedy will use, can we remove this prop and just delegate the hardcoding to the component? Would centralize the different hardcoded values into a single component.

assets/src/components/v2/pre_fare_single_screen_alert.tsx
```diff
@@ -42,43 +37,44 @@ interface StandardLayoutProps {
...
-        effect={effect}
-        remedy={remedy}
-        contentTextSize={contentTextSize}
-      />
+    <div className="alert-card__content-block" >
+      <div ref={contentBlockRef}>
+        <StandardIssueSection
+          issue={issue}
+          location={location}
+          contentTextSize={effect === "station_closure" ? contentTextSize : "large"}
```
cmaddox5, line 70: Similar question as my other about the remedy. Removing this prop would also mean an effect prop would be needed, so not sure if there's much benefit for this one.

cmaddox5 (approved):
Looks good! Have a couple of questions about consolidating some hardcoded values. Not blocking though so I'll approve now.

assets/src/components/v2/pre_fare_single_screen_alert.tsx
hannahpurcell: We do need a way of passing the text size one way or another. StandardLayout uses font size large for the remedy, and DownstreamLayout uses font size medium for this component, so the component alone can't dictate which font size it uses

cmaddox5: Gotcha. Makes sense.

PR #1952, opened 2024-01-02
tweak: Flex Zone takeover layout

cmaddox5:
Also need a layout where the top screen is normal and the body screen has a flex takeover.


hannahpurcell (approved):
We'll consolidate these templates in the future, but for now, gotta get ready for the surge!

PR #1956, opened 2024-01-04
fix: Prefare polish -- Use FreeText to prevent station text wrapping in subheaders

hannahpurcell:
[Notion](https://www.notion.so/mbta-downtown-crossing/fd75b83b5904405fbc71ffbcdd0cec47?v=d4971a59c7eb45ae9841e522c53ad368&p=0f0b87cc98f349e3a4469b7d6849bca9&pm=s)

😐 ok maybe this is janky a little, but maybe it still works??

Screenshot 2024-01-04 at 5 28 50 PM

### DIFF
--- a/lib/screens/v2/widget_instance/reconstructed_alert.ex
+++ b/lib/screens/v2/widget_instance/reconstructed_alert.ex
@@ -12,7 +12,7 @@ defmodule Screens.V2.WidgetInstance.ReconstructedAlert do
   alias Screens.V2.WidgetInstance.ReconstructedAlert
   alias Screens.V2.WidgetInstance.Serializer.RoutePill
   alias ScreensConfig.Screen
-  alias ScreensConfig.V2.FreeTextLine
+  alias ScreensConfig.V2.{FreeText, FreeTextLine}
 
   require Logger
 
--- a/lib/screens/v2/widget_instance/reconstructed_alert.ex
+++ b/lib/screens/v2/widget_instance/reconstructed_alert.ex
@@ -12,6 +12,7 @@ defmodule Screens.V2.WidgetInstance.ReconstructedAlert do
   alias Screens.V2.WidgetInstance.ReconstructedAlert
   alias Screens.V2.WidgetInstance.Serializer.RoutePill
   alias ScreensConfig.Screen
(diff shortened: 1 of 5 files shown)


lib/screens/v2/widget_instance/reconstructed_alert.ex
```diff
@@ -258,6 +259,29 @@ defmodule Screens.V2.WidgetInstance.ReconstructedAlert do
...
+  defp format_station_name_list([s1, s2]),
+    do: [
+      %{format: :nowrap, text: "#{s1}"},
+      " & ",
+      %{format: :nowrap, text: "#{s2}"}
+    ]
+
+  defp format_station_name_list(list) do
+    list
+    |> List.update_at(length(list) - 1, &" & #{&1}")
```
jzimbel-mbta, line 278: `List.update_at` (and most other elixir fns that take an index and act on a collection) lets you use negatives to index from the end, for convenience:
```ex
iex> List.update_at([1, 2, 3], -1, &(&1 + 10))
[1, 2, 13]
```

So this can be
```suggestion
    |> List.update_at(-1, &" & #{&1}")
```

Also, I love the `&" & #{&1}"` expression, it looks so goofy 😄

lib/screens/v2/widget_instance/reconstructed_alert.ex
```diff
@@ -258,6 +259,29 @@ defmodule Screens.V2.WidgetInstance.ReconstructedAlert do
...
+    |> Enum.join(", #")
+    |> String.split("#")
+    |> Enum.map(fn string -> %{format: :nowrap, text: "#{string}"} end)
```
jzimbel-mbta, line 281: Should be able to use `string` directly, since it's already a string:
```suggestion
    |> Enum.map(fn string -> %{format: :nowrap, text: string} end)
```

jzimbel-mbta (approved):
Two small things, otherwise LGTM! The comments help since this is a weird bit of extra logic to satisfy a design requirement.

lib/screens/v2/widget_instance/reconstructed_alert.ex
hannahpurcell: PFFT ha good catch 😅

PR #1957, opened 2024-01-05
fix: Prefare alerts polish -- Fix GL headsign in various cases

hannahpurcell:
[Notion](https://www.notion.so/mbta-downtown-crossing/fd75b83b5904405fbc71ffbcdd0cec47?v=d4971a59c7eb45ae9841e522c53ad368&p=61cf9d560fe74ee28607cd05c25e8c55&pm=s)

### DIFF
--- a/lib/screens/v2/widget_instance/reconstructed_alert.ex
+++ b/lib/screens/v2/widget_instance/reconstructed_alert.ex
@@ -102,7 +102,8 @@ defmodule Screens.V2.WidgetInstance.ReconstructedAlert do
     "Green-B" => ["Boston College", "Government Center"],
     "Green-C" => ["Cleveland Circle", "Government Center"],
     "Green-D" => ["Riverside", "Union Square"],
-    "Green-E" => ["Heath Street", "Medford/Tufts"]
+    "Green-E" => ["Heath Street", "Medford/Tufts"],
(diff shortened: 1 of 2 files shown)


lib/screens/v2/widget_instance/reconstructed_alert.ex
```diff
@@ -167,6 +179,22 @@ defmodule Screens.V2.WidgetInstance.ReconstructedAlert do
     end
   end
 
+  # Given an entity and the directionality of the alert from the home stop,
+  # return a tuple with the affected direction_id and route id
+  defp get_direction_and_route_from_entity(%{direction_id: nil, route: route}, location)
+       when location == :downstream,
```
cmaddox5, line 185: This and the pattern under it might be a little more readable without the guard:

```suggestion
  defp get_direction_and_route_from_entity(%{direction_id: nil, route: route}, :downstream),
```

cmaddox5 (approved):
Looks good. Just one refactoring suggestion.

lib/screens/v2/widget_instance/reconstructed_alert.ex
hannahpurcell: oh duh, thanks!

PR #1958, opened 2024-01-05
fix: Prefare polish -- Alerts that break assumptions use the fallback case

hannahpurcell:
Notion [task 1](https://www.notion.so/mbta-downtown-crossing/fd75b83b5904405fbc71ffbcdd0cec47?v=d4971a59c7eb45ae9841e522c53ad368&p=05a81fd8095841cebfb4f3cb563f9f87&pm=s) and [task 2](https://www.notion.so/mbta-downtown-crossing/fd75b83b5904405fbc71ffbcdd0cec47?v=d4971a59c7eb45ae9841e522c53ad368&p=f2c21b4d7bca4254b774f5a686cc3d77&pm=s)

I'm not certain I've caught all the cases where assumptions break... but I've accounted for the ones Betsy caught during QA

### DIFF
--- a/assets/src/components/v2/pre_fare_single_screen_alert.tsx
+++ b/assets/src/components/v2/pre_fare_single_screen_alert.tsx
@@ -369,6 +369,16 @@ const PreFareSingleScreenAlert: React.ComponentType<
         />
       );
       break;
+    case !disruption_diagram:
+      layout = (
+        <FallbackLayout
+          issue={issue}
+          remedy={remedy}
+          remedyBold={remedy_bold}
+          effect={effect}
+        />
+      );
+      break;
     case effect === "station_closure" && region === "here":
       layout = (
         <MultiLineLayout
@@ -388,7 +398,6 @@ const PreFareSingleScreenAlert: React.ComponentType<
           disruptionDiagram={disruption_diagram}
         />
       );
-
       break;
     case (region === "boundary" || region === "here") &&
       (effect === "shuttle" || effect === "suspension"):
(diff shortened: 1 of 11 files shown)


assets/src/components/v2/pre_fare_single_screen_alert.tsx
```diff
@@ -177,28 +180,33 @@ const FallbackLayout: React.ComponentType<FallbackLayoutProps> = ({
...
+      <NoServiceIcon className="alert-card__fallback__icon" />
     );
 
   return (
-    <div className="alert-card__pio-text">
+    <div className="alert-card__fallback">
       {icon}
-      {issue && <div className="alert-card__pio-text__main-text">{issue}</div>}
+      {issue && <div className="alert-card__fallback__issue-text">{issue}</div>}
       {remedy && (
```
cmaddox5, line 194: I know you didn't touch this so not blocking, but is this necessary considering `remedy` is not optional? Wondering if it can be simplified a bit since this condition should always be true.

cmaddox5: Nevermind. After looking at the `WidgetInstance`, I remembered that it can be `null` and that's what the conditional is for. My bad.

assets/src/components/v2/pre_fare_single_screen_alert.tsx
```diff
@@ -360,6 +369,16 @@ const PreFareSingleScreenAlert: React.ComponentType<
         />
       );
       break;
+    case !disruption_diagram:
```
cmaddox5, line 372: Might be unnecessarily nitpicky, but can this be `disruption_diagram === undefined` instead so all the cases consistently check for equality? This gets the point across though so totally up to you.

cmaddox5 (approved):
Looks good! Just one non-blocking suggestion.

cmaddox5 (approved):
Tests look good so approval still stands. 🎸

PR #1961, opened 2024-01-10
fix: Prefare polish -- New abbreviation rule

hannahpurcell:
[Notion](https://www.notion.so/mbta-downtown-crossing/fd75b83b5904405fbc71ffbcdd0cec47?v=d4971a59c7eb45ae9841e522c53ad368&p=2c3e2b67a4b04748a7371d066b003817&pm=s)

Only abbreviate station names if they are a part of a short list of stations. 

Along the way, this fixed also revealed a situation where I noticed the diagram was getting mis-sized, and a long station name was getting cut off. Turns out that the font can take so long to render Inter that the useEffect hook thinks the diagram is smaller than it is and scales it up too much. Added a lil setTimeout on the whole hook to delay it a tiny bit longer so the font can render.

### DIFF
--- a/assets/src/components/v2/disruption_diagram/disruption_diagram.tsx
+++ b/assets/src/components/v2/disruption_diagram/disruption_diagram.tsx
@@ -604,6 +604,8 @@ const DisruptionDiagram: ComponentType<DisruptionDiagramData> = (props) => {
 
   const [doAbbreviate, setDoAbbreviate] = useState(false);
   const [scaleFactor, setScaleFactor] = useState(1);
+  const [isDone, setIsDone] = useState(false);
+
   const numStops = slots.length;
   const spaceBetween = Math.min(
     60,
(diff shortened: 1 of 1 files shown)


assets/src/components/v2/disruption_diagram/disruption_diagram.tsx
```diff
@@ -645,45 +658,44 @@ const DisruptionDiagram: ComponentType<DisruptionDiagramData> = (props) => {
...
-        const factor = Math.min(
-          xScaleFactor,
-          yScaleFactor
-        );
-        setScaleFactor(factor);
-        setTimeout(() => {
-          setIsDone(true);
-        }, 200);
+    setTimeout(() => {
+      // Get updated dimensions each time this hook runs
```
cmaddox5, line 668: The nested callbacks are getting a little hard to follow. Can you make the original callback its own function and call that function in the `setTimeout`?

hannahpurcell: Good call. Did this help?

cmaddox5 (approved):
🎸

PR #1962, opened 2024-01-10
fix: Prefare polish -- Reorder audio with alerts before subway status

hannahpurcell:
[Notion](https://www.notion.so/mbta-downtown-crossing/fd75b83b5904405fbc71ffbcdd0cec47?v=d4971a59c7eb45ae9841e522c53ad368&p=cab9f47a242e4d7e9d878c8f9cf76dab&pm=s)


cmaddox5 (approved):
Makes sense! If this feature ends up being deployed during the Surge, we'll want to double check ordering of the static content at Gov't Ctr. That's just in configuration though so not a biggie. 

I see tests are failing but I imagine those are also just integer changes. I'll go ahead and approve so it's ready to merge after fixing.

PR #1965, opened 2024-01-16
feat: Pending screen endpoint(s)

jzimbel-mbta:
**Asana task**: [Add pending endpoint to Screens](https://app.asana.com/0/1185117109217413/1205700210661296/f)

This ballooned a bit because it was really 4 endpoints:

| | |
| - | - |
| Pending screen page | /v2/screen/pending/:id |
| Pending screen _simulation_ page | /v2/screen/pending/:id/simulation |
| Pending screen data | /v2/api/screen/pending/:id |
| Pending screen _simulation_ data | /v2/api/screen/pending/:id/simulation |

This adds a new file needed by local dev: `priv/local_pending.json`. Its absence doesn't prevent the server from starting up like `local.json` does, but you'll get errors if you try to do anything related to pending screens without it.
The minimal contents are:
```json
{"screens":{}}
```

# ❓ Logging? ❓
I didn't add any new log messages specifically for pending screen requests, should I do that?

- [ ] Tests added?


assets/src/apps/v2/bus_eink.tsx
```diff
@@ -96,6 +96,13 @@ const App = (): JSX.Element => {
   return (
     <Router>
       <Switch>
+        <Route exact path={["/v2/screen/:id", "/v2/screen/pending/:id"]}>
```
jzimbel-mbta, line 99: (I moved the main route to the top for visibility. Looks like it was at the bottom only because it was matching paths like `/v2/screen/:id/simulation`; adding `exact` fixes that.)

lib/screens/pending_config.ex
```diff
@@ -0,0 +1,18 @@
...
+        }
+
+  @enforce_keys [:screens]
+  defstruct @enforce_keys
+
+  use ScreensConfig.Struct, children: [screens: {:map, Screen}]
+
+  defp value_from_json(_, value), do: value
+  defp value_to_json(_, value), do: value
+end
```
jzimbel-mbta, line 18: Just realized this should probably go over in screens-config-lib since the other [top level `Screens.Config`](https://github.com/mbta/screens-config-lib/blob/main/lib/config/config.ex) module has now been moved there as well.

Let me know if that sounds good and I'll do it.

hannahpurcell: Ha! Took the words out of my mouth. (That was something I have queued up in my review.) Yeah, sounds good to me. Edit: oh shoot, GH won't let me send comments while I have a review pending. That's annoying!

hannahpurcell (commented):
Why were the `/v2/screen/pending/:id` endpoints needed?

jzimbel-mbta:
> Why were the `/v2/screen/pending/:id` endpoints needed?

This was how I decided to make it possible for the client to tell the server to look up the screen's config in a different place from usual. E.g. pending-screens-prod.json instead of screens-prod.json.

This is how it works:
1. Someone's browser requests the page at `/v2/screen/pending/<pending id>`. (or .../simulation)
2. Server responds with html identical to that of the normal screen app page, except:
   - `data-refresh-rate` on the `#app` div is set to 0 since we don't want the pending screen previews to continually update
   - `data-is-pending="true"` is added to the `#app` div
3. Browser loads up the page & starts the react app.
4. React app looks for the `is-pending` data attribute, sees it's set to true, and as a result, requests screen data from `/v2/api/screen/pending/<pending id>` instead of the usual API path.
5. Because of the different route, the server looks in the pending config instead of the main config to answer this data request.

Since the only difference in behavior is that refreshes are turned off and data is requested from a different path, maybe this could work better as just a query param on the page url? Something like `/v2/screen/<pending id>?is_pending=true`.

What do you think?

hannahpurcell:
Great breakdown! But I realized my question was incredibly vague. I mean that we shouldn't need to be looking up pending screens in the browser, only via simulation, so I was thinking that the non-simulation endpoints are probably not needed.

jzimbel-mbta:
Ah I see! The non-simulation page endpoint is there because:
1. Ana wanted the ability to see the pending screen previews full-size via a button, in some discussion Mary and I had with her while adjusting the final designs for the "review pending" part of the new screen workflow.
2. I figured it would be worth adding for debug purposes. But I guess it doesn't really provide anything the simulation view doesn't already...

image

hannahpurcell:
Ahhh yes, the preview! Ok, fantastic. Thanks for covering that

jzimbel-mbta:
Pending config shape added to the config lib: https://github.com/mbta/screens-config-lib/pull/13

lib/screens_web/controllers/v2/screen_controller.ex
```diff
@@ -83,41 +83,81 @@ defmodule ScreensWeb.V2.ScreenController do
...
     render_not_found(conn)
   end
 
+  def index_pending(conn, %{"id" => screen_id} = params) do
+    # Should we log info for these requests?
+    # Current precedent is to not, for simulation requests.
+    # But it could be helpful, and we don't have to come up with a way to
+    # distinguish between requests from "real screens" vs not.
+    # (Because real screens should never hit this endpoint.)
+    # _ = Screens.LogScreenData.log_pending_screen_page_load(screen_id)
```
jzimbel-mbta, line 108: Oop I left this code comment that's more like a PR comment/question, want me to remove it?

hannahpurcell: Ohhhhhh I forgot to comment on this, also in the PR description. I think you're wise to point out the logging potential, and might be worth adding here and now

hannahpurcell: Actually, I've let the Screenplay PRs back up a ton because I was so slow to review. Let's remember this as a #todo and count it as a follow-on item?

hannahpurcell (approved):
Fantastic! 🎉

PR #1974, opened 2024-02-14
 feat: Migrate from Cognito to Keycloak

cmaddox5:
**Asana task**: [migrate screens from Cognito to Keycloak](https://app.asana.com/0/1185117109217413/1206141963656810/f)

This PR does two big things: upgrade Erlang and Elixir and replace all Cognito configuration with Keycloak. The login screen may look a tad different, but no other functionality should be changed.


hannahpurcell (approved):
🎉

PR #1978, opened 2024-02-29
tweak: Multi-Screen Route order

cmaddox5:
**Asana task**: ad-hoc

Adjusted route order so the multi-screen route uses the correct component.

- [ ] Tests added?

### DIFF
--- a/assets/src/apps/v2/bus_eink.tsx
+++ b/assets/src/apps/v2/bus_eink.tsx
@@ -96,6 +96,12 @@ const App = (): JSX.Element => {
   return (
     <Router>
       <Switch>
+        <Route exact path="/v2/screen/bus_eink_v2">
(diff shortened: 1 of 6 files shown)


hannahpurcell:
Oh interesting! I approved the [original reordering](https://github.com/mbta/screens/pull/1965#discussion_r1453948260) and I'm wondering why we both missed the fact that this introduced a bug? Does it have to be in this order, or everything breaks?

cmaddox5:
It does because it navigates to the first route in the tree that matches. Even having `exact` doesn't work because it's second in the list. What it was doing is matching to `/v2/screen/:id` where `:id` is mapping to `pre_fare_v2`.

hannahpurcell (approved):
Aha! Thanks 👍

PR #1981, opened 2024-03-04
feat: Show PreFare takeover alert as single screen

cmaddox5:
**Asana task**: [Work to support single screen alert for Pre-fare display](https://app.asana.com/0/1185117109217413/1206567004613748/f)

This PR uses the new boolean config field `pair_takeover_with_cr_widget` to force a takeover alert to use the single screen template. This will eventually be paired with, you guessed it, the CR widget. 

**Deploying this will take some strategy because all existing PreFare configs will need to be edited. 

- [ ] Tests added?


hannahpurcell (approved):
I'm so satisfied that this solution was so straightforward 👍 👍

PR #1982, opened 2024-03-07
Move value in alerts config to cr widget config; set cr widget in right slot sometimes

hannahpurcell:
**Asana task**: [Work to support CR widget placement on right screen](https://app.asana.com/0/1185117109217413/1206567004613747)

Moved the config flag to pair the CR widget with a condensed alert widget to the CR config instead of the alert one. Then let the CR widget overflow to the righthand screen if there is a takeover alert on the left.

- [ ] Tests added?


cmaddox5 (approved):
Config changes look good! Gonna approve for that but will wait for another approval for the wording changes.

PR #1985, opened 2024-03-12
Setup updates

digitalcora:
### `doc: remove devcontainer setup`

Per @jzimbel-mbta, no one on the team currently uses this approach, and it has fallen out of date in the years since it was created.

### `doc: revise setup guide and add config script`

This is based on my experience running through the steps. There is one major addition, the `pull_configs` script, which cuts out some manual work previously required to update a developer's local configuration.

Further details on specific changes:

* Explicitly installing `autoconf` and `gnupg` no longer appears to be required; I did not do this, from a blank Homebrew slate, and did not encounter any issues.

* `asdf-nodejs` no longer requires manually importing a keyring, per its current README.

* The `asdf-erlang` issue with "You are natively building for a later version" appears to have been fixed upstream, after some turbulence (see https://github.com/asdf-vm/asdf-erlang/issues/161 and https://github.com/asdf-vm/asdf-erlang/issues/228). This issue also only ever occurred with Erlang 23 and below, which we're well beyond now, so we probably don't need to keep these detailed docs. If we run into it again, the solution is easily searched-for.

* `/screen/1` does not exist (anymore?), so this path was not a good test of workingness.

* Setting up persistent environment variables and AWS credentials are moved into their own sections after the main guide, since neither is required to reach a state where you can test that the app works.

* The environment variable section recommends `direnv` for managing these values. Conveniently, this repo had already gitignored `.envrc`.

* The AWS section makes it clearer what this setup is needed for (since it's not required just to get things running), and recommends storing keys in 1Password as per Infrastructure best practice, explained in the linked Notion doc.


jzimbel-mbta (approved):
This is fantastic, thank you! Huge readability improvement to the plaintext view from (I assume) using a utility to split the lines at a consistent length.

digitalcora:
"Hard wrap at ruler" is a built-in command in Sublime, I'd be surprised if other editors didn't also have it. I recommend it for committed docs!

PR #2006, opened 2024-03-19
Initial `busway_v2` screen setup

digitalcora:
**Asana**:
* [Widgetize solari into Busway App](https://app.asana.com/0/1206602639264164/1206581985806628) (multiple subtasks)
* [Header widget: set up and update design](https://app.asana.com/0/1206602639264164/1206645864760371)

Supporting config change: https://github.com/mbta/screens-config-lib/pull/18

### Summary

This renames the existing `solari_v2` "skeleton" and brings it up to functional and design parity with bus shelter screens, but does not include any features _unique_ to busway screens, as those have been tasked separately.

The major elements:

* Loading and "no data" states are configured.

* Styles for departures and unusual states are added to busway screens, and those shared with bus shelter screens are moved into `lcd_common`.

* The "no data" state for the `lcd` version of the departures widget is changed to use the wording "Live departure updates" instead of "Live _bus_ departure updates", as this state is no longer specific to the (bus-only) bus shelter screens.

* The above state also gains handling for `show_alternatives` with no `stop_id`. Previously this was technically allowed but the component would assume these props were always provided together, and display an unhelpful URL if not.

* The `lcd` header (used on bus shelter and now busway screens) gains better auto-resize behavior when text would overflow. Previously it allowed two lines of "large" text before resizing, but the font size does not actually allow this to fit comfortably within the header. The icon would also be misaligned in all cases except for a single line of "large" text. (No bus shelter screens currently use an icon or wrap to a second line.)

### Screenshots

Normal | Loading | Error
--- | --- | ---
Screen Shot 2024-03-19 at 16 21 49 | Screen Shot 2024-03-19 at 10 42 29 | Screen Shot 2024-03-19 at 10 38 46

### Header variants

This was a bit of an adventure, so I wanted to document that the various combinations of icon and text length are all working.

&nbsp; | Large | Small | Multi-line
--- | --- | --- | ---
🍵 | Screen Shot 2024-03-19 at 16 05 03 | Screen Shot 2024-03-19 at 16 04 21 | Screen Shot 2024-03-19 at 16 04 44
🚫 | Screen Shot 2024-03-19 at 16 07 55 | Screen Shot 2024-03-19 at 16 06 49 | Screen Shot 2024-03-19 at 16 07 07

### Example config

```
    "busway-test": {
      "app_id": "busway_v2",
      "app_params": {
        "departures": {
          "sections": [
            {
              "filter": null,
              "headway": {
                "headway_id": null,
                "override": null,
                "sign_ids": []
              },
              "query": {
                "opts": {
                  "include_schedules": false
                },
                "params": {
                  "direction_id": "both",
                  "route_ids": [],
                  "route_type": null,
                  "stop_ids": ["1060", "1123", "72", "102"]
                }
              }
            }
          ]
        },
        "header": {
          "stop_name": "Central"
        }
      },
      "device_id": "TEST",
      "disabled": false,
      "name": "",
      "refresh_if_loaded_before": null,
      "tags": [],
      "vendor": "solari"
    },
```


assets/css/v2/bus_shelter/departures_no_data.scss
jzimbel-mbta, line 1: Since both files' contents are now so short, would it make sense to merge this one into `bus_shelter/departures.scss`? Same for the new busway `departures.scss` and `departures_no_data.scss` files.

assets/css/v2/lcd_common/normal_header.scss
```diff
@@ -39,8 +37,19 @@
...
-  &--with-icon {
-    width: 668px;
+  /*
+  Hack: when displaying the title with an icon, roughly cancel out the space
+  reserved for descenders on the last line, making it "optically centered" with
+  the icon instead of bounding-box-centered. Will need to change if the font or
+  font size changes.
+  */
+
+  &--large#{&}--with-icon {
```
jzimbel-mbta, line 47: That's some syntax I haven't seen before. So this translates to the following selector?
```scss
.normal-header-title__text--large.normal-header-title__text--with-icon
```

Separately: did you try giving this element a `height` and `line-height` that matches the height of the Ⓣ? That's usually how we manage to vertically align icons with text. (If it didn't work, this is fine, vertically aligning text and graphics in CSS is a pain.)

lib/screens/v2/candidate_generator/busway.ex
```diff
@@ -1,37 +1,39 @@
...
   alias ScreensConfig.Screen
+  alias ScreensConfig.V2.Busway
   alias ScreensConfig.V2.Header.CurrentStopName
-  alias ScreensConfig.V2.Solari
+
+  defmodule Deps do
+    @moduledoc false
+    defstruct now: &DateTime.utc_now/0,
+              departures_instances: &Widgets.Departures.departures_instances/1
+  end
```
jzimbel-mbta, line 16: Ooh I like this. Much neater than adding a separate argument for each dependency. 👍

jzimbel-mbta (approved):
Good to go, nice work! Just two very minor thoughts/suggestions.

assets/css/v2/bus_shelter/departures_no_data.scss
digitalcora: I do think it's nice to keep the correspondence between these CSS files, the same-named "common" CSS files they're meant to be paired with, and the same-named React components (well, the Departures component is actually called `NormalDepartures`... though there isn't an <i>ab</i>normal departures component). It ain't CSS Modules, but it's something!

assets/css/v2/lcd_common/normal_header.scss
digitalcora: You're right about the compilation result. You could say `#{...}` is the general way of accessing variables (of which `&` is a built-in one), and `&` as the first character of a selector is a special case where the interpolation syntax can be left out.

About the alignment, the flexbox properties actually take care of that and also work for any number of text lines. The issue is the "optical" alignment I mention in the comment. This is what it looks like without the extra negative margin, but with everything else the same:

Screenshot 2024-03-25 at 1 51 38 PM

This is not visually centered, which is clearest when you compare the center line of the "C" to the circle around the Ⓣ. But the _bounding box_ of the text is perfectly centered:

Screenshot 2024-03-25 at 1 51 28 PM

That's because the bounding box includes the space below the baseline reserved for descenders, which becomes clearer (though it doesn't make the visual centering any better) if you change the header to use a character with a descender:

Screenshot 2024-03-25 at 1 53 02 PM

Screenshot 2024-03-25 at 1 53 17 PM

It turns out it's [pretty involved](https://iamvdo.me/en/blog/css-font-metrics-line-height-and-vertical-align) to compensate for this in a font-size-agnostic way; there's no standard ([yet](https://medium.com/microsoft-design/leading-trim-the-future-of-digital-typesetting-d082d84b202)) that changes the bounding box to align with the "optical" height of the text. So I went for some easier, but more fragile, manual adjustments.

Screenshot 2024-03-25 at 1 53 27 PM

We only need to deal with this because the text may run onto multiple lines, as in the "Station with longer name" example in Figma, so it's not sufficient to add the logo as an inline(-block) element alongside the text. If you do that, and set the line-height to the height of the image (as you mentioned), then it all works out because the image _also_ has that extra space reserved below it — even though there's no "text" in it, it's laid out as a "text box".

digitalcora: ...and, as is often the way of things, now that I've written all this out, it occurs to me there's no reason we couldn't still exploit rendering the icon as a "text box" to automatically get the right spacing. The icon itself just needs to also know what size the text is, which we could do by moving the size class up a level to the containing box.

digitalcora: This approach seems to work without having to use any manual offsets 🎉 It's a tiny bit off in the case where there are two lines of small text (see below); I'm not sure where the extra few pixels are coming from, but it looks acceptable to me and we likely won't have many screens with enough title text to cause this.

Screenshot 2024-03-25 at 3 30 03 PM

Screenshot 2024-03-25 at 3 29 57 PM

PR #2009, opened 2024-03-22
ci: parallelize jobs using composite setup action

digitalcora:
This allows each check to execute and report its status separately and in parallel, without duplicating the setup and cache-restoring logic.

I gave these jobs "Elixir"/`ex` prefixes (and removed the "Elixir" prefix from the workflow itself) in anticipation of restoring working frontend linting, type-checking, etc., which would see further benefits from this approach as they wouldn't delay each other or the existing jobs.

Note in the event anything invalidates the various caches, it will waste some runner time (not real time) in proportion to the number of jobs, as each job independently starts, finds there is no matching cache, and does the work to populate it. In this case it doesn't cost us anything, since Actions is free for open-source repos. If we care about this, we could introduce a "pre-cache" job which runs the setup and does nothing else, and have all the other jobs require that one to finish running first — but in testing I found this incurs a ~20s time penalty in the common case that the cache is fully up-to-date, which is a 1.5x to 2x slowdown depending on the job.


panentheos (approved):
Looks good!

PR #2015, opened 2024-04-16
Fix TypeScript errors and enable `tsc` in CI

digitalcora:
This adds type-checking to CI and resolves all existing type errors that are not due to implicit `any` types — in other words, all cases where the code or the type annotations are wrong in a way the compiler can prove without needing full type annotations. (This is more limited than it sounds; to catch all errors, explicit annotations are needed on most function arguments and return values, which is why `noImplicitAny` exists and is enabled by default in strict mode. For now, I've disabled it here, since we have such a huge number of implicit `any`s that it would explode the scope of this task.)

The "magic comment" `@ts-expect-error` suppresses any error that would be reported on the next line, and is used here to punt on situations that would be non-trivial to resolve, particularly in non-v2 code where we may never have to resolve them. Note this differs from `@ts-ignore` in that if there is _no_ error on the next line, this is reported as an error — so if a future improvement to our types or the compiler results in the error being resolved, we'll know and can remove the comment.

There are a few general issues which are all fixed in similar ways:

* When the compiler can't prove the conditions of an `if` or `switch` cover all possibilities, and there is no fallback `else` or `default`, we have to add one. In most cases we can just throw an error, as we expect the code path to never be reached.

* Several places in existing type definitions were trying to use `object` as though it meant "object with unspecified fields of unspecified types". However, `object` actually means an "opaque" object: it only permits operations that would be allowed on _any_ object, which does not include accessing any specific properties (which might not exist). The correct way to do this is `Record<string, any>`, though in some cases involving React props the best we can do without filling in a more detailed type is `any`.

* A type annotation is required on variables initialized to `null` or `undefined` and later assigned, or arrays initially declared as `[]` and later populated. The compiler does not backtrack to fill in the variable's type based on later usage; it only looks at the initial value.

* Similar to the above, `useState` requires a type parameter for the values that will later be assigned, if the type cannot be inferred from the initial value.

* Also similar to the above, `useRef` and `forwardRef` require a type parameter for the DOM element the ref will later be associated with. Due to a quirk of how the types are implemented for this function, `null` must also be explicitly provided as the initial value (we were already doing this in most places).

* The value in a ref is `null` during the initial render, so code that accesses properties of `ref.current` needs to account for this, even if we expect it to never happen (e.g. inside a `useLayoutEffect`).

---
  - https://app.asana.com/0/0/1207055911773551

### DIFF
--- a/assets/src/components/admin/admin_cells.tsx
+++ b/assets/src/components/admin/admin_cells.tsx
@@ -59,7 +59,7 @@ const EditableNumberInput = ({
     if (!isNaN(value)) {
       doUpdate(index, mutator || id, value);
     } else {
-      alert(`Integer value expected in ${id} for Screen ID ${rowValues.id}`);
+      alert(`Integer value expected in ${id}`);
     }
   };
 
@@ -119,6 +119,7 @@ const EditableSelect = ({
       disabled={!editable}
     >
       {options.map((opt) => (
+        // @ts-expect-error
         <option value={opt} key={opt}>
           {opt}
         </option>
@@ -154,10 +155,10 @@ const EditableTextarea = ({
 };
 
 const IndeterminateCheckbox = ({ indeterminate, ...rest }) => {
-  const ref = useRef();
+  const ref = useRef<HTMLInputElement>(null);
 
   useEffect(() => {
-    ref.current.indeterminate = indeterminate;
+    if (ref.current) ref.current.indeterminate = indeterminate;
   }, [ref, indeterminate]);
 
   return (
--- a/assets/tsconfig.json
+++ b/assets/tsconfig.json
@@ -27,7 +27,7 @@
 
     /* Strict Type-Checking Options */
     "strict": true /* Enable all strict type-checking options. */,
-    "noImplicitAny": true /* Raise error on expressions and declarations with an implied 'any' type. */,
+    "noImplicitAny": false /* Raise error on expressions and declarations with an implied 'any' type. */,
     "strictNullChecks": true /* Enable strict null checks. */,
     "strictFunctionTypes": true /* Enable strict checking of function types. */,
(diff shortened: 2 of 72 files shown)


assets/tsconfig.json
```diff
@@ -27,7 +27,7 @@
 
     /* Strict Type-Checking Options */
     "strict": true /* Enable all strict type-checking options. */,
-    "noImplicitAny": true /* Raise error on expressions and declarations with an implied 'any' type. */,
+    "noImplicitAny": false /* Raise error on expressions and declarations with an implied 'any' type. */,
```
cmaddox5, line 30:
```suggestion
    "noImplicitAny": false /* Do not raise error on expressions and declarations with an implied 'any' type. */,
```

assets/src/components/admin/admin_cells.tsx
```diff
@@ -59,7 +59,7 @@ const EditableNumberInput = ({
     if (!isNaN(value)) {
       doUpdate(index, mutator || id, value);
     } else {
-      alert(`Integer value expected in ${id} for Screen ID ${rowValues.id}`);
```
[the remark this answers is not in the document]
digitalcora: It seems to have been a copy-paste error? There is another alert like this in the same file where the variable does exist.

assets/tsconfig.json
digitalcora: In my mind the comments are "static" documentation of what the options do when enabled, so under that logic I'd keep it the same, if that makes sense. (Actually I'd prefer not having the comments at all, since they're noisy, duplicative of the `tsconfig` documentation, and can easily get out of sync with it, but thought to save that idea for another day.)

cmaddox5: Yep, makes sense.

PR #2029, opened 2024-04-19
fix: missing route pills on Solari screens

digitalcora:
The refactor of this logic to address type errors in 1f411cb was incorrect and ended up breaking text route pills (e.g. bus numbers, commuter rail track numbers). The variable mistakenly named `routeImg` here was intended to be an image-or-string.


jzimbel-mbta (approved):
Good catch!

PR #2030, opened 2024-04-19
TEMP: busway v2 migration script

digitalcora:
⚠️ Not intended to be merged, just a place for this to live (and be reviewed) until we're ready to do the migration.

As before, see here for documentation of the v1 configuration: https://github.com/mbta/screens/blob/jz/solari-departures-documentation/delete_me/solari-config-structure.ts

~~As of right now, the migrated configs mostly don't produce fully-working screens, since we don't yet support any of the config values unique to Busway: multiple sections, section headers, layout constraints, bidirectional mode, etc. But subject to those caveats, they do appear to "work". We can revisit this periodically as we implement those features.~~

---
  - https://app.asana.com/0/0/1206740440094011

### DIFF
--- a/scripts/busway-config-migrate.jq
+++ b/scripts/busway-config-migrate.jq
@@ -1,0 +1,141 @@
+# Migrates `solari` screens to `busway_v2` screens. The new screens have the
+# IDs of the old screens with "-V2" appended, leaving the old screens as-is.
+# New screens are overwritten if they already exist. The output is the entire
+# updated configuration file.
+#
+# Usage:
+#   jq -f scripts/busway-config-migrate.jq priv/local.json > priv/new.json
+
+.screens += (
+  .screens | with_entries(
+    .value.app_params as $params
+    | select(.value.app_id == "solari")
+    | select(.value.disabled == false)
+    | select($params.overhead == false)
+    | {
+      key: "\(.key)-V2",
+      value: (.value + {
+        app_id: "busway_v2",
+        app_params: {
+          header: { stop_name: $params.station_name },
+          evergreen_content: [],
+          departures: {
+            sections: [
+              $params.sections[] | {
+                header: {
+                  # v1 has the field `section_headers`, which if set to "none"
+                  # prevents headers from appearing visually even when a title
+                  # is defined for sections (and in that case the title still
+                  # provides the audio readout for the header if no alternate
+                  # readout is given). In v2 it is just the presence of a title
+                  # that determines whether a section has a visual header. This
+                  # also means a value for `read_as` is required to read out a
+                  # header with no visual title.
+                  title: (
+                    if $params.section_headers == "none" or .name == "" then
+                      null
+                    else
+                      .name
+                    end
+                  ),
+                  arrow: (
+                    if $params.section_headers == "none" then
+                      null
+                    else
+                      .arrow
+                    end
+                  ),
+                  read_as: (
+                    .audio.wayfinding as $wayfinding
+                    | (
+                        if $params.section_headers == "none" then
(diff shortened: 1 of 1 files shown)


scripts/busway-config-migrate.jq
```diff
@@ -0,0 +1,88 @@
...
+                    # NOTE: The v1 config cannot specify a base and max size
+                    # separately; `visible_rows` (if Later Departures enabled)
+                    # or `num_rows` (if not) fills both roles. This sometimes
+                    # means extra space on the screen goes unused. We should
+                    # manually change `max` to `null` on sections where it
+                    # makes sense for them to use any extra space available.
+                    .layout.opts
+                    | (
+                        # `visible_rows` counts Later Departures as one "row".
+                        if .paged then .visible_rows - 1 else .num_rows end
```
jzimbel-mbta, line 73: Does Later Departures get counted differently/separately in the new logic?

scripts/busway-config-migrate.jq
```diff
@@ -0,0 +1,88 @@
...
+                      ) as $max
+                    | { min: 1, max: $max, base: $max, include_later: .paged }
+                  else
+                    { min: 1, max: null, base: null, include_later: false }
```
jzimbel-mbta, line 77: Is the `layout` object even needed for bidirectional if all of these values are fixed in that case? Would it make sense for it to just be `layout: null`? Or, `layout: "bidirectional"` and omit the separate `bidirectional` boolean field?

If someone manually changed these values for an existing bidirectional Busway section, e.g. setting `min: 2`, would the logic be able to handle it gracefully?

(Having a value be either an object or a string might be a bit weird, I guess...)

jzimbel-mbta (approved):
Looks good! Just a couple questions/suggestons.

This is much more advanced jq usage than the basics that I know, but I don't see any obvious issues, and the output looks right. ✅

Separately, good call to keep the old Solari configs around for now. They will make it easier to visually check new vs old.

scripts/busway-config-migrate.jq
digitalcora: The short answer is that it doesn't count at all. To quote myself in `ScreensConfig.V2.Departures.Layout`:

>   If `include_later` is set, the section includes a paging "Later Departures" component, which shows departures that would have otherwise been dropped due to space constraints (plus as many further-out departures as it has space for). The other fields do not "count" departures shown here; in other words, a section with `%{base: 4, max: 6, include_later: true}` is treated as desiring 4-6 departures that are visible at all times, and if more are available, however many will fit in "Later Departures" in addition to that.

Of course, so far this is only a description of the _intent_, as the section sizing logic has not been implemented yet.

scripts/busway-config-migrate.jq
digitalcora: In terms of the implementation `bidirectional` is a server-side filter on the departures returned to the client, whereas `layout` controls the client-side behavior when resizing sections to fit. I didn't envision the client looking at `bidirectional` at all. Which brings up a good point: we should migrate existing `bidirectional` sections to have `"min": 2` in the new format, or else it's technically possible for them to get shrunk down to the default minimum of 1 departure, which we probably don't want.

So why is `bidirectional` not part of `Filters`, you ask? Probably just because it was already there in the root `Section` configuration and I didn't think to move it. From what Dave was saying in the Slack threads where we discussed this, we want to develop a more robust and universally-applicable notion of "bidirectional" in the future (which would probably be called something else), so this field may be destined for eventual relocation, renaming, or outright removal — which puts it in a weird spot right now.

digitalcora:
Config migration is complete and in production, closing this out.

PR #2033, opened 2024-04-25
Bidirectional mode and time filtering

digitalcora:
Best reviewed per-commit, as there is a non-trivial refactoring and also a sneaky extra feature here (filtering by time, which is used primarily in `include_schedules` sections to avoid showing schedules too far in the future — we didn't have a standalone Asana task for this).

Note the frontend still doesn't support multiple sections, so converted Solari configs (e.g. using #2030) will continue to have odd behavior. Bidirectional mode and time filtering do work correctly on single sections, if you trim down a config to use only one section.

---
  - https://app.asana.com/0/0/1206932080790888


cmaddox5 (approved):
Yeah this makes sense and is very clean. Glad it wasn't too much of a chore to pull out of the DUP module.

PR #2034, opened 2024-04-26
Prettier frontend code 🧹

digitalcora:
Huge diff is huuuuuge.

The biggest category of change is the addition of trailing commas when elements of list-like constructs are placed on their own lines. This can look a little weird at first, especially with function arguments and especially when there's only one (and when the function only takes one argument anyway, though Prettier doesn't know that). I do think this approach is preferable in the general case, since it means adding or removing elements doesn't generate diffs on unrelated lines to add or remove trailing commas.

---
  - https://app.asana.com/0/0/1207171929398149


PaulJKim (approved):
I scanned through each file and looks good to me

PR #2038, opened 2024-04-30
WIP: Triptych app changes

cmaddox5:
**Asana task**: [[Timebox] Investigate babel config to possibly transpile from react to plain JS](https://app.asana.com/0/1185117109217413/1206932108214250/f)

~~Working on some webpack changes for getting the packaged triptych app compiled down to something the hardware can handle. Will update PR as OFM gives app feedback.~~

App has been running well on OFM lab screens so this may be the best we can do at the moment. Some key changes:

- A new webpack entry has been created for Triptych packaging. This eliminates the need to change webpack when we are packaging up the app. The babel config for the new entry targets the oldest browsers possible (this is done by default when no targets are [provided](https://babeljs.io/docs/options#no-targets)). 
- All images in Triptych widgets are imported by default so that webpack preloads them in specific folders. We then only zip those folders up in the OFM package. This helps keep package size down. 
- Any changes made to the existing webpack entry are there to allow backwards compatibility. This should only be a `file-loader` config for images. I found that if one entry configured `file-loader` for images, others needed to as well. Regardless, it seemed like a good idea to explicitly configure them.

- [ ] Tests added?


assets/src/util/util.tsx
```diff
@@ -25,7 +25,10 @@ export const formatTimeString = (timeString: string) =>
   moment(timeString).tz("America/New_York").format("h:mm");
 
 export const imagePath = (fileName: string): string =>
-  isOFM() ? `images/${fileName}` : `/images/${fileName}`;
+  isOFM() ? outfrontImagePath(fileName) : `/images/${fileName}`;
```
digitalcora, line 28: If we're still using `imagePath` for OFMs in addition to `import`ing images into components, in which scenarios do we use which one?

assets/webpack.config.js
```diff
@@ -7,115 +7,268 @@ const TerserPlugin = require("terser-webpack-plugin");
...
+  },
+  {
+    resolve: {
+      extensions: [".ts", ".tsx", ".js", ".jsx"],
+      alias: {
+        // Please also update the "paths" list in tsconfig.json when you add aliases here!
+        Components: path.resolve(__dirname, "src/components"),
+        Hooks: path.resolve(__dirname, "src/hooks"),
+        Util: path.resolve(__dirname, "src/util"),
+        Constants: path.resolve(__dirname, "src/constants"),
```
digitalcora, line 145: Since this is a JS file and we can define variables and such, can we extract out common parts of these configs, so it's clear where they differ and we can change shared things in one place?

docs/triptych-react-app-testing.md
```diff
@@ -0,0 +1,18 @@
+# Triptych React app testing
```
digitalcora, line 1: Should we keep this committed in the final changeset? It seems more like a note specific to the process of handing off this PR.

assets/src/util/util.tsx
cmaddox5: The `import` should only have to be done once in `outfront_evergreen_content.tsx`. It gives webpack the ability to preload the images and save them to the [correct folder](https://github.com/mbta/screens/blob/a2839ceea385ea916fa470dfdfc528783dfcab4b/assets/webpack.config.js#L251). Because this folder only exists in the packaged app, we only need to reference that path on live triptychs. Eventually, DUPs should be treated similarly, but I wanted to leave them alone for now.

assets/src/components/v2/outfront_evergreen_content.tsx
```diff
@@ -5,6 +5,11 @@ import useIsOnScreen from "Hooks/v2/use_is_on_screen";
 import { imagePath } from "Util/util";
 import { TRIPTYCH_VERSION } from "./triptych/version";
 import { usePlayerName } from "Hooks/outfront";
+import { isOFM } from "Util/outfront";
+
+if (isOFM()) {
+  require.context("../../../static/images/triptych_psas", true, /\.(webp)$/);
```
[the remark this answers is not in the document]
digitalcora: I think you answered a different question than the one I asked? I was interested in why, in the regex `/\.(webp)$/`, why the `webp` is in a capture group. (in other words, why not `/\.webp$/`?)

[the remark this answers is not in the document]
cmaddox5: You're totally right, sorry about that.

digitalcora (approved):
Tentative thumbs-up, let's see if it works on the real thing!

digitalcora:
Oh, it looks like this has some conflicts, since unfortunately it branched off before my linting changes 😓 Sorry for the extra work!

digitalcora:
ah... "WIP" commit on main 😅

cmaddox5:
> ah... "WIP" commit on main 😅

Dangit. I even looked right at those letters and thought "Yeah looks good"...

PR #2039, opened 2024-04-30
chore: Screenplay simulation support for Busway v2

digitalcora:
`lcd_common` continues to not be the greatest name for things we're sharing between (only) `busway` and `bus_shelter`, but I haven't thought of anything better yet. See also: https://github.com/mbta/screenplay/pull/332

### DIFF
--- a/assets/css/v2/bus_shelter/simulation.scss
+++ b/assets/css/v2/bus_shelter/simulation.scss
@@ -1,40 +1,4 @@
 .simulation-screen-scrolling-container {
-  scrollbar-color: #607180 #2e3f4d;
-
-  &::-webkit-scrollbar {
-    width: 12px;
-  }
-  &::-webkit-scrollbar-thumb {
-    background-color: #607180;
-    border-radius: 12px;
-    border: 2px solid #2e3f4d;
-  }
-  &::-webkit-scrollbar-track {
-    -webkit-box-shadow: inset 0 0 6px rgba(0, 0, 0, 0.3);
-    border-radius: 12px;
-    background-color: #2e3f4d;
-  }
-}
-
-.simulation-screen-scrolling-container {
-  display: flex;
-  overflow-y: hidden;
-  overflow-x: auto;
-  padding-bottom: 20px;
-
-  .simulation__full-page {
-    height: 344px;
-    width: 193.5px;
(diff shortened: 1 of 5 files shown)


assets/css/v2/bus_shelter/simulation.scss
```diff
@@ -1,40 +1,4 @@
 .simulation-screen-scrolling-container {
```
PaulJKim, line 1: I'm not very confident about my knowledge on CSS so pardon if this is a silly question, but we don't need to be concerned about this class name also being used in `lcd_common/simulation.scss` do we? Like is it possible for classes to clash/overwrite each other?

digitalcora: > Like is it possible for classes to clash/overwrite each other?

Yes, but on a more granular level than selectors; this is the "Cascade" of Cascading Style Sheets. You can [read all about it here](https://developer.mozilla.org/en-US/docs/Web/CSS/Cascade), but a quick and inaccurate summary might be that all defined styles are "merged" together. If your styles contain both `li { font-weight: bold }` and, in some other place, `li { color: red }`, then list items will be both bold and red; there is no conflict.

If an element matches multiple selectors that both define the _same_ property, like `color`, the one with the greater "specificity" applies — which has a complex definition but sort of corresponds to how "narrow" the selector is. For example, `li` applies to all list items, but `li.highlight` only applies to list items with the `highlight` class, so the latter is "more specific" and wins if they both define the same property. You've probably seen `!important` before; that's an escape hatch for when you want to ignore specificity and say "this one always wins". (Having to use this is often a signal that the styles could be better-organized.)

The other thing to keep in mind is that "nesting" selectors like this is enabled by [Sass](https://sass-lang.com/) rather than CSS itself. Sass compiles this:

```scss
.card {
  .highlight {
    background-color: yellow;
  }
}
```

into this:

```css
.card .highlight {
  background-color: yellow;
}
```

The main benefit is to avoid duplicating the "containing" selector (`.card` here) when you have a lot of styles you want to only take effect within a certain element.

PaulJKim: Thank you! This was a very helpful explainer :)

PR #2041, opened 2024-05-02
Migrate to ESLint and resolve lint errors

digitalcora:
TSLint was deprecated in favor of ESLint several years ago: https://github.com/palantir/tslint/issues/4534

Additionally, running the linter has not been part of CI, so (much like type-checking and formatting) we've accumulated many lint issues in the code over time.

Rules where we have a huge number of outstanding issues (primarily `no-explicit-any` and `prop-types`) are set to `warn` for the time being, so they are still visible but won't fail CI. Remaining issues have been cleaned up.

---
  - https://app.asana.com/0/0/1207222640929033


cmaddox5 (approved):
Awesome, thank you so much for all of the code health work you have done so far.

PR #2043, opened 2024-05-09
fix: Crowding App alert logic

cmaddox5:
**Asana task**: ad-hoc

When looking at alerts in the crowding app CG, we were not filtering out upcoming. This means `any_alert_makes_this_a_terminal?/2` was returning true if an upcoming alert made the stop a terminal. 

- [ ] Tests added?


digitalcora (approved):
⚠️ 🕒 ✅

PR #2047, opened 2024-05-16
Gracefully handle no departures for Line Map

sloanelybutsurely:
**Asana task**: [[Screens 🐞] Gracefully handle no departures for Line Map](https://app.asana.com/0/1207094698667699/1206864414986293/f)

### Description

Handles the case when there fewer than two departures with predictions but also no departures _without_ a prediction.
- Previously this code would pass `nil` to `Screens.V2.Departure.time/1` which raised an exception 

#### Other changes

- Refactors counting of departures with predictions to use a single `Enum.count/2`
  - I'm happy to remove this if you all feel like it is out-of-place in this changeset

#### Known issues / thoughts

This doesn't eliminate all possible errors related to the `nil`ability of departures and the arrival and departure fields within schedules and predictions but is instead focused on the specific error reported to Sentry. A refactor to account for that class of errors would have a much larger surface area. Happy to address this if needed!

### DIFF
--- a/lib/screens/v2/widget_instance/line_map.ex
+++ b/lib/screens/v2/widget_instance/line_map.ex
@@ -265,6 +265,9 @@ defmodule Screens.V2.WidgetInstance.LineMap do
   def serialize_scheduled_departure(_departures, _direction_id, _stops, true = _is_terminal?),
     do: nil
 
+  def serialize_scheduled_departure([] = _departures, _direction_id, _stops, _is_terminal?),
+    do: nil
+
   def serialize_scheduled_departure(departures, direction_id, stops, _is_terminal?) do
     # Number of departures with predictions (not just schedules) in this direction
     prediction_count =
--- a/lib/screens/v2/widget_instance/line_map.ex
+++ b/lib/screens/v2/widget_instance/line_map.ex
@@ -273,23 +273,19 @@ defmodule Screens.V2.WidgetInstance.LineMap do
         &match?(%Departure{prediction: %Prediction{trip: %Trip{direction_id: ^direction_id}}}, &1)
       )
 
-    if prediction_count < 2 do
+    departure = Enum.find(departures, &is_nil(&1.prediction))
+
+    if prediction_count < 2 and not is_nil(departure) do
       %{name: origin_stop_name} = Enum.at(stops, 0)
 
-      if departure = Enum.find(departures, &is_nil(&1.prediction)) do
-        {:ok, local_time} =
-          departure
-          |> Departure.time()
-          |> DateTime.shift_zone("America/New_York")
+      {:ok, local_time} =
+        departure
+        |> Departure.time()
+        |> DateTime.shift_zone("America/New_York")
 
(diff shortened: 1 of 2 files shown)


lib/screens/v2/widget_instance/line_map.ex
```diff
@@ -268,27 +268,26 @@ defmodule Screens.V2.WidgetInstance.LineMap do
...
-      |> Stream.reject(fn %Departure{prediction: p} -> is_nil(p) end)
-      |> Stream.reject(fn %Departure{prediction: %Prediction{trip: t}} -> is_nil(t) end)
-      |> Stream.filter(fn %Departure{prediction: %Prediction{trip: %Trip{direction_id: d}}} ->
-        d == direction_id
-      end)
-      |> Enum.count()
+      Enum.count(
+        departures,
+        &match?(%Departure{prediction: %Prediction{trip: %Trip{direction_id: ^direction_id}}}, &1)
+      )
```
sloanelybutsurely, line 274: **note:** As called out in the PR description this isn't really related to this bug. I'm happy to remove this if you'd like.

digitalcora:
> I'm not a huge fan of nesting `if` expressions. This function could further refactored to use a `with` expression which would also return `nil` in cases where there are no passed `stops`, `DateTime.shift_zone/2` fails, or `Timex.format/2` fails. However, I wasn't sure if those possible runtime errors were desired so I left them in.

`with` expressions can include `=` assignments, which behave like "normal" assignments and crash if the pattern doesn't match. So if it improves clarity, it should be possible to reformat it as a `with` and not change the existing behavior. (I'd say crashes are desirable if they would always indicate a "programmer error", so they will cause a test failure or a Sentry report and we can fix them — and I think at least some of these fall into that category.)

sloanelybutsurely:
@digitalcora hmmm there were more `=` than there were `<-` _but_ i realized that the two if expressions could be combined so i went with that: [44487c8](https://github.com/mbta/screens/pull/2047/commits/44487c8dd41da637758f2d9ec7fdbebad9105457)

cmaddox5 (commented):
This looks great. Definitely like the most recent commit over the initial solution. I'm fighting myself on whether the other errors could be in scope just because the screen can still be useful if we don't have this piece. If another error happened, it would be good to log it but should let the screen render what it can. I agree with @digitalcora on the definition of a good crash, but don't think the date functions would error because of anything but a data issue. Would love to hear other thoughts though.

digitalcora:
Looking at the trade-off of the added complexity to log an error and continue anyway in this specific case, versus how likely the errors are to occur... I'd say it's questionably worth it. It's not like we do this in any disciplined fashion everywhere we use e.g. `Departure.time` currently, and if we suspected this was a data issue worth guarding against, we should build it into the return value of that function. (Given the arguments are in fact valid datetimes and zone names / format strings, I don't think `DateTime.shift_zone` or `Timex.format` would ever error.)

cmaddox5 (approved):

PR #2048, opened 2024-05-16
feat: departures multi-section auto-sizing 

digitalcora:
Best reviewed per-commit, there are a couple independent refactors.

* Update departures widget generation to pass layout parameters from configuration to the frontend.

* Replace existing auto-sizing implementation, which only worked for a single section, with a version that handles multiple sections using the layout parameters.

* The auto-sizing itself (independent of the React component) is unit-tested. These are the first real frontend tests in Screens, so we take the opportunity to:

  * Update Jest to its latest version.

  * Replace the third-party `@types/jest` package with a direct dependency on `@jest/globals`, which is already part of Jest. The only advantage of the former is that it allows using `describe`, `test`, etc. without importing them.

  * Add `eslint-plugin-jest` to our ESLint setup.

  * Add `fishery` for defining factories to build test data.

---
  - https://app.asana.com/0/0/1206932080790882

### DIFF
--- a/assets/src/components/v2/departures/departure_row.tsx
+++ b/assets/src/components/v2/departures/departure_row.tsx
@@ -1,17 +1,27 @@
-import React from "react";
+import React, { ComponentType } from "react";
 
-import RoutePill from "Components/v2/departures/route_pill";
+import RoutePill, { Pill } from "Components/v2/departures/route_pill";
 import Destination from "Components/v2/departures/destination";
-import DepartureTimes from "Components/v2/departures/departure_times";
+import DepartureTimes, {
+  TimeWithCrowding,
+} from "Components/v2/departures/departure_times";
 import DepartureAlerts from "Components/v2/departures/departure_alerts";
 
-const DepartureRow = ({
+type DepartureRow = {
+  id: string;
+  route: Pill;
+  headsign: Destination;
+  times_with_crowding: TimeWithCrowding[];
+  // currently never used, will be removed
+  inline_alerts: any[];
+};
+
+const DepartureRow: ComponentType<DepartureRow> = ({
   headsign,
   route,
   times_with_crowding: timesWithCrowding,
   inline_alerts: inlineAlerts,
 }) => {
-  const routeText = Number(route.text);
   return (
     <div className="departure-row">
       <div // Keep pill aligned to top if there is a variation for the headsign.
@@ -20,10 +30,7 @@ const DepartureRow = ({
           "departure-row__route" + (headsign.variation ? "" : " center")
         }
       >
-        <RoutePill
-          {...route}
-          size={isNaN(routeText) || routeText > 200 ? "small" : "large"}
-        />
+        <RoutePill {...route} />
       </div>
       <div className="departure-row__destination">
         <Destination {...headsign} />
--- a/assets/src/components/v2/departures/departure_row.tsx
+++ b/assets/src/components/v2/departures/departure_row.tsx
@@ -12,7 +12,7 @@ type DepartureRow = {
   route: Pill;
   headsign: Destination;
   times_with_crowding: TimeWithCrowding[];
-  // currently never used, will be removed
+  // currently the server always sets this to an empty array; will be removed
   inline_alerts: any[];
 };
 
--- a/assets/src/components/v2/departures/departure_time.tsx
+++ b/assets/src/components/v2/departures/departure_time.tsx
@@ -1,4 +1,4 @@
-import React from "react";
+import React, { ComponentType } from "react";
 import { classWithModifier } from "Util/util";
 
 const TextDepartureTime = ({ text }) => {
@@ -21,7 +21,7 @@ const TimestampDepartureTime = ({ hour, minute }) => {
   return <div className="departure-time__timestamp">{timestamp}</div>;
 };
 
-type Props =
+type DepartureTime =
   | (TextDeparture & { type: "text" })
   | (MinutesDeparture & { type: "minutes" })
   | (TimestampDeparture & { type: "timestamp" });
@@ -37,7 +37,7 @@ interface TimestampDeparture {
   minute: number;
 }
 
-const DepartureTime = ({ type, ...data }: Props) => {
+const DepartureTime: ComponentType<DepartureTime> = ({ type, ...data }) => {
   let inner;
   if (type === "text") {
     inner = <TextDepartureTime {...(data as TextDeparture)} />;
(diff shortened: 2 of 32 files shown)


assets/src/components/v2/departures/departure_row.tsx
```diff
@@ -1,17 +1,27 @@
...
 import DepartureAlerts from "Components/v2/departures/departure_alerts";
 
-const DepartureRow = ({
+type DepartureRow = {
+  id: string;
+  route: Pill;
+  headsign: Destination;
+  times_with_crowding: TimeWithCrowding[];
+  // currently never used, will be removed
+  inline_alerts: any[];
```
sloanelybutsurely, line 16: **question:** if this isn't used is there a reason to keep it in the type? passing additional properties to a react component should not raise a type error. then the `DepartureRow` component is explicit about only the things it _needs_.

assets/src/components/v2/departures/section.ts
```diff
@@ -0,0 +1,121 @@
...
+export const trimSections = (sections: Section[]): Section[] => {
+  const trimmed = _.cloneDeep(sections);
+
+  const sortedSizedSections = trimmed
+    .map((section) => ({ section, length: sectionLength(section) }))
+    .sort(({ length: a }, { length: b }) => b - a);
+
+  for (const key in trimStages) {
+    if (trimStages[key](sortedSizedSections)) return trimmed;
+  }
```
sloanelybutsurely, line 26: **question (non-blocking):** are we able to use `Object.values` here?

```suggestion
  for (const trimStage of Object.values(trimStages)) {
    if (trimStage(sortedSizedSections)) return trimmed;
  }
```

assets/src/components/v2/departures/section.ts
```diff
@@ -0,0 +1,121 @@
...
+  allToMax: (sections) => {
+    return sections.reduce((didTrim, { section, length }) => {
+      if (section.type == "normal_section" && section.layout.max) {
+        for (let l = length; l > section.layout.max; l--) {
+          if (trimSection(section)) {
+            // Trimmed this section; may be able to trim more, so keep going.
+            didTrim ||= true;
+          } else {
+            // This section can't be trimmed further, so return early.
+            return didTrim || false;
```
sloanelybutsurely, line 67: **question:** is this the same as `return didTrim`?

assets/src/components/v2/departures/departure_row.tsx
digitalcora: By "never used" I meant the server currently never returns anything in this field other than an empty array — it is still accessed in the frontend code (in this component, in fact), and there is a "property does not exist" error if this line is removed from the type. Because of this current "non-use" and expected future removal, and because it's somewhat outside the scope of the task, I didn't think it was important to fill in a detailed type for it. The wording of the comment could definitely be improved. 😅

assets/src/components/v2/departures/departure_time.tsx
```diff
@@ -37,7 +37,7 @@ interface TimestampDeparture {
   minute: number;
 }
 
-const DepartureTime = ({ type, ...data }: Props) => {
+const DepartureTime: ComponentType<DepartureTime> = ({ type, ...data }) => {
```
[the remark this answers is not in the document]
digitalcora: I used this pattern in places where a distinct "object" returned by the server corresponds exactly to the props of the component which renders an instance of that object. A `DepartureTime` (value) is a component which renders a `DepartureTime` (type). Within their namespaces, I think it makes sense to call each one of these by that name. But things don't always line up this way, as with e.g. the component one level above this, `DepartureTimes`: in this case the "props type" has a distinct name and is not exported.

```ts
export type TimeWithCrowding = { /* ... */ };

type Props = { timesWithCrowding: TimeWithCrowding[] };

const DepartureTimes: ComponentType<Props> = ({ timesWithCrowding }) => { /* ... */ };
```

I'd hesitate to say there's an "established pattern", since unfortunately a lot of the existing codebase just doesn't have explicit types at all (which is why we currently allow implicit-any). If anything it currently leans towards using separately-named "props types" in all cases. So I guess I'm proposing that we establish this as a pattern 🙂 What do you think?

In any case, thanks for calling this out, since I realized we should be using `import type` in places where we only use the type. It looks like the linter [has a rule to enforce this](https://typescript-eslint.io/rules/consistent-type-imports) but it's off by default. I'll see what happens when we turn it on.

assets/src/components/v2/departures/section.ts
digitalcora: Definitely. I think it was this way originally because I was logging `key` itself for debugging. Technically `trimStages` could just be an array instead of an object, but I do like the ability for the functions to have names.

assets/src/components/v2/departures/section.ts
digitalcora: Hah, yep. I put the `true` and `false` here to correspond with the possible returns of `trimSection`, wanting to make sure it was OR'd with the accumulator in both cases, not thinking about the fact that `|| false` is a no-op.

assets/src/components/v2/departures/departure_time.tsx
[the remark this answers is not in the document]
digitalcora: Okay, the lint rule gives 162 errors on the current codebase... it looks like it's less sophisticated than I thought, and just errors _whenever_ you import a type without using `type`, even if the name has no overlap with a value. (Though it seems like this could have advantages for bundle size anyway, so maybe we should think about turning it on, separately?)

I looked at all the imports across changed files in this PR, but as far as I can see, there actually isn't any place where I imported a value+type name and only used the type, so I haven't added any `type` keywords here.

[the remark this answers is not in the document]
sloanelybutsurely: > If anything it currently leans towards using separately-named "props types" in all cases. So I guess I'm proposing that we establish this as a pattern 🙂 What do you think?

This is my personal preference so I'm on-board!

As an aside, I also prefer to use something like [zod](https://zod.dev/) to parse external data to ensure that types for it are accurate and not just wishful thinking. 📌 We can put a pin in that for now though.

assets/src/components/v2/departures/departure_row.tsx
sloanelybutsurely: Ah okay. That makes sense. Thanks for the clarification!

sloanelybutsurely (approved):
LGTM (as far as I can tell that is...). It might still be a good idea to have someone else take a look because I'm really just looking at code with little context. I've made very little judgement based on behavior here.

We can type names as-is and have that discussion more broadly whenever we feel like it ("never" also an option for when to chat about it)

cmaddox5 (approved):
Really digging the approach here. And the code is organized in a way that makes this complicated operation very approachable.

PR #2051, opened 2024-05-17
fix: resolve webp file-loader warnings

cmaddox5:
**Asana task**: ad-hoc

The Triptych graphics are a little special in that they are always named after the panel they display on. The existing file-loader config didn't pull in parent folder so there was a conflict between the different PSA folders. Breaking out these files and adding `[folder]` to the `name` will allow the images to coexist.

- [ ] Tests added?

### DIFF
--- a/assets/src/util/util.tsx
+++ b/assets/src/util/util.tsx
@@ -28,7 +28,7 @@ export const imagePath = (fileName: string): string =>
   isOFM() ? outfrontImagePath(fileName) : `/images/${fileName}`;
 
 export const outfrontImagePath = (fileName: string): string =>
-  isTriptych() ? `triptych_images/${fileName}` : `/images/${fileName}`;
+  isTriptych() ? `triptych_images/${fileName}` : `images/${fileName}`;
 
 export const pillPath = (fileName: string): string =>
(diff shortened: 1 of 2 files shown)


sloanelybutsurely (approved):
lgtm! is there a reason we don't use things like `[contenthash]` or just add `/[folder]/` to the general image loader config?

digitalcora:
Yeah, it feels like this is not actually specific to `webp` images, just that `webp` images happen to be the one place where we run into this currently — but not necessarily in the future.

cmaddox5:
> lgtm! is there a reason we don't use things like `[contenthash]` or just add `/[folder]/` to the general image loader config?

My hesitation on using `contenthash` is due to one of our app types, `dup_v2`. It has a unique deployment process similar to `triptych_v2`. Eventually, I would like to break out the "packaged" dup app into it's own webpack entry (similar to `packaged_triptych` entry), but will leave that for a later task. When that happens, I believe we should be able to move to using hashes.

Side note: while testing the most recent commit, I found a bug in the packaged DUP image paths. Made that change here due to how small it is.

sloanelybutsurely (approved):

PR #2052, opened 2024-05-20
dev: add Stylelint and fix issues

digitalcora:


assets/stylelint.config.mjs
```diff
@@ -0,0 +1,33 @@
...
+    // We use blank lines to separate "groups" of imports.
+    "at-rule-empty-line-before": null,
+
+    // Docs: "We recommend turning this rule off if you use a lot of nesting."
+    "no-descending-specificity": null,
+
+    // Since we use SCSS exclusively, our `@import`s are all Sass imports
+    // (which are allowed to appear anywhere in a file) and not native CSS
+    // imports (which aren't, as this rule warns about).
+    "no-invalid-position-at-import-rule": null,
```
sloanelybutsurely, line 18: **question:** Does SCSS transpile font imports like these?

https://github.com/mbta/screens/pull/2052/files#diff-b08c54a442d55c9fb9eecde799b9acd6b6c717c63b3f0280fc56e9687486dddeL1-R1

sloanelybutsurely (approved):
I'm in favor of linting and standardization. In matters of style I have few opinions beyond "so long as a machine does it for me".

panentheos (approved):
Looks good to me, assuming we've spot-checked the result to make sure everything still looks okay.

assets/stylelint.config.mjs
digitalcora: It appears these are being compiled to use `url` syntax. The following:

```css
@import "https://rsms.me/inter/inter.css";
```

...comes out as this in the final CSS:

```css
@import url(https://rsms.me/inter/inter.css);
```

PR #2053, opened 2024-05-21
Sloane/wayfinding departures headers

sloanelybutsurely:
**Asana task**: [Departures widget: Implement "wayfinding" headers with new styling](https://app.asana.com/0/1185117109217413/1206602639264155/f)

#### Description

Following @digitalcora _very_ helpful notes, this does the following:

- Forwards `header` information from `ScreensConfig` through the API
- Updates the Screens React app to render the newly returned `header` information
- [ ] Updates the audio readout for the departures widget 
- Adds tests for server changes


assets/src/components/v2/departures/header.tsx
```diff
@@ -0,0 +1,40 @@
...
+        "rotate-90": direction === "ne",
+        "rotate-135": direction === "e",
+        "rotate-180": direction === "se",
+        "rotate-225": direction === "s",
+        "rotate-270": direction === "sw",
+        "rotate-315": direction === "w",
+      })}
+    />
+  );
+};
```
sloanelybutsurely, line 29: **note:** There is already a [similar component for "Solari"](https://github.com/mbta/screens/blob/main/assets/src/components/solari/arrow.tsx); however, while that component uses both `arrow.svg` and `arrow-45.svg`, this only uses `arrow-45.svg`. This produces slightly larger N, E, S, and W arrows (that are the same size as the oblique arrows but with a larger bounding box). This matches the [designs in Figma](https://www.figma.com/design/uyXu91sE66baBZMg81XOmk/ScreenKit?node-id=1212-2806&t=AF8TMdu5kySSqM5Z-4).

Screenshot 2024-05-22 at 8 32 56 AM
image

assets/src/components/v2/departures/header.tsx
```diff
@@ -0,0 +1,40 @@
...
+      data-direction={direction}
+      className={cx("direction-arrow", {
+        "rotate-45": direction === "n",
+        "rotate-90": direction === "ne",
+        "rotate-135": direction === "e",
+        "rotate-180": direction === "se",
+        "rotate-225": direction === "s",
+        "rotate-270": direction === "sw",
+        "rotate-315": direction === "w",
+      })}
```
sloanelybutsurely, line 26: **note:** I've pulled in `classnames` because it makes handling dynamic classes for components much easier. If there's opposition to introducing this library I'm happy to do this by hand.

assets/css/v2/busway/departures/header.scss
```diff
@@ -0,0 +1,19 @@
...
+  display: flex;
+  flex-direction: row;
+  align-items: center;
+  justify-content: space-between;
+  font-family: Inter, sans-serif;
+  font-size: 56px;
+  font-weight: 700;
+  line-height: 67.77px;
+  text-transform: uppercase;
+  letter-spacing: 3px;
```
sloanelybutsurely, line 11: **question:** Do we have shared classes for these kinds of typographic styles?

assets/css/v2/busway/departures/header.scss
```diff
@@ -0,0 +1,19 @@
...
+  background-color: #cccbc8;
+  border-top: 4px solid #b2b1af;
```
sloanelybutsurely, line 13: **question:** Do we have Sass/SCSS variables for colors that I should reference?

assets/css/busway_v2.scss
```diff
@@ -19,10 +19,10 @@
 @import "v2/lcd_common/no_data";
 @import "v2/lcd_common/page_load_no_data";
 
-@import "v2/busway/departures";
 @import "v2/lcd_common/departures";
-@import "v2/busway/departures_no_data";
 @import "v2/lcd_common/departures_no_data";
+@import "v2/busway/departures";
+@import "v2/busway/departures_no_data";
```
sloanelybutsurely, line 25: **note:** I reordered these so that busway rules would override conflicting lcd_common rules.

assets/src/components/v2/departures/header.tsx
```diff
@@ -0,0 +1,40 @@
...
+};
+
+const Header = ({ title, direction }: Props) => {
+  return (
+    <header className="departures-header">
+      {(title || direction) && <span>{title}</span>}
+      {direction && <DirectionArrow direction={direction} />}
+    </header>
+  );
+};
```
sloanelybutsurely, line 38: **note:** I believe the intent is for a header with a `read_as` value but no `title` to be audio only. If that's not the case this will need to be updated to fallback to `read_as` if `title` is not set.

lib/screens_web/views/v2/audio/departures_view.ex
```diff
@@ -17,6 +17,11 @@ defmodule ScreensWeb.V2.Audio.DeparturesView do
     ~E|<p><%= text %></p>|
   end
 
+  defp render_section(%{type: :normal_section, header: header, departure_groups: departure_groups})
+       when is_binary(header) do
+    ~E|<p><%= header %></p><%= Enum.map(departure_groups, &render_departure_group/1) %>|
+  end
```
sloanelybutsurely, line 23: **note:** I wasn't sure _how_ the header should be rendered for audio but this made sense to me.

test/screens/v2/candidate_generator/widgets/departures_test.exs
sloanelybutsurely, line 1: **note:** I've replaced a lot of the `==` in asserts with `=` pattern matches to make the assertions of each test more focused.

assets/src/components/v2/departures/header.tsx
digitalcora: [Jon would be happy](https://app.asana.com/0/1185117109217422/1201783781811458) 😄

assets/css/v2/busway/departures/header.scss
digitalcora: We already have `colors.scss` which would be a logical place to put them. Currently we do a lot of duplication of values even when they're explicitly a common design token (in Figma), but I for one would be interested in moving away from that tendency and I wouldn't expect many objections.

This particular color doesn't seem to be associated with a design token. It might be worth checking Screenkit to see if a corresponding value is actually in there and this instance just isn't linked to it.

assets/css/v2/busway/departures/header.scss
digitalcora: Things like font sizes and line heights tend to be fairly specific to each combination of component and screen type. I'm not aware of anything other than the mixins in `fonts.scss`, though some of this could be put in there.

assets/src/components/v2/departures/header.tsx
digitalcora: Correct. In reality I don't think we have any such screens currently (if an audio version of the header is configured, it always has a visual title as well), so we're free to decide how we want this to work. And I can see audio-only headers being useful in scenarios where we don't need the wayfinding but do want an audio "separator" between sections to match the visual separator that is still present when there's no title.

lib/screens_web/views/v2/audio/departures_view.ex
digitalcora: Seems reasonable. We will get a better feel for it once we do the full audio equivalence implementation for this screen type.

digitalcora:
@sloanelybutsurely Are those last 3 screenshots from a different version of the styles? The headers seem to have more vertical padding than they should. Other than that, looks very nice!

sloanelybutsurely:
> Are those last 3 screenshots from a different version of the styles? The headers seem to have more vertical padding than they should.

@digitalcora Ahh! Not sure how what happened but setting `box-sizing: border-box;` on the headers fixed this. Something must have been setting this higher up before. 093010f6e30eb22b9fc430d3dbd3b6e0a3d57817

assets/css/v2/busway/departures/header.scss
sloanelybutsurely: All set! f15e3fdbe3245191620e04967ef37cca05dd6294

assets/css/v2/busway/departures/header.scss
sloanelybutsurely: Also handled here: a9c7110212892826b5bfea10290cc28ebd3788f9

test/screens/v2/candidate_generator/widgets/departures_test.exs
digitalcora: This hurts my soul a tiny bit, because I really disagree with how the formatter indents assignments, but I guess there's not a whole lot we can do about it. 🫠

assets/css/busway_v2.scss
digitalcora: I think I'd prefer to set things up such that we minimize "overrides" — if a style differs between bus shelter and busway, then the things that are different should be in their respective screen-type-specific stylesheets, and there should not be any "default" in `lcd_common` that gets overridden. That said, if this were true, it wouldn't matter what order we import things in, so I wouldn't have any objection to this anyway.

test/screens_web/views/v2/audio/departures_view_test.exs
```diff
@@ -0,0 +1,71 @@
+defmodule ScreensWeb.V2.Audio.DeparturesViewTest do
+  use ScreensWeb.ConnCase, async: true
+
+  alias ScreensWeb.V2.Audio.DeparturesView
+
+  describe "no sections" do
+    setup :empty
```
digitalcora, line 7: This feels like slightly pre-emptive refactoring given we only use each of the values once. I think understanding these tests would be easier if, within each one, we defined the value and then did our assertion — no [mystery guests](http://xunitpatterns.com/Obscure%20Test.html#Mystery%20Guest) coming in from elsewhere that must be cross-referenced.

```elixir
assigns = %{
  sections: [
    %{type: :normal_section, departure_groups: []},
    %{type: :normal_section, departure_groups: []},
    %{type: :normal_section, departure_groups: []}
  ]
}

assert render(assigns: assigns) =~ "There are no upcoming trips at this time"

File: assets/css/v2/busway/departures.scss [hunk 9]
```diff
@@ -1,3 +1,7 @@
 .departures-container {
   height: 100%;
+  padding-top: 0;
 }
+
+@import "departures/header";
+@import "departures/arrow";
```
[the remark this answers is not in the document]
sloanelybutsurely: Hmmm 🤔. Moving these imports to the top of the file won't break anything so I'll go ahead and do that. However, the ordering and location of imports does impact the application of rules whose selectors have the same specificity. I'm not sure if we could move all existing not-at-the-top-of-the-file `@import`s to the top of the file and produce without introducing visual bugs.

We are probably best off asking for this in new code but leaving existing stuff as-is.

---

FWIW I was referencing the patterns established by `v2/lcd_common/departures.scss` when I did it this way:

https://github.com/mbta/screens/blob/ba872ea4d3e878122fc5b604a72877d436ba0f61/assets/css/v2/lcd_common/departures.scss#L1-L11

assets/src/components/v2/departures/header.tsx
```diff
@@ -0,0 +1,40 @@
...
+  title: string | null;
+  direction: CardinalDirection | null;
+}
+
+const DirectionArrow = ({ direction }: { direction: CardinalDirection }) => {
+  const src = useMemo(() => imagePath("arrow-45.svg"), []);
+  return (
+    <img
+      src={src}
+      data-direction={direction}
```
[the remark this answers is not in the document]
sloanelybutsurely: Not anymore, no. I used it to verify that the arrows were visually pointing the direction they're supposed to. I'll remove it.

assets/src/components/v2/departures/header.tsx
```diff
@@ -0,0 +1,40 @@
...
+
+type CardinalDirection = "n" | "ne" | "e" | "se" | "s" | "sw" | "w" | "nw";
+
+interface Props {
+  title: string | null;
+  direction: CardinalDirection | null;
+}
+
+const DirectionArrow = ({ direction }: { direction: CardinalDirection }) => {
+  const src = useMemo(() => imagePath("arrow-45.svg"), []);
```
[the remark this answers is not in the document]
sloanelybutsurely: All set! a75760ce

assets/src/components/v2/departures/header.tsx
```diff
@@ -0,0 +1,40 @@
...
+        "rotate-315": direction === "w",
+      })}
+    />
+  );
+};
+
+const Header = ({ title, direction }: Props) => {
+  return (
+    <header className="departures-header">
+      {(title || direction) && <span>{title}</span>}
```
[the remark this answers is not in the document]
sloanelybutsurely: Yes that's correct. The empty span ensures that a direction arrow where there is no title text is still rendered on the right side of the header. React does not render `null` values.

assets/src/components/v2/departures/normal_section.tsx
```diff
@@ -3,26 +3,37 @@ import weakKey from "weak-key";
...
 export type Layout = {
   base: number | null;
   max: number | null;
   min: number;
 };
 
+export type HeaderData = {
+  title: string | null;
+  arrow: Direction | null;
+  read_as: string | null;
```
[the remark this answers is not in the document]
sloanelybutsurely: 🫡 34aaf2fb

lib/screens/v2/widget_instance/departures.ex
```diff
@@ -20,7 +21,8 @@ defmodule Screens.V2.WidgetInstance.Departures do
   @type section :: %{
           type: :normal_section,
           rows: list(Departure.t() | notice()),
-          layout: Layout.t()
+          layout: Layout.t(),
+          header: Header.t() | nil
```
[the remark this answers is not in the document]
sloanelybutsurely: 🙅‍♀️ `nil` 🙅‍♀️ 

b1ada784

test/screens/v2/candidate_generator/widgets/departures_test.exs
```diff
@@ -85,37 +80,44 @@ defmodule Screens.V2.CandidateGenerator.Widgets.DeparturesTest do
...
-        %DeparturesWidget{
-          screen: config,
-          section_data: [%{type: :normal_section, layout: layout, rows: []}]
-        }
-      ]
+      assert [
+               %DeparturesWidget{
+                 screen: ^config,
+                 section_data: [%{layout: ^layout}]
+               }
```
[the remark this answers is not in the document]
sloanelybutsurely: ec9b6ade

test/screens_web/views/v2/audio/departures_view_test.exs
sloanelybutsurely: e1542793

assets/css/busway_v2.scss
sloanelybutsurely: I'm in a bit of a pickle with this then. Let's chat and figure it out!

<details>
<summary>🥒</summary>

| When removing order change | Padding | CSS |
| - | - | - |
| Screenshot 2024-05-24 at 9 25 33 AM | Screenshot 2024-05-24 at 9 25 48 AM | Screenshot 2024-05-24 at 9 26 09 AM |

</details>

assets/css/v2/busway/departures.scss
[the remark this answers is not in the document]
digitalcora: > I'm not sure if we could move all existing not-at-the-top-of-the-file `@import`s to the top of the file and produce without introducing visual bugs.

Yes, I couldn't easily verify whether this was the case when I was setting up Stylelint, so I chose to disable the rule (since the problem it's meant to warn you about doesn't actually exist in SCSS anyway). But I did note this in the SCSS docs:

> ⚠️ **Heads up!**
>
> The Sass team discourages the continued use of the `@import` rule. Sass will [gradually phase it out](https://github.com/sass/sass/blob/main/accepted/module-system.md#timeline) over the next few years, and eventually remove it from the language entirely. Prefer the [`@use` rule](https://sass-lang.com/documentation/at-rules/use) instead.

Sass `@use` has the same restriction as vanilla CSS `@import`: it can appear only as the first thing in a file. So thinking ahead to a possible future where we migrate to `@use`, it's probably for the best if we pre-emptively follow this restriction and make sure our styles work correctly that way.

assets/css/busway_v2.scss
digitalcora: Summary of huddle convo: Let's move the top-padding that's currently shared in `lcd_common` to only apply to `bus_shelter`. Then, when a busway screen (only) has a first section with no visual header, the header component adds this padding (instead of its border). This way, when there _is_ a visual header, it's flush with the main screen header, and when there isn't, the padding is the same as with the bus shelter design.

sloanelybutsurely: 7b748aa2 all set!

| First empty header | Other empty header |
| - | - |
| Screenshot 2024-05-24 at 12 10 01 PM | Screenshot 2024-05-24 at 12 10 12 PM |

assets/css/v2/busway/departures.scss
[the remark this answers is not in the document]
sloanelybutsurely: Oh good to know! I tried to swap out the few uses of `@import` I added but ran into issues with referencing the `font--inter--bold` mixin I added (even after also including a `@use` for `fonts.scss` in the file using the mixing). 

Looks like there's a migration tool though so we can probably make the switch using that in another PR: https://sass-lang.com/documentation/cli/migrator/

digitalcora (approved):
🚀

PR #2054, opened 2024-05-21
fix: Auth request fallback

cmaddox5:
**Asana task**: ad-hoc

Every once in a while, we get a Sentry error like [this](https://mbtace.sentry.io/issues/5046527403/events/0be23ee4507a4079bf7d2d99198dc695/) one. Adding this fallback will prevent an error while also keeping out unauthorized requests.

- [ ] Tests added?


digitalcora (approved):
Looks 👍. Maybe we could add an explanatory comment, like "Respond with 404 instead of crashing when the path doesn't match a supported provider". That way if we decide to do something different in the future, we know why this is here and what behavior we want to remain in place.

PR #2055, opened 2024-05-22
feat: V3 API Response Caching

cmaddox5:
**Asana task**: [title](url)

This PR adds Nebulex for API response caching. The `V3Api.get_json` function now maintains its own cache using the `if-modified-since` header. This will drastically reduce the number of network calls for screens. The motivation here was to reduce the amount of DUP timeouts. OFM controls app loading and every DUP loads at the same time. This causes a network request queue to form and any screen unlucky enough to be at the end of the queue never makes a request. 

This PR is likely only one piece to that puzzle, but testing on dev-green showed that this alone is a major improvement. Next step is to either consider using a different and more modern HTTP client or to tweak the hackney pool settings so that less connections are fighting over a spot in the queue.

- [ ] Tests added?


sloanelybutsurely (commented):
LGTM! I like that this also made it possible to simplifies some call sites as well.

Depending on the memory headroom of the servers running this application we might want to do some investigation into how large of an in-memory cache we are able to maintain and tune the Nebulex cache settings to match (or rather just to ensure there still headroom)

config/config.exs
```diff
@@ -412,6 +412,10 @@ config :screens, :screens_by_alert,
   screens_last_updated_ttl_seconds: 3600,
   screens_ttl_seconds: 40
 
+config :screens, Screens.ScreenApiResponseCache,
+  # => 1 hr
+  gc_interval: 3_600_000
```
digitalcora, line 417: Formatting came out a little weird here with the comment. As an alternative, can we do `:timer.hours(1)`?

lib/screens/v3_api.ex
```diff
@@ -59,6 +66,15 @@ defmodule Screens.V3Api do
     end
   end
 
+  defp update_response_cache(url, response, headers) do
+    date =
+      headers
+      |> Enum.into(%{})
+      |> Map.get("last-modified")
+
+    ScreenApiResponseCache.put(url, {response, date}, ttl: :timer.hours(1))
```
digitalcora, line 75: I think in theory, because we are always checking with the server whether our data is still fresh, the only reason cache entries should ever expire is because it's possible to reconfigure screens (or take them offline) such that a given URL is no longer accessed at all, in which case the cached response is just taking up space. But per the Nebulex [docs](https://hexdocs.pm/nebulex/Nebulex.Adapters.Local.html), `ttl` is not a mechanism for evicting unused data, as it's only evaluated when the data is retrieved, which in this scenario it never would be:

> The `:ttl` is evaluated on-demand when a key is retrieved, and at that moment if it's expired, then remove it from the cache, hence, it can not be used as eviction method, it is more for keep the integrity and consistency in the cache.

So I don't think we need to set a `ttl` here; the garbage collection should take care of cleaning up responses that aren't being actively used, and the If-Modified-Since checking should ensure we never use a stale response.

What we might want to do is set a maximum `allocated_memory` just to make sure we don't hit the instance's limit — the docs recommend setting either this or `max_size`, and neither is set by default. It appears we have 1GB to work with (in all environments) and currently use about 40% of it, so maybe 250MB for the cache?

cmaddox5: Interesting, good catch. 250MB sounds good, at least for now. If we see issues with this, we can adjust as needed.

PR #2057, opened 2024-05-24
fix: revert addition of generic font fallbacks

digitalcora:
This reverts the specific changes in 83d806c that added generic font fallbacks, and disables the Stylelint rule that required them. We found this caused problems with (at least) DUP and Bus e-Ink rendering on real hardware.

### DIFF
--- a/assets/css/connection_error.scss
+++ b/assets/css/connection_error.scss
@@ -10,7 +10,7 @@
   position: absolute;
   margin-top: 40px;
   margin-left: 1000px;
-  font-family: Inter, sans-serif;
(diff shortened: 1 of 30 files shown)


digitalcora:
@jzimbel-mbta Are you able to point your test hardware at dev-green to confirm whether this fixes the font issue?

cmaddox5 (approved):
Had a quick glance at all screen types and all looks good. I did notice a styling issue on DUPs on `main` that this branch resolves. We'll still want a live screen check, but I think the local dev improvement is better to have in `main` until then.

PR #2058, opened 2024-05-28
fix: DUP Section header support

cmaddox5:
**Asana task**: ad-hoc

Quick fix for DUPs. None of the `serialize_section` patterns were matching because the header was missing from the map. Added an empty struct similar to how `Layout` is handled.

- [ ] Tests added?


sloanelybutsurely (approved):
oops! this is my fault i think

PR #2061, opened 2024-05-31
Widgetize Later Departures

sloanelybutsurely:
**Asana task**: [Departures: Widgetize "Later departures"](https://app.asana.com/0/1207094698667699/1206645864760369/f)

#### Description

- Tracks `trimmedRows` in sections layout helpers
- Renders `trimmedRows` in a "Later Departures" component when `include_later === true`

<details>
<summary>Recordings 📹</summary>

https://github.com/mbta/screens/assets/1699281/dc5cb717-7cdf-4fb2-b51b-5c8030c078d9

https://github.com/mbta/screens/assets/1699281/884fb4be-4b9c-4e61-87d6-57111ac245db

</summary>

### DIFF
--- a/assets/css/v2/lcd_common/departures/later_departures.scss
+++ b/assets/css/v2/lcd_common/departures/later_departures.scss
@@ -1,0 +1,102 @@
+@use "sass:math";
+
+.later-departures__carousel {
+  position: relative;
+  display: flex;
+  flex-direction: row;
+  width: 100%;
+  transform: translateX(calc(-100% * var(--later-departures-offset)));
+}
+
+.later-departures {
+  box-sizing: border-box;
+  display: flex;
+  flex-direction: column;
+  flex-shrink: 0;
+  width: 100%;
+  background-color: $warm-neutral-90;
+
+  &__header {
+    box-sizing: inherit;
+    display: flex;
+    flex-direction: row;
+    justify-content: space-between;
+    margin: 0 32px;
+    border-bottom: 4px solid $true-grey-70;
+
+    h3 {
+      box-sizing: inherit;
+      flex-shrink: 0;
+      padding: 18px 0;
+      margin: 0;
+      font-family: Inter;
+      font-size: 40px;
+      font-weight: 500;
+      line-height: 43px;
+      color: $cool-black-15;
+      text-transform: uppercase;
+      letter-spacing: 2px;
+    }
+
+    ol {
+      display: flex;
+      flex-direction: row;
+      gap: 24px;
+      justify-content: space-between;
+      list-style: none;
+
+      .route-pill {
+        // RoutePill is using some position: absolute/relative stuff to center
+        // text vertically; however, it also relies on fixed widths. To get
+        // these route pills to grow horizontally when needed (and simplify
+        // doing that) the smaller RoutePills used in the Later Departures
+        // component use flexbox to center things instead. `position: static`
+        // is the default value so this rule resets the default behavior.
+        position: static;
+        box-sizing: border-box;
+        display: flex;
+        align-items: center;
+        justify-content: center;
+        width: auto;
+        min-width: 86px;
+        height: 44px;
+      }
+
+      .route-pill--yellow {
+        box-shadow: 1px 1px 1px 0 rgba(0 0 0 / 25%);
+      }
+
+      .route-pill__text,
+      .route-pill__slashed-text {
+        position: static;
+        padding: 0 14px;
+        font-size: 36px;
+        line-height: 36px;
+      }
+
+      .route-pill__slashed-text {
+        display: flex;
+      }
+    }
+  }
+
+  &__route--selected {
+    position: relative;
+
+    &::after {
+      $caret-size: 16px;
+
+      position: absolute;
+      bottom: -25px;
+      left: calc(50% - math.hypot($caret-size, $caret-size) / 2);
+      box-sizing: border-box;
+      width: $caret-size;
+      height: $caret-size;
+      content: " ";
+      background-color: $warm-neutral-90;
+      border-top: 4px solid $true-grey-70;
+      border-left: 4px solid $true-grey-70;
+      transform: rotate(45deg);
+    }
+  }
+}
--- a/assets/src/components/v2/departures/later_departures.tsx
+++ b/assets/src/components/v2/departures/later_departures.tsx
@@ -1,0 +1,87 @@
+import React, {
+  useEffect,
+  useLayoutEffect,
+  useMemo,
+  useRef,
+  useState,
+} from "react";
+import _ from "lodash";
+import cx from "classnames";
+
+import { hasOverflowX } from "Util/util";
+import useRefreshRate from "Hooks/v2/use_refresh_rate";
+
+import DepartureRow from "./departure_row";
+import RoutePill from "./route_pill";
+
+const MAX_LATER_DEPARTURES = 5;
+
+// "Magic" number carried over from Solari. Prevents a Later Departures
+// component from paging through departures very slowly when there are only a
+// few later departures
+const MAX_PAGE_RATE_MS = 3750;
+
+const LaterDepatures = ({ rows }: { rows: DepartureRow[] }) => {
+  const { refreshRateMs } = useRefreshRate();
+  const [limit, setLimit] = useState(MAX_LATER_DEPARTURES);
+  const departures = useMemo(() => _.take(rows, limit), [rows, limit]);
+
+  const ref = useRef<HTMLDivElement>(null);
+  const [currentDepartureIdx, setCurrentDepartureIdx] = useState(0);
+
+  useLayoutEffect(() => {
+    if (hasOverflowX(ref)) {
+      setLimit(Math.max(0, limit - 1));
+    }
+  }, [limit]);
+
+  useEffect(() => {
+    if (departures.length < 2) return;
+    const pageDurationMs = Math.min(
+      MAX_PAGE_RATE_MS,
+      refreshRateMs / departures.length,
+    );
+
+    const interval = setInterval(() => {
+      setCurrentDepartureIdx((i) => (i + 1) % departures.length);
+    }, pageDurationMs);
+
+    return () => clearInterval(interval);
+  }, [refreshRateMs, departures.length]);
+
+  return (
+    <div
+      className="later-departures__carousel"
+      style={
+        {
+          "--later-departures-offset": currentDepartureIdx,
+        } as React.CSSProperties
+      }
+    >
+      {departures.map((departure, i) => (
+        <div className="later-departures" key={departure.id}>
+          <div className="later-departures__header" ref={ref}>
+            <h3>Later Departures</h3>
+            <ol>
+              {departures.map((d, j) => (
+                <li
+                  key={d.id}
+                  className={cx({
+                    "later-departures__route--selected": i === j,
+                  })}
+                >
+                  <RoutePill {...d.route} outline={i !== j} />
+                </li>
+              ))}
+            </ol>
+          </div>
+          <div>
+            <DepartureRow {...departure} />
+          </div>
+        </div>
+      ))}
+    </div>
+  );
+};
+
+export default LaterDepatures;
--- a/assets/src/components/v2/departures/normal_section.tsx
+++ b/assets/src/components/v2/departures/normal_section.tsx
@@ -4,24 +4,35 @@ import weakKey from "weak-key";
 import DepartureRow from "./departure_row";
 import NoticeRow from "./notice_row";
 import Header from "./header";
+import LaterDepatures from "./later_departures";
 
 export type Layout = {
   base: number | null;
   max: number | null;
   min: number;
+  include_later: boolean;
 };
 
 export type Row =
   | (DepartureRow & { type: "departure_row" })
   | (NoticeRow & { type: "notice_row" });
 
-type NormalSection = {
+export interface NormalSection {
   layout: Layout;
   header: React.ComponentProps<typeof Header>;
   rows: Row[];
-};
+}
 
-const NormalSection: ComponentType<NormalSection> = ({ rows, header }) => {
+export interface NormalSectionWithLaterRows extends NormalSection {
+  laterRows: DepartureRow[];
+}
+
+const NormalSection: ComponentType<NormalSectionWithLaterRows> = ({
+  header,
+  layout,
+  rows,
+  laterRows,
+}) => {
   return (
     <div className="departures-section">
       <Header {...header} />
@@ -32,6 +43,9 @@ const NormalSection: ComponentType<NormalSection> = ({ rows, header }) => {
           return <NoticeRow row={row} key={weakKey(row)} />;
         }
       })}
+      {layout.include_later && laterRows.length > 0 && (
+        <LaterDepatures rows={laterRows} />
+      )}
     </div>
   );
 };
--- a/assets/src/components/v2/departures/route_pill.tsx
+++ b/assets/src/components/v2/departures/route_pill.tsx
@@ -90,6 +90,8 @@ const SlashedRoutePill: ComponentType<SlashedPill> = ({ part1, part2 }) => {
 const RoutePill: ComponentType<Pill> = (pill) => {
   const modifiers: string[] = [pill.color];
 
+  if (pill.outline) modifiers.push("outline");
+
   let innerContent: JSX.Element | null = null;
   switch (pill.type) {
     case "text": {
--- a/assets/css/colors.scss
+++ b/assets/css/colors.scss
@@ -10,4 +10,8 @@ $line-color-ferry: #008eaa;
 $alert-yellow: #fd0;
 
 $true-grey-70: #b2b1af;
+
 $warm-neutral-80: #cccbc8;
+$warm-neutral-90: #e5e4e1;
+
+$cool-black-15: #171f26;
--- a/assets/css/v2/lcd_common/departures.scss
+++ b/assets/css/v2/lcd_common/departures.scss
@@ -8,3 +8,4 @@
 @import "departures/time";
 @import "departures/crowding";
 @import "departures/alert";
+@import "departures/later_departures";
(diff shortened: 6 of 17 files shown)


assets/src/components/v2/departures/normal_section.tsx
```diff
@@ -19,9 +21,15 @@ type NormalSection = {
   layout: Layout;
   header: React.ComponentProps<typeof Header>;
   rows: Row[];
+  trimmedRows?: Row[];
```
digitalcora, line 24: `trimmedRows` makes sense as a name for the output of the "trimming" process, but from the perspective of this component, which isn't really involved in that, it feels odd to me. Maybe `laterRows`? (to align with `include_later` — if we "include later", then we include the "later rows")

Also, if there's a way we could make this non-optional, it seems like it would simplify some of the logic here and in the trimming process, especially since there's not a meaningful difference between the key being absent and being set to `[]`.

assets/src/components/v2/departures/later_departures.tsx
```diff
@@ -0,0 +1,80 @@
...
+const MAX_LATER_DEPARTURES = 5;
+
+const LaterDepatures = ({ rows }: { rows: Row[] }) => {
+  const [limit, setLimit] = useState(MAX_LATER_DEPARTURES);
+  const departures = useMemo(
+    () =>
+      _.take(
+        rows.filter(
+          (r): r is DepartureRow & { type: "departure_row" } =>
+            r.type === "departure_row",
```
digitalcora, line 27: Could we instead specify that the type of this component's prop (and of `trimmedRows`) is `DepartureRow[]`? Since ultimately we only ever trim / "later"-ify departure rows.

assets/src/components/v2/departures/later_departures.tsx
```diff
@@ -0,0 +1,80 @@
...
+        ),
+        limit,
+      ),
+    [rows, limit],
+  );
+
+  const ref = useRef<HTMLDivElement>(null);
+  const [idx, setIdx] = useState(0);
```
digitalcora, line 35: To be clear about what this is an index into, maybe something like `currentDepartureIndex`?

assets/src/components/v2/departures/later_departures.tsx
```diff
@@ -0,0 +1,80 @@
...
+
+  useLayoutEffect(() => {
+    if (hasOverflowX(ref)) {
+      setLimit(Math.max(0, limit - 1));
+    }
+  }, [limit]);
+
+  useEffect(() => {
+    if (departures.length < 2) return;
+    const refreshMs = Math.min(3750, 15000 / departures.length);
```
digitalcora, line 45: Can we extract the `15000` into a descriptively-named constant and calculate the `3750` from it? It'd be nice if we could actually use whatever the configured refresh rate of the screen is, but maybe not essential to untangle that right now.

assets/src/components/v2/departures/later_departures.tsx
```diff
@@ -0,0 +1,80 @@
...
+
+    const interval = setInterval(() => {
+      setIdx((i) => (i + 1) % departures.length);
+    }, refreshMs);
+
+    return () => clearInterval(interval);
+  }, [departures.length]);
+
+  const departure = departures[idx % departures.length];
```
digitalcora, line 54: There are two instances of `% departures.length`, here and on L48; do we need them both?

assets/src/components/v2/departures/later_departures.tsx
```diff
@@ -0,0 +1,80 @@
...
+              className={cx({
+                "later-departures--departure--selected": d === departure,
+              })}
+            >
+              <RoutePill {...d.route} outline={d !== departure} />
+            </li>
+          ))}
+        </ol>
+      </div>
+      <div className="later-departures__departures">
```
digitalcora, line 73: Maybe we should name this `__departure` or `__current-departure`, since it only displays one? (this also appears un-referenced in the CSS, not sure if intentional)

assets/src/components/v2/departures/later_departures.tsx
```diff
@@ -0,0 +1,80 @@
...
+  const departure = departures[idx % departures.length];
+
+  return (
+    <div className="later-departures">
+      <div className="later-departures__header" ref={ref}>
+        <h3>Later Departures</h3>
+        <ol>
+          {departures.map((d) => (
+            <li
+              key={weakKey(d)}
```
digitalcora, line 63: `DepartureRow`s have an `id` we could use.

assets/css/v2/lcd_common/departures/later_departures.scss
```diff
@@ -0,0 +1,83 @@
+@use "sass:math";
+
+.later-departures {
+  box-sizing: border-box;
```
digitalcora, line 4: We stan `border-box` 🔲

assets/css/v2/lcd_common/departures/later_departures.scss
```diff
@@ -0,0 +1,83 @@
...
+  display: flex;
+  flex-direction: column;
+  width: 100%;
+  min-height: 280px;
```
digitalcora, line 8: I saw the commit message about "departures with additional information" but wasn't sure what that meant; why is the explicit height needed?

assets/css/v2/lcd_common/departures/later_departures.scss
```diff
@@ -0,0 +1,83 @@
...
+        min-width: 86px;
+        height: 44px;
+      }
+
+      .route-pill--yellow {
+        box-shadow: 1px 1px 1px 0 rgba(0 0 0 / 25%);
+      }
+
+      .route-pill__text {
+        position: static;
```
digitalcora, line 56: Curious what the story behind this line is.

assets/src/components/v2/departures/later_departures.tsx
```diff
@@ -0,0 +1,80 @@
...
+              className={cx({
+                "later-departures--departure--selected": d === departure,
```
digitalcora, line 65: BEM-wise this would probably be like... `later-departures__route--selected`? Not gonna stress too much about the exact naming, though, feel free to ignore 🙂

assets/css/v2/lcd_common/departures/later_departures.scss
```diff
@@ -0,0 +1,83 @@
...
+        line-height: 36px;
+      }
+    }
+  }
+
+  &--departure--selected {
+    position: relative;
+
+    &::after {
+      $carot-size: 16px;
```
digitalcora, line 68: I think you meant `caret`? 🥕

assets/css/v2/lcd_common/departures/later_departures.scss
```diff
@@ -0,0 +1,83 @@
...
+
+      position: absolute;
+      bottom: -25px;
+      left: calc(50% - math.hypot($carot-size, $carot-size) / 2);
```
digitalcora, line 72: I have no doubt I'm going to be impressed by this, but I will first need an explanation of how and why the math works 😄

digitalcora (commented):
I apparently had a bunch of comments but it's really all small things — super impressed with how well you've integrated this and made it look perfect!

assets/src/components/v2/departures/normal_section.tsx
sloanelybutsurely: > Also, if there's a way we could make this non-optional, it seems like it would simplify some of the logic here and in the trimming process, especially since there's not a meaningful difference between the key being absent and being set to `[]`

The challenge here is that this type is used as both a value for React component props and as a representation of API response data. The `trimSections` function transforms this from one type to another. If we separated API data types from component props types we could easily have `trimSections` return the latter or just do this operation from `APINormalSection` -> `ClientNormalSection` before running `trimSections` so `trimSections` is only working on `ClientNormalSection`.

I guess what I'm trying to say is that removing "optional" from this key passes the type checker but is actually not an accurate representation of the data flowing through the system.

Would you feel okay about leaving this for now and revisiting how we handle data from t he API and its types more holistically?

assets/src/components/v2/departures/route_pill.tsx
```diff
@@ -90,6 +90,8 @@ const SlashedRoutePill: ComponentType<SlashedPill> = ({ part1, part2 }) => {
 const RoutePill: ComponentType<Pill> = (pill) => {
   const modifiers: string[] = [pill.color];
 
+  if (pill.outline) modifiers.push("outline");
```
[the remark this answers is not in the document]
sloanelybutsurely: There are both `.route-pill--outline` and `.route-pill__text--outline`. Am I missing something?

#### route-pill--outline
https://github.com/mbta/screens/blob/9c4c0711acbb70d6d9e52072bfa23aefe48aaecd/assets/css/v2/lcd_common/route_pill.scss#L40-L44

#### route-pill__text--outline
https://github.com/mbta/screens/blob/9c4c0711acbb70d6d9e52072bfa23aefe48aaecd/assets/css/v2/lcd_common/route_pill.scss#L70-L72

assets/src/components/v2/departures/later_departures.tsx
sloanelybutsurely: Nope, we only need one of them!

assets/css/v2/lcd_common/departures/later_departures.scss
sloanelybutsurely: I was referring to stuff like this:

image

Without the `min-height` these caused layout shift

assets/css/v2/lcd_common/departures/later_departures.scss
sloanelybutsurely: I had to change the layout for the smaller route pills so they could expand based on their contents.

The existing route pill component uses `position: relative` and `position: absolute` hacks to achieve horizontal and vertical centering of their text. This relied on fixed route pill widths. `position: static` resets position to the default content flow. Then I use flexbox to get the same overall centering.

assets/css/v2/lcd_common/departures/later_departures.scss
sloanelybutsurely: I did! I _vividly_ remember fixing this misspelling already but alas my memory is fallible 🙃

assets/src/components/v2/departures/later_departures.tsx
sloanelybutsurely: Happy to make that change. I've never really used a class name convention like this. I've been a CSS modules (or similar) advocate for a long time but I can learn new tricks!

assets/src/components/v2/departures/normal_section.tsx
digitalcora: > The `trimSections` function transforms this from one type to another. If we separated API data types from component props types

We do already in some places, for example `DepartureTimes`, and I think it would be a reasonable approach here. If you think that separation will be too hairy to perform as part of this task, it's probably fine, I just think it could be confusing that (as you say) we're conceptually going from one distinct type to another but using the same type in both cases.

assets/src/components/v2/departures/route_pill.tsx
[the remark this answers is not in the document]
digitalcora: Ah, I see, you added both of the styles in this PR — the `__text--outline` was already present in the main branch but had no styles applied to it. Ignore me~

assets/src/components/v2/departures/later_departures.tsx
sloanelybutsurely: They do but `trimSections` "flattens" `DepartureRow`s with multiple times into multiple `DepartureRow`s reusing the same `id`. I do this to reduce layout shift when rotating through later departures. It does have the unfortunately mean that there can now be multiple `laterRows` with the same `id` ☹️

assets/css/v2/lcd_common/departures/later_departures.scss
digitalcora: Ah, now I see (literally) what you mean. It would be nice if we didn't have to hard-code this... total armchair-engineering here, what if all "later departures" were rendered at all times, horizontally next to each other but all taking up 100% width, so only one is visible at a time? Then "rotating" the array by `currentDepartureIdx` places the current departure in the first position, making it visible — but the heights of all departures are always contributing to the height of the container.

```
visible | hidden in overflow
A       | B C D E
[rotate]
B       | C D E A
```

Anyway I do think the current solution is fine, but if we stick with that, maybe we should add a comment explaining why it's here.

assets/css/v2/lcd_common/departures/later_departures.scss
sloanelybutsurely: If I remove the `left` and `transform` values we have a square that is aligned with the left side of the bounding box around the route pill

Screenshot 2024-06-03 at 10 59 04 AM

Then if I set `left: 50%` the square isn't centered within the bounding box of the route pill but instead its left side is.

Screenshot 2024-06-03 at 10 59 18 AM

So I need to set a left value slightly less than `50%`: `50% - half of the width of my caret`. This would simply be `calc(50% - $caret-size / 2)` _if_ I wasn't also going to rotate it 45deg...

Screenshot 2024-06-03 at 10 59 23 AM

By rotating it its width is now something other than `$caret-size` (`16px`)

Screenshot 2024-06-03 at 11 04 52 AM

I _could_ just copy this computed value into the CSS but that makes it difficult to ever change the size of the caret. Luckily this value is actually the hypotenuse of one of the two triangles that make up this square! $\sqrt(16px) + \sqrt(16px) = \sqrt(box width)$. Good ol' Pythagoras helping us out!

Finally, SCSS's `math.hypot` will solve this equation for us. Now if we change the size of the caret it will remain horizontally centered beneath the route pill.

---

An aside: Is this the _best_ way to do this? I'm not sure! Is it the way I thought up first? Yes, it is. If this is too strange I'm happy to noodle on other approaches **or** if you have an idea I'm happy to try that out too!!

assets/css/v2/lcd_common/departures/later_departures.scss
digitalcora: Would it be feasible to backport your flexbox improvements to route pills outside this context? If not that's fine, though I think we should add a comment here explaining what's going on (that this is overriding a value used to do centering on route pills elsewhere, so the centering can be done in a different way).

assets/src/components/v2/departures/later_departures.tsx
digitalcora: Hah, I think CSS modules would more qualify as the "new thing", at least by my reckoning of how old BEM is at this point. I think I've already said so in another PR but I would also be interested in how we might adopt CSS modules, if and when the opportunity arises.

My understanding: BEM is a naming convention (family of conventions?) that divides UI into Blocks, which contain Elements, which may have Modifiers. Blocks are complete reusable components, Elements are the individual parts of them, Modifiers are different "states" or variants the same block/element could have. Classes are then named like `block-name__element-name--modifier-name` (if there are multiple modifiers, each gets its own class, with the full block/element name repeated for each one). There's other nuances, but it seems to me this project uses BEM less as part of an opinionated system and more just as a handy way to avoid global namespace collisions.

assets/css/v2/lcd_common/departures/later_departures.scss
digitalcora: You know it's a good explanation when you have to break out the LaTeX 😄 Thanks for going over this, and it's good to also have it documented here with visuals.

assets/src/components/v2/departures/later_departures.tsx
digitalcora: Ah, right, makes sense. I think flattening the times is the intended behavior anyway (or at least how v1 behaves currently), otherwise Later Departures would consume even more vertical space, when the point of it is to save space.

This does seem like a case where someone (maybe future-us) could come across this, have the same thought I did, and "improve" it to use the `id`, unwittingly introducing a rendering bug. Maybe we could instead fix it by generating new IDs for the synthesized rows when we create them? Thus having basically the same effect as using `weakKey`, but preserving the contract that `id` is unique.

assets/css/v2/lcd_common/departures/later_departures.scss
sloanelybutsurely: We can totally get all of the route pills using flexbox instead but without an easy way to verify every possible route pill still looks correct I'm not comfortable making those changes. 📌 

This would be a great opportunity for Storybook or something similar.

Do we keep track of "tech debt" or "improvement" tasks any differently than other kinds of tasks in Asana?

sloanelybutsurely: I've added a comment explaining this code for now 32ca43f3

assets/css/v2/lcd_common/departures/later_departures.scss
sloanelybutsurely: Oh I love this idea! Very clever. I'll give it a shot.

assets/src/components/v2/departures/later_departures.tsx
digitalcora: Erm, I see you've pushed a change which now keeps times grouped when they're sent into later-departures, which is not the thing we want? Or at least, it's different from how v1 works, and I'm not sure we should change that in the absence of a design saying we should.

> I think flattening the times is the intended behavior anyway (or at least how v1 behaves currently)

assets/css/v2/lcd_common/departures/later_departures.scss
digitalcora: Huh, I was thinking of just the departure rows themselves being rendered all next to each other, rather than a full copy of Later Departures for each rotation. This works too though!

assets/src/components/v2/departures/later_departures.tsx
sloanelybutsurely: Oops! Easy enough to undo that. I had originally flattened the times because I ran into those layout issues I mentioned

assets/src/components/v2/departures/later_departures.tsx
sloanelybutsurely: ✅ [`056b6b5..232379d`](https://github.com/mbta/screens/pull/2061/files/056b6b57066a5fb1e14a28f15d6d1c48aa2a7f4c..232379d324176aece6ddefd06210fa3ce39da421)

digitalcora (approved):
Looks great! If it's not too much work to untangle, it'd be nice to have the `useRefreshRate` change as a separate refactor commit, then a commit with the main work to add Later Departures. (But I can see that being a bit of work to untangle.)

sloanelybutsurely:
> If it's not too much work to untangle, it'd be nice to have the `useRefreshRate` change as a separate refactor commit, then a commit with the main work to add Later Departures.

easy peasy!

PR #2063, opened 2024-06-05
No-departures state for Sectional sections

digitalcora:
* Add basic style support for "notice rows", containing FreeText data, to the `lcd_common` screen types (bus shelter and sectional/busway).

* Implement the "no departures" message as a notice row added to sections that would otherwise be empty, on screens with multiple sections (currently only Sectional, discounting DUPs which have their own candidate generator). Single-section screen types retain whatever their current behaviors are.

* Filter out departures during fetching if their route type is devops-disabled, resulting in the same message appearing if a section is displaying data exclusively for a devops-disabled mode. This should only affect Sectionals, as DUPs and the other (single-section) screen types have their own logic for handling devops-disabled modes that is upstream of departure fetching.

Given some of the code already has the concept of "notice sections" and "no-data sections", the choice to use a normal section with a "notice row" for this feature is mostly one of convenience and flexibility: Notice rows were already in a fully-working state on `lcd_common` screens other than being unstyled (including audio serialization), and this leaves open the possibility that free-text notices could mingle with other kinds of rows in the same section.

<details>
Left: an "organic" instance of no departures (no trips currently stop at the Harvard lower busway). Right: the effect of setting `subway` to devops-disabled in my local config.
<table>
<tr>
<td></td>
<td></td>
</tr>
</table>
</details>

---
  - https://app.asana.com/0/0/1206971590492831


sloanelybutsurely (approved):
looks good to me!

lib/screens/v2/candidate_generator/widgets/departures.ex
```diff
@@ -79,24 +108,80 @@ defmodule Screens.V2.CandidateGenerator.Widgets.Departures do
...
+          stop_ids: stop_ids
+        },
+        fn {_key, value} -> value == [] end
+      )
+
+    route =
+      case route_fetch_fn.(fetch_params) do
+        {:ok, [route]} -> route
+        :error -> nil
+      end
```
sloanelybutsurely, line 140: This is raising a `CaseClauseError` in dev. Looks like we just need to adjust the pattern matches.

[Sentry - SCREENS-A5](https://mbtace.sentry.io/issues/5459051453/)

PR #2065, opened 2024-06-07
fix: expand patterns to prevent CaseClauseError

sloanelybutsurely:
**Sentry issue**: [SCREENS-A5](https://mbtace.sentry.io/issues/5459051453/)
**Asana:** [Fix crash on MUL-108](https://app.asana.com/0/1207094698667699/1207495619155907/f)

Updates `case` to take any non-empty list and return the first value and return `nil` for anything else.

---
  - https://app.asana.com/0/0/1207495619155907


digitalcora (approved):
Good catch, I had already noticed the crash this was causing and [created a subtask for it](https://app.asana.com/0/1185117109217413/1207495619155907) but hadn't dug into the exact cause. (Edit: and I see you already found that subtask and linked to it!)

PR #2067, opened 2024-06-11
chore: replace custom metadata plug for remote_ip

sloanelybutsurely:
See also: https://github.com/mbta/screenplay/pull/366


sloanelybutsurely:
@digitalcora i decided to just write out both values just in case

PR #2072, opened 2024-06-13
fix: use CR abbreviations in Later Departures

digitalcora:
"Route abbreviations" (formally named in this changeset) are short text codes that stand in for the names of routes which would otherwise be represented by a "mode icon" in route pills. Currently this only applies to Commuter Rail, though could conceivably apply to ferry in the future.

Previously, whether a "route abbreviation" was used was controlled at the time the route pill was serialized — the result from the client's perspective would either be an "icon pill" with the mode icon or a "text pill" with the abbreviation. However, the "Later Departures" feature requires that the client know both of these pieces of information at once, since which one is used depends on context that is not known until the widget is rendered.

Accordingly, this adds a field `route_abbrev` to the serialized form of icon and text pills (needed for text pills as well because Commuter Rail departures where the train has a track assignment are represented with a "TRx" text pill). The client's `RoutePill` component gains a prop that determines whether to use the abbreviation in the pill data. If true, it renders a text pill with the abbreviation as the content.

Screenshot 2024-06-13 at 11 45 46 AM

---
  - https://app.asana.com/0/0/1207495619155884

### DIFF
--- a/assets/src/components/v2/alert.tsx
+++ b/assets/src/components/v2/alert.tsx
@@ -48,7 +48,11 @@ const BaseAlert: ComponentType<BaseAlertProps> = ({
             )}
           >
             {routePills.map((pill) => (
-              <RoutePill {...pill} key={routePillKey(pill)} />
+              <RoutePill
+                pill={pill}
+                useRouteAbbrev={true}
+                key={routePillKey(pill)}
+              />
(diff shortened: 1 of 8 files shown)


sloanelybutsurely (approved):
looks great!

PR #2073, opened 2024-06-17
fix: use Later Departures when it saves space

digitalcora:
We may not always want to display "Later Departures" even when it is enabled and some rows have been trimmed, because the carousel itself can take up more space than just displaying the departures normally would have.

The "magic constant" 3 is an imperfect guess that errs on the side of not introducing Later Departures unless it will almost certainly save space (it also happens to be the same constant used in the v1 Solari logic). Using layout effects to measure how much space is actually taken up by each approach is left as a potential future improvement.

#### About the refactor commit...

At some point I was convinced that to do this correctly, the section component would need access to the _original_ departure rows, without the destructive trimming applied. That prompted a refactor where the trio of original rows, trimmed rows, and "rows that were trimmed" were all available as part of the processed sections. To be clear about which were which, I named these the "original", "above-the-fold", and "below-the-fold" departures, and then thought it made sense to name the sections themselves "folded sections" (also more concise than "sections with later rows").

Eventually I worked out that not only was having the original rows unnecessary, it also prevented auto-sizing from working as expected with the `max` field set. So, this refactor is now purely one of naming/structure ("original rows" not included), and could be reverted without affecting anything. I think it's still a (minor) improvement in legibility, but please review it in that light!

### Example

This is a representative case where the new logic allows the two departures that would have otherwise been in Later Departures to be displayed at the same time (and even saves a few pixels of space).

<table>
<tr><th>Before</th><th>After</th></tr>
<tr>
<td>

</td>
<td>

</td>
</tr>
</table>

---
  - https://app.asana.com/0/0/1207519791642956

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -102,6 +102,15 @@ jobs:
       - working-directory: ./assets
         run: npm run styles:check
 
+  js-tests:
+    name: Frontend Tests
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: ./.github/actions/setup
+      - working-directory: ./assets
+        run: npm test
+
   js-types:
     name: Frontend Types
     runs-on: ubuntu-latest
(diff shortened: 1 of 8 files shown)


cmaddox5 (approved):
Looks good! I agree, the refactor does help a bit with legibility.

PR #2056, opened 2024-05-23
Add Park Street and Arlington to DUP headsign configs

sloanelybutsurely:
**Asana task**: [Add Park Street and Arlington to DUP headsign configs](https://app.asana.com/0/1185117109217422/1207171583176322/f)

#### Description

<details>
<summary>Non-exhaustive examples 🖼️</summary>

Screen Shot 2024-05-30 at 07 59 16
Screen Shot 2024-05-30 at 08 00 46
Screen Shot 2024-05-30 at 08 02 27
Screen Shot 2024-05-30 at 08 04 26
Screen Shot 2024-05-30 at 08 05 16

</details>

PR #2081, opened 2024-06-20
feat: Automate `dup_alert_headsign_matchers` configuration generation

sloanelybutsurely:
**Asana task**: [[Retro] Document DUP headsign config and adding new route to PA/ESS](https://app.asana.com/0/1185117109217413/1207495619156084/f)

### Description

In attempting to document how to add new stations to the DUP alert headsign configuration I realized that the process required **A LOT** of complicated, easy to mess up cross-referencing of V3 API requests. While working on #2056 I imported static GTFS into a sqlite database to join across the required relations but that was a bit much to document.

I figured a script that output the correct config for you was probably just as helpful if not more so!

---
  - https://app.asana.com/0/0/1207348699397552

### DIFF
--- a/scripts/add_dup_headsign_config.exs
+++ b/scripts/add_dup_headsign_config.exs
@@ -27,14 +27,18 @@ defmodule AddDUPHeadsignConfig do
           using the `--headsign_0` and/or `--headsign_1` options.
 
 
+  NOTE: Pay extra attention when dealing with branching stations! You will likely
+  have to manually alter the sets of `informed` and `not_informed` stations to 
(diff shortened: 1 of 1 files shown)


cmaddox5 (approved):
Nice! Thanks for putting this together!

One caveat to this is that this probably wouldn't be enough for branching GL stations like Kenmore. We have matchers there that look at sets of informed vs not informed so we can give a rider friendly branch combo headsign. Those cases will always be too complicated for a script so it should be enough to add a note to this instead of trying to make that work.

PR #2082, opened 2024-06-20
fix: rogue semicolon on screen pages

digitalcora:
Huh, that's weird... why is the screen page 1939.2px tall, rather than 1920px?

Screenshot 2024-06-20 at 5 50 12 PM

Nothing seems out of place here...

Wait... zoom in...

THERE

Screenshot 2024-06-20 at 5 46 09 PM


cmaddox5 (approved):
_**ENHANCE**_

PR #2089, opened 2024-06-25
fix: replace remaining prop spreads into RoutePill

sloanelybutsurely:
#### Description

passes the correct props to remaining `<RoutePill />` instances following #2072

### DIFF
--- a/assets/src/components/v2/cr_departures/cr_departures_header_free.tsx
+++ b/assets/src/components/v2/cr_departures/cr_departures_header_free.tsx
@@ -11,7 +11,7 @@ const CRDeparturesHeaderFree = ({ headerPill }) => {
       <div className="departures-card__header-text">
         <span>Free Commuter Rail during</span>
         <span className="pill-container">
-          <RoutePill {...headerPill} />
+          <RoutePill pill={headerPill} />
         </span>
         <span>disruption</span>
       </div>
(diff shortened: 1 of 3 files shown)


digitalcora (approved):
Yikes! I thought I had gotten all of these... evidently the compiler didn't have my back. (I'm a little mystified as to why, though. In this case, aren't we failing to provide a required prop?)

sloanelybutsurely:
> I'm a little mystified as to why, though. In this case, aren't we failing to provide a required prop?

@digitalcora Because these components do not have types for their own props their props have an implicit `any` type. Passing `any` in a spread like this basically disables any meaningful type checking.

[Minimal reproduction of the issue](https://www.typescriptlang.org/play/?#code/JYWwDg9gTgLgBAJQKYEMDG8BmUIjgcilQ3wG4AoctCAOwGd4EIBXGJABWABsu4BeOAAoA3mG5cAvgC44wuGJ4zh1LtBkMowGgHMJEgJT8AfEPJw4AHjpgUNI6PEA6FdAkWA9NdtHy+ytXp4AFkAT3YUbSR+ITAcMDoZWxDDPhNhMzgiGGYoGlNzcwsmVg5xWUcK2Ih4xwVJOHcfcz8JIA) (explicit `any` here because the TypeScript playground has "no implicit any" turned on by default)

digitalcora:
> Because these components do not have types for their own props their props have an implicit `any` type.

Oof. I suppose things like this will be with us for a bit, then. Gotta keep whittling away at those `any`s...

PR #2092, opened 2024-06-27
fix: use `runtime.exs`, start `:httpoison` and `:hackney_telemetry` in prod (release)

sloanelybutsurely:
- Reverts #2093 
    - Installs and configures `:hackney_telemetry`
    - renames `handle_span` to `log_span` in `Screens.Telemetry`
    - Logs hackney telemetry data
- Moves from deprecated `releases.exs` file to `runtime.exs`
    - Moves existing contents into `config_env() == :prod` check to match `releases.exs` behavior (as much as we were using that behavior)
- Adds `{:ok, _} = ` matches to calls to `Application.ensure_all_started` in `runtime.exs`
    - Starts `:httpoison` instead of `:poison` which always failed
- Starts `:hackney_telemetry` before `:hackney` in `runtime.exs` to prevent crashing when calling `ExAws.SecretsManager`

<details>
<summary>dev-green results</summary>

image

</details>


mix.exs
```diff
@@ -81,14 +81,15 @@ defmodule Screens.MixProject do
       {:stream_data, "~> 1.1", only: :test},
       {:memcachex, "~> 0.5.5"},
       {:aja, "~> 0.6.2"},
-      {:telemetry, "~> 1.2"},
+      {:telemetry, "~> 1.2", override: true},
```
digitalcora, line 84: What is this overriding? Maybe we should have a comment explaining it.

sloanelybutsurely: 🫡 https://github.com/mbta/screens/pull/2092/commits/eeaf975d956ddf7bb67f4d7f848cbd384bfb5ca0

digitalcora (approved):
Nice sleuthing, and a great upgrade to `runtime.exs` as well!

PR #2095, opened 2024-06-27
tweak: Suppress Alewife platform closure alert.

cmaddox5:
**Asana task**: ad-hoc

The Alewife platform is being bypassed this weekend. The alert was created using a new Alerts UI feature that the Pre-Fare screens cannot handle yet. We will use a static image in its place.

Note: the scenario we are testing here is suppressing the above alert and letting another shuttle alert render normally. There is a flex zone alert bug keeping this branch from displaying the alert properly. That is being resolved in #2094.

- [ ] Tests added?

### DIFF
--- a/lib/screens/v2/widget_instance/reconstructed_alert.ex
+++ b/lib/screens/v2/widget_instance/reconstructed_alert.ex
@@ -1270,7 +1270,9 @@ defmodule Screens.V2.WidgetInstance.ReconstructedAlert do
   def alert_ids(%__MODULE__{} = t), do: [t.alert.id]
 
   def valid_candidate?(%__MODULE__{} = t) do
-    t.alert.id not in ["197140", "580015"]
+    test_alert_ids = ["197140"]
+    prod_alert_ids = ["580015"]
+    t.alert.id not in (test_alert_ids ++ prod_alert_ids)
   end
 
   defimpl Screens.V2.WidgetInstance do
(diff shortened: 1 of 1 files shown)


lib/screens/v2/widget_instance/reconstructed_alert.ex
```diff
@@ -1270,11 +1270,7 @@ defmodule Screens.V2.WidgetInstance.ReconstructedAlert do
   def alert_ids(%__MODULE__{} = t), do: [t.alert.id]
 
   def valid_candidate?(%__MODULE__{} = t) do
-    suppressed =
-      t.alert.id in ["549108", "555079", "555080"] and
-        t.location_context.home_stop === "place-gover"
-
-    not suppressed
+    t.alert.id not in ["197140", "580015"]
```
digitalcora, line 1273: Where did these IDs come from? When I try to look up either of them in the API, I get a 404 error. Looking at the alerts currently in the feed, it seems like these should be `571855` and `578253`.

cmaddox5: `197140` is the alert ID that Dave created in test. I can put the values in variables so that they are more obviously labeled. The other ID is in prod so I'm not sure why the API isn't returning it. I found it on the `Future` tab in AlertsUI. Based on the discussion in [this thread](https://mbta.slack.com/archives/C03K6NLKKD1/p1718222171213539), the alert that will be used for the closure will be a station closure instead of a service change. That may be why `571855` starts and ends at the same time. They are only using it in notifications but will switch to `580015` when the alert is in effect. I will triple check with Dave.

digitalcora: Huh, okay, I think I see what they're doing there... it doesn't really make sense to me (why two alerts? how is the second one being hidden from the feed?), but the ID is at least correct to what they've told us.

Incidentally, it looks like 571855 is now active _currently_, until tonight. Maybe we should add that one here in addition to 580015?

cmaddox5: Because `571855` is a Service Change, it won't show on these screens anyway.

digitalcora (approved):

PR #2096, opened 2024-06-27
feat: log error with context on timeout

sloanelybutsurely:
I'm still stuck on #2091 but I think this will be helpful in the mean time.

### DIFF
--- a/lib/screens/v2/candidate_generator/dup/departures.ex
+++ b/lib/screens/v2/candidate_generator/dup/departures.ex
@@ -225,9 +225,21 @@ defmodule Screens.V2.CandidateGenerator.Dup.Departures do
             create_station_with_routes_map_fn,
             now,
             ctx
-          )
+          ),
+          on_timeout: :kill_task
         )
-        |> Enum.map(fn {:ok, data} -> data end)
+        |> Enum.flat_map(fn
+          {:ok, data} ->
+            [data]
+
+          {:exit, reason} ->
+            ctx =
+              Screens.Telemetry.context()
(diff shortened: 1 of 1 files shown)


lib/screens/v2/candidate_generator/dup/departures.ex
```diff
@@ -225,9 +225,21 @@ defmodule Screens.V2.CandidateGenerator.Dup.Departures do
...
+          {:ok, data} ->
+            [data]
+
+          {:exit, reason} ->
+            ctx =
+              Screens.Telemetry.context()
+              |> Enum.map_join(" ", fn {k, v} -> "#{k}=#{v}" end)
+
+            Logger.error(["event=get_section_data.exit reason=#{reason} ", ctx])
+            []
```
digitalcora, line 241: The previous behavior was that the parent process (the whole screen data request) would crash on a timeout, resulting in the screen displaying an "offline" message. This instead acts like the offending section doesn't exist / isn't configured, which I'm hesitant about since it seems like it could have worse rider-facing consequences, like the screen/slot just being blank (and not showing any error message). Ideally I think we'd want to generate a `no_data_section`, but that currently requires a `route` to construct a pill to show with the message, and figuring out what that route should be requires API requests... so maybe we should keep the existing behavior for now, and crash. @cmaddox5 can weigh in here if I'm being too conservative, though.

sloanelybutsurely: fwiw this does still end up showing the "offline" message in my testing but i'm happy to put an explicit raise in here instead

digitalcora (approved):
I don't think the Credo failure is significant here... we could suppress it. Yay for better logging!

PR #2103, opened 2024-07-01
tweak: Prevent delays from serializing diagrams.

cmaddox5:
**Asana task**: ad-hoc

When a delay is a single page alert, it will try to serialize the diagram. Problem is that the diagram will always consider delays an invalid effect and spams Splunk with warnings for it. Since we know delays never produce a diagram, we can prevent `DisruptionDiagram.serialize(t)` from being called at all.

Note: I left [validate_effect/1](https://github.com/mbta/screens/blob/6f253595c7caa582b15be1996954e37ff18754c9/lib/screens/v2/disruption_diagram/validator.ex#L26) alone because this code is organized in such a way that other widgets can call it. Didn't want to make a `ReconstructedAlert` assumption in the module because of that. Feel free to push back on that though. 

- [ ] Tests added?


sloanelybutsurely (approved):
sounds good to me. did you want to add a test to this?

cmaddox5:
> sounds good to me. did you want to add a test to this?

I think existing tests should cover it. `disruption_diagram_test` already had a test for delays and assertions for the existing full screen delay tests in `reconstructed_alert_test` will be the same.

PR #2107, opened 2024-07-03
fix: Handle platform closures

cmaddox5:
**Asana task**: [Handle single-child-stop alerts on pre-fares](https://app.asana.com/0/1185117109217413/1207532857226987/f)

This PR adds support for platform closures on Pre-Fare alerts and SubwayStatus. A few key changes:
1. A platform closure that affects the screen's home stop shows a single screen alert instead of a takeover.
2. A full screen alert uses the fallback template where only the OIO text is displayed.
3. A flex zone alert shows the platform that is closed.
4. The subway status uses generic text that directs the user to dotcom.

Note: I definitely want to get code review started to catch an obvious mistakes, but I think a lot of this testing needs to be in dev-green using a variety of real alerts. I think I covered all the bases in tests but will let Dave work out any others. Feel free to do the same.

- [x] Tests added?


lib/screens/alerts/alert.ex
```diff
@@ -592,4 +592,16 @@ defmodule Screens.Alerts.Alert do
...
     do: List.first(informed_entities).direction_id
+
+  def is_child_stop_closure?(%__MODULE__{
+        effect: :station_closure,
+        informed_entities: informed_entities
+      }) do
+    affected_platform_ids =
+      Enum.reject(informed_entities, &String.starts_with?(&1.stop, "place-"))
+
+    length(affected_platform_ids) == 1
```
dmaltzan, line 603: I broke this logic by doing a Station Closure at Alewife - it only has one child stop so you hit this logic even though the entire station is closed.

dmaltzan: Honestly though, I'm not sure if I even care - this would be an incredibly rare situation and the outcome is fine (the fallback is very reasonable)

dmaltzan: If it solves the Alewife problem (and doesn't break things at Downtown Crossing) I'd be fine with switching back to that

lib/screens/alerts/alert.ex
```diff
@@ -592,4 +592,37 @@ defmodule Screens.Alerts.Alert do
...
   def direction_id(%__MODULE__{informed_entities: informed_entities}),
     do: List.first(informed_entities).direction_id
+
+  def informed_subway_platforms(%__MODULE__{
+        informed_entities: informed_entities
+      }) do
+    Enum.reject(
+      informed_entities,
+      &(String.starts_with?(&1.stop, "place-") or &1.route_type != 1)
+    )
```
sloanelybutsurely, line 602: **question:** is there a way to rewrite this as a `filter` and explicitly include things instead of excluding things we don't want?

lib/screens/v2/widget_instance/reconstructed_alert.ex
```diff
@@ -1061,6 +1070,36 @@ defmodule Screens.V2.WidgetInstance.ReconstructedAlert do
...
+    issue =
+      case Alert.informed_subway_platforms(alert) do
+        [informed_platform] ->
+          platform =
+            Enum.find(all_platforms_at_informed_station, &(&1.id == informed_platform.stop))
+
+          "Bypassing #{platform.platform_name} platform at #{informed_station}"
+
+        informed_subway_platforms ->
+          "Bypassing #{length(informed_subway_platforms)} platform at #{informed_station}"
```
sloanelybutsurely, line 1089: **question:** did you want to pluralize this string?

lib/screens/stops/stop.ex
```diff
@@ -361,6 +363,22 @@ defmodule Screens.Stops.Stop do
...
+      {:ok, %{"included" => child_stop_data}} ->
+        child_stop_data
+        |> Enum.filter(fn %{
+                            "attributes" => %{
+                              "location_type" => location_type,
+                              "vehicle_type" => vehicle_type
+                            }
+                          } ->
+          location_type == 0 and vehicle_type == 1
+        end)
```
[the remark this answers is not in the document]
cmaddox5: Nice, thank you!

lib/screens/alerts/alert.ex
cmaddox5: Yeah sure. Platform IDs will always be integers so I can check that `stop` is an integer and `route_type == 1`.

lib/screens/v2/widget_instance/reconstructed_alert.ex
cmaddox5: Ah yes thank you.

lib/screens/alerts/alert.ex
digitalcora: We're very much not guaranteed that platform stop IDs will always look like integers — it would be better to look at the `location_type` of the stop, which tells you explicitly whether a stop is a specific "stopping location" (type 0, a.k.a. a platform) or not.

digitalcora: (That said, all _current_ subway platform stop IDs do look like integers, and at this point it's possible enough other teams have made this assumption — contra the documented best practice of treating all IDs as opaque — that introducing a new platform stop ID that _didn't_ look like an integer would be considered a breaking change.)

cmaddox5: Hmm yeah I see your concern. The issue is `location_type` is not added to `informed_entities`. I could add a bit more stability by passing the list of subway platforms to the function and find a union of the lists. The only problem with that is it makes the function dependent on a separate API call. That's not really a problem in this PR considering we are guaranteed to have that list when making the call. It might be better if I turn this into a `Stop` module function. Then I can just map `informed_entities` into their IDs and do the union that way.

lib/screens/v2/widget_instance/reconstructed_alert.ex
```diff
@@ -1165,12 +1217,25 @@ defmodule Screens.V2.WidgetInstance.ReconstructedAlert do
...
+          alert: %Alert{effect: effect} = alert,
+          all_platforms_at_informed_station: all_platforms_at_informed_station
+        } = t,
+        log_fn
+      ) do
+    location = LocalizedAlert.location(t)
 
-    main_data = pick_layout_serializer(t, diagram_data, effect, dual_screen_alert?(t))
+    if Alert.is_partial_station_closure?(alert, all_platforms_at_informed_station) do
+      t |> serialize_single_screen_fallback_alert(location)
```
digitalcora, line 1231: Pipeline not needed here for a single function call. (I keep wondering why Credo isn't complaining about this on various projects, so I finally looked it up — turns out it's tagged as "controversial" and thus opt-in. Why this is "controversial", I'm not sure.)

lib/screens/v2/widget_instance/reconstructed_alert.ex
```diff
@@ -1233,18 +1299,42 @@ defmodule Screens.V2.WidgetInstance.ReconstructedAlert do
...
+      ) do
+    cond do
+      Alert.is_partial_station_closure?(alert, all_platforms_at_informed_station) ->
+        [:paged_main_content_left]
+
+      dual_screen_alert?(t) ->
+        [:full_body]
+
+      true ->
+        [:paged_main_content_left]
```
digitalcora, line 1316: Would it make more sense to change the criteria for `dual_screen_alert?` to exclude partial closures? That seems like it does describe what we're doing, and then this conditional would remain simple to follow rather than a "maybe A, or maybe B, or maybe A after all" thing.

lib/screens/v2/widget_instance/reconstructed_alert.ex
```diff
@@ -1233,18 +1299,42 @@ defmodule Screens.V2.WidgetInstance.ReconstructedAlert do
...
+      ) do
+    cond do
+      Alert.is_partial_station_closure?(alert, all_platforms_at_informed_station) ->
+        :single_screen_alert
+
+      dual_screen_alert?(t) ->
+        :reconstructed_takeover
+
+      true ->
+        :single_screen_alert
```
digitalcora, line 1336: Same comment as above.

lib/screens/v2/widget_instance/subway_status.ex
```diff
@@ -13,7 +13,7 @@ defmodule Screens.V2.WidgetInstance.SubwayStatus do
 
   @type t :: %__MODULE__{
           screen: Screen.t(),
-          subway_alerts: list(Alert.t())
+          subway_alerts: list(%{alert: Alert.t(), context: %{}})
```
digitalcora, line 16: Could this be a struct rather than a map? This would give us better compile-time checks and would make it clearer what the overall data structure "is"/represents.

lib/screens/v2/widget_instance/subway_status.ex
```diff
@@ -421,9 +444,11 @@ defmodule Screens.V2.WidgetInstance.SubwayStatus do
   end
 
   def serialize_green_line_branch_alert(
```
digitalcora, line 446: This seems like a good opportunity to add a typespec to this public function.

lib/screens/alerts/alert.ex
digitalcora: > The only problem with that is it makes the function dependent on a separate API call.

Yep, that's pretty much the only fully-correct way to do this, as there's no way to fetch related resources on alert informed entities as part of the same call. Fetch alert -> fetch stops matching the IDs in the alert -> get attributes of stops. Of course, we _are_ already doing a stop fetch here to get the full set of child stops for the parent station, so this doesn't strictly need to introduce a new API call, just change the parameters of an existing one.

That said, the shortcut of assuming `place-` is a parent station (and that parent stations and child stops are the only kinds of stop that will appear in a station closure alert) is probably going to be durable enough, kinda like how we assume a route is Commuter Rail iff the ID starts with `CR-`. It pains me a tiny bit, though.

lib/screens/stops/stop.ex
```diff
@@ -361,6 +363,25 @@ defmodule Screens.Stops.Stop do
...
+  def fetch_subway_platforms_for_stop(stop_id) do
+    case Screens.V3Api.get_json("stops/" <> stop_id, %{"include" => "child_stops"}) do
+      {:ok, %{"included" => child_stop_data}} ->
+        child_stop_data
+        |> Enum.filter(
+          &match?(
+            %{
+              "attributes" => %{
+                "location_type" => 0,
+                "vehicle_type" => 1
```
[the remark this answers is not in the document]
cmaddox5: That's a good idea. I don't _think_ GL will have any alerts like this anytime soon but no reason not to make this function support GL.

lib/screens/v2/widget_instance/reconstructed_alert.ex
cmaddox5: Hmm, maybe we should opt-in. Any time I commit these, it's almost always on accident. Just the aftermath of a logic change where I didn't need the other pipes anymore.

lib/screens/v2/widget_instance/subway_status.ex
cmaddox5: Hmm, this function shouldn't be public actually. Will switch to private. I still think this and `serialize_alert/2` could benefit from typespecs so will add them.

lib/screens/v2/candidate_generator/widgets/reconstructed_alert.ex
```diff
@@ -157,21 +159,43 @@ defmodule Screens.V2.CandidateGenerator.Widgets.ReconstructedAlert do
...
       %ReconstructedAlert{
         screen: config,
         alert: alert,
         now: now,
         location_context: location_context,
         informed_stations: get_stations(alert, fetch_stop_name_fn),
         is_terminal_station: is_terminal_station,
-        is_full_screen: is_full_screen
+        is_full_screen: is_full_screen,
+        all_platforms_at_informed_station: all_platforms_at_informed_station
```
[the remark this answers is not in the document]
cmaddox5: Yeah this is something I struggled with while writing it and can't find think of a solution I'm super happy with. Would having a `context` like in `SubwayStatus` make this any clearer?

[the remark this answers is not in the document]
digitalcora: Could this struct have a `is_partial_station_closure` field instead, and we can make that determination all in one place? "Exporting" the final value seems easier to follow than doing half the logic here, "exporting" an intermediate value with a highly specific meaning, then doing the other half later in another place.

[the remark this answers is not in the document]
cmaddox5: The issue is needing the `platform_name` [here](https://github.com/mbta/screens/blob/288958ab56590f92becccd910ab47f4e91deb9aa/lib/screens/v2/widget_instance/reconstructed_alert.ex#L1094). So just knowing if the alert is a partial closure isn't enough for this one case.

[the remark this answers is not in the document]
digitalcora: Then maybe `partial_closure_platform_names`? It seems we're already encoding "presentational" data in this struct since `informed_stations` is just a list of station names, so this would be in line with that. And the name would make it clearer that this is only populated when the alert is a partial station closure. Whatever we can do to put the "final result" we need into the struct here, in one place, I think will be better than splitting up the logic.

Somewhat related thought: what happens when an alert partially-closes multiple stations? It's a definite possibility within Alerts UI.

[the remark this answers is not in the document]
cmaddox5: Had a chat with Dave about the multiple stations question. He is ok working under the assumption that the scenario won't happen. If it does, it can use the full station closure logic.

lib/screens/v2/widget_instance/subway_status.ex
```diff
@@ -360,6 +360,9 @@ defmodule Screens.V2.WidgetInstance.SubwayStatus do
     Map.merge(%{route_pill: serialize_route_pill(route_id)}, serialize_alert(alert, route_id))
   end
 
+  @spec serialize_alert(Alert.t() | nil, String.t()) :: alert()
```
digitalcora, line 381: We could use the type `Route.id()` here to be clear about the meaning (and same below).

lib/screens/v2/widget_instance/subway_status.ex
```diff
@@ -8,12 +10,26 @@ defmodule Screens.V2.WidgetInstance.SubwayStatus do
...
   alias ScreensConfig.V2.{Footer, GlEink, PreFare}
 
+  defmodule SubwayStatusAlert do
+    @type t :: %__MODULE__{
+            alert: Alert.t(),
+            context: context()
+          }
+
+    @enforce_keys [:alert]
+    defstruct @enforce_keys ++ [context: nil]
```
digitalcora, line 22: Should be `context: %{}`, going by the types anyway.

lib/screens/v2/candidate_generator/widgets/subway_status.ex
```diff
@@ -53,11 +53,11 @@ defmodule Screens.V2.CandidateGenerator.Widgets.SubwayStatus do
...
 
-    %{
+    %SubwayStatus.SubwayStatusAlert{
       alert: alert,
       context: %{all_platforms_at_informed_station: all_platforms_at_informed_station}
     }
   end
 
-  defp append_context(alert, _), do: %{alert: alert, context: %{}}
+  defp append_context(alert, _), do: struct(SubwayStatus.SubwayStatusAlert, alert: alert)
```
digitalcora, line 62: Since the struct is known here we can say `%SubwayStatus.SubwayStatusAlert{alert: alert}` and get a bit more compile-time verification.

digitalcora (approved):
Some minor comments. I think the latest refactor turned out well! Approving this assuming we re-test in dev-green with some real alerts, just to be sure the various changes didn't introduce a bug.

lib/screens/v2/candidate_generator/widgets/subway_status.ex
cmaddox5: Woops, I had the commit to change that locally but forgot to push it up 😬

PR #2110, opened 2024-07-15
Add log to catch nil CR schedule.

cmaddox5:
**Asana task**: ad-hoc

Every once in a while, the Porter Pre-Fare screen crashes with the following error 
```
7c59130100be ** (FunctionClauseError) no function clause matching in Screens.V2.Departure.select_arrival_time/1

7c59130100be (screens 0.1.0) lib/screens/v2/departure.ex:255: Screens.V2.Departure.select_arrival_time(nil)
```

If I'm following the code correctly, this only happens when `schedule == nil`. This should not happen though, so I think this is a data problem. Before suggesting something is wrong, I want to try to catch the departure data when it happens next. 

- [ ] Tests added?

### DIFF
--- a/lib/screens/v2/widget_instance/cr_departures.ex
+++ b/lib/screens/v2/widget_instance/cr_departures.ex
@@ -133,14 +133,19 @@ defmodule Screens.V2.WidgetInstance.CRDepartures do
          } = departure,
          now
        ) do
-    if is_nil(schedule) do
-      Logger.error("[cr_departures serialize_time] schedule is nil: #{inspect(departure)}")
-    end
-
     {:ok, scheduled_departure_time} =
-      %Departure{schedule: schedule}
-      |> Departure.time()
-      |> DateTime.shift_zone("America/New_York")
+      try do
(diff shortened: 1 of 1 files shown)


digitalcora:
This seems like it should work, though just to be sure we're actually catching the case where this crashes, we could wrap the code that sometimes crashes in a try/rescue/reraise. Like this:

```elixir
{:ok, scheduled_departure_time} =
  try do
    %Departure{schedule: schedule}
    |> Departure.time()
    |> DateTime.shift_zone("America/New_York")
  rescue ex ->
    Logger.error("oh dear")
    reraise ex, __STACKTRACE__
  end
```

cmaddox5:
> This seems like it should work, though just to be sure we're actually catching the case where this crashes, we could wrap the code that sometimes crashes in a try/rescue/reraise.

Good call. Tried to rush through the solution and ended up with something that might not actually catch the issue 😬 Thank you.

digitalcora (approved):
🔍

PR #2111, opened 2024-07-15
fix: show "BRD" only if the vehicle is at the stop

sloanelybutsurely:
compares the vehicle's stop to the prediction's stop and only shows "BRD" if they match. this prevents "BRD" from showing when stops are very close to each other (such that a prediction is within the "BRD" window) and a vehicle is boarding at the previous stop

**Asana task**: [Fix "BRD" when vehicle is boarding at a previous stop](https://app.asana.com/0/1185117109217422/1207697513391322/f)

Description

- [x] Tests added

### DIFF
--- a/lib/screens/v2/departure.ex
+++ b/lib/screens/v2/departure.ex
@@ -186,8 +186,6 @@ defmodule Screens.V2.Departure do
     stop_id
   end
 
-  def stop_id(_), do: nil
-
   def stop_type(%__MODULE__{
         prediction: %Prediction{arrival_time: arrival_time, departure_time: departure_time}
       }) do
--- a/lib/screens/v2/departure.ex
+++ b/lib/screens/v2/departure.ex
@@ -186,6 +186,8 @@ defmodule Screens.V2.Departure do
     stop_id
   end
 
+  def stop_id(_), do: nil
+
   def stop_type(%__MODULE__{
         prediction: %Prediction{arrival_time: arrival_time, departure_time: departure_time}
       }) do
--- a/lib/screens/v2/widget_instance/departures.ex
+++ b/lib/screens/v2/widget_instance/departures.ex
@@ -3,6 +3,7 @@ defmodule Screens.V2.WidgetInstance.Departures do
 
   alias Screens.Alerts.Alert
   alias Screens.Departures.Departure
+  alias Screens.Predictions.Prediction
   alias Screens.Routes.Route
   alias Screens.Schedules.Schedule
   alias Screens.Util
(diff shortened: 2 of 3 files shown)


lib/screens/v2/departure.ex
```diff
@@ -186,6 +186,8 @@ defmodule Screens.V2.Departure do
     stop_id
   end
 
+  def stop_id(_), do: nil
```
digitalcora, line 189: When do we encounter this case? I don't think it's possible in practice to have a `Departure` where `prediction` and `schedule` are both nil, so that leaves the `stop` field of the prediction/schedule being `nil` (or some other value) as a possible match failure — but going by the typespec for both of these modules, it should only ever be a `Stop`.

sloanelybutsurely: the stop is omitted in many tests

lib/screens/v2/widget_instance/departures.ex
```diff
@@ -414,9 +415,12 @@ defmodule Screens.V2.WidgetInstance.Departures do
        ),
        do: %{time: %{type: :icon, icon: :overnight}}
 
+  # credo:disable-for-lines:30 Credo.Check.Refactor.CyclomaticComplexity
```
digitalcora, line 418: I'm not sure I've ever seen this check really be useful, in terms of the subjective experience of reading/understanding/maintaining the function... I wouldn't complain if we disabled it entirely, and it looks like there is precedent for this on other projects. If we don't do that, though, I wonder if this would work with just `disable-for-next-line`, as the single line with the `defp` is the one the issue is reported on (and then we don't have to deal with the line count inevitably falling out of sync).

lib/screens/v2/departure.ex
digitalcora: Absent fixing the tests themselves (which I think would be best in the long term, but maybe that's a lot of extra churn to add to this one small fix, given we don't currently use factories for test data in many places), maybe we could explain why this is here? Like:

```elixir
# Expected to never occur with real data, but accommodates tests that don't fill in a stop field
# for predictions/schedules.
def stop_id(_), do: nil
```

lib/screens/v2/widget_instance/departures.ex
sloanelybutsurely: @digitalcora ah! i thought `disable-for-next-line` wasn't working there earlier so i changed it to the `for-lines:30`... turns out i had just typoed `disable-for-next-line`. fixed!

sloanelybutsurely: if we'd like to disable this altogether (i'm in favor of that!): https://github.com/mbta/screens/pull/2112

lib/screens/v2/departure.ex
sloanelybutsurely: i've added `Stop` structs in the tests that were missing them

digitalcora (approved):
🚋 🛑 ✅

PR #2114, opened 2024-07-25
fix: only set cached=true when a 304 is received

sloanelybutsurely:
Previously our telemetry logging would show `cached=true` whenever we had a possible value for a request in the cache **not** when we actually used the cache. This was misleading and is corrected in this change.

Being able to return a meta value in a `Screens.Telemetry` span required a new function that expects the same return value as `:telemetry.span/3` which is a tuple where the second element is a map of metadata to include in the `*.stop` event.

### DIFF
--- a/lib/screens/telemetry.ex
+++ b/lib/screens/telemetry.ex
@@ -74,6 +74,29 @@ defmodule Screens.Telemetry do
     end)
   end
 
+  @doc """
+  Like `span/3` but requires a tuple of `{<return value>, <stop meta>}` to be
+  returned from the passed function. The `<stop meta>` map will be merged into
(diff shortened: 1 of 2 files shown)


digitalcora (approved):
Tested this out locally and the logging now seems to match what I'd expect!

> `push-nmrsxoowxuyx`

Zalgo, is that you? 😅

sloanelybutsurely:
> > `push-nmrsxoowxuyx`
> 
> Zalgo, is that you? 😅

@digitalcora lol not zalgo, [`jj`](https://martinvonz.github.io/jj/latest/)'s generated branch name based on change ID

PR #2115, opened 2024-07-29
refactor: add types for DUP widgets

digitalcora:
This ended up being an exercise in stumbling upon a bunch of little things that could be refactored, so best reviewed per-commit (there are some explanations in the commit messages). The main point here is the final commit!

---
  - https://app.asana.com/0/0/1207574972639085


cmaddox5 (approved):
Looks good. Thanks for adding the helpful comments.

PR #2116, opened 2024-07-31
refactor: remove unused routes-at-stop agent

digitalcora:
d52ead8b, which introduced more comprehensive V3 API request caching, removed the only instance where `put` was ever called on this cache, so it was effectively unused.

After removing a few layers of indirection:

* `Stop.create_station_with_routes_map` is renamed and relocated to `Route.serving_stop`. Unclear why the previous name suggested it would "create a map".

* The similar-in-nature `Route.fetch_routes_by_stop` is renamed to `Route.serving_stop_with_active` so these names are aligned.

* Logging is added to the retry logic of `serving_stop` so we can tell whether it's still being used. This seems like a workaround for a possible bug in the V3 API which may have since been fixed, and if it hasn't been, we'll know and can report it.

* `serving_stop_with_active` is fixed to conform to its typespec (and the spec of `LocationContext`) by returning a map containing _only_ `route_id` and `active?` fields, rather than a map version of the entire `Route` struct with the `id` field renamed. It appears nothing was accessing the extra fields.


cmaddox5 (approved):
Thanks for taking care of this!

PR #2117, opened 2024-07-31
deps: install server_sent_event_stage

sloanelybutsurely:
Installs server_sent_event_stage which requires also adding `castore` with explicitly to the `mix.exs` and specifying `override: true`.

server_sent_event_stage expects `{:castore, "~> 0.1"}` but castore 1.0.8 is already installed by another dependency. We can remove the `override: true` once server_sent_event_stage has been updated to support newer versions of castore.

The move to from `0.1.x` to `1.0.x` versions in castore was a non-breaking change[^1]. Subsequent patch versions of castore are simply updates to the certificates included (removing old/expired certs and adding new ones).

[^1]: castore 0.1.22..1.0.0: https://diff.hex.pm/diff/castore/0.1.22..1.0.0


cmaddox5 (approved):

digitalcora:
Since `server_sent_event_stage` is ours (...comrade), I went ahead and [updated it](https://github.com/mbta/server_sent_event_stage/pull/17). In the next release, we won't have to override `ca_store`.

PR #2118, opened 2024-07-31
feat: in-memory Alert storage

sloanelybutsurely:
Uses ServerSentEventStage to implement a simple in-memory store of the current alerts from the V3 API.

[Server Sent Events](https://developer.mozilla.org/en-US/docs/Web/API/Server-sent_events) are parsed into `Screens.Alerts.Alert` structs and stored in a `Map` which is kept in the process state of the `GenStage` consumer.

> [!NOTE]  
> This implementation is limited to serial access because the data is kept in process state. This is a known limitation. I'd like to keep this as-is knowing that I will come back through and swap this for a more parallelizable implementation after filtering logic (#2119) is solid.

### DIFF
--- a/lib/screens/alerts/cache.ex
+++ b/lib/screens/alerts/cache.ex
@@ -1,0 +1,77 @@
+defmodule Screens.Alerts.Cache do
+  @moduledoc """
+  GenStage Consumer of Alert server sent event data
+  """
+  use GenStage
+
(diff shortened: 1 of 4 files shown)


sloanelybutsurely:
Closing in favor of #2121 which uses ETS to allow concurrent access

PR #2120, opened 2024-08-01
fix: refactor of routes-serving-stop fetching

digitalcora:
The version of this function prior to the refactor in 297c302 returned only a list of routes, returning an empty list if an error occurred. The moved/renamed version returns an `:ok` tuple or `:error`, consistent with other data fetching functions, but the call sites were not updated to reflect this (😓), and no tests failed because the mocks for the function still used the old return type (or, in the case of `ElevatorClosures`, the module lacked any tests at all).

This keeps the return type of the fetching function the same, so errors can be distinguished if needed, but changes the existing call sites to perform the "default to empty list" logic themselves, so there should be no change in behavior.

Deployed this branch to dev-green so we should see errors if there are still any, from the self-refresh runner.

### DIFF
--- a/lib/screens/v2/candidate_generator/widgets/elevator_closures.ex
+++ b/lib/screens/v2/candidate_generator/widgets/elevator_closures.ex
@@ -52,7 +52,7 @@ defmodule Screens.V2.CandidateGenerator.Widgets.ElevatorClosures do
     |> MapSet.new()
     |> MapSet.put(home_parent_station_id)
     |> Enum.map(fn station_id ->
-      {station_id, station_id |> Route.serving_stop() |> routes_to_icons()}
+      {station_id, station_id |> routes_serving_stop() |> routes_to_icons()}
     end)
     |> Enum.into(%{})
   end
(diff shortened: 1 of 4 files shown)


lib/screens/v2/candidate_generator/widgets/elevator_closures.ex
```diff
@@ -66,6 +66,13 @@ defmodule Screens.V2.CandidateGenerator.Widgets.ElevatorClosures do
     end)
   end
 
+  defp routes_serving_stop(stop_id) do
+    case Route.serving_stop(stop_id) do
+      {:ok, routes} -> routes
+      :error -> []
+    end
+  end
```
sloanelybutsurely, line 74: **off-topic:** maybe someday we'll have something like [`maybe` from `Data.Maybe`](https://hackage.haskell.org/package/base-4.20.0.1/docs/Data-Maybe.html#v:maybe)... _:: sigh ::_

sloanelybutsurely (approved):

lib/screens/v2/candidate_generator/widgets/elevator_closures.ex
digitalcora: Can I interest you in some [Gleam](https://gleam.run/)? 😺

PR #2121, opened 2024-08-01
feat: in-memory Alert storage

sloanelybutsurely:
Uses ServerSentEventStage to consume alerts events from the V3 API and store them in an ETS table keyed by alert ID.

### DIFF
--- a/lib/screens/alerts/cache.ex
+++ b/lib/screens/alerts/cache.ex
@@ -1,0 +1,82 @@
+defmodule Screens.Alerts.Cache do
+  @moduledoc """
+  GenStage Consumer of Alert server sent event data
+  """
+  use GenStage
+
+  require Logger
+
+  alias Screens.Alerts
+  alias ServerSentEventStage.Event
+
+  @table __MODULE__
+
+  defstruct [:table]
+
+  def start_link(opts) do
+    {name, init_arg} = Keyword.pop(opts, :name, __MODULE__)
+    GenStage.start_link(__MODULE__, init_arg, name: name)
+  end
+
+  @impl true
+  def init(init_arg) do
+    subscribe_to = Keyword.get(init_arg, :subscribe_to, [Screens.Streams.Alerts.Producer])
+
+    table = @table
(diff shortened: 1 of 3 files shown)


lib/screens/streams.ex
```diff
@@ -0,0 +1,24 @@
...
+    Supervisor.start_link(__MODULE__, init_arg, name: name)
+  end
+
+  @impl true
+  def init(_init_arg) do
+    children = [
+      Screens.Streams.Alerts
+    ]
+
+    Supervisor.init(children, strategy: :one_for_one)
```
panentheos, line 22: Do we need this layer of supervisor, or could we flatten it out and start `Screens.Streams.Alerts` directly under the application?

lib/screens/alerts/cache.ex
```diff
@@ -0,0 +1,89 @@
...
+  defp handle_event(event, _state) do
+    Logger.warning(fn -> "Unknown event: #{inspect(event)}" end)
+
+    :ok
+  end
+
+  defp decode_data(%Event{data: encoded} = event) do
+    case Jason.decode(encoded) do
+      {:ok, decoded} -> %{event | data: decoded}
+      _ -> :error
```
panentheos, line 86: nit: On the off chance this fails, we'll end up logging `:error` above, which won't be helpful in diagnosing the issue.

sloanelybutsurely: Good point! I've changed it to use `Jason.decode!/1` so it raises an exception if it cannot decode the event data. I've also removed the catch-all clause in `handle_event/2` so any unrecognized event will also cause an error.

Finally, I changed the strategy for the `Screens.Streams.Alerts` supervisor to `:one_for_all` so if the consumer crashes, the producer is also restarted.

I've made all of these changes because the server sent event stream is stateful. Any missed event would result in drift in the alerts cache. Now if an error happens in consuming events the entire pipeline will be restarted resulting in a new `reset` event being sent from the v3 api.

PR #2122, opened 2024-08-01
feat: switch to in-memory Alerts cache instead of V3 API

sloanelybutsurely:
Relies on the in-memory Alerts cache when a request for Alerts data has filters supported by the cache. Uses the existing V3 API client otherwise.

---
  - https://app.asana.com/0/0/1207848480396826

### DIFF
--- a/lib/screens/alerts/alert.ex
+++ b/lib/screens/alerts/alert.ex
@@ -175,8 +175,18 @@ defmodule Screens.Alerts.Alert do
   ]
 
   @spec fetch(keyword()) :: {:ok, list(t())} | :error
-  def fetch(opts \\ [], get_json_fn \\ &V3Api.get_json/2) do
-    Screens.Telemetry.span([:screens, :alerts, :alert, :fetch], fn ->
+  def fetch(opts \\ [], get_json_fn \\ &V3Api.get_json/2, get_all_alerts \\ &Cache.all/0) do
+    Screens.Telemetry.span_with_stop_meta([:screens, :alerts, :alert, :fetch], fn ->
+      if supported_by_cache?(opts) do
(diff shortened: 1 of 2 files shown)


test/screens/alerts/alert_test.exs
```diff
@@ -58,6 +58,7 @@ defmodule Screens.Alerts.AlertTest do
       }
     end
 
+    @tag skip: "needs reconciled with changes to alert fetching via cache"
```
sloanelybutsurely, line 61: suggestions on how to handle this more than welcome! 🙏

lib/screens/alerts/alert.ex
```diff
@@ -226,6 +238,25 @@ defmodule Screens.Alerts.Alert do
 
   defp format_cache_filter(filter), do: filter
 
+  defp supported_by_cache?(opts) do
+    Enum.all?(opts, fn {key, _} -> supported_cache_filter?(key) end)
+  end
+
+  for supported_filter <- ~w[routes
```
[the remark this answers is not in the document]
sloanelybutsurely: It is possible but would require more than just adding `include=facilities`. I'd like to hold off on that for now to limit the scope of this change.

test/screens/alerts/alert_test.exs
cmaddox5: Hmm, this is tough. Could `fetch_by_stop_and_route/3` be adjusted so it always fetches from cache? We know based on the parameters that the filtering is supported by the cache as-is and there is currently no way to add additional options. So basically the third parameters can be changed to `get_all_alerts \\ &Cache.all/0`. Another idea is to seed the cache in `setup` before running the test but that feels less elegant.

PR #2124, opened 2024-08-02
fix: Add envrc and remove SecretsManager dependency

cmaddox5:
There's no reason for us to read directly from SecretsManager. These secrets are added as ENV variables in TF. I also added `.envrc.template` and updated some docs.


config/runtime.exs
```diff
@@ -28,23 +28,7 @@ if config_env() == :prod do
   {:ok, _} = Application.ensure_all_started(:ex_aws)
   {:ok, _} = Application.ensure_all_started(:ex_aws_secretsmanager)
```
sloanelybutsurely, line 29: You can remove all of these now, too!

config/runtime.exs
```diff
@@ -28,23 +28,7 @@ if config_env() == :prod do
...
-    |> ExAws.SecretsManager.get_secret_value()
-    |> ExAws.request!()
-    |> Map.fetch!("SecretString")
-
-  screens_auth_secret =
-    (eb_env_name <> "-screens-auth-secret")
-    |> ExAws.SecretsManager.get_secret_value()
-    |> ExAws.request!()
-    |> Map.fetch!("SecretString")
+  eb_env_name = System.get_env("ENVIRONMENT_NAME")
```
sloanelybutsurely, line 31: This is already set on line 7. No need to do it again, right?

cmaddox5: Ahh yes thanks. Didn't see that up there.

sloanelybutsurely (approved):

PR #2130, opened 2024-08-14
feat: switch alert requests for specific routes to cache

sloanelybutsurely:
Replaces calls to `Alert.fetch/1` and `Alert.fetch_or_empty_list/1` that only filter on `route_id` with calls to `Alert.fetch_from_cache/1` and `Alert.fetch_from_cache_or_empty_list/1` respectively.

### DIFF
--- a/lib/screens/alerts/alert.ex
+++ b/lib/screens/alerts/alert.ex
@@ -204,6 +204,12 @@ defmodule Screens.Alerts.Alert do
     {:ok, Screens.Alerts.Cache.Filter.filter_by(alerts, filters)}
   end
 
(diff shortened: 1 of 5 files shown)


cmaddox5 (approved):
Excited to see how this turns out.

digitalcora:
I'm a bit late here, but when fetching based on route ID from the cache, shouldn't we also specify a route type? An alert that affects a whole route type (without specifying individual routes) affects all routes of that type. In the V3 API if I request alerts for `route=Red`, I'll also get any alerts with `route_type=1`.

PR #2131, opened 2024-08-15
feat: Expand stops filters to include routes at stops

sloanelybutsurely:
Adds `fetch_routes_for_stops/1` and uses that to expand a `stops:` filter similar to the V3 API.


lib/screens/stops/stop.ex
```diff
@@ -310,6 +310,19 @@ defmodule Screens.Stops.Stop do
     end
   end
 
+  @callback fetch_routes_for_stops(stop_ids :: String.t() | [String.t()]) ::
+              {:ok, [Route.t()]} | {:error, term()}
+  def fetch_routes_for_stops(stop_ids) do
```
digitalcora, line 315: We already have a similar `Route.serving_stop` function; maybe these could be combined. I think it may be better to go with that function's approach — just asking for routes with a `filter[stop]` and letting the API decide what it means for a route to serve a stop. At least, that seems like the behavior we'd want for fetching alerts? (to not limit ourselves to "canonical" service only)

test/screens/alerts/alert_test.exs
```diff
@@ -110,6 +113,13 @@ defmodule Screens.Alerts.AlertTest do
 
   describe "fetch_from_cache/2" do
     setup do
+      default_stops_mod = Application.get_env(:screens, :alerts_cache_filter_stops_mod)
+      Application.put_env(:screens, :alerts_cache_filter_stops_mod, Stop.Mock)
```
digitalcora, line 117: This may cause inconsistent test failures due to this test module being `async: true`. Is there any problem with always using the mock version of the module in `test.exs`? I'm not sure if we'd ever want it to be the "real" one since that would imply making real network requests.

digitalcora (commented):
I may have missed or misunderstood something, but I was under the impression from previous discussions that re-implementing the API's "alert matching" logic within our cache was specifically something we wanted to avoid — with the reasoning that, in most cases where we fetch alerts, we already have the needed context of route type/route/stop/direction handy, and could pass those as explicit filters to the cache functions (which would be taken literally, not "expanded"). I don't _disagree_ with this approach, just wondering how we arrived here.

lib/screens/routes/route.ex
```diff
@@ -62,6 +62,7 @@ defmodule Screens.Routes.Route do
   end
 
   @doc "Fetches routes that serve the given stop."
+  @callback serving_stop(Stop.id()) :: {:ok, [t()]} | :error
```
digitalcora, line 65: I'm surprised this actually works, though it feels vaguely like a subversion of the Mox philosophy ("No ad-hoc mocks. You can only create mocks based on behaviours"). I guess you could say it's an "inline behavior"? Not an objection, just a "huh, that's interesting".

Experimentally, it seems to be sufficient to _only_ have the `@callback` — if you remove the duplicated `@spec`, Dialyzer still throws an error if you e.g. assign the return value to a pattern that can never match.

lib/screens/alerts/cache/filter.ex
```diff
@@ -78,12 +86,31 @@ defmodule Screens.Alerts.Cache.Filter do
   end
 
   defp build_matcher({:stops, values}, acc) when is_list(values) do
-    matchers_for_values(acc, :stop, values)
-  end
+    route_mod = route_mod()
```
digitalcora, line 89: Since this is now always the same per-Mix-env, it can be defined at compile time:

```elixir
@route_mod Application.compile_env(:screens, :alerts_cache_filter_routes_mod, Screens.Routes.Route)
```

This has the major advantage of preserving all the compile-time protections we'd normally get when calling a function directly, and allows Dialyzer to still work.

sloanelybutsurely: The issue with defining this at compile time is that the Mock module is not defined at compile time so we run into compile warnings when running tests.

```
$ mix test
Compiling 5 files (.ex)
warning: Screens.Routes.Route.Mock.serving_stop/1 is undefined (module Screens.Routes.Route.Mock is not available or is yet to be defined)
  lib/screens/alerts/cache/filter.ex:94: Screens.Alerts.Cache.Filter.build_matcher/2
```

lib/screens/routes/route.ex
sloanelybutsurely: Our dependency on the V3 API makes testing cumbersome. Probably a nicer solution would be to use use an adapter pattern for the V3 API.

For now I'll remove the redundant `@spec` attribute here.

digitalcora (approved):
🚉 🔍 ✅

PR #2139, opened 2024-08-15
refactor: remove redundant Sentry-disabling code

digitalcora:
Simplifies the multiple overlapping mechanisms for enabling and disabling Sentry, hopefully without changing existing behavior.

* If a `SENTRY_DSN` environment variable is defined, and the Mix env is `prod`, Sentry is enabled on the backend.
* If the above is true _and_ there is not a `disable_sentry` query param in the URL, Sentry is also enabled on the frontend.

There do not appear to be any screens actively using `disable_sentry` at the time of writing, but we'll leave this feature alone for now.

### DIFF
--- a/assets/src/components/v2/widget_tree_error_boundary.tsx
+++ b/assets/src/components/v2/widget_tree_error_boundary.tsx
@@ -65,7 +65,7 @@ class WidgetTreeErrorBoundary extends React.Component<Props, State> {
       return;
     }
 
-    if (getDataset().disableSentry) {
+    if (!getDataset().sentry) {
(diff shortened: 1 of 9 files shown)


sloanelybutsurely (approved):

cmaddox5 (approved):
Wow. We really went for the overkill method with this. I think you're right about no screen using `disable_sentry`. That was originally put in place because one of our screen types (GDS E-Ink maybe?) would crash if it tried to initialize it. Might be a good idea to leave in just in case we get another screen with a similar issue.

PR #2140, opened 2024-08-16
chore: Add dependabot groups

cmaddox5:
I want to try out Dependabot's `groups` option so we can reduce the number of PRs it opens. Feel free to call out groups that should be added or groups I added that don't look right.


cmaddox5:
Yep, that makes sense. I can convert this to draft for now.

PR #2143, opened 2024-08-16
chore: clean up admin dashboard navigation

digitalcora:
Was thinking of adding a new item to this dashboard related to the DUP project, so it was a good time to clean house.

* Remove all v1 screen tables except Solari, since we should no longer have any active screens of these types
* Split nav items into groups and improve styling a bit
* Refactor to generate nav links and routes from a constant
* Remove "admin layout" which was an exact duplicate of the v2 layout

Screenshot 2024-08-16 at 6 15 07 PM

---
  - https://app.asana.com/0/0/1208083484684124


cmaddox5 (approved):
😍

PR #2144, opened 2024-08-19
fix: error reporting on speech synthesis failures

digitalcora:
Over the weekend, we hit a "maximum text length exceeded" error when trying to synthesize audio for a pre-fare ([Sentry](https://mbtace.sentry.io/issues/5728118365/)). Unfortunately we don't know exactly what the text was, because A) the code to log the error itself contained an error, and B) even if it had worked, we weren't logging the input text. Also, we probably wouldn't know it had happened at all, since this report would have only gone to the logs and not to Sentry.

This fixes the string interpolation in the logging, adds some needed context, and adds a Sentry report so we'll be notified when this happens again.

The [current audio readout](https://screens.mbta.com/v2/audio/PRE-110/debug) for this pre-fare is indeed rather close to the [limit](https://docs.aws.amazon.com/polly/latest/dg/limits.html#limits-synthesizespeech) of 3,000 "billed characters" (exclusive of characters in SSML tags). I'll create a task for us to look into splitting up our speech somehow or cutting down on the number of characters.

### DIFF
--- a/lib/screens/audio.ex
+++ b/lib/screens/audio.ex
@@ -20,7 +20,8 @@ defmodule Screens.Audio do
 
   @type departure_group :: {departure_group_key(), [map()]}
 
-  def synthesize(ssml_string, is_screen, text_type) do
+  @spec synthesize(String.t(), String.t(), keyword()) :: {:ok, binary()} | :error
+  def synthesize(ssml_string, text_type, log_meta) do
     result =
       ssml_string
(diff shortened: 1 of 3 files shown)


cmaddox5 (approved):
Wow, that is close. Is the whitespace working against us here? I wonder if a quick win is to trim all strings in the views? Not something to do here, though.

digitalcora:
I'm not seeing anything that indicates whitespace wouldn't be counted (it is "spoken", after all, though multiple spaces are not spoken any differently from a single one). So that could be something, though it would only save a couple dozen characters.

PR #2145, opened 2024-08-19
fix: apply filters to alerts cache correctly

sloanelybutsurely:
The alerts cache was missing a step that's present in the V3 API which meant that filters were not being applied to the list of all alerts correctly producing the wrong results.

This change adds a test suite to verify that filters work as expected and adds the additional step of expanding built matchers so partial matches are considered a match.

### DIFF
--- a/test/screens/alerts/cache/filter_test.exs
+++ b/test/screens/alerts/cache/filter_test.exs
@@ -3,69 +3,126 @@ defmodule Screens.Alerts.Cache.FilterTest do
 
   import Mox
 
+  alias Screens.Alerts.Alert
   alias Screens.Alerts.Cache.Filter
   alias Screens.Routes.Route
 
   setup :verify_on_exit!
 
-  describe "build_matchers/1" do
-    test "passes through empty filters" do
-      assert [] == Filter.build_matchers(%{})
+  describe "filter_by/2" do
+    setup do
+      alerts = [
+        alert("stop only", ie(stop: "stop-only")),
+        alert("route serving stop", ie(route: "serving-stop")),
+        alert("route only", ie(route: "route-only", route_type: 0)),
+        alert("stop and route", ie(stop: "stop-and-route-stop", route: "stop-and-route-route")),
+        alert("route only 2", ie(route: "route-only-2", route_type: 1)),
+        alert("route type only", ie(route_type: 0)),
+        alert(
+          "route in one direction",
+          ie(route: "route-in-direction", direction_id: 0, route_type: 3)
+        ),
+        alert(
+          "route in the other direction",
+          ie(route: "route-in-direction", direction_id: 1, route_type: 3)
+        ),
+        alert("route in either direction", ie(route: "route-in-direction", route_type: 3)),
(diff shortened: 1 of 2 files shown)


test/screens/alerts/cache/filter_test.exs
```diff
@@ -66,4 +67,120 @@ defmodule Screens.Alerts.Cache.FilterTest do
...
+               Filter.filter_by(alerts, %{route_types: [0, 1]})
+    end
+
+    test "filters for a specific direction_id", %{alerts: alerts} do
+      stub(Route.Mock, :by_id, fn
+        "route-in-direction" -> {:ok, %Route{id: "route-in-direction", type: :bus}}
+      end)
+
+      assert [%Alert{id: "route in one direction"}] =
+               Filter.filter_by(alerts, %{routes: ["route-in-direction"], direction_id: 0})
```
digitalcora, line 144: Shouldn't this also include all alerts with a "non-directional" informed entity? Since its absence implies the alert affects both directions of service. What I'd expect the direction filter to do is basically reject any alerts that _only_ affect the opposite direction. (Though I don't know off-hand if this is how the API does it...)

test/screens/alerts/cache/filter_test.exs
```diff
@@ -3,6 +3,7 @@ defmodule Screens.Alerts.Cache.FilterTest do
 
   import Mox
```
digitalcora, line 4: Not directly related to this change, but I'd suggest including `setup :verify_on_exit!` whenever we're using Mox, so tests fail if we set up an `expect` that's never used (alerting us that either something is wrong or the mock is redundant and can be removed). As-is it wouldn't have any effect on these tests, since `stub`s are never verified.

digitalcora (commented):
Compared to these new tests which are testing the "public API", the old tests for `build_matchers` seem like they're testing an implementation detail — are they still providing some value above and beyond the `filter_by` tests?

test/screens/alerts/cache/filter_test.exs
sloanelybutsurely: 🫡 here's the suggested change as its own PR https://github.com/mbta/screens/pull/2147

test/screens/alerts/cache/filter_test.exs
sloanelybutsurely: You're correct. That is how this code and the API handle that. I've updated the test case to reflect that.

sloanelybutsurely:
> the old tests for `build_matchers` seem like they're testing an implementation detail — are they still providing some value above and beyond the `filter_by` tests?

@digitalcora You're totally right. `build_matchers` was a private function a while back. I've made it private again and removed those test cases. They were helpful while I wrote that code but isn't really behavior we need to test anymore.

test/screens/alerts/cache/filter_test.exs
digitalcora: Ah, thanks for the update, I was a little confused about how the test data intersected with the specific test case.

PR #2146, opened 2024-08-19
refactor: remove unused audio component/endpoint

digitalcora:
Nothing seems to be calling this endpoint in production, and the only reference to it in the code is this `Audio` component, which itself is not imported anywhere.

Since this was the only case where we sent Polly requests using plain text instead of SSML, the parameter for text type can also be removed.


cmaddox5 (approved):
🧹

PR #2149, opened 2024-08-20
dev: enforce React hooks rules in ESLint

digitalcora:
Inspired by mbta/screenplay#459. Pleasantly, there was only one extant "rule-breaker", which was easy to correct.


sloanelybutsurely (approved):
gorge

PR #2156, opened 2024-08-26
feat: admin screen inspector

digitalcora:
Presenting a ~~proof that I should not be left unsupervised~~ new Screens admin tool: the **Inspector**.

This is, I promise, pursuant to [Set up parallel running of new DUP logic](https://app.asana.com/0/1185117109217413/1207992837169564). The train of thought went something like this:

* Rather than hunting down and typing in DUP screen IDs and query params, it'd be nice to have a little tool that lets you select any screen you want, display it, and flip between the code paths. That should only take a couple days to build, it'll just expand the scope of this task a little bit.
* Okay, neither "base" screen pages nor Screenplay simulations are very readable without zooming them out/in... we should be able to resize the screen.
* And we'll probably want a way to reload the screen data immediately when we switch code paths.
  * Oh, and it'd be nice if we could suspend the automatic data updates. Hate when I'm fiddling with something in the web inspector and the DOM keeps changing out from under me.
* Repeat for a few more "it'd be nice if we had this" things which are increasingly unrelated to DUP departures logic.
* Hey, where'd that week go?

Feature creep aside, I'm hoping folks agree this is a useful tool in general and a decent "attractor" for developer/admin features that might otherwise be screen page query params or other less-friendly mechanisms. I know I'd personally use it a lot for debugging in preference to the "raw" screen pages. Of course, if you don't agree, please say so 🙂 Considering this wasn't explicitly planned at any point, I'm very willing to rework it if there are concerns.

If it's not clear: the implementation of the "DUP logic switch" within this framework would be extremely similar to the existing Refresh button in the Data section. Instead of signaling the data hook to perform a refresh, it would signal to update a bit of internal state to use a different URL for fetching the data. The screen data API controller would then pick up on this and from there the approach would be identical to the originally-proposed one.

[Preview on dev-green](https://screens-dev-green.mbtace.com/admin/inspector?id=MUL-114-V2).

Screenshot 2024-08-26 at 1 36 53 PM

---
  - https://app.asana.com/0/0/1208138805469601

### DIFF
--- a/assets/src/apps/admin.tsx
+++ b/assets/src/apps/admin.tsx
@@ -25,8 +25,10 @@ import AdminScreenConfigForm from "Components/admin/admin_screen_config_form";
 import AdminTriptychPlayerForm from "Components/admin/admin_triptych_player_form";
 import ImageManager from "Components/admin/admin_image_manager";
 import Devops from "Components/admin/devops";
+import Inspector from "Components/admin/inspector";
 
 const routes: [string, string, ComponentType][][] = [
+  [["inspector", "🔍", Inspector]],
   [["", "All Screens", AllScreensTable]],
   [
     ["bus-eink-v2-screens", "Bus E-ink", BusEinkV2ScreensTable],
--- a/assets/css/admin.scss
+++ b/assets/css/admin.scss
@@ -84,6 +84,12 @@
   width: 240px;
 }
 
+.admin-table__inspector-link {
+  color: unset;
+  text-decoration: none;
+  border-bottom: 1px dotted;
+}
+
 .admin-modal__background {
   position: fixed;
   top: 0;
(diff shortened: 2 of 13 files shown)


assets/src/apps/admin.tsx
```diff
@@ -25,8 +25,10 @@ import AdminScreenConfigForm from "Components/admin/admin_screen_config_form";
 import AdminTriptychPlayerForm from "Components/admin/admin_triptych_player_form";
 import ImageManager from "Components/admin/admin_image_manager";
 import Devops from "Components/admin/devops";
+import Inspector from "Components/admin/inspector";
 
 const routes: [string, string, ComponentType][][] = [
+  [["inspector", "🔍", Inspector]],
```
cmaddox5, line 31: Is this our first emoji? What a milestone.

sloanelybutsurely (approved):
😍 it's.... it's beautiful! holy smokes!! 👏

cmaddox5 (approved):
Holy moly this is crazy. The kind of scope creep I will gladly support.

PR #2164, opened 2024-08-27
chore: enable frontend source maps in Sentry

digitalcora:
* Upload source maps in parallel as part of the deploy Action.
* Removes unneeded options from TerserPlugin. There is no such option as `cache` (this did nothing?), and `parallel` defaults to true.
* Removes an old script that appears to have been previously used to upload source maps, but was not referenced anywhere.

Unfortunately the source maps this gives us aren't very _good_ yet, because the source code as seen by Sentry appears to be just the pre-minification JS, which is downstream from TypeScript, all the polyfills, etc. But this is an unplanned task I didn't want to sink too much time into, and it's strictly better than what we had before.

---
  - https://app.asana.com/0/0/1208162809395304


cmaddox5 (approved):
🙌

PR #2165, opened 2024-08-27
chore: remove unused alert caching code

sloanelybutsurely:
Ultimately the cache wasn't definitively faster than just fetching from the API directly. This removes the unused code.


cmaddox5 (approved):
🎻

I see a compilation error in a test but will go ahead so you can merge right after fixing.

PR #2168, opened 2024-08-28
fix: handle unscheduled commuter rail departures

sloanelybutsurely:
The commuter rail departures widget made the assumption that there would never be a departure that had a prediction but no matching schedule. We have observed that this isn't always the case.

Adds handling of a `nil` schedule to the commuter rail departures widget. The only thing relying on a present schedule in `serialize_departure/5` was calculating if a departure was delayed. If there is no schedule we do not flag the departure as delayed; otherwise, the delayed calculation remains unchanged.

**Asana task**: [[Screens 🐞] Fix Departure prediction/schedule merging](https://app.asana.com/0/1185117109217413/1207953256080348/f)

> [!NOTE]
> The associated ticket references an issue with filtering out past schedules before merging them with predictions into departures. To verify that this is not actually the case I have added tests verifying that past schedules are still merged with their matching predictions.

### DIFF
--- a/test/screens/v2/widget_instance/cr_departures_test.exs
+++ b/test/screens/v2/widget_instance/cr_departures_test.exs
@@ -4,9 +4,9 @@ defmodule Screens.V2.WidgetInstance.CRDeparturesTest do
   # alias Screens.Alerts.Alert
   # alias Screens.Config.Dup.Override.FreeTextLine
   # alias Screens.Config.Screen
+  alias Screens.Schedules.Schedule
   alias Screens.Predictions.Prediction
   # alias Screens.Routes.Route
-  # alias Screens.Schedules.Schedule
   alias Screens.Trips.Trip
   # alias Screens.Vehicles.Vehicle
   alias Screens.V2.{Departure, WidgetInstance}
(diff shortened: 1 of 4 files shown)


test/screens/v2/widget_instance/cr_departures_test.exs
```diff
@@ -53,6 +53,86 @@ defmodule Screens.V2.WidgetInstance.CRDeparturesTest do
...
+                 departure,
+                 "Somewhere",
+                 %{},
+                 "place-smwhr",
+                 now
+               )
+    end
+  end
+
   # describe "serialize_times_with_crowding/2" do
```
cmaddox5, line 136: I'm curious why these are here 🤔 It's been here for 2 years so it feels safe to remove. Will leave it up to you.

cmaddox5 (approved):

test/screens/v2/widget_instance/cr_departures_test.exs
sloanelybutsurely: Easy enough: https://github.com/mbta/screens/pull/2170

PR #2169, opened 2024-08-28
chore: suppress alert 590467 on pre-fare screens

sloanelybutsurely:
Alert #590467 represents suspension of red line service on the Braintree branch with shuttles connecting at Ashmont instead of JFK/UMass. Because of this abnormal shuttle pattern, the generated line maps will not be accurate. Evergreen content will be used to display the correct information to riders.

**Asana task**: [Custom graphics during Braintree Branch diversion](https://app.asana.com/0/1185117109217422/1208136071296475/f)

- [x] Tests added?

### DIFF
--- a/lib/screens/v2/widget_instance/reconstructed_alert.ex
+++ b/lib/screens/v2/widget_instance/reconstructed_alert.ex
@@ -1313,10 +1313,10 @@ defmodule Screens.V2.WidgetInstance.ReconstructedAlert do
 
   def alert_ids(%__MODULE__{} = t), do: [t.alert.id]
 
-  def valid_candidate?(%__MODULE__{} = t) do
-    test_alert_ids = ["197140"]
-    prod_alert_ids = ["580015"]
-    t.alert.id not in (test_alert_ids ++ prod_alert_ids)
+  @suppressed_alert_ids ~w[590467]
+
+  def valid_candidate?(%__MODULE__{alert: %{id: alert_id}}) do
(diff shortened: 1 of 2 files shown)


lib/screens/v2/widget_instance/reconstructed_alert.ex
```diff
@@ -1313,10 +1313,11 @@ defmodule Screens.V2.WidgetInstance.ReconstructedAlert do
 
   def alert_ids(%__MODULE__{} = t), do: [t.alert.id]
 
-  def valid_candidate?(%__MODULE__{} = t) do
-    test_alert_ids = ["197140"]
-    prod_alert_ids = ["580015"]
-    t.alert.id not in (test_alert_ids ++ prod_alert_ids)
+  @test_alert_ids ~w[197140]
+  @prod_alert_ids ~w[580015 590467]
```
cmaddox5, line 1317: Harmless to leave them, but the IDs that were already here can be removed. Alerts can't be reopened so we know those IDs won't ever be active again.

cmaddox5 (approved):

lib/screens/v2/widget_instance/reconstructed_alert.ex
sloanelybutsurely: [Done!](https://github.com/mbta/screens/compare/90b443cc218239221a03ab279021a00c63079094..9841db120da6be30070a3ce50ddac4c527c89a92) 🚮

PR #2174, opened 2024-08-29
fix: filter cancelled & skipped departures

sloanelybutsurely:
Parses the `schedule_relationship` field for `Predictions` and uses that data to remove departures with a SKIPPED or CANCELLED schedule relationship when merging `Prediction`s and `Schedule`s.

Some Predictions now have a SKIPPED schedule relationship **AND** arrival/departure times (indicating what time the vehicle is predicted to pass through the stop). These Predictions need to be filtered out when we fetch departures. Previously these didn't have times so our other filtering logic would remove them.

For completeness we are also filtering out Predictions with a CANCELLED schedule relationship.

**Asana task**: [Filter out cancelled/skipped departures on Screens](https://app.asana.com/0/1185117109217413/1207498013010848/f)

### DIFF
--- a/lib/screens/predictions/parser.ex
+++ b/lib/screens/predictions/parser.ex
@@ -1,5 +1,6 @@
 defmodule Screens.Predictions.Parser do
   @moduledoc false
+  alias Screens.Predictions.ScheduleRelationship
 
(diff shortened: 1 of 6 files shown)


digitalcora (approved):
Looks good!

PR #2175, opened 2024-08-29
chore: Comment Mercury logic

cmaddox5:
Some of the Mercury specific logic is not at all self-explanatory without context so I added some comments explaining why it exists and the Asana task used to track the work.


sloanelybutsurely (approved):
you love to see it

PR #2178, opened 2024-08-30
fix: always use API_V3_URL environment variable

sloanelybutsurely:
Our ECS tasks all set a `API_V3_URL` pointing to the desired V3 API environment but we were ignoring that value and instead deriving the V3 API URL based on `ENVIRONMENT_NAME`. This is confusing.

With this change the URL used for the V3 API is always fetched from the environment as `API_V3_URL` (unless `MIX_ENV=test`).

This has the added benefit of picking up the changes in https://github.com/mbta/devops/pull/2122

**Asana task**: [Point screens-dev-green at api-dev-blue](https://app.asana.com/0/1185117109217413/1207488840910528/f)

### DIFF
--- a/config/runtime.exs
+++ b/config/runtime.exs
@@ -4,23 +4,15 @@
 # remember to add this file to your .gitignore.
 import Config
 
-eb_env_name = System.get_env("ENVIRONMENT_NAME")
-
-api_v3_url =
-  case eb_env_name do
-    "screens-prod" -> "https://api-v3.mbta.com/"
-    "screens-dev" -> "https://api-dev.mbtace.com/"
-    "screens-dev-green" -> "https://api-dev-green.mbtace.com/"
-    _ -> System.get_env("API_V3_URL", "https://api-v3.mbta.com/")
-  end
-
 unless config_env() == :test do
   config :screens,
-    api_v3_url: api_v3_url,
+    api_v3_url: System.get_env("API_V3_URL", "https://api-v3.mbta.com/"),
     api_v3_key: System.get_env("API_V3_KEY")
 end
 
 if config_env() == :prod do
+  eb_env_name = System.get_env("ENVIRONMENT_NAME")
+
   config :sentry,
     dsn: System.get_env("SENTRY_DSN"),
     environment_name: eb_env_name


cmaddox5 (approved):

digitalcora:
Seems odd that we have two ways of doing this. If we have `API_V3_URL` in deployed environments, any reason we shouldn't just always look at that, and delete this extra condition based on `ENVIRONMENT_NAME`?

sloanelybutsurely:
@digitalcora that's a great point! we _do_ have this as an environment variable in [prod](https://github.com/mbta/devops/blob/95d9646b6f809d8d7d781bbb479709ff7f2f3da7/terraform/prod/screens.tf#L11), [dev](https://github.com/mbta/devops/blob/95d9646b6f809d8d7d781bbb479709ff7f2f3da7/terraform/aws-tid-main-dev/screens.tf#L11), and [dev-green](https://github.com/mbta/devops/blob/95d9646b6f809d8d7d781bbb479709ff7f2f3da7/terraform/aws-tid-main-dev/screens.tf#L21). i'll go ahead and refactor this code to always use `API_V3_URL`.

---

the signs_ui s3 bucket is also configured based on `ENVIRONMENT_NAME` but those values are available as a terraform variable (would require exposing them via environment variable in the ECS task definition). i'll open separate PRs for those changes.

https://github.com/mbta/screens/blob/c7d9904d868224e4feacc458a7d5eb0a6c60ff33/config/runtime.exs#L28-L34

digitalcora:
I suppose we could also say that `ENVIRONMENT_NAME` is the one true way, and derive everything from that. Advantage would be that it doesn't require Terraform changes to change how our environments are connected. But then that's kind of putting "infrastructural" data in the app code.

sloanelybutsurely:
> I suppose we could also say that `ENVIRONMENT_NAME` is the one true way, and derive everything from that. Advantage would be that it doesn't require Terraform changes to change how our environments are connected. But then that's kind of putting "infrastructural" data in the app code.

@digitalcora my preference is to always use the environment variables for each individual parameter. this keeps that kind of configuration in terraform as a single source of truth for connections between applications

config/runtime.exs
```diff
@@ -6,17 +6,9 @@ import Config
...
-    "screens-prod" -> "https://api-v3.mbta.com/"
-    "screens-dev" -> "https://api-dev.mbtace.com/"
-    "screens-dev-green" -> "https://api-dev-green.mbtace.com/"
-    _ -> System.get_env("API_V3_URL", "https://api-v3.mbta.com/")
-  end
-
 unless config_env() == :test do
   config :screens,
-    api_v3_url: api_v3_url,
+    api_v3_url: System.get_env("API_V3_URL") || raise("Must set API_V3_URL"),
```
digitalcora, line 11: `System.fetch_env!/1` does the equivalent of this. (oops, commented this on the commit itself before, which was then rewritten out of existence)

sloanelybutsurely: yep i can do that and keep the default value. the evolution of this change is that `api_v3_url` was originally calculated outside of this `unless` so i removed the default to make it harder to accidentally run tests against they actual V3 API.

cmaddox5 (approved):
Sorry for missing what Cora caught. And thanks for fixing it.

PR #2179, opened 2024-08-30
chore: test logging output of serialize_departure/5

sloanelybutsurely:
`Screens.V2.WidgetInstance.CRDepartures.serialize_departure/5` logs and error for departures with a prediction but no schedule. This updates tests to make assertions on that log output and also prevents that logging from showing when running tests, keeping things nice and clean! 🧼


digitalcora (approved):
Just noticed this myself, but you were faster 😄

PR #2180, opened 2024-08-30
chore: use SIGNS_UI_S3_BUCKET environment variable

sloanelybutsurely:
References the new environment variable from https://github.com/mbta/devops/pull/2129 instead of deriving this value based on `ENVIRONMENT_NAME`

> [!WARNING]
> ~~https://github.com/mbta/devops/pull/2129 must be deployed first!~~ **Update: This is done ✅**

### DIFF
--- a/config/runtime.exs
+++ b/config/runtime.exs
@@ -17,21 +17,13 @@ if config_env() == :prod do
     dsn: System.get_env("SENTRY_DSN"),
     environment_name: eb_env_name
 
-  signs_ui_s3_bucket =
-    case eb_env_name do
-      "screens-prod" -> "mbta-signs"
-      "screens-dev" -> "mbta-signs-dev"
-      "screens-dev-green" -> "mbta-signs-dev"
-      _ -> nil
-    end
-
   config :screens, ScreensWeb.Endpoint,
     http: [:inet6, port: String.to_integer(System.get_env("PORT") || "4000")],
     secret_key_base: System.get_env("SECRET_KEY_BASE")
 
   config :screens,
     environment_name: eb_env_name,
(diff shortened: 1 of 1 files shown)


cmaddox5:
A small thing I noticed. SignsUI dev and dev-green share the same bucket but use different config files (which we store in `config.exs` under `:signs_ui_s3_path`), with the only difference being `dev -> config.json` and `dev-green -> config-dev-green.json`. Can we also make the change to reflect that here?

digitalcora:
Given the existing logic was to always use the config for dev and not dev-green when pointed at the "dev bucket", I'd be fine with leaving that as it is for this change — no need to introduce more environment variables / configuration unless we have a known use case for them.

sloanelybutsurely:
> SignsUI dev and dev-green share the same bucket but use different config files (which we store in `config.exs` under `:signs_ui_s3_path`), with the only difference being `dev -> config.json` and `dev-green -> config-dev-green.json`.

@cmaddox5 Where do you see this? I only see one place where `signs_ui_s3_path` is set.

https://github.com/mbta/screens/blob/b3523ed157dede73e204b4e969e4bf3927f68065/config/config.exs#L71

[🔍 Code search for `signs_ui_s3_path`](https://github.com/search?q=repo%3Ambta%2Fscreens+signs_ui_s3_path&type=code)

cmaddox5:
@sloanelybutsurely I was referencing the actual config name for each environment in SignsUI. So technically, we aren't matching up our dev-green environment to the correct config. But I agree with Cora, this is a different issue so it can be deferred to a later change if needed.

cmaddox5 (approved):

sloanelybutsurely:
Just waiting on the infra changes to be applied to `prod` before merging

PR #2185, opened 2024-09-03
refactor: reorganize screen data API

digitalcora:
Key changes:

* Moves responsibility for the top-level structure of the "screen data API" from the `ScreenData` module into the API controller. The former now only has the responsibility of generating serialized widgets from screen configuration.

* Moves widget generation and placement from the `ScreenData` module into a new `Layout` module. This is a complex piece of logic with a small public interface, which is already shared with another module (`ScreenAudioData`), so it was a good candidate for extraction. _(Not sure if this is the best name for it?)_

* Simplifies the interface of `ScreenData` to two functions, one for "normal" screen data and one for Screenplay simulation data. Pending config is handled by passing it in as an option. Updating the cache of "visible alerts" is also handled with an explicit option rather than implicitly whenever non-pending, non-simulation data is requested.


digitalcora:
It was part of trying to figure out how the "DUP logic switch" would fit into all this, and then moving things towards a state where that would be easier to add. As sometimes happens, I then got a little carried away with repeatedly seeing little opportunities for further refinement. Glad you also think it's an improvement, though!

PR #2190, opened 2024-09-05
feat: Sectional Audio

cmaddox5:
**Asana task**: [Create new Sectional audio](https://app.asana.com/0/1185117109217413/1206730847640130/f)

This PR adds the code necessary to enable audio on Sectionals. Luckily, the departures widget is already configured to readout so there were no widget changes necessary. The default for the config is set up so audio is always enabled. We shouldn't need to do anything after deploying.

- [ ] Tests added?


lib/screens/v2/widget_instance/departures.ex
```diff
@@ -280,18 +286,26 @@ defmodule Screens.V2.WidgetInstance.Departures do
       {ref, [notice]} when is_reference(ref) ->
         {:notice, notice}
 
-      {_key, departure_group} ->
-        {:normal, Enum.take(departure_group, max_entries_per_group)}
+      {_key, [first_departure | _] = departure_group} ->
+        departures =
+          if first_departure |> Departure.time() |> DateTime.diff(now, :minute) <= 2 do
```
digitalcora, line 291: Unless I'm missing something, this will reduce audio readouts from 2 departures per route to 1 (in the typical case) on _all_ screen types that use the shared departures widget, including bus shelters and e-ink. Admittedly the requested changes in Asana don't specifically say "only on Sectionals" but that might have been the implicit context. I can definitely imagine that in some cases we'd want to keep the current "always 2" readouts — GL e-ink never even displays more than two departures, so the readouts are already quite short and the extra information is probably appreciated. Maybe we should ask Dave to be sure.

lib/screens_web/views/v2/audio/departures_view.ex
```diff
@@ -96,10 +102,14 @@ defmodule ScreensWeb.V2.Audio.DeparturesView do
       {route_text, &render_route/1},
       {vehicle_type || "trip", fn v -> ~E|<%= v %>| end},
       {headsign, fn h -> ~E|to <%= render_headsign(h) %>| end},
-      {track_number, fn tn -> ~E|on track <%= tn %>| end}
+      {track_number, fn tn -> render_track_number(tn, vehicle_type) end}
     ])
   end
 
+  defp render_track_number(nil, _), do: ~E""
+  defp render_track_number(track_number, :bus), do: ~E|at berth <%= track_number %><break/>|
```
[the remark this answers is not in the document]
cmaddox5: Thank you for calling this out. I thought the same thing and meant to call out that I will make this change after this PR is merged. Wanted to limit the diff to just the new audio. 

And will add tests now.

lib/screens_web/views/v2/audio/departures_view.ex
```diff
@@ -64,8 +64,14 @@ defmodule ScreensWeb.V2.Audio.DeparturesView do
     ~E|<s><%= content %></s>|
   end
 
-  defp render_time_with_crowding({%{crowding: crowding, time: time}, _}, route, _headsign) do
-    route_headsign_rendered = render_route_headsign(route, nil)
+  defp render_time_with_crowding({%{crowding: crowding, time: time}, _}, route, headsign) do
+    route_headsign_rendered =
+      if route.vehicle_type == :train do
+        render_route_headsign(route, headsign)
```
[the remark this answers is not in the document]
cmaddox5: Hmm yeah thank you. Will correct this.

lib/screens/v2/widget_instance/departures.ex
```diff
@@ -259,18 +264,24 @@ defmodule Screens.V2.WidgetInstance.Departures do
...
+          DateTime.t(),
+          integer(),
+          Screen.app_id()
+        ) ::
           list(
             {:normal, list(Departure.t())}
             | {:notice, notice}
           )
-  def group_all_departures(departures, max_entries_per_group) do
+  def group_all_departures(departures, now, max_entries_per_group, app_id) do
```
digitalcora, line 284: * Do we really need `max_entries_per_group` as a parameter here? In the one place this function is called, the value is hard-coded to `2`. So it seems like we could just push the hard-coding a level down and not have the parameter. (we have `Enum.take(departure_group, 2)` below, and that `2` is hard-coded, which I think is fine)
* As far as I can tell, this could be a `defp`; nothing else calls it.

lib/screens_web/views/v2/audio/departures_view.ex
```diff
@@ -64,8 +64,14 @@ defmodule ScreensWeb.V2.Audio.DeparturesView do
     ~E|<s><%= content %></s>|
   end
 
-  defp render_time_with_crowding({%{crowding: crowding, time: time}, _}, route, _headsign) do
-    route_headsign_rendered = render_route_headsign(route, nil)
+  defp render_time_with_crowding({%{crowding: crowding, time: time}, _}, route, headsign) do
+    route_headsign_rendered =
+      if String.starts_with?(route.route_text, "Green Line") do
```
digitalcora, line 69: Any way we can flag this special case at a level where we still have access to the route ID? Making the behavior conditional on the user-visible (or rather user-audible) name of the route feels fragile.

test/screens/v2/widget_instance/departures_test.exs
```diff
@@ -950,11 +966,171 @@ defmodule Screens.V2.WidgetInstance.DeparturesTest do
...
+              route: %Route{type: :subway},
+              trip: %Trip{headsign: "Test"},
+              stop: %Stop{id: "place-test"}
+            }
+          }
+        ],
+        header: %Header{title: "Section Header"}
+      }
+
+      screen = struct(Screen, %{app_id: :busway_v2})
```
digitalcora, line 1011: Any particular reason you used `struct` for these? It seems equivalent to `%Screen{app_id: :busway_v2}`, but with less compile-time safety.

lib/screens/v2/widget_instance/departures.ex
cmaddox5: Yeah I suppose it's a bit unnecessary.

test/screens/v2/widget_instance/departures_test.exs
cmaddox5: Just to automatically apply the enforced keys that we don't care about. Creating the `struct` without some hardware specific fields throws a compilation error.

lib/screens/v2/candidate_generator/widgets/departures.ex
```diff
@@ -54,7 +53,7 @@ defmodule Screens.V2.CandidateGenerator.Widgets.Departures do
         Keyword.get(options, :departure_fetch_fn, &Departure.fetch/2),
         Keyword.get(options, :post_process_fn, fn results, _config -> results end),
         Keyword.get(options, :route_fetch_fn, &Route.fetch/1),
-        now
+        options[:now]
```
digitalcora, line 56: It seems like this is now effectively a required argument, rather than an "option" with a default value — if you didn't provide it, downstream code would crash when it expected a `DateTime` but got `nil`. I think if it's optional it should be a keyword with a default value, but if it's required it should be its own argument. No preference as to which.

digitalcora (approved):
One last comment, but looking good. One step closer to dropping v1 screens!

lib/screens/v2/candidate_generator/widgets/departures.ex
cmaddox5: Yeah good point, didn't think about that. Making it optional seems fine and lines up with the rest of the list.

cmaddox5: Actually, I think I'm changing my mind. We already default `now` in the app CGs so there's already the variable available to pass to this.

PR #2191, opened 2024-09-05
feat: track last train of the day

sloanelybutsurely:
Tracks trips marked as `last_trip: true` and recent departures from all route-direction-stop tuples.

Fetches TripUpdates and VehiclePositions data from concentrate S3 bucket to populate two caches (one for last trips and one for a list of recent departures from a stop on a given route going a direction).

Both caches have TTLs of one hour so old trips and departures will fall off over time. Recent departures are a special case because they are tracked as a list of recent departures. Departures (tuples of trip_id and departure time in unix epoch) are upserted into the cache key for a given RDS. During update departures more than one hour in the past are removed from the list for that RDS.

Caches are reset every day after 3:30am EST.

> [!NOTE]
> The realtime_signs code uses `If-Modified-Since` and `Last-Modified` headers presumably in an attempt to reduce unnecessary API requests. I added support for that behavior into this flow as well but because the requested data changes so frequently it almost _always_ had to fetch all of the data. I've removed that functionality to keep the code simpler given that there is almost no benefit to tracking those time stamps / previous responses.

**Asana Task:** [Start tracking LTOTD state for use by screens][task]

[task]: https://app.asana.com/0/1185117109217413/1207574972639078/f

### DIFF
--- a/.envrc.template
+++ b/.envrc.template
@@ -9,3 +9,9 @@ export API_V3_URL="https://api-v3.mbta.com"
 # export AWS_SECRET_ACCESS_KEY=
 
 export SECRET_KEY_BASE="local_secret_key_base_at_least_64_bytes_________________________________"
+
+# # Should point to the Enhanced JSON feeds. Default values are provided, only
+# # set these if you need something different.
+# # https://github.com/mbta/gtfs-documentation/blob/master/reference/gtfs-realtime.md#enhanced-json-feeds
+# export TRIP_UPDATES_URL=
+# export VEHICLE_POSITIONS_URL=
--- a/lib/screens/last_trip.ex
+++ b/lib/screens/last_trip.ex
@@ -1,0 +1,41 @@
+defmodule Screens.LastTrip do
+  @moduledoc """
+  Supervisor and public interface for fetching information about the last trips
+  of the day (AKA Last Train of the Day, LTOTD).
+  """
+  alias Screens.LastTrip.Cache
+  alias Screens.LastTrip.Poller
+  use Supervisor
+
+  @spec start_link(any()) :: Supervisor.on_start()
+  def start_link(_) do
+    Supervisor.start_link(__MODULE__, [], name: __MODULE__)
+  end
+
+  @impl true
+  def init(_) do
+    children = [
+      Cache.LastTrips,
+      Cache.RecentDepartures,
+      Poller
+    ]
+
+    Supervisor.init(children, strategy: :one_for_one)
+  end
+
+  @spec last_trip?(trip_id :: String.t()) :: boolean()
+  defdelegate last_trip?(trip_id), to: Cache
+
+  @spec service_ended_for_rds?(Cache.rds()) :: boolean()
+  def service_ended_for_rds?({_r, _d, _s} = rds, now_fn \\ &DateTime.utc_now/0) do
+    now_unix = now_fn.() |> DateTime.to_unix()
+
+    rds
+    |> Cache.get_recent_departures()
+    |> Enum.any?(fn {trip_id, departure_time_unix} ->
+      seconds_since_departure = now_unix - departure_time_unix
+
+      seconds_since_departure > 3 and last_trip?(trip_id)
+    end)
+  end
+end
--- a/lib/screens/last_trip/poller.ex
+++ b/lib/screens/last_trip/poller.ex
@@ -1,0 +1,86 @@
+defmodule Screens.LastTrip.Poller do
+  @moduledoc """
+  GenServer that polls predictions to calculate the last trip of the day
+  """
+  alias Screens.LastTrip.Cache
+  alias Screens.LastTrip.Parser
+  alias Screens.LastTrip.TripUpdates
+  alias Screens.LastTrip.VehiclePositions
+  use GenServer
+
+  defstruct [:next_reset]
+
+  @polling_interval :timer.seconds(1)
+
+  def start_link(_) do
+    GenServer.start_link(__MODULE__, [], name: __MODULE__)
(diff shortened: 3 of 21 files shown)


digitalcora (commented):
This looks solid, though I am a little concerned it doesn't have any tests that would _assert_ it is solid. No typespecs for public functions, also, which made things a little harder to follow than they could've been (I especially would have appreciated types for the cache keys and values). Is the idea that this is copied closely enough from RTS that our "version" of it doesn't really need its own tests?

mix.exs
```diff
@@ -90,7 +90,8 @@ defmodule Screens.MixProject do
       {:nebulex, "~> 2.6"},
       {:remote_ip, "~> 1.2"},
       {:hackney_telemetry, "~> 0.2.0"},
-      {:ex_cldr_messages, "~> 1.0"}
+      {:ex_cldr_messages, "~> 1.0"},
+      {:req, "~> 0.5.6"}
```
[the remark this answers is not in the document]
sloanelybutsurely: oops! i used this for a moment but swapped it out for HTTPoison

lib/screens/last_trip/poller.ex
```diff
@@ -0,0 +1,123 @@
...
+  GenServer that polls predictions to calculate the last trip of the day
+  """
+  alias Screens.LastTrip.Cache
+  alias Screens.LastTrip.TripUpdates
+  alias Screens.LastTrip.VehiclePositions
+  use GenServer
+
+  defstruct [:next_reset]
+
+  @polling_interval :timer.seconds(1)
```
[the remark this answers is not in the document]
sloanelybutsurely: I took this interval from [realtime_signs](https://github.com/mbta/realtime_signs/blob/ca2822395970f23daf1533c9cc302fa593b241c0/lib/engine/last_trip.ex#L110-L112). These feeds are updated much more frequently than every 5 seconds.

Polling for changes shows that the feed can often change every second.

```elixir
[
  ["Mon, 16 Sep 2024 15:05:42 GMT"],
  ["Mon, 16 Sep 2024 15:05:43 GMT"],
  ["Mon, 16 Sep 2024 15:05:44 GMT"],
  ["Mon, 16 Sep 2024 15:05:46 GMT"],
  ["Mon, 16 Sep 2024 15:05:47 GMT"],
  ["Mon, 16 Sep 2024 15:05:48 GMT"],
  ["Mon, 16 Sep 2024 15:05:50 GMT"],
  ["Mon, 16 Sep 2024 15:05:51 GMT"],
  ["Mon, 16 Sep 2024 15:05:52 GMT"]
]
```

<details>
<summary>script 📜</summary>

```elixir
# Run as: iex --dot-iex path/to/notebook.exs

# Title: Feed change frequency

Mix.install([
  {:req, "~> 0.5.6"}
])

# ── Section ──

defmodule TripUpdatesEnhanced do
  @trip_updates_url "https://cdn.mbta.com/realtime/TripUpdates_enhanced.json"

  def get do
    Req.get!(@trip_updates_url)
  end
end

Stream.repeatedly(fn -> 
  Process.sleep(250)
  TripUpdatesEnhanced.get().headers["last-modified"]
end)
|> Enum.take(25)
|> Enum.uniq()

```

I'm happy to dial this back the polling interval back regardless. Are you wanting to poll every 5 seconds?

.envrc.template
```diff
@@ -9,3 +9,6 @@ export API_V3_URL="https://api-v3.mbta.com"
 # export AWS_SECRET_ACCESS_KEY=
 
 export SECRET_KEY_BASE="local_secret_key_base_at_least_64_bytes_________________________________"
+
+# export TRIP_UPDATES_URL=
+# export VEHICLE_POSITIONS_URL=
```
cmaddox5, line 14: Very nitpicky: It might be helpful to add the link to [this section](https://github.com/mbta/gtfs-documentation/blob/master/reference/gtfs-realtime.md#enhanced-json-feeds) here for reference when a dev is onboarding. Cora's comment is actually my first time seeing this.

lib/screens/last_trip.ex
```diff
@@ -0,0 +1,41 @@
+defmodule Screens.LastTrip do
+  @moduledoc """
+  Supervisor and public interface for fetching information about the last trips
+  of the day (Last Train of the Day, LTOTD).
```
cmaddox5, line 4: Very nitpicky: This looks to be the only place where it's called `Last Train` instead of `Last Trip`. Trip seems a little better because this isn't mode specific.

cmaddox5 (approved):
This looks great! Can't wait to see it in action. Have just a couple non-blocking nitpicks.

lib/screens/last_trip.ex
sloanelybutsurely: hmm i put these parenthetical in here so someone searching for "last **train**" or "LTOTD" would still find the relevant code. would you still like me to remove these?

.envrc.template
sloanelybutsurely: [np!](https://github.com/mbta/screens/compare/251e06a912af136a72606c623d7066023a42c067..0c9ca14d32f20fa5ad47b1999608bc9af8b7404f)

lib/screens/last_trip.ex
sloanelybutsurely: how about [this?](https://github.com/mbta/screens/compare/0c9ca14d32f20fa5ad47b1999608bc9af8b7404f..413ff59c88929bcf9465b04022b3bcb383538f94)

cmaddox5: Yeah, was actually mid-comment suggesting something like that. Nice!

PR #2203, opened 2024-09-16
chore: Remove Triptychs

cmaddox5:
The Triptych app is dead. Removing all code.


digitalcora:
Looks like we've got quite the batch of conflicts here. Sorry for the shake-up, I think that's my fault.

cmaddox5:
No problem! I think I resolved all of it.

assets/src/components/v2/dup/README.md
```diff
@@ -44,7 +44,7 @@
     ```
     This sets up a fake MRAID object that emulates the real one available to the client when running on Outfront screens.
     The MRAID object gives our client info about which screen it's running on.
-  - replace the definition of `OUTFRONT_BASE_URI` in `assets/src/hooks/v2/use_api_response.tsx` with `"http://localhost:4000"`.
+  - replace the definition of `getOutfrontAbsolutePath` in assets/src/hooks/v2/use_api_response.tsx with `const getOutfrontAbsolutePath = () => isDup() ? "http://localhost:4000" : "";`.
```
digitalcora, line 47: Rebase mishap? I think the `main` version of this is the correct one.

assets/src/util/outfront.tsx
```diff
@@ -44,7 +24,7 @@ export const getRotationIndex = (): RotationIndex | null => {
 /**
  * Gets Outfront's unique name/ID for the media player we're running on.
  *
- * For DUPs, this is just a single ID. For triptychs, each of the 3 panes has its own player name.
+ * For DUPs, this is just a single ID.
```
digitalcora, line 27: Can probably remove this "paragraph" entirely as it was only meaningful in the context of the comparison between DUPs and Triptychs.

lib/screens/v2/widget_instance/evergreen_content.ex
```diff
@@ -31,12 +31,6 @@ defmodule Screens.V2.WidgetInstance.EvergreenContent do
 
   def priority(%__MODULE__{} = instance), do: instance.priority
 
-  def serialize(%__MODULE__{
-        screen: %Screen{app_params: %Triptych{show_identifiers: show_identifiers}},
-        asset_url: asset_url
-      }),
-      do: %{asset_url: asset_url, show_identifiers: show_identifiers}
```
digitalcora, line 38: The `show_identifiers` field doesn't seem to be used aside from Triptychs. Can we remove it from this module's struct+type outright?

lib/screens_web/controllers/v2/screen_api_controller.ex
```diff
@@ -90,8 +89,7 @@ defmodule ScreensWeb.V2.ScreenApiController do
             update_visible_alerts?: true,
             logging_options: %{
```
digitalcora, line 90: This whole option (`logging_options`) was only used on Triptychs, so we should be able to completely remove it. For that matter, I think it was the only reason options needed to be passed down from `ScreenData` into candidate generators at all.

assets/src/components/v2/dup/README.md
cmaddox5: Ahh, yes thank you.

lib/screens_web/controllers/v2/screen_api_controller.ex
cmaddox5: Nice catch. I didn't notice that.

digitalcora (approved):
Let's deploy this to dev-green and make sure nothing crashes, and maybe also give the DUP packaging process another run-through and ensure it still works as a `file://`. Other than that, this looks good. Any day where you get to delete 3,500 lines is a good day!

cmaddox5:
DUP packaging looks good: 
[dup-app-0.zip](https://github.com/user-attachments/files/17015569/dup-app-0.zip)
[dup-app-1.zip](https://github.com/user-attachments/files/17015572/dup-app-1.zip)
[dup-app-2.zip](https://github.com/user-attachments/files/17015573/dup-app-2.zip)

Will let the branch sit on dev-green for a bit before merging.

PR #2204, opened 2024-09-16
Initial widgets for new DUP logic

digitalcora:
Rather than being entirely placeholders, the `new_departures` DUP variant now generates header and evergreen content widgets that should work identically to the base variant. The "departures" area is stubbed to always render a "no data" message. (This PR incidentally fixes an existing bug with that message where it would render a duplicate header, pushing down and cutting off the content.)

Example of a `new_departures` DUP being rendered with an evergreen content item enabled, and without:

Screenshot 2024-09-16 at 2 04 44 PM

---
  - https://app.asana.com/0/0/1207992837169563


cmaddox5 (approved):
Good call on the telemetry additions.

PR #2211, opened 2024-09-18
fix: Filter out predictions w/ arrival_time but no departure_time.

cmaddox5:
This PR adds additional filtering to predictions returned from the API. When a prediction has an `arrival_time` but no `departure_time`, it is an API best practice to filter it out. 

---
  - https://app.asana.com/0/0/1208058010923825

### DIFF
--- a/lib/screens/v2/departure/builder.ex
+++ b/lib/screens/v2/departure/builder.ex
@@ -17,6 +17,7 @@ defmodule Screens.V2.Departure.Builder do
   def get_relevant_departures(predictions_or_schedules, now \\ DateTime.utc_now()) do
     predictions_or_schedules
     |> Stream.reject(&in_past_or_nil_time?(&1, now))
+    |> Stream.reject(&prediction_has_no_departure_time?/1)
     |> Stream.reject(&multi_route_duplicate?/1)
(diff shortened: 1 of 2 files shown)


sloanelybutsurely (approved):

lib/screens/v2/departure/builder.ex
```diff
@@ -58,6 +59,12 @@ defmodule Screens.V2.Departure.Builder do
 
   defp vehicle_already_departed?(_), do: false
 
+  defp prediction_has_no_departure_time?(%Schedule{}), do: false
```
digitalcora, line 62: We should also do this for scheduled departures, I think (filtering them out if they have no departure time). We don't want to display an "arrival-only" stop that won't be accepting passengers, even if it's not a live prediction.

cmaddox5:
Turns out that we actually do already have this implemented for both [predictions](https://github.com/mbta/screens/blob/6f99840d662f5ecba995cb292302117c62b6b35e/lib/screens/predictions/prediction.ex#L45) and [schedules](https://github.com/mbta/screens/blob/6f99840d662f5ecba995cb292302117c62b6b35e/lib/screens/schedules/schedule.ex#L41).

PR #2218, opened 2024-09-26
Refactors from RDS generation

digitalcora:
Implements the updated ["RDS generation" algorithm](https://www.notion.so/mbta-downtown-crossing/Real-time-departure-states-13dddc816adb43e79a3d952c246b08f7?pvs=4#347394ea5da94f4684ef0580783b09bd) and stubs out a `NoDepartures` state for all RDSes. Rather than doing this specifically "from DUP screen configs", the module accepts a `Departures` widget configuration struct, which is common to all screen types (though DUPs are a little odd in that they have two independent instances of this, for the multiple rotations — some glue logic will be needed in the candidate generator to accommodate this).

Currently this is not connected to anything "real" but will shortly be hooked up to the `new_departures` DUP variant as specified in the dependent Asana task.

---
  - https://app.asana.com/0/0/1207992837169561

### DIFF
--- a/lib/screens/alerts/alert.ex
+++ b/lib/screens/alerts/alert.ex
@@ -5,6 +5,7 @@ defmodule Screens.Alerts.Alert do
   alias Screens.Routes.Route
   alias Screens.RouteType
   alias Screens.Stops.Stop
+  alias Screens.Trips.Trip
   alias Screens.V3Api
 
   defstruct id: nil,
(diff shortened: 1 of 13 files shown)


cmaddox5 (approved):
Looks good! Big fan of the approach.

digitalcora:
Reverting to draft since I think we may be changing direction on this somewhat — it's emerged from design and engineering conversations that making the "RDS" the core of this algorithm may not quite result in the behavior we want in cases like the two branches of the Red Line or the multiple routes of the Green Line. Apologies if it took a bunch of your time to review, @cmaddox5 🙇‍♀️ I'm hoping a good chunk of this can be reused in the end.

digitalcora:
To "make the most" of this approval, I'm tearing all the RDS-specific work off this branch and leaving only the refactor commits, which are still useful in isolation. The rework of the "destinations" logic will be a new PR.

PR #2224, opened 2024-10-02
chore: Split DUP packages into separate webpack entry

cmaddox5:
As we start working towards upgrading webpack, we want to split out DUP packages into a separate build. We know that the current configuration works for live DUPs so we want to ensure it remains unchanged as we upgrade webpack. 

---
  - https://app.asana.com/0/0/1208162115483285

### DIFF
--- a/assets/webpack.config.js
+++ b/assets/webpack.config.js
@@ -98,11 +98,11 @@ const common_plugins = [
 ];
 
 module.exports = (env, argv) => {
-  // Upload source maps to Sentry for prod builds. Must be the last plugin.
-  const appendPlugins =
+  const plugins =
     argv.mode == "production"
       ? [
           ...common_plugins,
+          // Upload source maps to Sentry for prod builds. Must be the last plugin.
           sentryWebpackPlugin({
             authToken: env.SENTRY_AUTH_TOKEN,
             org: env.SENTRY_ORG,
@@ -150,7 +150,7 @@ module.exports = (env, argv) => {
           ...getCommonRules(false),
         ],
       },
-      plugins: appendPlugins,
+      plugins: plugins,
     },
     {
       ...common_export_body,
@@ -185,7 +185,7 @@ module.exports = (env, argv) => {
           ...getCommonRules(true),
         ],
       },
-      plugins: appendPlugins,
+      plugins: plugins,
     },
   ];
 };
--- a/assets/webpack.config.js
+++ b/assets/webpack.config.js
@@ -8,6 +8,62 @@ const OptimizeCSSAssetsPlugin = require("optimize-css-assets-webpack-plugin");
 const CopyWebpackPlugin = require("copy-webpack-plugin");
 const { sentryWebpackPlugin } = require("@sentry/webpack-plugin");
 
+const common_export_body = {
+  resolve: {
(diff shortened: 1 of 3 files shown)


assets/webpack.config.js
```diff
@@ -71,48 +109,76 @@ module.exports = (env, argv) => {
...
+                  name: "[name].[ext]",
+                  outputPath: "fonts/",
+                  publicPath: "../fonts/",
+                  useRelativePaths: true,
+                },
               },
+            ],
+          },
+          {
+            test: /\.(png|jpe?g|gif|webp)$/i,
```
digitalcora, line 132: This rule seems entirely duplicated — we could extract it into `common_rules`?

assets/webpack.config.js
```diff
@@ -71,48 +109,76 @@ module.exports = (env, argv) => {
...
           },
+          ...common_rules,
           {
-            enforce: "pre",
-            test: /\.js$/,
-            loader: "source-map-loader",
-          },
-          {
-            test: /\.s?css$/,
+            test: /\.(woff(2)?|ttf|eot)(\?v=\d+\.\d+\.\d+)?$/,
```
digitalcora, line 118: I wonder if we could also de-duplicate this rule, since the DUP config only has one parameter that's different. Maybe a little helper function?

assets/webpack.config.js
```diff
@@ -143,7 +209,9 @@ module.exports = (env, argv) => {
       },
       plugins: [
         new MiniCssExtractPlugin({ filename: "../css/[name].css" }),
-        new CopyWebpackPlugin({ patterns: [{ from: "static/", to: "../" }] }),
+        new CopyWebpackPlugin({
+          patterns: [{ from: "static/fonts", to: "../fonts" }],
```
digitalcora, line 213: Remind me why we copy all of `static` in the non-DUP config but only `static/fonts` here?

cmaddox5: I thought at first it was a way to reduce package size but a quick test proved that wrong. No reason for them to be different.

digitalcora: Excellent — then I think some further deduplication is possible with these plugins.

cmaddox5:
Meant to attach this when opening the PR, but here is a DUP package using the most recent commit:

[dup-app-0.zip](https://github.com/user-attachments/files/17246379/dup-app-0.zip)

assets/webpack.config.js
```diff
@@ -8,40 +8,112 @@ const OptimizeCSSAssetsPlugin = require("optimize-css-assets-webpack-plugin");
...
+];
+
+const common_plugins = [
+  new MiniCssExtractPlugin({ filename: "../css/[name].css" }),
+  new CopyWebpackPlugin({ patterns: [{ from: "static/", to: "../" }] }),
+];
+
 module.exports = (env, argv) => {
   // Upload source maps to Sentry for prod builds. Must be the last plugin.
   const appendPlugins =
```
digitalcora, line 102: Perhaps this is now better described as just "plugins", since it includes everything. The above comment could be relocated to make it clear what it applies to, as well.

digitalcora (approved):
Looking better than ever!

PR #2236, opened 2024-10-08
chore: Cleanup Admin columns and default values

cmaddox5:
Main goal of this PR is to correct default value for `evergreen_content` when creating a screen use the `Add New Screen` modal. I also saw an opportunity to clean up/fix columns shown on the app filter tabs.

---
  - https://app.asana.com/0/0/1208461183523595

### DIFF
--- a/assets/src/components/admin/admin_add_modal.tsx
+++ b/assets/src/components/admin/admin_add_modal.tsx
@@ -90,8 +90,7 @@ const AddModal = ({ setData, closeModal }): JSX.Element => {
   const addScreen = () => {
     const newRow = {
       app_id: formValues.app_id,
-      // @ts-expect-error
-      app_params: defaultAppParamsByAppId[formValues.app_id],
+      app_params: defaultAppParamsByAppId[formValues.app_id ?? 0],
       device_id: formValues.device_id,
       disabled: false,
       id: formValues.id,
--- a/assets/src/components/admin/admin_add_modal.tsx
+++ b/assets/src/components/admin/admin_add_modal.tsx
@@ -36,47 +36,48 @@ const fields = [
   { key: "device_id", label: "Device ID", FormCell: FormTextCell },
 ];
 
-const commonV2Params = {
-  header: { stop_id: "" },
-  evergreen_content: [],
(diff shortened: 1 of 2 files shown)


assets/src/components/admin/admin_add_modal.tsx
```diff
@@ -42,52 +36,47 @@ const fields = [
   { key: "device_id", label: "Device ID", FormCell: FormTextCell },
 ];
 
+const commonV2Params = {
+  header: { stop_id: "" },
```
digitalcora, line 40: `stop_name` is the only header type supported by all V2 screens. Busway in particular doesn't support `stop_id`.

assets/src/components/admin/admin_add_modal.tsx
```diff
@@ -90,8 +90,7 @@ const AddModal = ({ setData, closeModal }): JSX.Element => {
   const addScreen = () => {
     const newRow = {
       app_id: formValues.app_id,
-      // @ts-expect-error
-      app_params: defaultAppParamsByAppId[formValues.app_id],
+      app_params: defaultAppParamsByAppId[formValues.app_id ?? 0],
```
digitalcora, line 93: I'm confused as to what this would do in the error case, or why the compiler is fine with this — there is no key `0` in the object.

digitalcora (approved):
Nice cleanup, a couple non-blocking comments.

assets/src/components/admin/admin_add_modal.tsx
cmaddox5: Hmm, thanks for catching that. Pre-Fare is actually the opposite and doesn't support `stop_name`. I may just need to apply these individually to match what the app accepts.

assets/src/components/admin/admin_add_modal.tsx
cmaddox5: This was a bad way to handle this in hindsight. I changed it a bit so `useState` is called with the type of `initialFormValues`. And since each field in `fields` is a string, I changed the initial values from `undefined` to `string`.

PR #2237, opened 2024-10-08
chore: Fix vendors in config

cmaddox5:
This PR updates `screens-config-lib` so we can correctly flag invalid vendors. I also found that the admin tool was using `lg-mri` instead of `lg_mri` that the config expects.


digitalcora (approved):
Good catch with the `-`/`_`!

PR #2238, opened 2024-10-10
Remove all DUP v1 code/references

cmaddox5:
We only have three DUP v1 configs and none of them are used by Outfront or have logged anything since July. I'm confident that this code is completely unused now.


lib/screens_web/templates/layout/app.html.eex
```diff
@@ -5,9 +5,6 @@
     <meta http-equiv="X-UA-Compatible" content="IE=edge"/>
     <meta name="viewport" content="width=device-width, initial-scale=1.0"/>
     <title>Screens</title>
-    <%= if @app_id != :dup do %>
```
digitalcora, line 8: Since the condition is "_not_ DUP", it seems like this stylesheet should now be included always, instead of never.

digitalcora (approved):
It Begins. ☠️🎉 Just one comment.

lib/screens_web/templates/layout/app.html.eex
cmaddox5: Oof, thank you.

PR #2240, opened 2024-10-11
Remove all V1 bus_eink references

cmaddox5:
There is only one `bus_eink` screen config in prod and it has no logs this year. Removing all the `bus_eink` code.


digitalcora (approved):
I see you're on a mission 😎

PR #2241, opened 2024-10-11
Remove gl_eink_double

cmaddox5:
We have no `gl_eink_double` configs so removing the code.


digitalcora (approved):
Good call on also removing the "double" from V2 GL E-inks, given it's the only kind.

PR #2242, opened 2024-10-11
Initial destinations framework

digitalcora:
This adds a module which generates "destinations" from screen configs, the first step in the proposed [real-time destinations](https://www.notion.so/mbta-downtown-crossing/Real-time-destination-states-114f5d8d11ea8077b79cef05225e5725) logic. Both "permanent" (from typical route patterns) and "live" (from upcoming departures) destinations are generated. Initially there is only one possible state for a destination, `NoDepartures`. Following tasks will add more states and wire these into the `DupNew` screen variant.

This required a lot of refactoring and expansion of how we fetch data. Individual commit messages have the explanations of this. 📝

---
  - https://app.asana.com/0/0/1207992837169561

### DIFF
--- a/lib/screens/route_patterns/route_direction_stops.ex
+++ b/lib/screens/route_patterns/route_direction_stops.ex
@@ -1,4 +1,4 @@
-defmodule Screens.RoutePatterns.Parser do
+defmodule Screens.RoutePatterns.RouteDirectionStops do
   @moduledoc false
 
   require Logger
--- a/config/test.exs
+++ b/config/test.exs
@@ -165,6 +165,11 @@ config :screens, Screens.V2.ScreenData,
 
 config :screens, Screens.V2.CandidateGenerator.DupNew, stop_module: Screens.Stops.MockStop
 
+config :screens, Screens.V2.RDS,
+  departure_module: Screens.V2.MockDeparture,
+  route_pattern_module: Screens.RoutePatterns.MockRoutePattern,
+  stop_module: Screens.Stops.MockStop
+
 config :screens, Screens.LastTrip,
   trip_updates_adapter: Screens.LastTrip.TripUpdates.Noop,
   vehicle_positions_adapter: Screens.LastTrip.VehiclePositions.Noop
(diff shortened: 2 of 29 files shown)


lib/screens/route_patterns/route_direction_stops.ex
```diff
@@ -0,0 +1,82 @@
...
+    Enum.map(stops_data, &parse_stop_data/1)
+  end
+
+  defp parse_stop_data(%{"id" => stop_id, "type" => "stop"}) do
+    stop_id
+  end
+
+  defp parse_data(data, included_data, route_id) do
+    filtered_data = filter_by_route(data, route_id)
+    [typical_data | _] = filtered_data
```
cmaddox5, line 41: Can this just be 
```suggestion
    [typical_data | _] = filter_by_route(data, route_id)
```

lib/screens/route_patterns/route_direction_stops.ex
```diff
@@ -0,0 +1,82 @@
...
+
+  defp parse_route_pattern(
+         %{
+           "relationships" => %{
+             "representative_trip" => %{"data" => %{"id" => trip_id, "type" => "trip"}}
+           }
+         },
+         included_data
+       ) do
+    # The only way this function output an empty array is if the trip data has an empty stop list
```
cmaddox5, line 65: nit: 
```suggestion
    # The only way this function outputs an empty array is if the trip data has an empty stop list
```

cmaddox5 (approved):
Looks good. The commit breakdown was very helpful. Just a couple non-blocking comments.

lib/screens/route_patterns/route_direction_stops.ex
digitalcora: Probably! I didn't write this at all — this is a rename of the original `Parser` that's not being detected as a rename, I assume because the "from" file still exists and has new contents. I think by splitting this up into a commit that does the rename, and a second commit that introduces the new `Parser`, it will be detected properly (and then blames and stuff will still work).

digitalcora: ...also these changes are incoherent, since I accidentally split the rename across 8f32b42 and 89cdbc0 🫠

PR #2246, opened 2024-10-15
chore: Add Notion doc link to README

cmaddox5:
I created a doc that goes into detail on the DUP package deployment process. Adding that link to the README.


digitalcora (approved):
✨

PR #2248, opened 2024-10-16
feat: Elevator Happy Path

cmaddox5:
Created the new Elevator screen widget with a hardcoded "happy path". At the moment, it only sends the configured ID, current time, and empty lists of alerts. I used [these](https://www.figma.com/design/UG6HV7HIsdR6B6QABOFm01/Elevator-Screens?node-id=1804-6453&t=fFiNnbTpyIavuBsH-4) designs to guide the implementation. 

https://screens-dev-green.mbtace.com/v2/screen/ELE-101

---
  - https://app.asana.com/0/0/1208494233405565


digitalcora (approved):
This approach seems fine to me — though I do wonder if we should be pre-emptively splitting the distinct "parts" we've already identified here (header, content, footer) into separate widgets on the backend, so e.g. we could have a piece of evergreen content replace only the main content area, and not have to add "hard-coded" copies of the header and footer to make it look right. Up to you, I think it would be the same amount of work to do that now or do it later.

cmaddox5:
That's a good point. I didn't think about evergreen enough when building this out. I'll go ahead and do that now so we have a finalized layout before starting the harder work later.

digitalcora (approved):

PR #2250, opened 2024-10-17
chore: Log requests to widget endpoints

cmaddox5:
We have no logs for these endpoints in any environment but Mercury uses them to create images for the e-inks. Adding some logging so we have a way to tell when something goes wrong.

### DIFF
--- a/lib/screens_web/controllers/v2/screen_controller.ex
+++ b/lib/screens_web/controllers/v2/screen_controller.ex
@@ -1,6 +1,6 @@
 defmodule ScreensWeb.V2.ScreenController do
   use ScreensWeb, :controller
-  require Logger
+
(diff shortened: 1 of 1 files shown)


cmaddox5:
The existing logs may make the `info` log I added redundant, but I think having some visibility into bad requests would be helpful. I would be curious to see if an error indicates an issue on the screen so being notified of them could help answer that. What do you think?

digitalcora:
Error logging sounds good — this is kind of an "exceptional" scenario rather than something we'd expect to hit regularly. If we want to be notified proactively about it, maybe a Sentry report instead? But I think just making it stand out a bit more in the logs is fine.

digitalcora (approved):

PR #2256, opened 2024-10-21
chore: use static per-screen-type audio config

digitalcora:
The central/motivating change here is mbta/screens-config-lib#38.

This creates a `Static` struct to encode/organize "static" parameters for each screen type — including candidate generators, refresh rate, periodic audio readout interval, and now various other audio settings.


digitalcora:
Jumped the gun on this one. It's now ready for review, for real 🤞

cmaddox5 (approved):
Love this. Really simplifies the config. Because this config is optional, I don't foresee any deployment issues due to config struct changes. Might still be a good idea to deploy to dev-green first just in case.

digitalcora:
Seems to work fine on dev-green — I deployed, then used the admin Config Editor to Validate+Save the config without changes, and it correctly updated the structure.

PR #2257, opened 2024-10-22
Support Mercury bus e-ink screens

digitalcora:
We'll soon be setting up our first Bus E-ink screen provided by Mercury. This requires a couple of changes:

* In places we previously had logic to include extra "Mercury fields" when the screen type was `gl_eink`, we should instead key on the `vendor` field.
* Since this new screen will have an audio readout button similar to GL E-ink, the Bus E-ink screen type now needs to support audio.

---
  - https://app.asana.com/0/0/1208343202037657

### DIFF
--- a/lib/screens/v2/widget_instance/alert.ex
+++ b/lib/screens/v2/widget_instance/alert.ex
@@ -271,7 +271,10 @@ defmodule Screens.V2.WidgetInstance.Alert do
     end
   end
 
-  def audio_valid_candidate?(%__MODULE__{screen: %Screen{app_id: :gl_eink_v2}}), do: true
+  def audio_valid_candidate?(%__MODULE__{screen: %Screen{app_id: app_id}})
+      when app_id in ~w[bus_eink_v2 gl_eink_v2]a,
+      do: true
+
   def audio_valid_candidate?(_instance), do: false
 
   def audio_view(_instance), do: ScreensWeb.V2.Audio.AlertView
--- a/lib/screens/v2/widget_instance/subway_status.ex
+++ b/lib/screens/v2/widget_instance/subway_status.ex
@@ -9,7 +9,7 @@ defmodule Screens.V2.WidgetInstance.SubwayStatus do
   alias Screens.Stops.Stop
   alias Screens.V2.WidgetInstance.SubwayStatus
   alias ScreensConfig.Screen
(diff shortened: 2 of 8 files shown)


lib/screens/v2/widget_instance/alert.ex
```diff
@@ -271,7 +271,10 @@ defmodule Screens.V2.WidgetInstance.Alert do
     end
   end
 
-  def audio_valid_candidate?(%__MODULE__{screen: %Screen{app_id: :gl_eink_v2}}), do: true
+  def audio_valid_candidate?(%__MODULE__{screen: %Screen{app_id: app_id}})
+      when app_id in ~w[bus_eink_v2 gl_eink_v2]a,
```
digitalcora, line 275: I'm wondering about this conditional. Are there places we use this widget on audio-supporting screens where we specifically don't want it to be included in audio readouts?

lib/screens/v2/widget_instance/subway_status.ex
```diff
@@ -135,7 +135,7 @@ defmodule Screens.V2.WidgetInstance.SubwayStatus do
     def audio_sort_key(_instance), do: [2]
 
     def audio_valid_candidate?(%{screen: %Screen{app_params: %app{}}})
-        when app in [PreFare, GlEink],
+        when app in [BusEink, GlEink, PreFare],
```
digitalcora, line 138: Same wondering here as above.

lib/screens/v2/widget_instance/alert.ex
cmaddox5: I think we first introduced audio for this widget when we started converting e-inks to Mercury. My guess is that we wanted to limit the scope of the change to just Mercury screens so we added this guard to exclude Bus Shelter. In hindsight, it seems overly cautious because bus alerts should read the same as GL. Problem is we never made a task to review the alert audio on Bus Shelters. It might be ok to just leave out the guard and ship alert audio to two screens at once. I feel confident in that but would understand leaving it to a future task.

lib/screens/v2/widget_instance/subway_status.ex
cmaddox5: This is likely done for the same reason as above. Something to think about here is whether it's not a bad experience to hear subway audio at a bus stop. That's the only thing I could see being an issue and even that is reaching.

PR #2260, opened 2024-10-24
chore: Add Recon

cmaddox5:
Memory usage seems to be higher than normal recently and we aren't sure why. Adding in a process that logs the top 20 memory usage processes every 5 minutes. Took inspiration from RTS: https://github.com/mbta/realtime_signs/blob/main/lib/engine/health.ex. Just left out network checks.


digitalcora:
Let's test this out on dev-green and see if the resulting info seems likely to give us some insights on production. Actual implementation looks okay to me.

cmaddox5:
This provides about the same info as our Phoenix Live Dashboard but gives the advantage of getting historical context. A decent preliminary query I made is `index="screens-dev-green-application" screens_process_health | bin _time span=1h | eventstats max(memory) as mem BY name, _time | table _time name mem` for reference.

digitalcora (approved):
I'm okay with trying this. I think we might end up wanting better visibility into the specific things _within_ e.g. the V3 API cache that are taking up a lot of space (like which resources?), but this would tell us whether we should look there in the first place.

PR #2266, opened 2024-10-25
chore: include contactless payments in E-ink copy

digitalcora:
---
  - https://app.asana.com/0/0/1208343202037652


cmaddox5 (approved):
Tough to get through, but looks good.

PR #2267, opened 2024-10-28
fix: Recon error workaround.

cmaddox5:
In the first PR, I removed this workaround because the author closed Cora's original [issue](https://github.com/ferd/recon/issues/95). This is definitely still an issue. I deployed the health check on Friday and saw the error occur about 60 times. Want to get this workaround in before deploying the check to prod.

### DIFF
--- a/lib/screens/health.ex
+++ b/lib/screens/health.ex
@@ -81,10 +81,20 @@ defmodule Screens.Health do
   defp process_metrics({pid, name, supervisor}) do
     metrics =
       pid
-      |> :recon.info(@process_metrics)
+      |> safe_recon_info(@process_metrics)
       |> Stream.map(fn {metric, value} -> "#{metric}=#{value}" end)
       |> Enum.intersperse(" ")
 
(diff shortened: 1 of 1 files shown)


lib/screens/health.ex
```diff
@@ -81,10 +81,20 @@ defmodule Screens.Health do
...
-      |> :recon.info(@process_metrics)
+      |> safe_recon_info(@process_metrics)
       |> Stream.map(fn {metric, value} -> "#{metric}=#{value}" end)
       |> Enum.intersperse(" ")
 
     {name, supervisor, metrics}
   end
+
+  # work around https://github.com/ferd/recon/issues/95
+  # The issue is closed but is still very much an issue.
```
digitalcora, line 92: Rather than putting an assertion like this in a comment where it can drift apart from the actual status of the issue, I'd prefer updating the issue itself, which I've just done. (someday, presumably, the issue will be closed _and_ fixed)

digitalcora (approved):

PR #2268, opened 2024-10-28
revert: React/TS Upgrades

cmaddox5:
The last prod deploy broke GDS E-Ink screens. These changes are the only ones that could possibly cause issues on these screens. For the sake of getting `main` in a deployable state, going to pull them out while we investigate.


digitalcora (approved):
🥲

PR #2280, opened 2024-10-31
chore: remove Solari screen type and "v1" code

digitalcora:
* Remove the `solari` screen type; we have no remaining screens of this type in the wild.

* Remove the "v1" screen page/API endpoints; there are now no remaining v1 screen types.

* Remove a substantial amount of code that was only used by v1 screens (likely not all of it, but there are few reliable tools available for detecting unused "public" code in Elixir).

* Remove the `solari_large_v2` screen type; this was added a long time ago in anticipation of the "large" Solari layout being ported to the v2 framework, but no real functionality was ever added to it, and the only screens of this type in the wild are now permanently offline and waiting to be decommissioned.

---
  - https://app.asana.com/0/0/1208651022136050


lib/screens_web/controllers/screen_controller.ex
```diff
@@ -1,107 +1,7 @@
...
-      nil ->
-        render_not_found(conn)
-    end
-  end
-
-  def index(conn, _params) do
-    render_not_found(conn)
-  end
 
   def show_image(conn, %{"filename" => filename}) do
```
cmaddox5, line 4: Do we know what uses this endpoint? Curious if we can just get rid of this file altogether.

cmaddox5 (approved):
Love it. Thanks for taking the time to do this.

lib/screens_web/controllers/screen_controller.ex
digitalcora: It is definitely being used somehow in the wild, based on logs (search for `GET /image/`), so I'm inclined not to mess with it for now.

PR #2283, opened 2024-11-01
refactor: remove unused parts of Signs UI config

digitalcora:
PA/ESS sign modes configured in Signs UI were only used by "v1" screen logic, so we no longer need to parse or store this data.


panentheos (approved):
Looks good to me.

PR #2285, opened 2024-11-04
fix: Incorrect types for TimeRepresentation

cmaddox5:
The `type` value we send to the client is always a lowercase atom converted to a string. For some reason, the client code was looking for an uppercase version of each value. This was causing the `div` that holds the time to be empty.

Before:
Screenshot 2024-11-04 at 8 17 49 AM

After:
Screenshot 2024-11-04 at 8 33 53 AM


panentheos (approved):

digitalcora:
I think this was my mistake — in #2280, as part of "inlining" a v1 E-ink component that was reused by Pre-Fares (😬), I removed code in these conditionals that was calling `toUpperCase` on the type field before comparing it. Since the type was defined as uppercase (which this PR fixes), I assumed that was an accurate reflection of what the server sent. In reality it was probably that the v1 code used uppercase and v2 used lowercase, and to make the component work with both screen types, this `toUpperCase` was thrown in.

PR #2288, opened 2024-11-05
feat: Elevator Closures list

cmaddox5:
This PR adds a simple list view of elevator closures throughout the system. It works very similarly to Pre-Fares except all dynamic paging is done on the client. The way I handled paging was to generate the pages before React drew anything on the screen and then switch to rendering these pages once it is done. This appears seamless and looks like it loads like any other screen. After all pages have been rendered, it executes a callback that tells React it is ready to replace it's data in the next refresh. The way I did this in code feels a little messy so please do suggest improvements.

### DIFF
--- a/assets/css/v2/elevator/elevator_closures.scss
+++ b/assets/css/v2/elevator/elevator_closures.scss
@@ -8,8 +8,9 @@
 
   .in-station-summary {
     display: flex;
+    gap: 82px;
     justify-content: space-between;
-    padding: 24px 58px;
+    padding: 24px 48px;
     font-size: 48px;
     font-weight: 400;
     line-height: 64px;
@@ -82,7 +83,15 @@
             .route-pill {
               width: 132px;
               height: 68.13px;
+
+              & > * {
+                height: 100%;
+              }
             }
+          }
+
+          &__elevator-name {
+            line-height: 64px;
           }
 
           &__elevator-name.list-item {
@@ -95,9 +104,16 @@
     }
 
     .paging-info-container {
+      position: absolute;
+      bottom: 0;
       display: flex;
       justify-content: space-between;
-      margin: 0 48px;
+      width: 100%;
+      height: 72px;
+
+      & > * {
+        margin: 0 48px 12px;
+      }
 
       .paging-indicators {
         display: flex;
@@ -111,7 +127,6 @@
     .alert-row__elevator-name {
       font-size: 48px;
       font-weight: 400;
-      line-height: 64px;
       color: $cool-black-30;
     }
   }
--- a/assets/css/v2/elevator/elevator_closures.scss
+++ b/assets/css/v2/elevator/elevator_closures.scss
@@ -82,6 +82,18 @@
       }
     }
 
+    .paging-info-container {
+      display: flex;
+      justify-content: space-between;
+
+      .paging-indicators {
+        display: flex;
+        align-items: center;
+        gap: 27px;
+        margin-right: 66px;
+      }
+    }
+
     .paging-info-container,
     .alert-row__elevator-name {
       font-weight: 400;
--- a/assets/css/v2/elevator/elevator_closures.scss
+++ b/assets/css/v2/elevator/elevator_closures.scss
@@ -16,13 +16,19 @@
     color: $cool-black-30;
   }
 
-  hr {
-    width: 100%;
+  hr.thick {
     min-height: 24px;
-    margin-top: 0;
-    margin-bottom: 0;
+    margin: 0;
     background-color: $cool-black-15;
     border: none;
+  }
+
+  hr.thin {
+    min-height: 2px;
+    margin: 24px 0 0;
+    background-color: $true-grey-45;
+    border: none;
+    opacity: 0.5;
   }
 
   .outside-alert-list {
@@ -31,24 +37,20 @@
     font-family: Inter;
     background-color: $warm-neutral-90;
 
-    hr {
-      min-height: 2px;
-      margin-top: 24px;
-      background-color: $true-grey-45;
-      border: none;
-      opacity: 0.5;
-    }
-
-    .header {
-      display: flex;
-      max-height: 432px;
+    .header-container {
       margin: 48px;
-      font-size: 112px;
-      font-weight: 700;
-      line-height: 112px;
+      margin-bottom: 0;
 
-      &__title {
-        word-spacing: 9999px;
+      .header {
+        display: flex;
+        max-height: 432px;
+        font-size: 112px;
+        font-weight: 700;
+        line-height: 112px;
+
+        &__title {
+          word-spacing: 9999px;
+        }
       }
     }
 
--- a/assets/css/v2/elevator/elevator_closures.scss
+++ b/assets/css/v2/elevator/elevator_closures.scss
@@ -28,7 +28,6 @@
   .outside-alert-list {
     position: relative;
     height: 100%;
-    padding: 48px;
     font-family: Inter;
     background-color: $warm-neutral-90;
 
@@ -43,7 +42,7 @@
     .header {
       display: flex;
       max-height: 432px;
-      margin-bottom: 48px;
+      margin: 48px;
       font-size: 112px;
       font-weight: 700;
       line-height: 112px;
@@ -54,11 +53,16 @@
     }
 
     .alert-list-container {
-      height: 904px;
+      overflow: hidden;
 
       .alert-list {
+        display: flex;
+        flex-flow: column wrap;
+        height: 904px;
+        transform: translateX(calc(-100% * var(--alert-list-offset)));
+
         .alert-row {
-          margin-top: 24px;
+          margin: 24px 48px 0;
 
           &__station-name {
             font-size: 62px;
@@ -85,6 +89,7 @@
     .paging-info-container {
       display: flex;
       justify-content: space-between;
+      margin: 0 48px;
 
       .paging-indicators {
         display: flex;
--- a/assets/css/v2/elevator/elevator_closures.scss
+++ b/assets/css/v2/elevator/elevator_closures.scss
@@ -26,27 +26,27 @@
   }
 
   .outside-alert-list {
-    height: 100%;
-    background-color: $warm-neutral-90;
-    padding: 48px;
     position: relative;
+    height: 100%;
+    padding: 48px;
     font-family: Inter;
+    background-color: $warm-neutral-90;
 
     hr {
-      background-color: $true-grey-45;
-      opacity: 50%;
       min-height: 2px;
       margin-top: 24px;
+      background-color: $true-grey-45;
       border: none;
+      opacity: 0.5;
     }
 
     .header {
-      max-height: 432px;
       display: flex;
+      max-height: 432px;
+      margin-bottom: 48px;
       font-size: 112px;
       font-weight: 700;
       line-height: 112px;
-      margin-bottom: 48px;
 
       &__title {
         word-spacing: 9999px;
@@ -61,16 +61,16 @@
           margin-top: 24px;
 
           &__station-name {
-            font-weight: 600;
             font-size: 62px;
+            font-weight: 600;
             line-height: 80px;
             color: $cool-black-15;
           }
 
           &__name-and-pills {
             display: flex;
-            align-items: center;
             gap: 24px;
+            align-items: center;
             margin-bottom: 14px;
 
             .route-pill {
@@ -88,16 +88,16 @@
 
       .paging-indicators {
         display: flex;
-        align-items: center;
         gap: 27px;
+        align-items: center;
         margin-right: 66px;
       }
     }
 
     .paging-info-container,
     .alert-row__elevator-name {
-      font-weight: 400;
       font-size: 48px;
+      font-weight: 400;
       line-height: 64px;
       color: $cool-black-30;
     }
--- a/assets/css/v2/elevator/elevator_closures.scss
+++ b/assets/css/v2/elevator/elevator_closures.scss
@@ -63,6 +63,17 @@
             height: 2px;
             margin-bottom: 24px;
           }
+
+          &__name-and-pills {
+            display: flex;
+            align-items: center;
+            gap: 24px;
+
+            .route-pill {
+              width: 132px;
+              height: 68.13px;
+            }
+          }
         }
       }
     }
--- a/assets/css/v2/elevator/elevator_closures.scss
+++ b/assets/css/v2/elevator/elevator_closures.scss
@@ -67,8 +67,8 @@
 
         .closure-row {
           max-width: 1080px;
-          margin-top: 24px;
           padding: 0 48px;
+          margin-top: 24px;
 
           &__station-name {
             font-size: 62px;
@@ -87,8 +87,17 @@
               height: 68.13px;
               margin-right: 24px;
 
-              & > * {
+              &__text {
+                line-height: 68.13px;
+              }
+
+              &__icon {
                 height: 100%;
+              }
+
+              &__icon-image {
+                height: 100%;
+                margin-top: 0;
               }
             }
           }
@@ -113,9 +122,11 @@
       justify-content: space-between;
       width: 100%;
       height: 72px;
+      margin-bottom: 20px;
 
-      & > * {
-        margin: 0 48px 12px;
+      .more-elevators-text {
+        align-self: center;
+        padding: 0 48px 20px;
       }
 
       .paging-indicators {
--- a/assets/css/v2/elevator/elevator_closures.scss
+++ b/assets/css/v2/elevator/elevator_closures.scss
@@ -18,10 +18,11 @@
 
   hr {
     width: 100%;
-    height: 24px;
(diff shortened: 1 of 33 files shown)


digitalcora:
The client-side resizing approach here seems _much_ more complicated than it needs to be — I thought we had been on the same page about how to implement this, so either that's my bad for not being clear enough, or maybe you ran into some issues?

Basically my idea was that Flexbox can almost entirely do the work for us. Throw this into a sandbox:

```html
<div class="container">
  <div class="closures">
    <div class="closure">
      Repellendus nisi ut adipisci aperiam enim. Itaque quam recusandae voluptatem eum asperiores mollitia.
    </div>
    <div class="closure">
      Ipsum ea et quidem. Nam quidem aut voluptas. Sed qui at quis doloribus. Facere neque voluptatum dolor temporibus a aliquam. Soluta aut eaque ipsam doloremque. Possimus accusantium occaecati reiciendis expedita quia.
    </div>
    <div class="closure">
      Itaque recusandae adipisci est harum possimus exercitationem. Odit voluptas consequatur tempore deserunt et officia. Quas tenetur quia veniam dolores vero.
    </div>
    <div class="closure">
      Distinctio distinctio voluptates quo. Recusandae non est natus ex velit dolores beatae. Est nisi sit nobis corrupti tempora. Debitis et autem voluptatem reprehenderit cum. Occaecati tempore aut praesentium mollitia omnis perferendis nam.
    </div>
    <div class="closure">
      In dicta ut explicabo aliquam distinctio. Et et neque et. Est architecto quaerat debitis.
    </div>
  </div>
</div>
```

```css
.container {
  width: 300px;
  border: 1px dotted black;
  overflow: hidden;
}

.closures {
  height: 200px;
  display: flex;
  flex-direction: column;
  flex-wrap: wrap;
}

.closure {
  border: 1px solid gray;
}
```

```js
const closures = document.getElementsByClassName("closure");

for (const closure of closures) {
  console.log(closure.offsetLeft);
}
```

And you'll get this (I commented out `overflow: hidden` so we can see what's "behind the overflow"):

Screenshot 2024-11-05 at 3 49 34 PM

Without any script, this already lays out how we want: when a closure is too tall to fit on the current page, it flexes to the next page (i.e. column). The script is demonstrating that by collecting the unique values of `offsetLeft`, we can tell how many pages there are (we'd do this inside a `useLayoutEffect`) — 9, 309, and 609 is three unique values, so we have three pages. Then the only missing piece is how to display a specific "page" other than the first one, which we already have a pretty clean example of in `LaterDepartures` that uses a CSS variable.

Let me know if that makes sense, or if you already tried this and it didn't work.

lib/screens/v2/candidate_generator/elevator/closures.ex
```diff
@@ -0,0 +1,160 @@
...
+  def elevator_status_instances(
+        %Screen{
+          app_params: %Elevator{
+            elevator_id: elevator_id
+          }
+        },
+        now \\ DateTime.utc_now()
+      ) do
+    with {:ok, %Stop{id: stop_id}} <- @facility.fetch_stop_for_facility(elevator_id),
+         {:ok, location_context} <- @stop.fetch_location_context(Elevator, stop_id, now),
```
digitalcora, line 31: It seems like the only part of the LocationContext we ultimately use here is the `home_stop`, which is simply the exact `stop_id` we're passing into this function. Can we save some data fetching by omitting this?

lib/screens/v2/candidate_generator/elevator/closures.ex
```diff
@@ -0,0 +1,160 @@
...
+    end)
+    |> Enum.into(%{})
+  end
+
+  defp get_parent_station_ids_from_entities(alerts) do
+    alerts
+    |> Enum.flat_map(fn %Alert{informed_entities: informed_entities} ->
+      informed_entities
+      |> Enum.map(fn %{stop: stop_id} -> stop_id end)
+      |> Enum.filter(&String.starts_with?(&1, "place-"))
```
digitalcora, line 78: We could use `InformedEntity.parent_station?` here, as below.

lib/screens/v2/candidate_generator/elevator/closures.ex
```diff
@@ -0,0 +1,160 @@
...
+
+  defp routes_to_labels(routes) do
+    routes
+    |> Enum.map(fn
+      %Route{type: :subway, id: id} -> id |> String.downcase() |> String.to_atom()
+      %Route{type: :light_rail, id: "Green-" <> _} -> :green
+      %Route{type: :light_rail, id: "Mattapan" <> _} -> :mattapan
+      %Route{type: :bus, short_name: "SL" <> _} -> :silver
+      %Route{type: :rail} -> :cr
+      %Route{type: type} -> type
```
digitalcora, line 98: This appears very similar to `Route.icon/1`. Can we replace it with that?

lib/screens/v2/candidate_generator/elevator/closures.ex
```diff
@@ -0,0 +1,160 @@
...
+    Enum.split_with(alerts, fn %Alert{informed_entities: informed_entities} ->
+      location_context.home_stop in Enum.map(informed_entities, & &1.stop)
+    end)
+  end
+
+  defp get_informed_facility(entities) do
+    entities
+    |> Enum.find_value(fn
+      %{facility: facility} -> facility
+      _ -> false
```
digitalcora, line 113: * The place where this is used in `alert_to_elevator_closure` doesn't seem to account for the possibility that it could be `nil`. Ultimately this could happen if there were an alert that impacted the `USING_WHEELCHAIR` activity but did not inform any facilities. We should explicitly handle that possibility in some way, whether it's to ignore/filter out that case, or crash on it.
* What happens if an alert affects multiple facilities? Again, maybe we could either explicitly ignore or crash.

lib/screens/v2/screen_data/parameters.ex
```diff
@@ -38,7 +38,7 @@ defmodule Screens.V2.ScreenData.Parameters do
     },
     elevator_v2: %Static{
       candidate_generator: CandidateGenerator.Elevator,
-      refresh_rate: 30
+      refresh_rate: 8
```
digitalcora, line 41: Huh, is our "persistent" component logic actually tied to the data refresh rate? That seems less than ideal, but maybe just a refactoring opportunity to keep in mind for the future. (Suppose your screen consists almost entirely of a paging component, and you expect it to normally have at least two pages: it can only "really" refresh once all the paging is done, so you're doing at least twice as many data refreshes as you need to.)

lib/screens/v2/widget_instance/elevator_closures.ex
```diff
@@ -1,22 +1,60 @@
...
+    alias Screens.V2.WidgetInstance.ElevatorClosures.Alert
+
+    @derive Jason.Encoder
+
+    defstruct ~w[id name routes alerts]a
+
+    @type t :: %__MODULE__{
+            id: Stop.id(),
+            name: String.t(),
+            routes: list(String.t()),
```
digitalcora, line 26: Should this be `list(Route.id())`? Might also be nice to distinguish between "routes" (the whole entity) and "route IDs", by naming this field `route_ids`.

assets/css/v2/elevator/elevator_closures.scss
```diff
@@ -18,22 +18,88 @@
...
     background-color: $cool-black-15;
+    border: none;
   }
 
   .outside-alert-list {
     height: 100%;
     background-color: $warm-neutral-90;
+    padding: 48px;
+    position: relative;
+    font-family: Inter;
```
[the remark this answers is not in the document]
cmaddox5: I think I just got copy-and-paste happy in Figma. Can remove.

assets/src/components/v2/elevator/elevator_closures.tsx
```diff
@@ -1,9 +1,65 @@
...
+} from "react";
+import cx from "classnames";
 import NormalService from "Images/svgr_bundled/normal-service.svg";
 import AccessibilityAlert from "Images/svgr_bundled/accessibility-alert.svg";
+import PagingDotUnselected from "Images/svgr_bundled/paging_dot_unselected.svg";
+import PagingDotSelected from "Images/svgr_bundled/paging_dot_selected.svg";
+import makePersistent, { WrappedComponentProps } from "../persistent_wrapper";
+import RoutePill, { routePillKey, type Pill } from "../departures/route_pill";
+
+type StationWithAlert = {
```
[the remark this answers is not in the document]
cmaddox5: Something I've struggled with during this task. Closures does fit better so I just chose incorrectly. Can adjust.

lib/screens/v2/candidate_generator/elevator/closures.ex
cmaddox5: Not sure why I insist on reinventing this wheel so often. Thanks.

lib/screens/v2/screen_data/parameters.ex
cmaddox5: It does currently, yes, and I just felt like it would be too far out of scope to adjust it here. Given that data refreshes are built into the framework, we may need to get around it by preventing it from refreshing at this level and adding data fetches to the widget component itself. Then where we call `onFinish()` currently could be replaced by that fetch.

lib/screens/v2/widget_instance/elevator_closures.ex
digitalcora: Ah, great, then the type can also be `Route.icon()`. (though I think we tend to call this a "route _pill_" in the frontend... inconsistencies)

assets/src/components/v2/elevator/elevator_closures.tsx
```diff
@@ -19,44 +75,138 @@ const InStationSummary = ({ alerts }: InStationSummaryProps) => {
...
+  useEffect(() => {
+    if (pageIndex === numPages - 1) {
+      onFinish();
+    }
+  }, [pageIndex]);
+
+  useLayoutEffect(() => {
+    const alertRows = Array.from(document.getElementsByClassName("alert-row"));
+    const screenWidth = 1080;
+    const totalXMargins = 48;
```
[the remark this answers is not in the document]
cmaddox5: Makes sense. Got the result I wanted so hope I didn't overcomplicate the transformation.

lib/screens/v2/candidate_generator/elevator/closures.ex
cmaddox5: No harm in handling them, but I think in both cases that would be a data issue. With that in mind, I think I'll just filter them out.

assets/src/components/v2/elevator/elevator_closures.tsx
```diff
@@ -1,9 +1,65 @@
...
+  useLayoutEffect,
+  useMemo,
+  useState,
+} from "react";
+import cx from "classnames";
 import NormalService from "Images/svgr_bundled/normal-service.svg";
 import AccessibilityAlert from "Images/svgr_bundled/accessibility-alert.svg";
+import PagingDotUnselected from "Images/svgr_bundled/paging_dot_unselected.svg";
+import PagingDotSelected from "Images/svgr_bundled/paging_dot_selected.svg";
+import makePersistent, { WrappedComponentProps } from "../persistent_wrapper";
```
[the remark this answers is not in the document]
cmaddox5: I tried to make that component work but it was created with the idea that the server would send up pages of widget data to render. With the new dynamic paging, we don't have a list of pages until we finish rendering and the pages we end up with don't have any widget data in them anyway. I do wonder if we can abstract out the hook portions of each though. At least get rid of some of the duplicate code. Will explore that.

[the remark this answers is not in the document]
cmaddox5: Yeah I think that worked out nicely. Let me know what you think.

assets/src/components/v2/elevator/elevator_closures.tsx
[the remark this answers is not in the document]
digitalcora: > a mapping of page number to item count `(0 => 2, 1 => 3, 2 => 1)`

...I think I just invented an array 😆 `[2, 3, 1]`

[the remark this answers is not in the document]
cmaddox5: 😂 We definitely both fell for that. I missed it too. So yes, I did overcomplicate the transformation.

lib/screens/v2/candidate_generator/elevator/closures.ex
```diff
@@ -61,8 +46,15 @@ defmodule Screens.V2.CandidateGenerator.Elevator.Closures do
     end
   end
 
-  defp relevant_alerts(alerts) do
-    Enum.filter(alerts, &(&1.effect == :elevator_closure))
+  defp relevant_closure?(closure) do
+    relevant_effect?(closure) and informs_one_facility?(closure)
+  end
+
+  defp relevant_effect?(closure), do: closure.effect == :elevator_closure
```
digitalcora, line 53: The "closure" renaming might have gone a bit too far here... I think when we're dealing with anything that's literally a `Screens.Alerts.Alert`, we should call that an `alert`. I thought it made sense to specifically call the entity that has an `elevator_name` and `header_text` and etc. a "closure" because the structure is very distinct from an alert and doesn't only include alert data (it also incorporates facility data).

test/screens/v2/candidate_generator/elevator/closures_test.exs
```diff
@@ -6,25 +6,18 @@ defmodule Screens.V2.CandidateGenerator.Elevator.ClosuresTest do
...
   alias Screens.V2.CandidateGenerator.Elevator.Closures, as: ElevatorClosures
   alias ScreensConfig.Screen
   alias ScreensConfig.V2.Elevator
 
-  describe "elevator_status_instances/5" do
-    test "Only returns alerts with effect of :elevator_closure" do
-      now = ~U[2024-10-01T05:00:00Z]
-
+  describe "elevator_status_instances/1" do
+    test "Only returns closures with effect of :elevator_closure" do
```
digitalcora, line 16: These test descriptions are another instance of "closures" when we really mean "alerts".

assets/src/components/v2/elevator/elevator_closures.tsx
```diff
@@ -19,44 +71,134 @@ const InStationSummary = ({ alerts }: InStationSummaryProps) => {
...
+  }, 0);
+
+  useLayoutEffect(() => {
+    if (!ref.current) return;
+
+    const offsets = Array.from(ref.current.children).map((closure) => {
+      return (closure as HTMLDivElement).offsetLeft;
+    });
+
+    const rowPageIndexes: number[] = [];
```
digitalcora, line 114: I guess I shouldn't be surprised that `number[]` is type-compatible with `{[key: number]: number}`, but it is a little bit confusing on first read. I think there's some opportunity for cleanup here, starting by replacing L92 with something like

```ts
const [rowCounts, setRowCounts] = useState<number[]>([]);
```

lib/screens/v2/location_context.ex
```diff
@@ -9,7 +9,7 @@ defmodule Screens.LocationContext do
   alias Screens.Stops.Stop
   alias Screens.Util
 
-  alias ScreensConfig.V2.{BusEink, BusShelter, Dup, GlEink, PreFare}
+  alias ScreensConfig.V2.{BusEink, BusShelter, Dup, Elevator, GlEink, PreFare}
```
digitalcora, line 12: The changes to this module aren't necessary anymore now that we aren't using it.

digitalcora (commented):
Looking pretty good! A few more minor comments.

lib/screens/v2/candidate_generator/elevator/closures.ex
```diff
@@ -0,0 +1,143 @@
...
+
+  defp alert_to_elevator_closure(%Alert{
+         id: id,
+         informed_entities: entities,
+         description: description,
+         header: header
+       }) do
+    facility = Enum.find_value(entities, fn %{facility: facility} -> facility end)
+
+    %{
```
digitalcora, line 108: Since we've defined `ElevatorClosures.Closure` we could use it here instead of a plain map.

lib/screens/v2/candidate_generator/elevator/closures.ex
```diff
@@ -0,0 +1,143 @@
...
+    |> Enum.group_by(&get_parent_station_id_from_informed_entities(&1.informed_entities))
+    |> Enum.map(fn {parent_station_id, closures} ->
+      closures_at_station = Enum.map(closures, &alert_to_elevator_closure/1)
+
+      route_pills =
+        station_id_to_routes
+        |> Map.fetch!(parent_station_id)
+        |> Enum.map(&RoutePill.serialize_icon/1)
+
+      %{
```
digitalcora, line 128: As above, with `ElevatorClosures.Station`.

digitalcora (approved):
A couple (really one) small comments, but I say this looks ready to go!

cmaddox5:
@digitalcora Just want to point out that I also tweaked the CSS so it uses compatible features. As of last night, the office screen now renders how we expect it to:
IMG_7835

PR #2289, opened 2024-11-05
refactor: extract functions from `Stop` module

digitalcora:
`Stop` has become a bit of a junk drawer, containing several entirely distinct responsibilities: fetching stop data from the V3 API, looking up data from a hard-coded list of subway stations, and building "location context" data (each with its own set of non-overlapping support functions). This splits those responsibilities out into appropriate modules.


cmaddox5 (approved):
Yeah this makes a lot more sense. I won't be scared to look in `Stop.ex` anymore.

PR #2290, opened 2024-11-07
chore: Add custom browserslist

cmaddox5:
Now that we have an idea what browsers our vendors use, we can add a browserslist config to help catch incompatibilities. I added this to ESLint and StyleLint so CI can help us find issues.

---
  - https://app.asana.com/0/0/1207719326862934


assets/.browserslistrc
```diff
@@ -0,0 +1,10 @@
+# DUP
+ChromeAndroid 61
+# Sectional
+Firefox 126
```
digitalcora, line 4: This should be `Firefox 90` since that seems to be what the Solari units have. Maybe we should be explicit in this file about which vendors' hardware we're referring to, since there is not a 1-to-1 mapping of "screen app" to hardware specs or vendors. For example we could have:

```
# Sectional (Solari)
Firefox 90
# Sectional (GDS)
Firefox 126
```

Then someday when we retire the Solaris, we know we should delete that line, but not the other one.

assets/.browserslistrc
```diff
@@ -0,0 +1,10 @@
...
+# Pre-Fare
+Chrome 126
+# Bus Shelter
+Firefox 126
+# Bus E-Ink
+ChromeAndroid 83
```
digitalcora, line 10: From our digging in the previous [task](https://app.asana.com/0/1185117109217413/1207719326862933) I thought this was going to be `Chrome 45`, based on GDS's version of qt-webkit.

assets/stylelint.config.mjs
```diff
@@ -38,5 +39,9 @@ export default {
     // Conflicts with Prettier, which breaks long lines involving operators by
     // putting operators at the ends of lines.
     "scss/operator-no-newline-after": null,
+    "plugin/no-unsupported-browser-features": [
+      true,
+      { ignore: ["intrinsic-width"] },
```
digitalcora, line 44: If we ignore something we should probably add a note explaining why. I assume it's because the only places we use this feature are on screen types that actually do support intrinsic-width?

digitalcora:
I don't see a change to our Webpack config file here; we have this in there:

```
["@babel/preset-env", { targets: "> 0.25%" }]
```

I assume we can now at least remove the `targets` configuration here and let browserslist take care of it — does `present-env` read that by default?

assets/.browserslistrc
cmaddox5: I must have been looking at an outdated list I put together. Sorry about that.

assets/stylelint.config.mjs
cmaddox5: Before adding the correct Bus E-Ink version, this was the only error we needed to deal with and the browser it is an issue for doesn't reach the particular CSS in question. After the update, there's a 919 errors so adding global ignores isn't a good idea. I'm curious how best to handle this. At a webpack level, that seems as easy as separating out the apps into their own entries and specifically calling out the browser each entry will use. I'm not sure about the linters though...

digitalcora: Hm... is this warning us about using features that Babel will ultimately compile down to something that does actually work? Or are those warnings truly about features that can't be supported automatically?

digitalcora: Follow-up question for investigation, does the Webpack build itself throw an error or warning when we use a feature Babel can't transpile to the target browsers, or does it silently allow such features?

cmaddox5: Hard to tell at the moment if it is failing silently or if it just so happens that our code can transpile for all configured browsers. I tried dropping the target down to `Chrome 4` and still see no errors.

cmaddox5: Yeah that sounds good. Will remove those and create a task.

cmaddox5: I'm curious if we should do the same with ESLint. The lint errors showing up right now are either false positives or handled by the polyfills we are already importing that ESLint has no knowledge of.

assets/.browserslistrc
```diff
@@ -0,0 +1,12 @@
...
+ChromeAndroid 61
+# Sectional (Solari)
+Firefox 90
+# Sectional (GDS)
+Firefox 126
+# Pre-Fare
+Chrome 126
+# Bus Shelter
+Firefox 126
+# Bus E-Ink
```
digitalcora, line 11: Maybe this should read "Bus E-ink (GDS)" — we're replacing bus e-ink with Mercury displays eventually, and then this would not be relevant to them.

assets/stylelint.config.mjs
digitalcora: Interesting. I would have expected the opposite based on [this thread](https://github.com/amilajack/eslint-plugin-compat/issues/602), which warns that the ESLint plugin is _too_ lenient in assuming which polyfills are available (suggesting there is some interaction with Babel).

cmaddox5: I may have just configured the plugin incorrectly (documentation is lackluster unfortunately). I made an adjustment and the errors are gone. I found in the [source code](https://github.com/amilajack/eslint-plugin-compat/blob/a0c853e260656cb60139f90e3ffc89d025f80969/src/rules/compat.ts#L91) for the plugin that it assumes polyfills are done in babel based on the presence of a babel config file. We didn't have one so I made it. Hopefully this doesn't mean I just silenced the errors by accident.

assets/babel.config.json
```diff
@@ -0,0 +1,16 @@
+{
+  "presets": [
+    // When no targets are specified: Babel will assume you are targeting the oldest browsers possible.
```
digitalcora, line 3: This is invalid JSON, though I guess Babel allows it? "Oldest browsers possible" def isn't accurate with how this is set up, so I think we should delete.

digitalcora (approved):
One small comment. I do like how it's possible to pull out the Babel config like this, simplifies the main Webpack config structure.

PR #2291, opened 2024-11-07
Config for push button audio for bus shelters

PaulJKim:
**Asana task**: [Add config option for push-button audio on Bus Shelters](https://app.asana.com/0/1185117109217413/1208437824376833/f)

Description
Uses the `interval_enabled` audio config flag to determine whether to disable audio readouts by returning `0` as the `audio_minutes` or use the existing logic to determine the readout.

- [x] Tests added?


lib/screens/v2/screen_data/parameters.ex
```diff
@@ -66,9 +66,24 @@ defmodule Screens.V2.ScreenData.Parameters do
...
+  def audio_interval_minutes(screen, static_params \\ @static_params)
+
+  def audio_interval_minutes(
+        %Screen{
+          app_id: app_id,
+          app_params: %{audio: %ScreensConfig.V2.Audio{interval_enabled: interval_enabled}}
+        },
+        static_params
+      ) do
+    if interval_enabled, do: fetch_interval_minutes(app_id, static_params), else: 0
```
digitalcora, line 79: By the existing typespec this function is allowed to return `nil`, so let's return `nil` in this case as well rather than using `0` to be clear that the intent is to have no interval. (I mentioned using 0 earlier since the client-side code does also disable readouts when the interval is 0, but I see I misremembered how we handled this on the server.)

test/screens/v2/screen_data/parameters_test.exs
```diff
@@ -64,12 +65,19 @@ defmodule Screens.V2.ScreenData.ParametersTest do
...
     end
 
     test "is the configured interval for a screen type with periodic audio" do
       static_params = build_params(periodic_audio: build_periodic_audio(interval_minutes: 7))
-      assert Parameters.audio_interval_minutes(@app_id, static_params) == 7
+      screen = build_screen()
+      assert Parameters.audio_interval_minutes(screen, static_params) == 7
+    end
+
+    test "is 0 for a screen type with audio interval disabled" do
```
digitalcora, line 78: I think this description should be "for a _screen_ with audio interval disabled", since it's the individual screens that have this flag, not the whole screen type.

test/screens/v2/screen_data/parameters_test.exs
```diff
@@ -32,7 +30,10 @@ defmodule Screens.V2.ScreenData.ParametersTest do
             alerts: %V2.Alerts{stop_id: "1"},
             departures: %V2.Departures{sections: []},
             footer: %V2.Footer{},
-            header: %V2.Header.CurrentStopId{stop_id: "1"}
+            header: %V2.Header.CurrentStopId{stop_id: "1"},
+            audio: %ScreensConfig.V2.Audio{
```
digitalcora, line 34: We have `V2` aliased here, so let's use it as `V2.Audio`.

test/screens/v2/screen_data/parameters_test.exs
PaulJKim: good point, agreed

digitalcora (approved):
🔉 🕙 ✅

digitalcora:
Oops, I approved this and then immediately remembered we also need a change to the admin UI to actually allow admins to set this field (`interval_enabled`) for bus shelter screens — otherwise the only way to change it is the "raw" config editor, which is much less friendly. Can you take care of that @PaulJKim?

PaulJKim:
@digitalcora Oh, right. Yep I can take care of that in a followup PR

PR #2292, opened 2024-11-08
Add admin field for audio interval enabled for bus shelters

PaulJKim:
**Asana task**: [[Bug] PA/ESS message Screenplay: Minute and AM/PM fields unexpectedly update in Chrome browser](https://app.asana.com/0/1185117109217413/1208573903710847/f)

Description
A follow-up to #2291

### DIFF
--- a/assets/src/components/admin/admin_tables.tsx
+++ b/assets/src/components/admin/admin_tables.tsx
@@ -258,6 +258,19 @@ const BusShelterV2ScreensTable = (): JSX.Element => {
       disableFilters: true,
       FormCell: FormTextCell,
     },
+    {
+      Header: "Audio Interval Enabled",
+      accessor: (row) => row.app_params.audio.interval_enabled,
+      mutator: (row, value) => {
+        console.log(value);
+        const newRow = structuredClone(row);
+        newRow.app_params.audio.interval_enabled = value;
+        return newRow;
+      },
+      Cell: EditableCheckbox,
+      disableFilters: true,
+      FormCell: FormBoolean,
+    },
     departuresColumn,
     footerColumn,
     alertsColumn,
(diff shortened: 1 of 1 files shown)


assets/src/components/admin/admin_tables.tsx
```diff
@@ -258,6 +258,19 @@ const BusShelterV2ScreensTable = (): JSX.Element => {
       disableFilters: true,
       FormCell: FormTextCell,
     },
+    {
+      Header: "Audio Interval Enabled",
+      accessor: (row) => row.app_params.audio.interval_enabled,
+      mutator: (row, value) => {
+        console.log(value);
```
digitalcora, line 265: Stray `console.log` ✂️

assets/src/components/admin/admin_tables.tsx
```diff
@@ -258,6 +258,19 @@ const BusShelterV2ScreensTable = (): JSX.Element => {
       disableFilters: true,
       FormCell: FormTextCell,
     },
+    {
+      Header: "Audio Interval Enabled",
```
digitalcora, line 262: Maybe this could just be `Audio Interval?` — the only content of the data cells is the checkbox, so it's best if we can have the column take up less space.

digitalcora (approved):
Looks good aside from a couple small things!

assets/src/components/admin/admin_tables.tsx
PaulJKim: sounds good to me. Did you mean to include the `?`?

digitalcora: Yep, since it's answering the "question" of whether the audio interval is enabled — like the Elixir convention to end booleans with a `?`. Probably not that important though.

PaulJKim: Cool, that's what I figured. Makes sense to me

PR #2293, opened 2024-11-08
refactor: fetch parents/children with stops

digitalcora:
Some stop functions were already using this relationship by effectively hand-rolling the relevant bits of the parsing. We can simplify by fully parsing the parent/child relationships in the main stop `Parser`, and having the special-purpose `Stop` functions delegate to a main `fetch` function as we do with other resources.

Since we may be "starting from" either a parent or a child, we need a way to encode the complete data without recursing infinitely through this bidirectional link. For this purpose we introduce `:unloaded` as a special value indicating a relationship was not populated, similar to `Ecto.Association.NotLoaded`. This allows ending the recursion once we know we have all the data. It also allows the relationships to not be fetched in the first place, for functions that don't need this data or in contexts where stops are being parsed as a relationship on another resource.

### DIFF
--- a/lib/screens/predictions/parser.ex
+++ b/lib/screens/predictions/parser.ex
@@ -31,7 +31,7 @@ defmodule Screens.Predictions.Parser do
         included
       ) do
     trip = included |> Map.fetch!({trip_id, "trip"}) |> Trips.Parser.parse_trip(included)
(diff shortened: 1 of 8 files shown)


cmaddox5 (approved):
Very cool. Thanks for all this nice refactor work.

PR #2295, opened 2024-11-14
chore: look up headway values by stop

digitalcora:
Instead of manually configuring a "headway key" for every departures section, we'd like to be able to translate stop IDs (which departures already must be configured with) into headway keys.

Currently the mapping of "headway key" to stop IDs only exists implicitly in the PA/ESS sign configuration in `realtime_signs`, and then only for stations that have PA/ESS signs. So for now, we need to encode this information ourselves within Screens.


cmaddox5 (approved):
Love it. I'm realizing now that the existing DUP headway logic is kinda meh. Your changes look good though.

PR #2303, opened 2024-11-18
refactor: simplify dependency injection approach

digitalcora:
Rather than adding an app configuration key for every mockable dependency of every module, we can standardize on a convention for naming and defining the mocks, assisted by a small macro.

### DIFF
--- a/lib/screens/inject.ex
+++ b/lib/screens/inject.ex
@@ -1,0 +1,26 @@
+defmodule Screens.Inject do
+  @moduledoc "Conveniences for dependency injection and mocking."
+
+  @doc """
+  When the Mix env is not `test`, resolves to the given module name. When it is, resolves to the
+  module name with `.Mock` appended.
+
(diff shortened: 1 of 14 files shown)


lib/screens/inject.ex
```diff
@@ -0,0 +1,26 @@
+defmodule Screens.Inject do
+  @moduledoc "Conveniences for dependency injection and mocking."
+
+  @doc """
+  When the Mix env is not `test`, resolves to the given module name. When it is, resolves to the
+  module name with `.Mock` appended.
+
+  The mock module is not automatically defined (see e.g. `test/support/mocks.ex`).
```
digitalcora, line 8: I tried for a bit to actually make this the case (auto-create the mock if not already defined), but kept running into issues with "cannot define module ... because it is currently being defined". This may require some higher-level compiler wizardry than I can conjure.

cmaddox5 (approved):
It's too bad about not being able to auto-definine mocks. Even without it, this should be helpful.

PR #2305, opened 2024-11-19
fix: Closure row margins

cmaddox5:
Closure rows were only taking up as much width as the longest elevator name. When the longest name can fit entirely between the margins, this caused dividers to be too short and for subsequent pages to appear on the current page. See [here](https://mbta.slack.com/archives/C039ARUM48H/p1731707189363699) for reference. I moved around some margins so that the rows are always 1080px wide.


assets/css/v2/elevator/elevator_closures.scss
```diff
@@ -66,8 +66,7 @@
         transform: translateX(calc(-100% * var(--closure-list-offset)));
 
         .closure-row {
-          max-width: 1080px;
-          padding: 0 48px;
```
digitalcora, line 70: Having a universal padding on a container feels more reliable than needing a (possibly duplicated) margin on each kind of item within the container, but maybe I'm overlooking a reason it has to be this way.

digitalcora (approved):

assets/css/v2/elevator/elevator_closures.scss
cmaddox5: I agree so I revisited. I think I came up with a better solution.

PR #2306, opened 2024-11-19
chore: add headway lookups to RDS logic

digitalcora:
...plus a couple other minor changes. This should be the last "foundation" task needed before we can finally implement some end-to-end DUP logic on top.


cmaddox5 (approved):
its_happening.gif

digitalcora:
@cmaddox5 Save your laser raves for when `DupNew` actually uses this 😉

PR #2307, opened 2024-11-21
refactor: remove unused departure alerts feature

digitalcora:
This feature has been "turned off" since 5e97a7f, three years ago. We plan to once again display alert information as part of departures in the near future, but in a considerably different form.


cmaddox5 (approved):
We were definitely a bit over optimistic thinking we would ever add these back 😅

PR #2310, opened 2024-11-22
fix: Handle LastTrip polling errors

cmaddox5:
We have been getting quite a few errors related to the `LastTrip.Poller`: https://mbtace.sentry.io/issues/5876255748/events/00f99016b99f4da49dbd9efd4191a83f/. It seems like we sometimes read the file and there is nothing in it. `Jason.decode` doesn't particular like this so throws an error which causes the process to crash. I'm not 100% sure if the file being empty is normal, but I also don't think we should crash constantly because of it. The timer for the process is every second so we will catch the new data as soon as it's published anyway.

### DIFF
--- a/lib/screens/last_trip/poller.ex
+++ b/lib/screens/last_trip/poller.ex
@@ -2,7 +2,6 @@ defmodule Screens.LastTrip.Poller do
   @moduledoc """
   GenServer that polls predictions to calculate the last trip of the day
   """
-  require Logger
   alias Screens.LastTrip.Cache
   alias Screens.LastTrip.Parser
   alias Screens.LastTrip.TripUpdates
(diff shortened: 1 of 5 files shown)


lib/screens/last_trip/poller.ex
```diff
@@ -32,14 +33,20 @@ defmodule Screens.LastTrip.Poller do
...
+           %{
+             trip_updates: trip_updates,
+             vehicle_positions: vehicle_positions
+           }} ->
+            update_last_trips(trip_updates)
+            update_recent_departures(trip_updates, vehicle_positions)
+
+          {:error, error} ->
+            Logger.warning(
+              "[last_trip_poller] failed while polling for new data: #{inspect(error)}"
```
digitalcora, line 47: In case there actually is a persistent error, I don't think we should necessarily just keep warning forever, letting this data drift arbitrarily far out of date without notifying us that something's wrong. Perhaps we could incorporate a "failure count" into the process state, and crash when consecutive failures are too high.

cmaddox5: Seems like a good use case for the `retry` macro. I can add that in no problem.

digitalcora (approved):
Looks okay, let's deploy it to dev-green and see if it does indeed handle the crashes.

PR #2313, opened 2024-11-22
Revert "chore: Add custom browserslist (#2290)"

cmaddox5:
GDS can't handle it so reverting it for now.


digitalcora:
Though I think it's likely, I'd like to test that this will properly unblock us on deploying `main` before actually merging it. My proposal would be:

* Deploy this branch to dev.
* Temporarily update just one of the GDS E-inks to point to the dev version of its URL.
* Monitor to see whether it continues making data requests. (Ideally, the only consequence is that the screen briefly shows "dev" in its header in small text, but otherwise displays the same information.)
* Revert the update to the E-ink device, pointing it back to production.
* Assuming this test was successful, merge this branch and deploy to production.

Let me know if that makes sense.

cmaddox5:
Picking EIB-101 to test this out. Copied its prod config into dev. Will let the branch sit for about 10 minutes and then update the URL.

cmaddox5:
Starting to see successful api requests: https://mbta.splunkcloud.com/en-US/app/search/search?sid=1732308914.7805057. I think this was the issue like we thought. Will leave this up for ~10 more minutes, switch the URL, and merge the change after approval. Then we can deploy `main` again.

cmaddox5:
Still looking good: https://mbta.splunkcloud.com/en-US/app/search/search?sid=1732309570.7807810. I'm now confident in this being the issue.

digitalcora (approved):
Sorry for the delay in approving, I misremembered that I had already approved it and so thought the ball was still in your court. Seems like this will fix things for now.

PR #2315, opened 2024-11-25
feat: Current elevator closure layout

cmaddox5:
This PR adds a new layout for elevator screens. When the current elevator is closed at all, we show [this](https://www.figma.com/design/UG6HV7HIsdR6B6QABOFm01/Elevator-Screens?node-id=2085-15738&node-type=canvas&t=VF0v2LngnA2jUdbk-0) layout that helps riders get to an alternate elevator. Maps will be created by Betsy and added to S3. Coordinates for the map marker will also be provided by Betsy.

Config that can be used for testing:
```json
{
  "ELE-101": {
    "disabled": false,
    "name": "",
    "app_id": "elevator_v2",
    "app_params": {
      "evergreen_content": [],
      "elevator_id": "842",
      "alternate_direction_text": "Alternate direction text goes here. This is the largest body text class. Use this for shorter sets of directions. Add up to 6 lines at this size",
      "accessible_path_image_url": "images/elevator/maps/forest-hills-map.png",
      "accessible_path_direction_arrow": "e",
      "accessible_path_image_here_coordinates": {
        "x": 151,
        "y": 182
      }
    },
    "device_id": "",
    "hidden_from_screenplay": false,
    "refresh_if_loaded_before": null,
    "tags": [],
    "vendor": "mimo"
  }
}
```

---
  - https://app.asana.com/0/0/1208677840815945


assets/css/v2/elevator/elevator_closures.scss
```diff
@@ -34,118 +61,238 @@
...
-              height: 68.13px;
-              margin-right: 24px;
+      .alternate-direction-text {
+        color: $cool-black-30;
+      }
 
-              &__text {
-                line-height: 68.13px;
-              }
+      .alternate-direction-text.small {
```
digitalcora, line 133: These could be nested to avoid the repetition, like

```scss
.alternate-direction-text {
  color: $cool-black-30;

  &.small {
    // ...
  }
```

assets/css/v2/elevator/elevator_closures.scss
```diff
@@ -34,118 +61,238 @@
...
-    .paging-info-container {
-      position: absolute;
-      bottom: 0;
+    .paging-indicators {
+      justify-content: end;
+      height: 120px;
+    }
+  }
+
+  .outside-closures-view {
```
digitalcora, line 165: Since `current-elevator-closed-view` and `outside-closures-view` are pretty independent of each other, what do you think about splitting them into separate stylesheets? This file is getting pretty long as it is.

assets/src/components/v2/elevator/elevator_closures.tsx
```diff
@@ -177,10 +196,92 @@ const OutsideClosureList = ({
...
+          <PagingIndicators numPages={numPages} pageIndex={pageIndex} />
+        </div>
+      )}
+    </div>
+  );
+};
+
+interface OutsideClosuresViewProps extends OutsideClosureListProps {}
+
+const OutsideClosuresView = ({
```
digitalcora, line 208: Same comment as with the stylesheets — these "views" seem so separate, and each complex enough, that they deserve their own modules. (This also makes me wonder about making them separate widgets on the backend, if that would make sense.)

assets/src/components/v2/elevator/elevator_closures.tsx
```diff
@@ -24,6 +31,50 @@ type ElevatorClosure = {
...
+    </div>
+  );
+};
+
+interface PagingIndicatorsProps {
+  numPages: number;
+  pageIndex: number;
+}
+
+const PagingIndicators = ({ numPages, pageIndex }: PagingIndicatorsProps) => {
```
digitalcora, line 53: Good refactor, making this a proper component.

lib/screens/v2/candidate_generator/elevator/closures.ex
```diff
@@ -31,10 +31,10 @@ defmodule Screens.V2.CandidateGenerator.Elevator.Closures do
 
       [
         %ElevatorClosures{
-          id: elevator_id,
           in_station_closures: Enum.map(in_station_alerts, &alert_to_elevator_closure/1),
           other_stations_with_closures:
-            format_outside_closures(outside_alerts, parent_station_map, routes_map)
+            format_outside_closures(outside_alerts, parent_station_map, routes_map),
+          elevator_config: config
```
digitalcora, line 37: `app_params` might be a better name for this — `config` is a little imprecise and has a lot of overlap with the general notion of a screen's whole "config".

lib/screens/v2/widget_instance/blue_bikes.ex
```diff
@@ -35,7 +34,7 @@ defmodule Screens.V2.WidgetInstance.BlueBikes do
   @type special_station :: %{
           status: :valet | :out_of_service,
           id: String.t(),
-          arrow: Station.arrow(),
+          arrow: Arrow.t(),
```
digitalcora, line 37: In the future I think it'd be nice to split changes like this into their own commit/PR if the change ends up being more than just updating the dependency — for anyone looking back at this work later, it'll be unclear why this change to the `BlueBikes` widget is included in a commit that's supposedly about elevator screens (assuming you plan to squash-merge all the work on this branch). No need to extract this particular change now, though.

cmaddox5: I did think about that when making this change and should have went with my gut. Will make sure to avoid this in the future.

assets/css/v2/elevator/elevator_closures.scss
cmaddox5: This was intentional due to some issues using the nesting selector when implementing the other layout. This is one of the places where having a working browserslist + stylelint configuration would be very helpful. I think I should add a comment to the top of this file communicating the modern styles that I know don't work on these screens.

digitalcora: I'm not sure what browser-specific issues could exist here? Sass is going to compile things like nesting and `&` down to the equivalent "flat" CSS, unless something has changed and they've started using the (_very_ new) native CSS nesting spec.

assets/css/v2/elevator/elevator_closures.scss
cmaddox5: Good call. Will work on splitting out the styles and React components.

assets/css/v2/elevator/elevator_closures.scss
```diff
@@ -34,118 +61,238 @@
...
+
+  .current-elevator-closed-view {
     position: relative;
     height: 100%;
-    background-color: $warm-neutral-90;
 
-    .header-container {
-      margin: 48px;
-      margin-bottom: 0;
+    .shape {
```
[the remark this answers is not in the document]
cmaddox5: "notch" is honestly the best term I can think of. I'll go with that.

lib/screens/v2/candidate_generator/elevator.ex
```diff
@@ -5,21 +5,14 @@ defmodule Screens.V2.CandidateGenerator.Elevator do
...
   def screen_template do
     {
       :screen,
       %{
-        normal: [
-          :header,
-          :main_content,
-          :footer
-        ]
+        normal: [:main_content]
```
cmaddox5, line 15: Want to call this change out. I am having trouble styling the header and footer when the current elevator is closed (`:has` is not supported on the screens). I made this change as more of an RFC because I'm not sure how I feel about it.

digitalcora: I think the `:has` solution was slightly sus to begin with — effectively the header and footer components "reaching outside themselves" to peek at what else is on the page, the CSS equivalent of a React component somehow directly accessing a parent component's props without having them passed in. This is _an_ option to address that, I'm not sure if it's the best.

One possible alternative: Maintain the separate Header and Footer widgets and slots on the backend, but have the candidate generator (`ElevatorClosures` that is) generate the variants of these widgets, depending on whether it also generates an "inside" or "outside" closure widget. In other words, the function either returns

```elixir
[%Header{}, %OutsideClosures{...}, %Footer{}]
```

or

```elixir
[%Header{variant: :closed}, %CurrentClosure{...}, %Footer{variant: :closed}]
```

all together.

cmaddox5: Looks like as of the current state of the designs, we will have 4 layouts but this layout is the only one that changes the header and footer. 

Your alternative was also something I considered but wasn't sure how I felt about the one CG creating 3 different widgets. The pro with it is the body takeover can still be done without needing a new layout. In my solution, we would need to create an additional layout.

digitalcora: Good point — we already have another combination(s) of header, footer, and body element, in the form of possible evergreen content.

> Your alternative was also something I considered but wasn't sure how I felt about the one CG creating 3 different widgets.

I don't think we need to adhere closely to a convention of "one generator function => one widget type". This might be the case in most circumstances, but if it were universally the case, there wouldn't even be a need for a separation between "generator function" and "widget type". In cases like this where the _combination_ of generated widgets legitimately depends on some higher-level criteria, it makes sense for that to be orchestrated by a single generator function.

cmaddox5: I think your alternative is better for this scenario. Nice to not have to juggle between layouts unnecessarily.

cmaddox5: Implemented your alternative. I _think_ it works but need to wait on the SSO issues to be resolved before confirming (can't make an alert right now).

cmaddox5: Actually, I see an existing elevator alert so I changed my screen to that ID (`935`). Should be good to go.

assets/src/components/v2/elevator/elevator_closures.tsx
```diff
@@ -1,211 +1,76 @@
...
+
   return (
     <div className="elevator-closures">
-      <InStationSummary closures={inStationClosures} />
-      <OutsideClosureList
-        stations={otherStationsWithClosures}
-        lastUpdate={lastUpdate}
-        onFinish={onFinish}
-      />
+      {currentElevatorClosure ? (
```
digitalcora, line 55: We might have already talked about this, but I think it could be worthwhile to move this conditional to the widget generation. This single "widget" is basically two widgets in a trenchcoat — we always send over all the fields for both, and then throw half of them away, depending on a condition the server itself could easily check. In other words, we could instead have a `CurrentElevatorClosed` widget and an `OutsideClosures` widget on the backend, each populated with only the fields needed for that "view".

I could be missing something about future plans for this widget that might require it to have access to both sets of information at the same time, though.

cmaddox5: Big fan of that approach. That would likely be a pretty big diff though. You think it makes sense here or should I create a followup PR?

digitalcora: I think within this PR is actually the best place for it, since this is where we're introducing the second widget. Else we'd have (in the final commits on `main`) one commit that adds the widget using this approach, and a second that completely changes the approach. That's a lot of extra, and ultimately pointless, information for anyone looking at history to sort through.

cmaddox5: Should be good to go if you want to take a look.

assets/src/components/v2/elevator/models/station.ts
```diff
@@ -0,0 +1,9 @@
+import { type Pill } from "Components/v2/departures/route_pill";
+import { type Closure } from "Components/v2/elevator/models/closure";
+
+export type StationWithClosures = {
```
digitalcora, line 4: I think these could both live in a `v2/elevator/types.ts`. Splitting them out at this point feels premature, and I don't love the name "model" when these don't have anything to them other than the type definition (it kinda implies some larger framework involving "models" as object-oriented entities that can be persisted or cached or etc.)

lib/screens/v2/widget_instance/outside_elevator_closures.ex
```diff
@@ -0,0 +1,58 @@
+defmodule Screens.V2.WidgetInstance.OutsideElevatorClosures do
+  @moduledoc false
+
+  alias Screens.Stops.Stop
+  alias Screens.V2.WidgetInstance.CurrentElevatorClosed.Closure
```
digitalcora, line 5: Since this is used from both "sibling" widget modules, defining it as a sub-module of one of them seems arbitrary. But then I also hesitate to put it anywhere else in the `WidgetInstance` namespace since it is not, itself, a Widget™. Maybe `v2/widget_instance/elevator/closure` would be okay?

test/screens/v2/widget_instance/current_elevator_closed_test.exs
```diff
@@ -0,0 +1,84 @@
...
+  end
+
+  describe "priority/1" do
+    test "returns 1", %{instance: instance} do
+      assert [1] == WidgetInstance.priority(instance)
+    end
+  end
+
+  describe "serialize/1" do
+    test "returns map with id, closure, and alternate direction info", %{instance: instance} do
```
digitalcora, line 35: I don't know if we really _need_ all these "it does the thing" tests for functions whose implementation is extremely basic... ideally the code paths would be covered by some higher-level test, but we have no such category of tests at the moment. Guess I won't make any specific recommendation here, just noting that it's an unfortunate situation.

digitalcora (approved):
Looking good!

test/screens/v2/widget_instance/current_elevator_closed_test.exs
cmaddox5: I have this thought every time I make tests for a new widget. Definitely a good idea to put end-to-end framework tests on some kind of radar. Not sure if that's big enough to be roadmap worthy or if a simple tech-debt task would be sufficient?

PR #2316, opened 2024-11-25
chore: Elevator footer style updates

cmaddox5:
Footer got a small design update. Because this affects all layouts, putting it in it's own small standalone branch. [Design](https://www.figma.com/design/UG6HV7HIsdR6B6QABOFm01/Elevator-Screens?node-id=2130-37201&node-type=frame&t=T11qgt4IXujdiHPl-0)

### DIFF
--- a/assets/css/v2/elevator/footer.scss
+++ b/assets/css/v2/elevator/footer.scss
@@ -1,11 +1,15 @@
 .footer {
-  display: flex;
-  align-items: center;
+  box-sizing: border-box;
   height: 100%;
-  padding: 0 48px;
-  font-family: Inter, sans-serif;
+  padding: 26px 48px;
   font-size: 48px;
+  font-weight: 500;
   line-height: 62px;
   color: white;
   background-color: $cool-black-30;
+
+  .bold {
+    font-weight: 700;
+    letter-spacing: -0.2px;
+  }
 }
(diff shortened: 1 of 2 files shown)


assets/css/v2/elevator/footer.scss
```diff
@@ -1,11 +1,15 @@
...
-  padding: 0 48px;
-  font-family: Inter, sans-serif;
+  padding: 26px 48px;
   font-size: 48px;
+  font-weight: 500;
   line-height: 62px;
   color: white;
   background-color: $cool-black-30;
+
+  .bold {
```
digitalcora, line 11: Since the class name isn't really adding any information here, could we just use `<b>`? This has the advantages that the markup would be less noisy; the font weight would be applied automatically as per the normal behavior of `b` tags; and `b` is semantically appropriate in this case (albeit this doesn't mean much since no one is programmatically going through the markup of our screens).

digitalcora (approved):
Looks fine, just a small optional suggestion.

assets/css/v2/elevator/footer.scss
cmaddox5: That makes sense. I'll still need to add the special `letter-spacing` but can just add it to the CSS for the `b` tag and let it handle `font-weight` on its own.

PR #2317, opened 2024-11-25
chore: Add flex_zone pages to Mercury screen requests

cmaddox5:
Mercury asked for us to append all flex zone pages to each data request. This will allow them to make less requests and improve battery life. Because the layout is required to get the flex zone data, I decided to make an alternative to `get/2` that works through `select_variant/3` with a different `then_fn`. I worry this might be a tad overcomplicated so happy to hear alternatives.

### DIFF
--- a/lib/screens/v2/screen_data.ex
+++ b/lib/screens/v2/screen_data.ex
@@ -16,6 +16,7 @@ defmodule Screens.V2.ScreenData do
 
   @type t :: %{type: atom()}
   @type simulation_data :: %{full_page: t(), flex_zone: [t()]}
+  @type data_with_flex_zone :: {t(), [t()]}
   @type variants(data) :: {data, %{String.t() => data}}
   @type screen_id :: String.t()
   @type options :: [
@@ -29,6 +30,12 @@ defmodule Screens.V2.ScreenData do
   @spec get(screen_id(), options()) :: t()
   def get(screen_id, opts \\ []) do
     select_variant(screen_id, opts, &layout_to_data/2)
+  end
+
+  @spec get_with_flex_zone(screen_id()) :: data_with_flex_zone()
+  @spec get_with_flex_zone(screen_id(), options()) :: data_with_flex_zone()
+  def get_with_flex_zone(screen_id, opts \\ []) do
+    select_variant(screen_id, opts, &layout_to_data_with_flex_zone/2)
   end
 
   @spec simulation(screen_id()) :: simulation_data()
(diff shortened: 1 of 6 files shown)


lib/screens/v2/screen_data.ex
```diff
@@ -31,6 +32,12 @@ defmodule Screens.V2.ScreenData do
     select_variant(screen_id, opts, &layout_to_data/2)
   end
 
+  @spec get_with_flex_zone(screen_id()) :: data_with_flex_zone()
+  @spec get_with_flex_zone(screen_id(), options()) :: data_with_flex_zone()
+  def get_with_flex_zone(screen_id, opts \\ []) do
```
digitalcora, line 37: This essentially is the same thing as `simulation`, it just returns the data in a slightly different format. It might be simpler to leave this module unchanged, call `simulation` from the controller in the `mercury` special case, and just reformat the data there. Like

```elixir
%{full_page: data, flex_zone: flex_zone} = ScreenData.simulation(...)
%{@base_response | data: data, flex_zone: flex_zone}
```

This makes the function name `simulation`, and the specific format of `simulation_data()`, a bit awkward because this is no longer specific to Screenplay simulations. I would be fine letting this awkwardness sit, since we'd expect to revisit it as we think more about client-side versus server-side paging.

lib/screens_web/controllers/v2/screen_api_controller.ex
```diff
@@ -8,7 +8,7 @@ defmodule ScreensWeb.V2.ScreenApiController do
   alias Screens.V2.{ScreenAudioData, ScreenData}
   alias ScreensConfig.Screen
 
-  @base_response %{data: nil, disabled: false, force_reload: false}
+  @base_response %{data: nil, disabled: false, force_reload: false, flex_zone: nil}
```
digitalcora, line 11: I'd be fine with omitting this from the base response, and having it only present at all for "Mercury requests", to reduce confusion about what its role is in our "normal" screen pipeline. I think our other Mercury special-case fields basically work the same way.

lib/screens/v2/screen_data.ex
cmaddox5: Wow, totally overlooked that in my first pass. I like that much better.

lib/screens_web/controllers/v2/screen_api_controller.ex
```diff
@@ -84,19 +84,29 @@ defmodule ScreensWeb.V2.ScreenApiController do
...
+            update_visible_alerts?: true
+          )
           |> put_extra_fields(screen_id, screen)
 
         json(conn, response)
     end
   end
 
-  defp screen_response(screen_id, "all", opts) do
+  defp screen_response(screen_id, "all", _, opts) do
```
digitalcora, line 97: Minor quibble: The order of arguments being `screen_id, variant, screen` feels off to me — I'd expect `screen_id, screen, variant`, so the "parts of the screen" are together.

digitalcora (approved):
Conditionally approving: I think we should have a controller test for this, to verify that A) it works as intended right now and B) we don't break Mercury screens in the future, since that kind of thing might otherwise slip past us in our own manual testing. It doesn't need a ton of detail, just that the `flex_zone` exists under the right circumstances.

PR #2328, opened 2024-12-03
feat: Filter out redundant outside elevator closures

cmaddox5:
Adds alert filtering based on presence of a redundant elevator. Only affects alerts outside the current station. CSV is [here](https://docs.google.com/spreadsheets/d/1lHogme-2SuDSgjrRK52k7yVgSFK-v_LMmUfkYgBJjIU/edit?gid=179933470#gid=179933470). Add some instructions to the script just in case the data is updated. 

---
  - https://app.asana.com/0/0/1208551608056342


priv/elevators/elevator_station_data.csv
```diff
@@ -0,0 +1,227 @@
+test,elevator_id,line,parent_station,Station (for Betsy),children_ids_served,elevator_type,escalator_direction,facility_name,alternate_elevator_ids,alternate_service_long_text,Entering System Categorization,Exiting System Categorization,What station to EXIT at,Short Text,Questions/Notes,Is direction important?,Is the line/mode important?,secondary_alternate_service_text,elevator_name,alternate_service_short_text,facility_lat,facility_lon
```
digitalcora, line 1: It doesn't seem like we need to commit this file; it's only read by the script, which produces another file which is then committed. The instructions could say to download the CSV to a temporary directory, outside the git repo.

priv/elevators/elevator_redundancy_data.json
```diff
@@ -0,0 +1,1057 @@
+[
+  {
+    "id": "814",
+    "station_id": "place-alfcl",
```
digitalcora, line 4: It appears we don't use this field at all in the current code. Is there a future need for it?

lib/screens/v2/candidate_generator/elevator/closures.ex
```diff
@@ -169,4 +172,19 @@ defmodule Screens.V2.CandidateGenerator.Elevator.Closures do
...
+       ) do
+    Enum.find(redundancy_data, &(&1.id == informed_facility_id))[:nearby_redundancy?]
+  end
+
+  defp fetch_elevator_redundancy_data do
+    :screens
+    |> :code.priv_dir()
+    |> Path.join("elevators/elevator_redundancy_data.json")
+    |> File.read!()
+    |> Jason.decode!(keys: :atoms)
```
digitalcora, line 188: We could do this once at compile time, instead of re-reading and decoding the file on every elevator screen request, by assigning a module attribute instead. As a bonus the app will then fail to compile if we get the path wrong or something, instead of crashing at runtime.

lib/screens/v2/candidate_generator/elevator/closures.ex
```diff
@@ -169,4 +172,19 @@ defmodule Screens.V2.CandidateGenerator.Elevator.Closures do
       ie -> if InformedEntity.parent_station?(ie), do: ie.stop
     end)
   end
+
+  defp has_nearby_redundancy?(
+         %Alert{informed_entities: [%{facility: %{id: informed_facility_id}} | _]},
+         redundancy_data
+       ) do
+    Enum.find(redundancy_data, &(&1.id == informed_facility_id))[:nearby_redundancy?]
```
digitalcora, line 180: * It's possible the ID we're looking for doesn't exist in the committed data (`find` may return `nil`), but we're not handling that case here.
* We could instead structure the data as a map where the elevator IDs are keys (`{"814": {"nearby_redundancy?": true}, ...}`), so instead of doing a linear-time search through the entire list for every active elevator closure, we could use a map lookup.

scripts/format_elevator_data.exs
```diff
@@ -0,0 +1,30 @@
+#!/usr/bin/env -S ERL_FLAGS=+B elixir
+
+# Script used to format the spreadsheet found at
+# https://docs.google.com/spreadsheets/d/1lHogme-2SuDSgjrRK52k7yVgSFK-v_LMmUfkYgBJjIU/edit?gid=179933470#gid=179933470
```
digitalcora, line 4: It occurs to me that this spreadsheet probably doesn't need to be public to the world, as the original copy is, given we are not programmatically downloading it. But we'd have to figure out what we're doing about access control on the team.

priv/elevators/elevator_station_data.csv
cmaddox5: I was going back and forth on this. You're right, I'll take it out of the commit. Can also add the path as a script param.

priv/elevators/elevator_redundancy_data.json
cmaddox5: I anticipated a need for it, but that may be premature. I assume it will be necessary in [this](https://app.asana.com/0/1185117109217422/1208677840815949/f) task, but so will other fields I didn't include. Can remove now and let future implementation add what is needed.

lib/screens/v2/candidate_generator/elevator/closures.ex
cmaddox5: Good idea. I feel like Betsy and Dave would want to know about elevators missing from the map so will ask them what they want to happen when that occurs. Might just add a log and return `false` in that case.

lib/screens/v2/candidate_generator/elevator/closures.ex
```diff
@@ -169,4 +176,15 @@ defmodule Screens.V2.CandidateGenerator.Elevator.Closures do
       ie -> if InformedEntity.parent_station?(ie), do: ie.stop
     end)
   end
+
+  defp has_nearby_redundancy?(%Alert{
+         informed_entities: [%{facility: %{id: informed_facility_id}} | _]
```
digitalcora, line 181: Here, we check for an affected elevator by matching on just the first informed entity in the alert (and actually, if this happens to not reference a facility, we'll crash ⚠️). But later in `alert_to_elevator_closure` we search through the informed entities for the first one that has a non-null facility. I wonder if we could move things around such that this logic only happens in one place and is therefore consistent — maybe first mapping the alerts to "closures" using `alert_to_elevator_closure`, then rejecting those with nearby redundancy, then constructing the route pills and `Station`s from those that remain.

lib/screens/v2/candidate_generator/elevator/closures.ex
```diff
@@ -169,4 +176,15 @@ defmodule Screens.V2.CandidateGenerator.Elevator.Closures do
...
+       }) do
+    if data = @elevator_redundancy_data[informed_facility_id] do
+      data["nearby_redundancy?"]
+    else
+      Logger.error("Elevator #{informed_facility_id} does not exist in redundancy data")
```
digitalcora, line 186: We're unlikely to see this unless we're actively checking for it, and the effect of an elevator incorrectly _not_ being filtered out is subtle enough that we maybe shouldn't count on it being reported. A Sentry report would be more attention-grabbing.

lib/screens/v2/candidate_generator/elevator/closures.ex
cmaddox5: Part of `relevant_alert?/1` is ensuring that all `informed_entities` include a facility and that the facility is the same for all. So we can be sure that any item in the list will include a facility in this case.

That said, I think there is room to clean up some of these helper functions. Will take a look.

lib/screens/v2/candidate_generator/elevator/closures.ex
cmaddox5: My idea was to report to Sentry but didn't think about calling `Sentry.capture_message`. Can change Sentry reporting only.

digitalcora: `Logger.error` currently doesn't report to Sentry; the "Logger integration" is for capturing process crashes only, by default. Capturing log _messages_ and reporting them to Sentry requires [enabling an option](https://docs.sentry.io/platforms/elixir/configuration/logging/#capture-log-messages). Ultimately I don't think it would be a bad idea to enable this: we'd be able to just `Logger.[error/warning]` and get console output in development, Sentry reports in production.

lib/screens/v2/candidate_generator/elevator/closures.ex
digitalcora: Ah, gotcha. So there is technically no possibility of a crash here, it's just not super obvious.

lib/screens/v2/candidate_generator/elevator/closures.ex
cmaddox5: Ahh ok not sure I realized that for some reason. Thank you.

lib/screens/v2/candidate_generator/elevator/closures.ex
cmaddox5: I think maybe the best way to make that more clear is to move this function call to `relevant_alert?` instead? There's no reason it can be done right after the facility check there.

lib/screens/v2/candidate_generator/elevator/closures.ex
```diff
@@ -75,8 +81,11 @@ defmodule Screens.V2.CandidateGenerator.Elevator.Closures do
     end
   end
 
-  defp relevant_alert?(alert) do
-    relevant_effect?(alert) and informs_one_facility?(alert)
+  # All alerts at home station are relevant but only elevators without redundancies are
+  # relevant at other stations.
+  defp relevant_alert?(alert, home_stop_id) do
+    relevant_effect?(alert) and informs_one_facility?(alert) and
+      (Alert.informs_stop_id?(alert, home_stop_id) or not has_nearby_redundancy?(alert))
```
digitalcora, line 88: When an alert informs a facility, does it also always inform the facility's parent station? This isn't necessarily true of GTFS in general but I'd believe that it's a guarantee of our alerts implementation.

digitalcora (approved):
One minor question. I'm also noticing there are no new/updated tests for this behavior; we should have something there.

lib/screens/v2/candidate_generator/elevator/closures.ex
cmaddox5: From what I can tell, yes. I think it always includes the parent + every child stop not flagged for exclusion in the `properties` field for the facility.

cmaddox5:
> I'm also noticing there are no new/updated tests for this behavior; we should have something there.

Thanks for the callout. Added a test for this. I'm not sure I'm happy with how I handled the file path now that we have a test fixture so any suggestions is appreciated.

PR #2332, opened 2024-12-04
temp: Force bottom filler for Mercury screens.

cmaddox5:
See [this thread](https://mbta.slack.com/archives/G012H4G4U06/p1733323450195799) for context.

Tl;dr: Our Mercury screens have a burn-in problem. Mercury has requested we use a static image on the bottom panel. They will then turn off the panel until the issue is resolved.

---
  - https://app.asana.com/0/0/1208994786336825


digitalcora:
@cmaddox5 Noticing this is still in Draft, does anything need to be handed off here?

digitalcora:
We no longer expect to need this since Mercury has deployed a firmware-level fix. [See Slack thread](https://mbta.slack.com/archives/G012H4G4U06/p1735850478028089).

PR #2334, opened 2024-12-05
fix: Map marker cords

cmaddox5:
Betsy pointed out that coordinates from Figma weren't matching up exactly in our code. I then realized I mapped `top` and `left` wrong 😅. Flipped them so coordinates work as expected.


digitalcora (approved):
Wow, can't believe I missed this too. ✅

PR #2338, opened 2024-12-09
feat: Elevator Closures at home station

cmaddox5:
This PR updates the list layout to include closures occurring at other elevators in the current station. When one of these closures exists, the summary header text changes and we add the closure row to the top of the list. Because this widget now shows closures at the current station, `outside` doesn't really make sense. Change some names and removed the `in_station_closures` field from the serialized response.

---
  - https://app.asana.com/0/0/1208839072659643


lib/screens/v2/candidate_generator/elevator/closures.ex
```diff
@@ -60,9 +60,8 @@ defmodule Screens.V2.CandidateGenerator.Elevator.Closures do
       {elevator_widget_instance, header_footer_variant} =
         if is_nil(current_elevator_closure) do
           {%ElevatorClosuresList{
-             in_station_closures: in_station_closures,
```
digitalcora, line 64: Since `in_station_closures` is now only used in the creation of `current_elevator_closure`, that logic could be merged into a single pipeline.

assets/src/components/v2/elevator/elevator_closures_list.tsx
```diff
@@ -39,33 +63,55 @@ const ClosureRow = ({ station }: ClosureRowProps) => {
...
+      <>
+        <b>This elevator is working.</b> Another elevator at this station is
+        down.
+      </>
+    );
+  } else {
+    text = (
+      <>
+        <b>This elevator is working.</b>{" "}
+        {_.capitalize(numberToWords.toWords(closures.length))} other elevators
```
digitalcora, line 89: There's a minor design issue we maybe didn't anticipate here, where certain "number words" are long enough to push the following word onto the next line, which cascades into creating a 3-line status message.

Screenshot 2024-12-09 at 10 46 50 AM

My suggestion would be to use the "numeric number" (e.g. "2 other elevators"), or just use some fixed language when more than one other elevator is down, like "Other elevators at this station". Let's see what Betsy thinks.

cmaddox5: To put a bow on this convo, Betsy agreed with the digit approach: https://mbta.slack.com/archives/C039ARUM48H/p1733760048513209?thread_ts=1733759798.319529&cid=C039ARUM48H

digitalcora (approved):

PR #2340, opened 2024-12-10
refactor: parse all elevator redundancy categories

digitalcora:
Rather than only storing whether an elevator's redundancy category is "nearby", we now store all categories, including any associated summary text. This is a prerequisite for displaying [redundancy summaries](https://app.asana.com/0/1185117109217413/1208677840815949) on the system-wide outages view.

Also extracts this logic into a dedicated module which can be mocked for testing, removing the need for a separate fixture file.


lib/screens/elevator.ex
```diff
@@ -0,0 +1,45 @@
...
+        |> File.read!()
+        |> Jason.decode!()
+
+  @callback get(Facility.id()) :: t() | nil
+  def get(id) do
+    case Map.get(@data, id) do
+      %{"redundancy" => 1} ->
+        %__MODULE__{id: id, redundancy: :nearby}
+
+      %{"redundancy" => 2} ->
```
PaulJKim, line 32: Category 2 should also be returning summary text as well shouldn't it? At least according to the task that this is a prerequisite for mentioned in the PR description.

digitalcora: I'm conceiving of this module as purely an interface to the data in the spreadsheet — "rules" apart from the raw data, like "when the redundancy category is 2, always use this specific text" would be implemented in the widget generator. I decided to draw the line here because of the rule where the summary text for _all_ categories changes if any of the "backup elevators" for a given elevator are down, which depends on data from outside the spreadsheet (active alerts).

lib/screens/v2/candidate_generator/elevator/closures.ex
```diff
@@ -188,15 +180,20 @@ defmodule Screens.V2.CandidateGenerator.Elevator.Closures do
...
+        true
+
+      %Elevator{} ->
+        false
+
+      nil ->
+        _ =
+          Sentry.capture_message(
+            "Elevator #{informed_facility_id} does not exist in redundancy data"
+          )
```
PaulJKim, line 194: Should we be logging to Splunk here as well or is this a case where only want to log to Sentry?

digitalcora: A previous discussion with Christian about this exact line is actually what prompted #2341. Once that's in, this will use `Screens.Log`. I think this will generally be the preferred way to log warnings/errors, since in development we'll see the printout in the console, and in deployed environments we'll get the Sentry report.

PaulJKim: ah, makes sense

PaulJKim (approved):
Looks good!

PR #2341, opened 2024-12-10
chore: consolidate error/warning logging

digitalcora:
Various modules were "manually" implementing the `key=value` log format we use for Splunk searchability, and most errors or warnings were only being logged rather than reported to Sentry (where they are easier to organize and we can be proactively notified about them). This change consolidates all existing errors/warnings into a shared function that logs to both `Logger` and Sentry.

* Sentry reports are also sent asynchronously, so they don't hold up returning a value to a client.

* All errors/warnings are given a unique underscored identifier rather than a free-text message, which improves Splunk ergonomics.

Existing logging at `info` level and below is not changed, but would be a logical next step.

### DIFF
--- a/lib/screens/log.ex
+++ b/lib/screens/log.ex
@@ -6,7 +6,7 @@ defmodule Screens.Log do
   def error(tag, data \\ []), do: log(:error, tag, data)
   def warning(tag, data \\ []), do: log(:warning, tag, data)
 
-  @spec log(:error | :warning, String.t(), Enumerable.t({term(), term()})) :: :ok
+  @spec log(:error | :warning, String.t(), Enumerable.t({String.Chars.t(), term()})) :: :ok
   defp log(level, tag, data) do
     Logger.log(
       level,
--- a/lib/screens/log.ex
+++ b/lib/screens/log.ex
@@ -1,0 +1,20 @@
+defmodule Screens.Log do
+  @moduledoc "Conveniences for logging to Splunk and Sentry."
+
+  require Logger
+
+  def error(tag, data \\ []), do: log(:error, tag, data)
+  def warning(tag, data \\ []), do: log(:warning, tag, data)
+
+  @spec log(:error | :warning, String.t(), Enumerable.t({term(), term()})) :: :ok
+  defp log(level, tag, data) do
+    Logger.log(
+      level,
+      tag <> " " <> Enum.map_join(data, " ", fn {key, value} -> "#{key}=#{inspect(value)}" end)
+    )
+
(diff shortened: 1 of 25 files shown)


lib/screens/log.ex
```diff
@@ -0,0 +1,20 @@
...
+  require Logger
+
+  def error(tag, data \\ []), do: log(:error, tag, data)
+  def warning(tag, data \\ []), do: log(:warning, tag, data)
+
+  @spec log(:error | :warning, String.t(), Enumerable.t({term(), term()})) :: :ok
+  defp log(level, tag, data) do
+    Logger.log(
+      level,
+      tag <> " " <> Enum.map_join(data, " ", fn {key, value} -> "#{key}=#{inspect(value)}" end)
```
PaulJKim, line 13: Probably unlikely, but technically anything can be passed in as a `key` right? Should we wrap it in an `inspect()` to be safe or maybe tighten the typespec to enforce that a keyword list is passed in?

digitalcora: You can also pass in a Map, or anything that implements Enumerable by returning `{key, value}` tuples — there are existing logs that use maps, so I wanted to preserve it. Though it's a good point that this will not accept literally anything, I should correct it to `Enumerable.t({String.Chars.t(), term()})`.

To be clear there is no real "type safety" here — `Enumerable.t(_anything)` is identical to `Enumerable.t()`, which is an alias to `term()`. Protocols are an Elixir invention so there is no way to express "value that implements protocol X" in Dialyzer's type system. The value is solely in the documentation of programmer intent.

lib/screens/route_patterns/route_direction_stops.ex
```diff
@@ -1,18 +1,13 @@
...
 
   def parse_result(%{"data" => data, "included" => included}, route_id) do
     included_data = parse_included_data(included)
     parse_data(data, included_data, route_id)
   end
 
-  def parse_result(_, _) do
-    Logger.warning("Unrecognized format of route_pattern data.")
-    :error
-  end
```
PaulJKim, line 14: Is this something we don't need to log anymore?

digitalcora: We haven't seen it at any time within log retention range, in any environment, and this is the only one of our many V3 API "parsers" that has something like this, so I feel okay about removing it. However I did notice that without this clause we're actually requiring an `included` element to be present, when it can technically be omitted from the response if there is no data at all — so thanks for the catch! I just updated this to align with what we're doing in other parsers.

PaulJKim (approved):
Looks good!

PR #2347, opened 2024-12-13
fix: unable to add elevator screens in admin

digitalcora:
When we added a few additional required fields to elevator screen configuration, we neglected to update the admin UI with these fields.

---
  - https://app.asana.com/0/0/1208978498455632

### DIFF
--- a/assets/src/components/admin/admin_add_modal.tsx
+++ b/assets/src/components/admin/admin_add_modal.tsx
@@ -80,7 +80,10 @@ const defaultAppParamsByAppId = {
     evergreen_content: [],
   },
   elevator_v2: {
-    elevator_id: "ELEVATOR_ID",
+    elevator_id: "",
+    alternate_direction_text: "",
+    accessible_path_image_here_coordinates: { x: 0, y: 0 },
+    evergreen_content: [],
   },
 };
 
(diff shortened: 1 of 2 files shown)


assets/src/components/admin/admin_add_modal.tsx
```diff
@@ -80,7 +80,11 @@ const defaultAppParamsByAppId = {
     evergreen_content: [],
   },
   elevator_v2: {
-    elevator_id: "ELEVATOR_ID",
+    elevator_id: "",
+    alternate_direction_text: "",
+    accessible_path_direction_arrow: "n",
```
robbie-sundstrom, line 85: Should this `accessible_path_direction_arrow` also have an empty default so that it forces a decision to be made?

assets/src/components/admin/admin_tables.tsx
```diff
@@ -385,6 +376,64 @@ const ElevatorV2ScreensTable = (): JSX.Element => {
...
+          FormCell: FormTextCell,
+        },
+        {
+          Header: "Dot X",
+          accessor: (row) =>
+            row.app_params.accessible_path_image_here_coordinates.x,
+          mutator: (row, value) => {
+            const newRow = structuredClone(row);
+            newRow.app_params.accessible_path_image_here_coordinates.x = value;
+            return newRow;
```
panentheos, line 418: nit: If we want to start taking advantage of lodash/fp, we could use `fp.set` here (and below) to do the nested update and handle the clone implicitly.

panentheos (approved):
Looks good, see small comment.

PR #2352, opened 2024-12-17
Add no data state for elevator screens

PaulJKim:
**Asana task**: ["No data" mode for elevator screens](https://app.asana.com/0/1185117109217413/1208859172526595/f)

Description
Shows a "no data" png on elevator screens when the api returns either a failure or disabled status.


assets/src/apps/v2/elevator.tsx
```diff
@@ -27,6 +28,7 @@ const TYPE_TO_COMPONENT = {
   evergreen_content: EvergreenContent,
   footer: Footer,
   normal_header: NormalHeader,
+  no_data: NoData,
```
digitalcora, line 31: There is also a `page_load_no_data` state (see `defaultResponseMapper` in `ScreenContainer`) which we should add here. You should be able to see it briefly when the screen page has first loaded but not yet received widget data from the server. We can just have this one also map to `NoData`.

assets/src/components/v2/elevator/no_data.tsx
```diff
@@ -0,0 +1,12 @@
+import React, { ComponentType } from "react";
+import { imagePath } from "Util/util";
+
+const NoData: ComponentType = () => {
+  return (
+    <>
+      <img src={imagePath(`elevator-status-no-data.png`)} />
+    </>
```
digitalcora, line 8: This enclosing fragment should be unneeded, since it only contains a single element.

assets/src/apps/v2/elevator.tsx
digitalcora: Ah yeah, just wasn't sure if you knew about it. I think it should be okay to have this as the `page_load_no_data` as well, just so we have _something_, and we can revisit it later if desired.

digitalcora (approved):

PR #2364, opened 2024-12-23
feat: audio readout for push-button Bus Shelters

digitalcora:
For Bus Shelter screens that use push-button audio (i.e. do not have a periodic audio readout), the Alert and Subway Status widgets should be included in the readout.

**Asana task**: https://app.asana.com/0/1185117109217413/1208591545042715/f

### DIFF
--- a/lib/screens/v2/widget_instance/alert.ex
+++ b/lib/screens/v2/widget_instance/alert.ex
@@ -7,6 +7,7 @@ defmodule Screens.V2.WidgetInstance.Alert do
   alias Screens.V2.WidgetInstance
   alias Screens.V2.WidgetInstance.Serializer.RoutePill
   alias ScreensConfig.Screen
+  alias ScreensConfig.V2.{Audio, BusShelter}
 
   defstruct screen: nil,
             alert: nil,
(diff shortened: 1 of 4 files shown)


lib/screens/v2/widget_instance/alert.ex
```diff
@@ -271,11 +272,12 @@ defmodule Screens.V2.WidgetInstance.Alert do
...
-  def audio_valid_candidate?(%__MODULE__{screen: %Screen{app_id: app_id}})
-      when app_id in ~w[bus_eink_v2 gl_eink_v2]a,
-      do: true
+  def audio_valid_candidate?(%__MODULE__{
+        screen: %Screen{app_params: %BusShelter{audio: %Audio{interval_enabled: true}}}
+      }),
+      do: false
 
-  def audio_valid_candidate?(_instance), do: false
+  def audio_valid_candidate?(_instance), do: true
```
PaulJKim, line 280: So does this mean now that any screen type that uses the Alert or Subway Status widgets will attempt an audio readout of them? Are there any implications of that we need to be concerned with or is that a non-issue?

digitalcora: This should be logically equivalent — the existing screen types that use the Alert widget are Bus E-ink, GL E-ink, and Bus Shelter, and the screen types that use Subway Status are the same plus PreFare. In other words the logic was previously written as "enable audio on [list of all screen types that use this widget except for Bus Shelter], disable it otherwise". I thought this would be better expressed by inverting the condition, making "Bus Shelter with periodic audio" the special case where we _don't_ read out, and otherwise enable it.

PaulJKim: I see. Makes sense!

PaulJKim (approved):
Looks good!

PR #2367, opened 2024-12-24
dev: fix module resolution issue with `util`

digitalcora:
We have a module `util` in a directory also named `util`. This happens to work fine for all current packages, but [Knip](https://knip.dev/) (which I'm interested in adding to our CI) is unable to resolve this specific module, complaining both that `util` is "unused" and that all references to `util` are "unlisted dependencies". This does not appear to be a bug in the library; it seems more likely our naming here is going against some convention of JS modules and doing something that isn't universally supported. In any case, it's easy to change.

[More information about Knip's module resolution](https://knip.dev/reference/faq#why-doesnt-knip-use-an-existing-module-resolver).


PaulJKim (approved):
Looks good to me. Seems like something handy for CI

PR #2369, opened 2024-12-24
dev: add Knip and address reported issues

digitalcora:
[Knip](https://knip.dev) is a tool that finds unused dependencies and exports in JS apps. This adds it to our CI checks and fixes or ignores all reported issues.

Because Knip resolves modules in a slightly different way than Webpack or TypeScript do natively, this required some tweaks to configuration.


assets/src/components/admin/admin_tables.tsx
```diff
@@ -417,7 +418,7 @@ const ElevatorV2ScreensTable = (): JSX.Element => {
             newRow.app_params.accessible_path_image_here_coordinates.x = value;
             return newRow;
           },
-          Cell: EditableCell,
+          Cell: EditableNumberInput,
```
PaulJKim, line 421: Is this a related change?

digitalcora: It's related in the sense that Knip reported `EditableNumberInput` as an unused export, and this caused me to realize there was actually a place we could be using it.

PaulJKim (approved):
neat! looks good to me

PR #2370, opened 2024-12-27
Add new U-turn arrow

PaulJKim:
**Asana task**: [Add "turn around" wayfinding arrow option](https://app.asana.com/0/1185117109217413/1208977593203657/f)

Description
Add the option to add a "turn around" arrow for current elevator closure widget. Went with `u` as in u-turn to represent that setting.


assets/src/components/v2/arrow.tsx
```diff
@@ -15,6 +15,7 @@ enum Direction {
   SW = "sw",
   W = "w",
   NW = "nw",
+  U = "u",
```
robbie-sundstrom, line 18: Either name `UTURN =  uturn` or add a comment here so that it's clear this `u` represents a u-turn. I think this would be unclear to someone looking at this without context

robbie-sundstrom (approved):
LGTM!

PR #2372, opened 2024-12-30
[Bugfix] Fix padding and locations of dividing lines for elevator closures

robbie-sundstrom:
**Asana task**: [Dividing Line Margin Issue on Elevator Screens](https://app.asana.com/0/1185117109217422/1209001244550628)

### Description
Fixes two issues with elevator closures separating lines on elevator screens, shown in the images below:
-  Padding: The dividing line for elevator closures had width set to 100%, which caused it to overflow to the right and appear on proceeding pages.
- Beyond the first page, there was no top line between the header and the closures.

image

image

- [ ] Tests added?

---
  - https://app.asana.com/0/0/1209061601128473

### DIFF
--- a/assets/css/v2/elevator/elevator_closures_list.scss
+++ b/assets/css/v2/elevator/elevator_closures_list.scss
@@ -65,7 +65,7 @@
 
           &:not(.current-station) {
             &::after,
-            &:first-child::before {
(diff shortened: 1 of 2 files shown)


PaulJKim (approved):
Looks good!

PR #2378, opened 2025-01-03
Ability for elevator wayfinding to have no arrow

robbie-sundstrom:
**Asana task**: [Elevator wayfinding: ability to have no arrow](https://app.asana.com/0/1208877354406742/1209001244550630/f)

No arrow will be shown when a direction isn't included in the wayfinding directions. Image of what this looks like below:
image

- [ ] Tests added?


assets/src/components/v2/elevator/current_elevator_closed.tsx
```diff
@@ -67,7 +67,9 @@ const CurrentElevatorClosed = ({
           <div className="subheading">Accessible Path</div>
           <div>
             <IsaNegative width={100} height={100} />
+            {accessiblePathDirectionArrow} ? (
             <Arrow direction={accessiblePathDirectionArrow} className="arrow" />
```
digitalcora, line 71: I think we should update

```ts
accessible_path_direction_arrow: Direction;
```

in the type definition above to instead be

```ts
accessible_path_direction_arrow: Direction | null;
```

to align with the reality of the API. That's ultimately why TS didn't raise an error on this line, forcing us to handle the possible `null`.

robbie-sundstrom: good callout! updated with this

assets/src/components/v2/elevator/current_elevator_closed.tsx
```diff
@@ -67,7 +67,12 @@ const CurrentElevatorClosed = ({
           <div className="subheading">Accessible Path</div>
           <div>
             <IsaNegative width={100} height={100} />
-            <Arrow direction={accessiblePathDirectionArrow} className="arrow" />
+            {accessiblePathDirectionArrow} ? (
+            <Arrow
+              direction={accessiblePathDirectionArrow!}
```
digitalcora, line 72: This should be:

```tsx
{accessiblePathDirectionArrow ? (
  <Arrow
    direction={accessiblePathDirectionArrow}
    className="arrow"
  />
) : null}
```

As-is, this compiles to

* the value of `accessiblePathDirectionArrow` as text
* the literal text `? (`
* the `<Arrow ... />` component
* the literal text `) : ()`

Screenshot 2025-01-03 at 2 28 16 PM

That's why this `!` non-null assertion seemed to be required — the "conditional branch" where the value was non-null didn't actually exist. (Some hints to this: the lack of indentation when formatted, `?` and `:` not being highlighted as operators.)

robbie-sundstrom: 🤦 thanks for catching! I should've retested after updating this code.

digitalcora (approved):
🚫 ➡️ ✅

PR #2379, opened 2025-01-03
refactor: introduce structs for departure sections

digitalcora:
Sections passed to the `Departures` widget for serialization were previously plain maps with a `type` field. In addition to lacking any compile-time verification of fields or other struct conveniences, this had some confusing overlap with the serialized form of sections, which are maps with a `type` field (of the same value, but with the rest of the fields being different!).

* Introduce `Departures.XSection` structs for each type of section the widget supports.

* Remove the plain-map indirection of "notice rows". Instead of `%{text: FreeTextLine.t()}` they are just `FreeTextLine.t()`.

* Does not remove some existing instances of `%{type: :x_section}` in the DUP departures candidate generator; these are part of an "internal API" of the module and are neither real Departures widget sections nor the serialized form.


lib/screens/v2/candidate_generator/dup/departures.ex
```diff
@@ -127,16 +134,22 @@ defmodule Screens.V2.CandidateGenerator.Dup.Departures do
...
   end
 
-  defp get_departure_instance_for_section(%{type: :no_data_section} = section, _, _, _, _),
-    do: section
+  defp get_departure_instance_for_section(
+         %{type: :no_data_section, route: route},
+         _is_only_section,
+         _now,
+         _fetch_schedules_fn,
+         _fetch_vehicles_fn
```
robbie-sundstrom, line 150: thanks for adding the variable names even though they're underscored!

PR #2393, opened 2025-01-10
refactor: new V3 API caching approach

digitalcora:
Investigating high memory usage in production suggested the V3 API connection pool was a major factor: each open connection consumes about 5MB, and with a pool size of 100, the pool could consume close to 500MB during peak usage, which is half the memory allotted to our instances. Add the 250MB limit on our API response cache and it becomes clear how we could start cutting things close.

Reducing the pool size would be an obvious step to take; but then we _do_ use all of those 100 connections (and still want more!) during peak times. If we could reduce the number of concurrent connections we need, then we could also reduce the pool size without impacting screen performance/stability.

To that end, this change:

* Reduces the V3 API connection pool size to 50 (the default). This should give us ~250MB of overhead.

* Splits API response caching into `Static` and `Realtime` buckets, based on which resources are included in the request, each with its own memory limit and GC settings. The static cache gets 150MB and the realtime cache 50MB, leaving some overhead from the ~250MB previously allotted to the shared cache. (Keeping in mind caches may exceed their memory limit for up to `gc_cleanup_min_timeout`, which defaults to 10 seconds.)

* Adds a layer of "unconditional" response caching: for a period after a response is cached (different for static vs. realtime), it is served from the cache without making an HTTP call to check whether it is outdated. Experimentally this allows ~25% of realtime data requests, and 80-99% of static data requests (depending on time of day), to be served without checking out an HTTP connection at all. This is what theoretically allows reducing the connection pool size without harming performance.

* Adds telemetry for cache statistics.

**Asana task**: https://app.asana.com/0/1185117109217413/1208614347699410

### DIFF
--- a/config/config.exs
+++ b/config/config.exs
@@ -471,13 +471,26 @@ config :screens, :screens_by_alert,
   screens_last_updated_ttl_seconds: 3600,
   screens_ttl_seconds: 40
 
-config :screens, Screens.ScreenApiResponseCache,
-  gc_interval: :timer.hours(1),
-  allocated_memory: 250_000_000
-
 config :screens, Screens.LastTrip,
   trip_updates_adapter: Screens.LastTrip.TripUpdates.GTFS,
   vehicle_positions_adapter: Screens.LastTrip.VehiclePositions.GTFS
+
+# Memory limits for V3 API response caches are based on ETS table memory usage measurements in a
+# deployed environment. To avoid thrashing and overloading the HTTP connection pool, these can and
+# should be tweaked as the V3 API usage patterns and requirements of the app change over time.
+
+config :screens, Screens.V3Api.Cache.Realtime,
+  allocated_memory: 50 * 1024 * 1024,
+  gc_cleanup_min_timeout: :timer.seconds(5),
+  gc_interval: :timer.seconds(30),
+  stats: true,
+  telemetry_prefix: ~w[screens v3_api cache]a
+
+config :screens, Screens.V3Api.Cache.Static,
+  allocated_memory: 150 * 1024 * 1024,
+  gc_interval: :timer.hours(1),
+  stats: true,
+  telemetry_prefix: ~w[screens v3_api cache]a
 
 # Import environment specific config. This must remain at the bottom
 # of this file so it overrides the configuration defined above.
--- a/lib/screens/application.ex
+++ b/lib/screens/application.ex
@@ -10,10 +10,9 @@ defmodule Screens.Application do
 
     # List all child processes to be supervised
     children = [
-      Screens.Telemetry,
       {Screens.Cache.Owner, engine_module: Screens.Config.Cache.Engine},
       {Screens.Cache.Owner, engine_module: Screens.SignsUiConfig.Cache.Engine},
-      :hackney_pool.child_spec(:api_v3_pool, max_connections: 100),
+      :hackney_pool.child_spec(:api_v3_pool, max_connections: 50),
       # Task supervisor for ScreensByAlert async updates
       {Task.Supervisor, name: Screens.ScreensByAlert.Memcache.TaskSupervisor},
       # ScreensByAlert server process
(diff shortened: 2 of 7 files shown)


config/config.exs
```diff
@@ -471,9 +471,18 @@ config :screens, :screens_by_alert,
...
-config :screens, Screens.ScreenApiResponseCache,
+config :screens, Screens.V3Api.Cache.Realtime,
+  allocated_memory: 50 * 1024 * 1024,
+  gc_cleanup_min_timeout: :timer.seconds(5),
+  gc_interval: :timer.seconds(30),
+  stats: true,
+  telemetry_prefix: ~w[screens v3_api cache]a
+
+config :screens, Screens.V3Api.Cache.Static,
+  allocated_memory: 150 * 1024 * 1024,
```
robbie-sundstrom, line 482: Could be helpful to add comments with reasoning to these config options so people can understand later (and now). Particularly the `allocated_memory` being three times as high for static vs. realtime caching is unclear.

lib/screens/telemetry.ex
```diff
@@ -50,7 +53,17 @@ defmodule Screens.Telemetry do
       :ok = :telemetry.attach_many(name, event_names, &Screens.Telemetry.handle_event/4, config)
     end
 
-    :ignore
+    children = [
+      {
+        :telemetry_poller,
+        measurements: [
+          {Screens.V3Api.Cache.Realtime, :dispatch_stats, []},
+          {Screens.V3Api.Cache.Static, :dispatch_stats, []}
```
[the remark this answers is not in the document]
digitalcora: Yep, here is an example log this produces:

```
[info] event=screens.v3_api.cache.stats cache=Screens.V3Api.Cache.Static hits=450 misses=506 writes=506 updates=0 evictions=0 expirations=0
```

lib/screens/application.ex
```diff
@@ -10,10 +10,9 @@ defmodule Screens.Application do
 
     # List all child processes to be supervised
     children = [
-      Screens.Telemetry,
       {Screens.Cache.Owner, engine_module: Screens.Config.Cache.Engine},
       {Screens.Cache.Owner, engine_module: Screens.SignsUiConfig.Cache.Engine},
-      :hackney_pool.child_spec(:api_v3_pool, max_connections: 100),
+      :hackney_pool.child_spec(:api_v3_pool, []),
```
robbie-sundstrom, line 15: I know `max_connections: 50` is the default, but I'd prefer it left in so that it's clear that it's intentional without looking through commit history for the line. 
Personal preference so I welcome disagreement

lib/screens/v3_api/cache.ex
```diff
@@ -0,0 +1,60 @@
+defmodule Screens.V3Api.Cache do
```
robbie-sundstrom, line 1: This whole module is really understandable with the structure of realtime vs. static and the comments to explain reasoning!!

lib/screens/application.ex
digitalcora: Makes sense!

lib/screens/telemetry.ex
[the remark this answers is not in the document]
robbie-sundstrom: hmmm it being cumulative is a little annoying, but I don't see any way around that based on the docs. Still seems helpful to leave on!

config/config.exs
digitalcora: I can try to write something useful here, but it was initially pretty vibes-based and then checked by looking at the real memory usage of the ETS tables in development. The static cache ended up at 130-140MB and the realtime at 15-30MB. The main reason I enabled the built-in stats telemetry was so we'd have some way of knowing whether we're evicting too eagerly in production and thus whether we should tweak these numbers. I wish there was a telemetry event emitted when a new generation is created since that would be much more obvious, but it seems like adapter-specific details like that don't really have telemetry.

robbie-sundstrom: Gotcha, that reasoning makes sense and I see how that's hard to fully put in a comment. Could maybe just add a comment that these values are based on initial testing and should be adjusted as we learn more in the future.

panentheos (approved):
Looks great! I'm excited to see how this does in prod.

robbie-sundstrom (approved):
Awesome work!

PR #2402, opened 2025-01-14
Dynamic paging intervals for elevator screens

robbie-sundstrom:
**Asana task**: [Set up dynamic paging intervals for elevator screens](https://app.asana.com/0/1208877354406742/1208811831018417)

Description

- [ ] Tests added?


assets/src/components/v2/persistent_wrapper.tsx
```diff
@@ -37,7 +37,7 @@ const PersistentWrapper: ComponentType<Props> = ({
       {...visibleData}
       onFinish={handleFinished}
       key={renderKey}
-      lastUpdate={lastFetch}
+      lastUpdate={lastFetch} // make this optional - either determined by lastFetch Context or by other interval
```
robbie-sundstrom, line 40: This was something I originally talked about with Cora, but there isn't really an equivalent for `lastUpdate` to trigger the hook in wrappedComponent. Not sure if this wrapper makes as much sense anymore to use for the elevator screen components

assets/src/hooks/v2/use_page_advancer.tsx
```diff
@@ -0,0 +1,77 @@
...
+  // Handle refresh-based advancement
+  useEffect(() => {
+    if (advanceOnDataRefresh && lastUpdate !== null) {
+      advancePage();
+    }
+  }, [lastUpdate]);
+
+  useEffect(() => {
+    if (pageIndex === 0) onFinish();
+  }, [pageIndex]);
```
robbie-sundstrom, line 72: This needs to be a separate hook relying on `pageIndex` because you cannot modify state within a component (ComponentWrapper) from a callback function in a different component. However, you can do this from a hook. So even though this seems like it would make more sense to be done in `advancePage`, it needs to be in its own hook.

digitalcora: In the abstract I think it does make sense even to use for paging components that don't trigger paging on data refresh — to avoid confusion or the appearance of a "glitchy" display we want them to get through all their pages before updated data potentially changes the arrangement of items or number of pages. For the concrete case of elevator screens, this data doesn't change much moment-to-moment, so maybe it's not that essential. Wrapped components are free to ignore / not match on this prop, of course, so in that sense it is already optional.

assets/src/components/v2/persistent_wrapper.tsx
```diff
@@ -3,7 +3,7 @@ import { LastFetchContext } from "Components/v2/screen_container";
 
 interface WrappedComponentProps {
   onFinish: () => void;
-  lastUpdate: number | null;
+  lastUpdate?: number | null;
```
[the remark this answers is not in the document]
robbie-sundstrom: agreed this should not have both undefined and null. I think null makes this easier, but I had the undefined/optional leftover from my TODO. Will clean up

assets/src/hooks/v2/use_page_advancer.tsx
```diff
@@ -0,0 +1,77 @@
...
+    if (cycleIntervalMs) {
+      intervalRef.current = setInterval(() => {
+        advancePage();
+      }, cycleIntervalMs);
+    }
+  }, [cycleIntervalMs, advancePage]);
+
+  // Cleanup the timer interval when the component unmounts or dependencies change
+  useEffect(() => {
+    if (!advanceOnDataRefresh && cycleIntervalMs) {
```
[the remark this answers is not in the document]
robbie-sundstrom: This hook did end up being more confusing than I was expecting to get working for both cases within the same hook. It would likely be quite a bit simpler split out into two hooks, which I think makes up for the ease of use of having one hook for all paging. 

I think going back to this approach makes sense!

assets/src/hooks/v2/use_interval_paging.tsx
```diff
@@ -0,0 +1,64 @@
...
+  cycleIntervalMs,
+  onFinish,
+}: UseIntervalPagingProps) {
+  const [pageIndex, setPageIndex] = useState(0);
+  // Use refs to keep stable references to numPages and interval without triggering re-renders
+  const intervalRef = useRef<NodeJS.Timeout | null>(null);
+  const numPagesRef = useRef(numPages);
+
+  useEffect(() => {
+    numPagesRef.current = numPages;
```
panentheos, line 24: Based on the current usage, it looks like we don't expect this value to change over the lifecycle of the component, so we could probably just use the argument directly without storing it in a ref.

assets/src/hooks/v2/use_interval_paging.tsx
```diff
@@ -0,0 +1,64 @@
...
+  }, [numPages]);
+
+  // Callback that handles changing the state of pageIndex
+  const advancePage = useCallback(() => {
```
panentheos, line 28: nit: Since `advancePage` and `startTimer` are only used once, we could inline them and avoid the ceremony of `useCallback`.

assets/src/hooks/v2/use_interval_paging.tsx
```diff
@@ -0,0 +1,64 @@
+import { useState, useEffect, useRef, useCallback } from "react";
+
+interface UseIntervalPagingProps {
+  numPages: number; // Total number of pages to cycle through
+  cycleIntervalMs?: number; // In milliseconds, the interval for cycling pages (only used if advanceOnDataRefresh is false)
```
panentheos, line 5: Currently this value is always present, and if we're okay with making it non-optional then we could remove a bunch of conditional logic below.

assets/src/hooks/v2/use_interval_paging.tsx
```diff
@@ -0,0 +1,64 @@
...
+  // Callback that handles changing the state of pageIndex
+  const advancePage = useCallback(() => {
+    setPageIndex((prevIndex) => {
+      return (prevIndex + 1) % numPagesRef.current;
+    });
+  }, []);
+
+  // Start the interval for time-based advancement
+  const startTimer = useCallback(() => {
+    if (intervalRef.current) clearInterval(intervalRef.current); // Clear any existing timer
```
panentheos, line 36: As long as we return a cleanup function from `useEffect` below, the interval should get cleared any time the effect needs to be re-run, so we shouldn't need to do that here as well.

assets/src/hooks/v2/use_interval_paging.tsx
```diff
@@ -0,0 +1,64 @@
...
+        if (intervalRef.current) clearInterval(intervalRef.current);
+      };
+    }
+    // Hooks must return void or a cleanup function,
+    // so we explicitly return undefined when no cleanup is necessary.
+    return undefined;
+  }, [cycleIntervalMs, startTimer]);
+
+  useEffect(() => {
+    if (pageIndex === 0) onFinish();
```
panentheos, line 58: This looks like it will call `onFinish` immediately when the component mounts since `pageIndex` is initialized to 0. That may be fine, but just calling it out.

assets/src/hooks/v2/use_interval_paging.tsx
panentheos: Actually, looking at this some more, I'm less sure about this. If this does change, what's the expected behavior?

assets/src/hooks/v2/use_interval_paging.tsx
robbie-sundstrom: This is overcomplicated from when I had the interval and refresh hooks combined and more logic was required. Good point that this shouldn't be needed anymore

assets/src/hooks/v2/use_interval_paging.tsx
robbie-sundstrom: `numPages` could change over the lifecycle of a component. We won't update the data being displayed in the component though until we reach page 0 after a refresh has been triggered during paging. This ref was necessary with the old functionality of a hook handling both refresh/interval paging, but I think I can get this working without a ref now.

assets/src/components/v2/persistent_wrapper.tsx
```diff
@@ -27,16 +26,14 @@ const PersistentWrapper: ComponentType<Props> = ({
   useEffect(() => {
     if (isFinished) {
       setVisibleData(data);
-      setRenderKey((n) => n + 1);
       setIsFinished(false);
     }
-  }, [lastFetch]);
+  }, [isFinished]);
```
robbie-sundstrom, line 31: Triggering the data refresh on `isFinished` instead of `lastFetch` allows this Wrapper to work the same way for both interval and refresh based pagination.

assets/src/components/v2/persistent_wrapper.tsx
```diff
@@ -27,16 +26,14 @@ const PersistentWrapper: ComponentType<Props> = ({
...
 
   return (
     <WrappedComponent
       {...visibleData}
       onFinish={handleFinished}
-      key={renderKey}
       lastUpdate={lastFetch}
```
digitalcora, line 37: Since in the new version this value is not used at all _within_ the wrapper component itself, it feels like it doesn't need to be included here at all. i.o.w. the "wrapped" component can pull in this value and use it directly or pass it to whatever actually needs it, _if_ anything needs it.

assets/src/hooks/v2/use_interval_paging.tsx
```diff
@@ -0,0 +1,57 @@
+import { useState, useEffect, useRef } from "react";
+
+interface UseIntervalPagingProps {
+  numPages: number; // Total number of pages to cycle through
+  cycleIntervalMs: number; // In milliseconds, the interval for cycling pagesß
```
digitalcora, line 5: `ß`?

assets/src/hooks/v2/use_interval_paging.tsx
```diff
@@ -0,0 +1,57 @@
...
+    // Cleanup interval on unmount
+    return () => {
+      if (intervalRef.current) {
+        clearInterval(intervalRef.current);
+      }
+    };
+  }, [cycleIntervalMs, numPages]);
+
+  // Handles calling the onFinish function in persistentWrapper.
+  // This needs to be done in a separate hook outside of the setInterval callback.
```
digitalcora, line 47: Why is this? Similar to the comment about `isFinished` in the persistent wrapper, it seems like we should be able to handle this by directly calling functions rather than setting internal bits of state that only exist to trigger effects.

assets/src/components/v2/persistent_wrapper.tsx
robbie-sundstrom: Nothing else actually needs this value! Now that page updates are handled purely within each hook, it's not necessary. To me, this makes more sense if hooks handle pagination, and this wrapper component is purely for persisting data and is called when pagination is complete

assets/src/components/v2/persistent_wrapper.tsx
robbie-sundstrom: This is a great callout! That's even simpler and I think should work

assets/src/hooks/v2/use_interval_paging.tsx
```diff
@@ -0,0 +1,57 @@
...
+function useIntervalPaging({
+  numPages,
+  cycleIntervalMs,
+  onFinish,
+}: UseIntervalPagingProps) {
+  const [pageIndex, setPageIndex] = useState(0);
+  const [isFinished, setIsFinished] = useState(false);
+
+  // Use ref to keep stable references to timer interval without triggering re-renders
+  const intervalRef = useRef<NodeJS.Timeout | null>(null);
```
digitalcora, line 22: This seems like the wrong type — we are not running on `NodeJS` in the browser. According to the docs for `setInterval` the return value is an integer ID, so the appropriate type here might just be `number`. (the `| null` is also not necessary — the type parameter of `useRef` is the type of the reference when it has a value)

assets/src/hooks/v2/use_interval_paging.tsx
robbie-sundstrom: The problem is that you can't update the state of persistent wrapper within the callback function of `setInterval`. This is a weird piece of React I wasn't familiar with. I got around this by having this boolean flag `isFinished` that serves only to trigger this hook.

assets/src/hooks/v2/use_interval_paging.tsx
robbie-sundstrom: whoops that's me misclicking a command-S to save

assets/src/hooks/v2/use_interval_paging.tsx
robbie-sundstrom: I actually did originally try this as an integer, but the `Timeout` value is set by the `@types/node` package. I think it's in browser environments that the return type is just a number. 

I think you're right on the point about null in Refs! Just testing now with the null initialization and typing removed

assets/src/hooks/v2/use_interval_paging.tsx
robbie-sundstrom: this was correct! Still getting familiar with React hooks, so I didn't realize this triggered on initialization but you were totally right. Modified this to avoid this issue

assets/src/hooks/v2/use_interval_paging.tsx
robbie-sundstrom: I updated the naming and comment here for more clarity.

assets/src/hooks/v2/use_interval_paging.tsx
digitalcora: Poking around it seems like this is a known issue with having `@types/node` installed in the same environment where you're targeting browsers, as the compiler doesn't know "which" `setInterval` you're referring to. Possible workarounds to avoid explicitly depending on the NodeJS type are:

* use `window.setInterval` (this is unambiguously the browser one, in NodeJS would be `global.setInterval`)
* specify the type as `ReturnType<typeof setInterval>`. when `@types/node` is installed this would ultimately resolve to `NodeJS.Timeout`, but when not, it wouldn't, so there's no dependency on it.

I don't have strong opinions about any of these alternatives (including leaving this as it is), so up to you!

assets/src/components/v2/elevator/current_elevator_closed.tsx
```diff
@@ -39,11 +40,15 @@ const CurrentElevatorClosed = ({
...
   accessible_path_image_here_coordinates: accessiblePathImageHereCoordinates,
-  onFinish,
-  lastUpdate,
+  updateVisibleData,
 }: Props) => {
   const numPages = accessiblePathImageUrl ? 2 : 1;
-  const pageIndex = useClientPaging({ numPages, onFinish, lastUpdate });
+  const pageIndex = useIntervalPaging({
+    numPages,
+    cycleIntervalMs: CURRENT_CLOSED_PAGING_INTERVAL_MS, // 12 seconds
```
digitalcora, line 48: Leftover comment from when this was directly `12000`?

assets/src/components/v2/persistent_wrapper.tsx
digitalcora: Resurfacing this since it doesn't seem to have been changed yet. Or were you suggesting it should stay here?

assets/src/components/v2/persistent_wrapper.tsx
```diff
@@ -50,7 +35,7 @@ const PersistentWrapper: ComponentType<Props> = ({
  * defining your `WrappedComponent`'s prop types.
  *
  * `WrappedComponent` must expect the following props:
- * - `onFinish`: A callback that `WrappedComponent` can call to indicate that it is ready to receive
+ * - `updateVisibleData`: A callback that `WrappedComponent` can call to indicate that it is ready to receive
```
digitalcora, line 38: Seems like we could revisit the wording of this documentation since it's not really specific to data refreshes anymore.

assets/src/hooks/v2/use_interval_paging.tsx
```diff
@@ -0,0 +1,59 @@
+import { useState, useEffect, useRef } from "react";
+
+interface UseIntervalPagingProps {
+  numPages: number; // Total number of pages to cycle through
+  cycleIntervalMs: number; // In milliseconds, the interval for cycling pages
```
digitalcora, line 5: "cycle interval" feels potentially confusing — without context I'd assume the "cycle" was the full set of pages, so maybe this would be the time the full "cycle" should take. It might actually be clearer as just `intervalMs`.

assets/src/hooks/v2/use_interval_paging.tsx
robbie-sundstrom: I updated this to `window.setInterval` so the number type can be used now, which also makes this state much easier to track than when it was a `NodeJS.Timeout` object

assets/src/components/v2/persistent_wrapper.tsx
robbie-sundstrom: Updated the documentation here, but not sure if I love my updated wording 

```
 * - `updateVisibleData`: A callback that `WrappedComponent` can call to refresh the data displayed by the child component
 *   with the most recently received data.
 * - `lastUpdate`: a **nullable** timestamp of the last successful data refresh, which `WrappedComponent` should use to
 *   sync its re-renders with the rest of the page. Used by components that use data refreshes to handle pagination.
```

assets/src/components/v2/persistent_wrapper.tsx
robbie-sundstrom: I was suggesting it should be removed, which it has been! I think we're agreeing here if I understand correctly? `renderKey` isn't being used anymore and is fully removed form the `WrappedComponent`

digitalcora: The line I'm commenting on is `lastUpdate`, which is the "this value" I was originally referring to. I guess that might not have been super clear. 😅

robbie-sundstrom: ahh that value isn't being used by the elevator screens anymore, but it is still used by the `PersistentCarousel` objects. Since they both use the same `PersistentWrapper`, I left it in even though it's optional. I don't see a good way to make including the `lastFetch` conditional without breaking it out into a separate Wrapper (which might be the best option)

robbie-sundstrom: actually testing out another idea here with optional props that would not require two different wrappers

digitalcora (approved):
📄 📄 📄 ✨

PR #2407, opened 2025-01-16
fix: show only active closures on elevator screens

digitalcora:
**Asana task**: https://app.asana.com/0/1185117109217413/1209154878642792


robbie-sundstrom (approved):
nice quick fix!

PR #2409, opened 2025-01-17
Add new arrow types for elevator screens

robbie-sundstrom:
**Asana task**: [Fix wayfinding Arrow sizing on Elevator Screens](Fix wayfinding Arrow sizing on Elevator Screens)

image
image

- [ ] Tests added?

### DIFF
--- a/assets/src/components/v2/arrow.tsx
+++ b/assets/src/components/v2/arrow.tsx
@@ -18,25 +18,42 @@ enum Direction {
   UTURN = "uturn",
 }
 
+enum ScreenType {
+  ELEVATOR = "elevator",
+}
+
 interface ArrowDescriptor {
   imageName: string;
   className: string;
 }
 
-const directionToArrow = (direction: Direction): ArrowDescriptor => {
+/**
+ * Returns an ArrowDescriptor containing an arrow SVG and CSS class that determines arrow rotation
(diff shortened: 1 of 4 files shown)


assets/src/components/v2/arrow.tsx
```diff
@@ -18,25 +18,42 @@ enum Direction {
   UTURN = "uturn",
 }
 
+enum ScreenType {
+  ELEVATOR = "elevator",
+}
+
```
robbie-sundstrom, line 24: I don't love this enum with only one value. If anyone knows of a good substitute for this sort of screen type that exists in TS, I couldn't find one

digitalcora: I'm not sure if we really need one — IMO "how does this arrow look" is a very low-level detail and "what type of screen is this" a very high-level one, and tying them together like this doesn't seem optimal. What is different about the arrow graphic itself? Does it have a different weight? A different viewbox/alignment? Maybe we could capture that difference in the naming of the option.

robbie-sundstrom: I like this framing! Much more intuitive. Updated the arrow type to be based on line weight.

digitalcora (approved):
↖️ ➡️ ⬇️ ✅

PR #2413, opened 2025-01-23
refactor: elevator closure header/footer handling

digitalcora:
Some refactoring on the way to future elevator closures.

* Rather than the main Elevator screen candidate generator creating the header and footer widgets and the Closures generator "modifying" them with variants as appropriate, now the Closures generator alone is responsible. This somewhat simplifies the implementation and testing on both ends.

* Tests for the Closures generator were more verbose and testing more than they strictly needed to be (for example, that the correct header and footer were present in every case). These are now cut down to only the relevant and unique details.

### DIFF
--- a/lib/screens/v2/candidate_generator/elevator.ex
+++ b/lib/screens/v2/candidate_generator/elevator.ex
@@ -5,12 +5,15 @@ defmodule Screens.V2.CandidateGenerator.Elevator do
   alias Screens.V2.CandidateGenerator.Elevator.Closures, as: ElevatorClosures
   alias Screens.V2.CandidateGenerator.Widgets.Evergreen
   alias Screens.V2.Template.Builder
-  alias Screens.V2.WidgetInstance.{Footer, NormalHeader}
-  alias ScreensConfig.Screen
-  alias ScreensConfig.V2.Elevator
 
   @behaviour CandidateGenerator
 
+  @instance_fns [
+    &ElevatorClosures.elevator_status_instances/2,
+    &Evergreen.evergreen_content_instances/2
+  ]
+
+  @impl true
   def screen_template do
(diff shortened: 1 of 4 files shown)


lib/screens/v2/candidate_generator/elevator.ex
```diff
@@ -26,29 +29,11 @@ defmodule Screens.V2.CandidateGenerator.Elevator do
...
-      elevator_closure_instances_fn.(
-        config,
-        header_instance(config, now),
-        footer_instance(config)
-      ),
-      evergreen_content_instances_fn.(config, now)
-    ])
+  @impl true
+  def candidate_instances(config, now \\ DateTime.utc_now(), instance_fns \\ @instance_fns) do
+    instance_fns |> Enum.map(& &1.(config, now)) |> Enum.concat()
```
robbie-sundstrom, line 34: I'm seeing an error happening on this line
`expected a function, got: {:elevator_status, &Screens.V2.CandidateGenerator.Elevator.Closures.elevator_status_instances/2}`

If I'm understanding correctly, it seems like this is happening because `instance_fns` is actually a list of tuples rather than just a list of functions.

robbie-sundstrom (changes requested):

lib/screens/v2/candidate_generator/elevator.ex
digitalcora: The perils of mocking... 😬

robbie-sundstrom: extremely relatable lol

robbie-sundstrom (approved):
Love this refactor!!! Having the header and footer into the widget definitely makes more sense 🚢

PR #2415, opened 2025-01-24
feat: support evergreen content on Sectionals

digitalcora:
It was possible to *configure* evergreen content on this screen type, but nothing actually looked at this configuration.

Other changes:

* Delete some old logic for generating placeholder widgets that would never be displayed.

* Simplify how candidate generator functions are specified, along the same lines as #2413.

* Introduce common styles for evergreen content on `bus(way|_shelter)` screens, and remove some redundant specification of pixel sizes that could instead use `100%` of the container's width/height.

**Asana task**: https://app.asana.com/0/1185117109217413/1209231912135278


lib/screens/v2/candidate_generator/widgets/evergreen.ex
```diff
@@ -5,13 +5,11 @@ defmodule Screens.V2.CandidateGenerator.Widgets.Evergreen do
...
   alias ScreensConfig.Screen
   alias ScreensConfig.V2.EvergreenContentItem
-  alias ScreensConfig.V2.{BusEink, BusShelter, Dup, Elevator, GlEink, PreFare}
 
   def evergreen_content_instances(
-        %Screen{app_params: %app{evergreen_content: evergreen_content}} = config,
+        %Screen{app_params: %_app{evergreen_content: evergreen_content}} = config,
         now \\ DateTime.utc_now()
-      )
-      when app in [BusEink, BusShelter, Dup, Elevator, GlEink, PreFare] do
```
robbie-sundstrom, line 14: This conditional doesn't seem like it was ever necessary right? It would never generate this widget for Sectionals, and `evergreen_content_instances` would only be called for these 6 screen types anyway

Mostly thinking out loud with this comment 😅

robbie-sundstrom (approved):
Code looks good to me! I haven't tested b/c I seem to have messed up my local environment on Friday. Assuming you've tested though, I don't want to hold you up.

lib/screens/v2/candidate_generator/widgets/evergreen.ex
digitalcora: Correct. It seems like perhaps more of a note/signal for developers that this is the set of screen types we've fully implemented evergreen content on (frontend component mapping and styles). Though IMO it's very unlikely we'd throw a call to generate evergreen widgets into a new screen type and, like, not test at all that it works end-to-end, so this comes across as overly-defensive programming to me.

PR #2416, opened 2025-01-24
fix: inspector showing `NaN` audio interval

digitalcora:
When a screen normally capable of periodic audio readouts (Bus Shelter) had them disabled, this was incorrectly loaded as an interval of `NaN` minutes, rather than no interval at all. This basically worked anyway on the screens themselves, but showed up oddly in the Inspector.

Screenshot 2025-01-24 at 5 03 06 PM


robbie-sundstrom (approved):
nice

PR #2419, opened 2025-01-27
refactor: parse elevator entering redundancy

digitalcora:
Previously the logic for elevator redundancy only cared about "exiting redundancy", so that was the only column of the spreadsheet we loaded. For the future elevator closures feature, we'll also have logic that depends on "entering redundancy", so both categories should be loaded and distinguishable.


PaulJKim (approved):
Looks good!

PR #2421, opened 2025-01-28
refactor: rework naming of elevator screen widgets

digitalcora:
Goals:

* Align naming of the "this elevator is working" widget with the idea that it isn't _only_ a "list" of closures. (This already wasn't the case, since this widget incorporates the "status bar" for elevators at the current station, but it will be even more so with the introduction of future elevator closure info.)

* Clean up duplicative "elevator" naming. When a module is already in an `elevator` namespace, it does not itself need to include `elevator` in its name.

I'd like feedback / possible alternatives for these names — I don't love them (in particular how the word _"closures"_ is getting thrown around a lot), I just think it's incrementally better than before. Part of the problem is that we also have the entirely-separate-yet-similar-in-function "elevator _status_" widget for Pre-Fare screens, so using the word "status" here could cause confusion.

### DIFF
--- a/assets/src/apps/v2/elevator.tsx
+++ b/assets/src/apps/v2/elevator.tsx
@@ -15,8 +15,8 @@ import EvergreenContent from "Components/v2/evergreen_content";
 import ScreenPage from "Components/v2/screen_page";
 import { MappingContext } from "Components/v2/widget";
 import MultiScreenPage from "Components/v2/multi_screen_page";
-import ElevatorClosuresList from "Components/v2/elevator/elevator_closures_list";
-import CurrentElevatorClosed from "Components/v2/elevator/current_elevator_closed";
+import Closures from "Components/v2/elevator/closures";
+import AlternatePath from "Components/v2/elevator/alternate_path";
 import SimulationScreenPage from "Components/v2/simulation_screen_page";
 import Footer from "Components/v2/elevator/footer";
 import NormalHeader from "Components/v2/normal_header";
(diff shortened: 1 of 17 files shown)


assets/src/apps/v2/elevator.tsx
```diff
@@ -25,8 +25,8 @@ import NoData from "Components/v2/elevator/no_data";
 const TYPE_TO_COMPONENT = {
   normal: NormalScreen,
   takeover: TakeoverScreen,
-  elevator_closures_list: ElevatorClosuresList,
-  current_elevator_closed: CurrentElevatorClosed,
+  elevator_closures: ElevatorClosures,
+  elevator_closed: ElevatorClosed,
```
robbie-sundstrom, line 29: An alternative could be something like `UnavailableElevators` for the list of all other elevators that are closed? That makes the `ElevatorClosed` component seem clearer and more distinct to me. 

Also like the word outages, could be something like `AllElevatorOutages`, or even just `AllElevatorClosures`. I think that naming could then be confusing in the future then if we do narrow down which closures are shown at each elevator

robbie-sundstrom: To follow up, I do really like the spirit of this PR, but agree that `ElevatorClosures` and `ElevatorClosed` are too similar. 24 hours to let it marinate though, and my new idea is incorporating that when the current elevator is closed, we include the alternate route map. Could be `ElevatorAlternateRoute` and `ElevatorClosures`

digitalcora: I went with `ElevatorAlternatePath` to avoid semantic overlap with `Route` as in a GTFS route, but otherwise took that suggestion. What do you think?

robbie-sundstrom: looking through the code with these changes, I really like it! Will be much clearer in the future.

PR #2424, opened 2025-01-31
On bus screen placeholder app

robbie-sundstrom:
**Asana task**: [[add-on] "hello world" version of on-bus screens](https://app.asana.com/0/1208877354406742/1209144532576373/f)

Description

- [ ] Tests added?


mix.exs
```diff
@@ -94,7 +94,7 @@ defmodule Screens.MixProject do
       {:telemetry_metrics, "~> 1.0"},
       {:screens_config,
        git: "https://github.com/mbta/screens-config-lib.git",
-       ref: "7dbb2b9b3da0154ca9128c7c2a4ab30d18501290"},
+       ref: "3e781e39e4dd8bd63a05825534847347e1cbffa9"},
```
robbie-sundstrom, line 97: Noting that this will need to be updated after merging `screens-config-lib` change into main

assets/src/components/v2/placeholder.tsx
```diff
@@ -7,6 +7,7 @@ interface Props {
 }
 
 const Placeholder: React.ComponentType<Props> = ({ color }) => {
+  console.log(classWithModifier("placeholder", color));
```
deanshi, line 10: Quick question(s) (sorry):

Are there rules on how we should be handling console logs as a team? Do we generally remove them from our code or is there a linter that shakes its fist at us when we keep them in? Or are there places we keep console logs for visibility/for people checking out inspect?

digitalcora: This seems like accidental leftover debugging — if there was something we wanted to log in production we'd probably use our Sentry utilities instead, since on real screens the browser console would be pretty difficult to access. Based on this PR passing CI, there is not currently a lint that warns about this, but I'm sure there is one [somewhere](https://eslint.org/docs/latest/rules/no-console) in our existing tools.

assets/css/v2/on_bus/body/normal.scss
```diff
@@ -0,0 +1,5 @@
+.body-normal {
+  position: relative;
+  width: 650px;
+  height: 432px;
```
digitalcora, line 4: Any reason we need to specify the overall width/height twice? (here and on `screen-container`)

assets/src/apps/v2/on_bus.tsx
```diff
@@ -0,0 +1,50 @@
...
+import ReactDOM from "react-dom";
+import { BrowserRouter as Router, Route, Switch } from "react-router-dom";
+import ScreenPage from "Components/v2/screen_page";
+import { MappingContext } from "Components/v2/widget";
+import MultiScreenPage from "Components/v2/multi_screen_page";
+import SimulationScreenPage from "Components/v2/simulation_screen_page";
+import Placeholder from "Components/v2/placeholder";
+
+const TYPE_TO_COMPONENT = {
+  body: Placeholder,
```
digitalcora, line 19: I'm a little confused about how this works, if it does. The generated `Placeholder` widget should serialize with a `type` of `placeholder` (see the implementation of `widget_type/1`), which should then appear in this mapping. Instead this is expecting a widget type of `body`, which I don't think any widget has.

lib/screens/v2/candidate_generator/on_bus.ex
```diff
@@ -0,0 +1,36 @@
...
+  alias Screens.V2.Template.Builder
+  alias Screens.V2.WidgetInstance.Placeholder
+
+  @behaviour CandidateGenerator
+
+  @impl true
+  @spec screen_template() ::
+          atom()
+          | {atom() | non_neg_integer() | {non_neg_integer(), atom()},
+             atom() | %{optional(atom()) => list()}}
```
digitalcora, line 14: This spec duplicates the one in `CandidateGenerator` — keeping it here could cause confusion, as it might actually override the behaviour's spec in the event they ever diverge. (Dialyzer does check behaviour implementations against the callback spec.)

lib/screens/v2/widget_instance/departures_no_service.ex
```diff
@@ -18,6 +18,7 @@ defmodule Screens.V2.WidgetInstance.DeparturesNoService do
 
     def priority(instance), do: DeparturesNoService.priority(instance)
     def serialize(instance), do: DeparturesNoService.serialize(instance)
+    @spec slot_names(Screens.V2.WidgetInstance.DeparturesNoService.t()) :: [:main_content, ...]
```
digitalcora, line 21: Unrelated change?

assets/src/apps/v2/on_bus.tsx
digitalcora, line 19: I noticed this isn't working as intended (the blue placeholder doesn't show up, only a blank white screen), and I think it's because this is conflating the layout name with the widget that's being placed in a slot of the layout. To try and break it down:

* The `Placeholder` widget on the backend has a `widget_type/1` of `:placeholder`. Normally, that would then appear in this mapping: `placeholder: Placeholder`. The widget type string is mapped to a React component, and the rest of the widget's serialized fields are passed as props to that component.
* So, why doesn't this error? Because on the frontend, layouts _are also_ mapped to components, and on this screen type we have a layout named `body` defined. When a layout is rendered on the frontend, the matching component receives each of the slot names as a prop, with the value being all the fields of the widget that was chosen for that slot (including the `type` field, so it can be passed straight into the generic `Widget` component).
* Putting it all together: The `body` layout is selected (it's the only layout), and the placeholder widget is chosen to go in the slot named `placeholder` (the only place it can go). `body` on the frontend is mapped to the component `Placeholder`. So this does the equivalent of `<Placeholder placeholder={{type: "placeholder", color: "blue", ...}} />` Then `Placeholder` has no prop named `placeholder`, so that's silently ignored and it doesn't render with any particular color.

In conclusion, even the most minimal possible layout requires two React components: One for the layout itself (which will receive as props a mapping of slot-name to widget-fields), and one for an actual widget that will be placed in the layout. You can see an example of one of these "layout-components" in `normal_body.tsx` in most screen types.

assets/src/apps/admin.tsx
```diff
@@ -36,6 +37,7 @@ const routes: [string, string, ComponentType][][] = [
     ["gl-eink-v2-screens", "GL E-ink", GLEinkV2ScreensTable],
     ["pre-fare-v2-screens", "Pre-Fare", PreFareV2ScreensTable],
     ["busway-v2-screens", "Sectional", BuswayV2ScreensTable],
+    ["on-bus-v2-screens", "On-Bus", OnBusScreensTable],
```
digitalcora, line 40: Let's keep these nav items alphabetized within their group?

assets/src/components/admin/admin_tables.tsx
```diff
@@ -451,6 +451,16 @@ const BuswayV2ScreensTable = (): JSX.Element => {
   return <AdminTable columns={columns} dataFilter={dataFilter} />;
 };
 
+const OnBusScreensTable = (): JSX.Element => {
+  const dataFilter = ({ app_id }) => {
+    return app_id === "on_bus_v2";
+  };
+
+  const columns = [...v2Columns];
```
digitalcora, line 459: I think we might now need to split out "Header" from `v2Columns`, since this is resulting in on-bus screens being configurable with a header, but they don't actually have such a field.

assets/src/apps/v2/on_bus.tsx
robbie-sundstrom: Thanks for breaking this down, this explanation brought it all together for me and I got it working again! Had been trying to debug why the Widget for placeholder was not receiving the correct props.

assets/css/v2/on_bus/body/normal.scss
robbie-sundstrom: `screen-container` is used for the iframe of the Inspector tool, and then I have another css sizing set up for the actual class that the placeholder widget falls under. I couldn't get the iframe working without specifying sizing for a top level element

digitalcora: Mm, so that does cover why we need an absolute size for `screen-container` (ultimately something has to define the size of the app) — then I guess what I'm asking is why things "below" that couldn't have width/height `100%`.

robbie-sundstrom: Ah good point, I've updated this to be `100%` for the classes below `screen-container` and moved them into the same file for simplicity for now

assets/src/apps/admin.tsx
robbie-sundstrom: updated!

assets/src/components/admin/admin_tables.tsx
robbie-sundstrom: this is a good point! Moved header out into a separate column and updated the other tables accordingly.

assets/src/components/admin/admin_tables.tsx
```diff
@@ -204,14 +206,21 @@ const BusEinkV2ScreensTable = (): JSX.Element => {
     return app_id === "bus_eink_v2";
   };
 
-  const columns = [...v2Columns, departuresColumn, footerColumn, alertsColumn];
+  const columns = [
+    ...v2Columns,
+    departuresColumn,
+    footerColumn,
+    alertsColumn,
+    headerColumn,
```
digitalcora, line 214: Maybe the header column should be first here (after `v2Columns`), since it was earlier in the order prior to the split. Though we are still unavoidably moving things around a bit (it used to go Header => Evergreen => ...).

assets/src/components/admin/admin_tables.tsx
```diff
@@ -446,7 +457,17 @@ const BuswayV2ScreensTable = (): JSX.Element => {
     return app_id === "busway_v2";
   };
 
-  const columns = [...v2Columns, departuresColumn];
+  const columns = [...v2Columns, departuresColumn, headerColumn];
```
digitalcora, line 460: Similar comment as above re: column ordering.

digitalcora (approved):
Looks good! Had a couple non-blocking nitpicks. Let's merge that `screens-config-lib` PR, then update this one with the proper hash.

assets/src/components/admin/admin_tables.tsx
robbie-sundstrom: updated this ordering!

PR #2427, opened 2025-02-05
Moved 288px height to normal_header.scss instead of normal.scss

deanshi:
**Asana task**: [[extra] bus e-ink header has incorrect appearance on widget endpoint](https://app.asana.com/0/1185117109217413/1209279595523778)

Description
- Fixed issue where header widget height was not being set to 288px
- Removed 288px from the other two spots in bus_eink and gl_eink

Let me know if removing them/putting them here has adverse effects, I tried to check the places that I could but I'm sure I missed places. 

- [ ] Tests added?

Screenshot 2025-02-05 at 3 59 24 PM

### DIFF
--- a/assets/css/v2/bus_eink/screen/normal.scss
+++ b/assets/css/v2/bus_eink/screen/normal.scss
@@ -10,7 +10,6 @@
   top: 0;
   left: 0;
   width: 1200px;
-  height: 288px;
 }
 
 .screen-normal__body {
--- a/assets/css/v2/bus_eink/screen/normal.scss
+++ b/assets/css/v2/bus_eink/screen/normal.scss
@@ -9,13 +9,10 @@
   position: absolute;
   top: 0;
   left: 0;
-  width: 1200px;
 }
 
 .screen-normal__body {
   position: absolute;
   top: 288px;
   left: 0;
-  width: 1200px;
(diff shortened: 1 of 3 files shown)


digitalcora:
This might fix the immediate problem Mercury was having but I think there is an opportunity here to clean up just a bit more. The problem in the abstract is that we're setting the absolute sizes of components by targeting the layout slot they're placed in, rather than the component's markup itself — when the component is then rendered in isolation, not in any slot, it doesn't have the right styles.

I think we should be able to:

* Completely remove `width` and `height` for `.screen-normal__header` and `.screen-normal__body`, in the `normal.scss` for Bus E-ink and GL E-ink
* Move the width value as well as the height value to `.normal-header` in `eink/normal_header.scss`
* Not have to do anything for `.body-normal`, since absolute width and height are already defined for this (redundantly, prior to this change)

digitalcora (approved):
🧹 ✨ thanks!

digitalcora:
> I would love a run down on how it all works from a higher level if possible so I can get a sense of how some of the code interacts with the different possible screens

I wasn't sure how much of the Screens Architecture™ Brett had already gone over but yes, I'd love to explain, at least to the extent I know it myself!

PR #2433, opened 2025-02-11
feat: Add test URL parameters for Hanover on-bus screens

robbie-sundstrom:
**Asana task**: [Test parameters for Hanover on-bus screens](https://app.asana.com/0/1185117109217413/1209111332947266)

A few different things going on in this PR
- Updates the `screen_controller` to pass along valid URL params to the `screen_api_controller` through `assigns
- `screen_api_controller` receives the URL params, passes to `ScreenData` module to be passed to candidate generation
- Updates `inspector.tsx` to pass query params to the iframe URL 
- Modifies `placeholder.tsx` widget to display text
- Adds a No Data State to the On Bus Screen app template

Quick video below shows modifying the URL parameters and having them updated live on the placeholder widget on the On Bus screen:
https://github.com/user-attachments/assets/ed07cc62-e204-451e-b5c8-860546f7e717

- [X ] Tests added?
Unit tests for `query_params.ex` added. Integration type tests not added.


assets/src/apps/v2/on_bus.tsx
```diff
@@ -16,23 +16,52 @@ import SimulationScreenPage from "Components/v2/simulation_screen_page";
...
+  switch (apiResponse.state) {
+    case "success":
+    case "simulation_success":
+      return apiResponse.data;
+    case "failure":
+      return LOADING_LAYOUT;
+    case "loading":
+      return LOADING_LAYOUT;
+    case "disabled":
+      return LOADING_LAYOUT;
```
robbie-sundstrom, line 46: not sure what to do with failure and disabled state for now.

lib/screens/v2/screen_data/query_params.ex
```diff
@@ -0,0 +1,40 @@
...
+  end
+
+  @spec get_url_param_list(Plug.Conn.t()) :: list()
+  def get_url_param_list(conn) do
+    QueryParams.valid_param_keys()
+    |> Enum.map(fn key ->
+      value = Map.get(QueryParams.get_url_param_map(conn), String.to_atom(key))
+      {String.to_atom(key), value}
+    end)
+    |> Enum.filter(fn {_key, value} -> value != nil end)
```
robbie-sundstrom, line 38: This resulted from me trying to convert the struct type to a list, but I see how this Elixir conversion logic is much more complicated than map -> list of tuples. I think I could either
1. Simplify logic
2. Just not DRY and instead of using `get_url_param_map`, rewrite logic that creates a map, not a struct

assets/src/hooks/v2/use_api_response.tsx
```diff
@@ -142,7 +142,10 @@ const useApiPath = (screenId: string, appendPath?: string): string => {
       last_refresh: getDatasetValue("lastRefresh"),
       requestor:
         getDatasetValue("requestor") ?? (isRealScreen() ? "real_screen" : null),
+      route_id: getDatasetValue("routeId"),
       screen_side: getScreenSide(),
+      stop_id: getDatasetValue("stopId"),
+      trip_id: getDatasetValue("tripId"),
```
digitalcora, line 148: From huddle: wonder if we're overcomplicating this a bit by "ingesting" query params through the initial `ScreenController` page load, which then embeds them as data attributes, which are then picked up by the API data hook, which then sends them to the API. Considering having this hook directly copy query params from the current page URL, determined by optional prop passed down through `ScreenPage`/`ScreenContainer` which would only be used in the on-bus entrypoint.

assets/src/apps/v2/on_bus.tsx
```diff
@@ -8,31 +8,68 @@ import "../../../css/on_bus_v2.scss";
 
 import React from "react";
 import ReactDOM from "react-dom";
-import { BrowserRouter as Router, Route, Switch } from "react-router-dom";
+import {
+  BrowserRouter as Router,
+  Route,
+  Switch,
+} from "react-router-dom";
```
digitalcora: Formatter weirdness? This seems to be exactly the same statement before and after.

assets/src/components/v2/screen_page.tsx
```diff
@@ -1,17 +1,35 @@
...
+interface ScreenPageProps {
   id?: string;
+  paramKeys?: string[];
 }
 
-const ScreenPage = ({ id }: Props) => {
+const ScreenPage = ({ id, paramKeys }: ScreenPageProps) => {
   const screenId = id ?? (useParams() as { id: string }).id;
+
+  const queryParams = paramKeys ? getQueryParams(paramKeys) : undefined;
```
digitalcora: `getQueryParams` calls `useLocation` which is a hook, and thus I think this technically breaks the [Rules of Hooks](https://react.dev/reference/rules/rules-of-hooks), namely that hook execution should never be conditional. In practice we _currently_ only assign a constant value that never changes to the `paramKeys` prop, so this sort-of isn't conditional, but in this case I think it's very easy to eliminate the conditional anyway (something like `const queryParams = getQueryParams(paramKeys ?? [])` should do it).

Also since `getQueryParams` qualifies as a "custom hook" by virtue of being a function that calls another hook, maybe we should name it starting with `use` to be clear about that — this is a convention that makes violations of the Rules of Hooks more obvious. `useQueryParams`?

assets/src/hooks/v2/use_api_response.tsx
```diff
@@ -147,29 +151,43 @@ const useApiPath = (screenId: string, appendPath?: string): string => {
...
+
+    for (const [key, value] of Object.entries(datasetParams)) {
       if (value) url.searchParams.append(key, value);
     }
 
     return url.toString();
   }, [screenId, appendPath]);
 };
 
+interface UseBaseApiResponseProps {
```
digitalcora, line 168: nit: These aren't "props" as `useBaseApiResponse` is not a React component. `UseApiResponseOpts`?

lib/screens_web/controllers/v2/screen_controller.ex
```diff
@@ -128,17 +128,17 @@ defmodule ScreensWeb.V2.ScreenController do
...
-      screen_side: screen_side(params),
+      refresh_rate_offset: calculate_refresh_rate_offset(screen_id, refresh_rate),
+      refresh_rate: refresh_rate,
       requestor: params["requestor"],
       rotation_index: rotation_index(params),
-      variant: params["variant"],
-      is_pending: false
+      screen_side: screen_side(params),
+      sentry_dsn: if(params["disable_sentry"], do: nil, else: Sentry.get_dsn()),
+      variant: params["variant"]
```
digitalcora, line 141: I do love a bit of alphabetization, but after the refactor this file is otherwise untouched by the changes, so maybe we should leave it be until we have some other reason to change it? Or it could be a separate refactor commit.

lib/screens/v2/screen_data/query_params.ex
```diff
@@ -0,0 +1,29 @@
...
+
+  # Valid keys for URL parameters to be passed into the screen app.
+  # To process a new URL parameter, it needs to be added to this list.
+  @valid_param_keys ["route_id", "stop_id", "trip_id"]
+
+  @doc "Returns a QueryParam struct of all valid query param keys and corresponding values."
+  @spec get_url_param_map(Plug.Conn.t()) :: struct()
+  def get_url_param_map(conn) do
+    conn
+    |> Plug.Conn.fetch_query_params()
```
digitalcora, line 22: It seems like we shouldn't need this since you've added `fetch_query_params` to the relevant pipelines in the Router. This also may not be the right place for this function anyway — it returns an updated Conn, with the expectation that that will be passed on to further router/controller processing, but here we're throwing it away.

lib/screens/v2/screen_data/query_params.ex
```diff
@@ -0,0 +1,29 @@
...
+    |> Map.get(:query_params, %{})
+    |> Map.take(@valid_param_keys)
+    |> Enum.filter(fn {_key, value} -> value != "" end)
+    |> Enum.into(%{}, fn {key, value} -> {String.to_atom(key), value} end)
```
digitalcora, line 26: ⚠️ We should use `String.to_existing_atom/1` unless there's a very good reason not to, as `String.to_atom/1` is allowed to create new atoms at runtime (which can be [problematic](https://til.hashrocket.com/posts/gkwwfy9xvw-converting-strings-to-atoms-safely)). In this case it's operating on a bounded set of keys determined at compile time so it's actually not a security risk, but there's also no reason we can't use `to_existing_atom` here since the atoms definitely already exist (they are used in the `defstruct` in this module, for one).

lib/screens_web/templates/v2/screen/index.html.eex
```diff
@@ -10,6 +10,15 @@
...
   data-sentry="<%= @sentry_dsn %>"
+  <%= if assigns[:route_id] do %>
+      data-route-id="<%= @route_id %>"
+  <% end %>
+  <%= if assigns[:stop_id] do %>
+      data-stop-id="<%= @stop_id %>"
+  <% end %>
+  <%= if assigns[:trip_id] do %>
+      data-trip-id="<%= @trip_id %>"
+  <% end %>
```
digitalcora, line 21: After the refactor these attributes are now never populated, and unused.

digitalcora (commented):
I like how this turned out! Apologies for the Even More Comments 😅

assets/src/components/admin/inspector.tsx
```diff
@@ -119,8 +120,17 @@ const Inspector: ComponentType = () => {
...
+                    isVariantEnabled ? "variant=all&" : "",
+                    urlParams.get("route_id")
+                      ? `route_id=${urlParams.get("route_id")}&`
+                      : "",
+                    urlParams.get("stop_id")
+                      ? `stop_id=${urlParams.get("stop_id")}&`
+                      : "",
+                    urlParams.get("trip_id")
+                      ? `trip_id=${urlParams.get("trip_id")}&`
+                      : "",
```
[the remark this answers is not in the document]
robbie-sundstrom: Great suggestion! I added a util file to create a Map like this that is then reused across Typescript, you can see in [this commit](https://github.com/mbta/screens/pull/2433/commits/54657b378167dee2f74c5eb78e20383051e557ab)

lib/screens/v2/screen_data/query_params.ex
robbie-sundstrom: good point, I added this before I learned about adding the `fetch_query_param` plug to the pipeline and then forgot to remove it

lib/screens/v2/screen_data/query_params.ex
robbie-sundstrom: This is very good to know about!!!! Thanks for pointing out

assets/src/components/v2/on_bus/no_data.tsx
```diff
@@ -0,0 +1,12 @@
+import React, { ComponentType } from "react";
+import Placeholder from "../placeholder";
+
+const NoData: ComponentType = () => {
+  return (
+    <>
+      <Placeholder color={"blue"} text={""} />
```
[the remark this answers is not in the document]
robbie-sundstrom: Updated with points 1&2! 

To your 3rd point, adding "No Data" makes it very obvious when the page is refreshed and url params are updated. I think this is a case where for this test verision it might be ok, but for the final implementation, we should update the Loading State to prevent that stutter on refresh. I think it could do something like store the previous component state in local storage, and then pass those as props to the final `on_bus` components on page reload.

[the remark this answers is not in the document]
digitalcora: I think in any case there's going to be a flicker/flash when the page is entirely reloaded, especially on a more resource-constrained device, because the HTML is just a blank white page and it takes time to parse and execute all the JS that ultimately renders something. It's going to be a compromised experience absent a way for vendors to communicate next-stop changes directly to our JS, which is definitely what we want to push for, but this option will hopefully be something they can work with initially.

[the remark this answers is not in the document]
robbie-sundstrom: Updated  the Placeholder and No Data components and associated CSS in [this commit](https://github.com/mbta/screens/pull/2433/commits/d6097aa2324cb63f6776b213985beef2545bfe8e)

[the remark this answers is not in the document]
robbie-sundstrom: Also added a header to indicate that it's a Test Mode based on a request from Dave.

assets/src/components/v2/screen_page.tsx
robbie-sundstrom: this is a good point, I should treat this more like a hook. I think I'm going to move `use_query_params` around to be in the hooks directory, even though I don't love exporting a const from there and I want to leave `URL_PARAMS_BY_SCREEN_TYPE` in the same location

digitalcora: IMO it's fine if those were in two different files. We're implementing them towards a common purpose in this moment, but they're independent concepts.

lib/screens_web/controllers/v2/screen_api_controller.ex
```diff
@@ -94,23 +95,29 @@ defmodule ScreensWeb.V2.ScreenApiController do
...
   end
 
-  defp screen_response(screen_id, _, "all", opts) do
+  defp screen_response(screen_id, _, "all", _conn, opts) do
     {default, variants} = ScreenData.variants(screen_id, opts)
     Map.put(%{@base_response | data: default}, :variants, variants)
   end
 
-  defp screen_response(screen_id, %Screen{vendor: :mercury}, variant, opts) do
+  defp screen_response(screen_id, %Screen{vendor: :mercury}, variant, _conn, opts) do
```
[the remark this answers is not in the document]
robbie-sundstrom: Addressed in [this commit](https://github.com/mbta/screens/pull/2433/commits/1590ffab1acfa252cf740ae895b0c57b1fca67d7)

lib/screens/v2/screen_data.ex
```diff
@@ -67,7 +67,7 @@ defmodule Screens.V2.ScreenData do
     end
 
     config
-    |> Layout.generate(selected_variant)
+    |> Layout.generate(selected_variant, Keyword.get(opts, :query_params))
```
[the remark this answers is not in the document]
robbie-sundstrom: Addressed in [this commit](https://github.com/mbta/screens/pull/2433/commits/1590ffab1acfa252cf740ae895b0c57b1fca67d7)

lib/screens/v2/screen_data/layout.ex
```diff
@@ -29,11 +30,17 @@ defmodule Screens.V2.ScreenData.Layout do
...
   def generate(config, variant \\ nil) do
+    generate(config, variant, %QueryParams{})
+  end
+
+  def generate(config, variant, query_params) do
     candidate_generator = @parameters.candidate_generator(config, variant)
     screen_template = candidate_generator.screen_template()
 
     config
+    |> Map.put("query_params", query_params)
```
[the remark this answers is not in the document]
robbie-sundstrom: I updated the `candidate_instances` callback definition, which did require me to modify quite a few files and tests. [Commit with that change here](https://github.com/mbta/screens/pull/2433/commits/b7fcc2f615a4cb7685fe948e91c6ea050c50bfc9)

assets/src/components/v2/screen_page.tsx
robbie-sundstrom: I did end up moving the Map into a util file

assets/src/components/admin/inspector.tsx
```diff
@@ -119,8 +141,12 @@ const Inspector: ComponentType = () => {
               ? new URL(
                   [
                     `/v2/screen/${screen.id}`,
-                    isSimulation ? "/simulation" : "",
-                    isVariantEnabled ? "?variant=all" : "",
+                    buildIframeUrlParams(
+                      screen.config.app_id,
+                      isSimulation,
+                      isVariantEnabled,
+                      urlParams,
```
digitalcora, line 148: It now feels a bit arbitrary that we're building the initial part of the URL outside of this function and the rest inside it — if this were `buildIframeUrl` and accepted the whole `screen` as an argument so it could build the entire URL, I think it'd be a nice simplification.

lib/screens/v2/screen_data/query_params.ex
```diff
@@ -0,0 +1,29 @@
+defmodule Screens.V2.ScreenData.QueryParams do
+  @moduledoc """
+  Encodes valid query parameters that are currently only used by on bus screens.
+  Contains methods for processing query paramaters used by screen and api controllers.
```
digitalcora, line 4: The bit here about "screen and api controllers" doesn't seem accurate anymore. (in general I try to avoid doc comments that assert things about where/how the documented entity is currently used, because they inevitably fall out of date)

lib/screens/v2/screen_data/query_params.ex
```diff
@@ -0,0 +1,29 @@
...
+
+  # Valid keys for URL parameters to be passed into the screen app.
+  # To process a new URL parameter, it needs to be added to this list.
+  @valid_param_keys ["route_id", "stop_id", "trip_id"]
+
+  @doc "Returns a QueryParam struct of all valid query param keys and corresponding values."
+  @spec get_url_param_map(Plug.Conn.t()) :: struct()
+  def get_url_param_map(conn) do
+    conn
+    # |> Plug.Conn.fetch_query_params()
```
digitalcora, line 22: ✂️

lib/screens/v2/screen_data/query_params.ex
```diff
@@ -0,0 +1,29 @@
...
+    |> Map.get(:query_params, %{})
+    |> Map.take(@valid_param_keys)
+    |> Enum.filter(fn {_key, value} -> value != "" end)
+    |> Enum.into(%{}, fn {key, value} -> {String.to_existing_atom(key), value} end)
```
digitalcora, line 26: Technically this could be `Enum.map` — `struct/2` accepts any enumerable of 2-element tuples, not just a map. It may be confusing why we're going to the extra trouble of building a map specifically.

digitalcora (approved):
The function signature change to `candidate_instances` actually ended up less of a churn than I feared. Nice! A few more small comments, address as you like, but ✅

lib/screens/v2/screen_data/query_params.ex
robbie-sundstrom: oh yes good call!

lib/screens/v2/screen_data/query_params.ex
robbie-sundstrom: that's a good philosophy! I did leave in the comment about it only being used for bus screens right now, b/c if it's updated again in the future that point seems very relevant to remember

assets/src/components/admin/inspector.tsx
robbie-sundstrom: moved the full url string building into a separate function! much more readable

PR #2436, opened 2025-02-12
Added elevator ID logging for displayed elevator outages

deanshi:
**Asana task**: [Investigate possible discrepancy with elevator alerts](https://app.asana.com/0/1185117109217413/1209274790976901)

- It was reported that different screens had different elevator outages
- Logging added to make sure that all of the screens are reporting the same elevators at the same time

Quick note about testing - I noticed that with `capture_logs` it seems to only work when the actual logs are outputted to console. I initially tried adding an `info` function to our `Log` utility helper class to keep things consistent, but realized during testing that we needed to either change the testing to show Info logs as well, or I needed to switch the added logging to be a warning instead. I chose the latter, but if there are options to keep `Info` and also have them not populate the console during testing, that would be ideal. 

- [x] Tests added?

### DIFF
--- a/lib/screens/v2/candidate_generator/elevator/closures.ex
+++ b/lib/screens/v2/candidate_generator/elevator/closures.ex
@@ -24,6 +24,7 @@ defmodule Screens.V2.CandidateGenerator.Elevator.Closures do
     defstruct @enforce_keys
   end
 
+  require Logger
   alias Screens.Alerts.{Alert, InformedEntity}
   alias Screens.Elevator
   alias Screens.Facilities.Facility
@@ -140,11 +141,8 @@ defmodule Screens.V2.CandidateGenerator.Elevator.Closures do
     closures
     |> Enum.filter(&relevant_closure?(&1, home_station_id, closures))
     |> Enum.group_by(& &1.station_id)
+    |> log_station_closures()
     |> Enum.map(fn {station_id, station_closures} ->
-      # https://app.asana.com/0/1185117109217413/1209274790976901
-      # Checking if all screens have the same elevator closure ids or not
-      Log.warning("station_closures", ids: Enum.map(station_closures, & &1.id))
-
       %ElevatorClosures.Station{
         id: station_id,
         name: Map.fetch!(station_names, station_id),
@@ -160,6 +158,20 @@ defmodule Screens.V2.CandidateGenerator.Elevator.Closures do
         summary: backup_route_summary(station_closures, closures)
       }
     end)
+  end
+
+  # https://app.asana.com/0/1185117109217413/1209274790976901
+  # Checking if all screens have the same elevator closure ids or not
+  defp log_station_closures(station_closures) do
+    closures =
+      station_closures
+      |> Enum.flat_map(fn {_, list_of_closures} ->
+        list_of_closures
+      end)
+      |> Enum.map_join(" ", & &1.id)
+
+    Logger.info("station_closures:" <> " " <> closures)
+    station_closures
   end
 
   # If we couldn't find alternate/redundancy data for an elevator, assume it's relevant.
--- a/lib/screens/v2/candidate_generator/elevator/closures.ex
+++ b/lib/screens/v2/candidate_generator/elevator/closures.ex
@@ -140,8 +140,8 @@ defmodule Screens.V2.CandidateGenerator.Elevator.Closures do
   defp build_stations_with_closures(closures, home_station_id, station_names, station_route_pills) do
     closures
     |> Enum.filter(&relevant_closure?(&1, home_station_id, closures))
-    |> Enum.group_by(& &1.station_id)
     |> log_station_closures()
+    |> Enum.group_by(& &1.station_id)
     |> Enum.map(fn {station_id, station_closures} ->
       %ElevatorClosures.Station{
         id: station_id,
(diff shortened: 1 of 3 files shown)


lib/screens/v2/candidate_generator/elevator/closures.ex
```diff
@@ -141,6 +141,10 @@ defmodule Screens.V2.CandidateGenerator.Elevator.Closures do
     |> Enum.filter(&relevant_closure?(&1, home_station_id, closures))
     |> Enum.group_by(& &1.station_id)
     |> Enum.map(fn {station_id, station_closures} ->
+      # https://app.asana.com/0/1185117109217413/1209274790976901
+      # Checking if all screens have the same elevator closure ids or not
+      Log.warning("station_closures", ids: Enum.map(station_closures, & &1.id))
```
robbie-sundstrom, line 146: To your point in the PR description about the `Log.info` cluttering up during testing, could you check the environment variable and only log if it's in production?

digitalcora: We probably don't want to log this using our `Log` module, as this goes to Sentry in deployed environments and the volume of logs would quickly eat into our quota.

I might have jumped the gun on thinking we should have a test case for this — as you've noticed (and as I temporarily forgot 🫠), logs are limited to only warnings and errors in the test environment, and `capture_log` can't be more permissive than that global setting. In the absence of a more convenient way to test this, and because this is all temporary anyway, I think it's sufficient to test manually by loading up an elevator screen and confirming that the expected logging is indeed emitted. We can just use `Logger.info` for this.

deanshi: Sounds good! Yeah, I think the `only production` logging would be great, but from my understanding we actually take what outputs to the terminal and use that as a part of `capture_log` in order to compare. If we suppress logging in testing it means we wouldn't get anything 😭. CORRECT ME IF I'M WRONG

It would be great if we could somehow hook into what's underneath and check without getting directly from the terminal, then it would work perfect but it's okay.

And I messed up, I confused Logger and Log again and I meant to use Logger.warning if there was one. Moot point anyway, I'll switch it to be `Logger.info`.

digitalcora: Maybe we should name that module something different. `Report`?

deanshi: I like it - I might put up the changes in a separate PR to keep this one concise then.

digitalcora: Feel free! If you also like `Report` then I think it's an improvement worth making.

lib/screens/v2/candidate_generator/elevator/closures.ex
```diff
@@ -158,6 +160,20 @@ defmodule Screens.V2.CandidateGenerator.Elevator.Closures do
...
+  # Checking if all screens have the same elevator closure ids or not
+  defp log_station_closures(station_closures) do
+    closures =
+      station_closures
+      |> Enum.flat_map(fn {_, list_of_closures} ->
+        list_of_closures
+      end)
+      |> Enum.map_join(" ", & &1.id)
+
+    Logger.info("station_closures:" <> " " <> closures)
```
digitalcora, line 173: `"station_closures:" <> " "` could be equivalently `"station_closures: "`.

lib/screens/v2/candidate_generator/elevator/closures.ex
```diff
@@ -158,6 +160,20 @@ defmodule Screens.V2.CandidateGenerator.Elevator.Closures do
     end)
   end
 
+  # https://app.asana.com/0/1185117109217413/1209274790976901
+  # Checking if all screens have the same elevator closure ids or not
+  defp log_station_closures(station_closures) do
+    closures =
+      station_closures
+      |> Enum.flat_map(fn {_, list_of_closures} ->
+        list_of_closures
```
digitalcora, line 169: This is a little roundabout — on L143 we're grouping the closures, then here we're un-grouping them again. If `log_station_closures` was positioned one step further up in the pipeline, then we'd already have the flat list of closures we want for the logging.

deanshi: Oops - in my mind `groupby` was doing something else, I should have double checked. moved `log_station_closures` up to avoid the flatmap.

lib/screens/v2/candidate_generator/elevator/closures.ex
```diff
@@ -158,6 +160,13 @@ defmodule Screens.V2.CandidateGenerator.Elevator.Closures do
     end)
   end
 
+  # https://app.asana.com/0/1185117109217413/1209274790976901
+  # Checking if all screens have the same elevator closure ids or not
+  defp log_station_closures(station_closures) do
+    Logger.info("station_closures: " <> Enum.map_join(station_closures, " ", & &1.id))
```
digitalcora, line 166: Just noticed we're not logging which screen these closures are being sent to, so if at any point they were different, we wouldn't know _which_ one was different. I think it'd be a good idea to include here the `elevator_id` from the screen configuration — the closest place this is available is in `elevator_closures/2` so we'd have to add it as an argument passed down into here.

Side note on logging: Splunk has some automatic parsing that makes it easier to query logs if they consist of `key=value` pairs (where `value` either doesn't contain spaces or is double-quoted). So it might be nicer if the overall log format was:

```
station_closures elevator_id=123 closures=121,514,849
```

With `closures` being comma-separated there should also be some querying functions that let us pull out the "multiple values" of this field.

digitalcora (approved):
Apologies for the back-and-forth, this looks fine to me assuming you've manually tested that the logging appears correctly!

deanshi:
> Apologies for the back-and-forth, this looks fine to me assuming you've manually tested that the logging appears correctly!

No worries! That's what PRs are for right hahaha. I've already learned so much about part of the code base, so thanks for the back and forth if anything.

PR #2441, opened 2025-02-18
Added suppression of alert 623609 for the Green Line diversion

deanshi:
**Asana task**: [Custom image for Govt Ctr pre-fare](https://app.asana.com/0/1209229552734401/1209391199110652)

Description
- Replaced suppression of old alert
- This in conjunction with the config changes will effectively replace the Alert with a custom image on the Govt Ctr prefare screens and takeovers for Kenmore and Haymarket on DUP screens

- [x] Tests added?

### DIFF
--- a/lib/screens/v2/widget_instance/reconstructed_alert.ex
+++ b/lib/screens/v2/widget_instance/reconstructed_alert.ex
@@ -1313,7 +1313,7 @@ defmodule Screens.V2.WidgetInstance.ReconstructedAlert do
 
   def alert_ids(%__MODULE__{} = t), do: [t.alert.id]
 
-  @suppressed_alert_ids ~w[590467]
+  @suppressed_alert_ids ~w[623609]
 
   def valid_candidate?(%__MODULE__{alert: %{id: alert_id}}) do
     alert_id not in @suppressed_alert_ids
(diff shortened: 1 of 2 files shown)


digitalcora (approved):
Looks fine, though there is a typo in the commit title: It says 623608, when the actual ID (which is correct in the code) is 623609.

deanshi:
> Looks fine, though there is a typo in the commit title: It says 623608, when the actual ID (which is correct in the code) is 623609.

Going to force push a commit title fix change for this - hopefully that doesn't break anything

digitalcora:
It definitely should not — I force-push branches quite a lot to get the commit history looking the way I want it before merging (though I also use "rebase and merge" which applies the commits exactly as they are on the branch to `main`, rather than squashing multiple commits into one).

Do note that if you use "squash and merge", the default title for the commit is the _pull request_ title, which may not be the same as your (single) commit title. You do also get an opportunity to edit that title before confirming the merge.

deanshi:
> It definitely should not — I force-push branches quite a lot to get the commit history looking the way I want it before merging (though I also use "rebase and merge" which applies the commits exactly as they are on the branch to `main`, rather than squashing multiple commits into one).
> 
> Do note that if you use "squash and merge", the default title for the commit is the _pull request_ title, which may not be the same as your (single) commit title. You do also get an opportunity to edit that title before confirming the merge.

I actually prefer rebase and merge (and clean up commits/squash them myself), but got used to squash and merge from my last job, so good to know that we're able to do either one.

PR #2443, opened 2025-02-20
feat: upcoming closures for elevator screens

digitalcora:
When the screen's configured elevator is not currently closed, but has an upcoming planned closure, include a special page that explains this alongside the list of active closures. When this page is not the current one, the in-station summary banner shows the same information in a more concise format, so it is always visible.

**Asana task**: https://app.asana.com/0/1185117109217413/1208859172526601

Screenshots below are from a test screen configured for elevator 920, which has a planned closure in the live alerts feed. The entering category of this elevator is 3A, which is neither "in-station" nor "shuttle bus", so we use the fallback summary.

&nbsp; | &nbsp;
---|---
Screen Shot 2025-02-20 at 11 53 39 | Screen Shot 2025-02-20 at 11 53 48


lib/screens/v2/candidate_generator/elevator/closures.ex
```diff
@@ -46,21 +49,31 @@ defmodule Screens.V2.CandidateGenerator.Elevator.Closures do
   @route injected(Route)
   @stop injected(Stop)
 
-  @fallback_summary "Visit mbta.com/elevators for more info"
+  @entering_summaries %{
+    in_station: "An accessible route will be available.",
+    shuttle: "An accessible route will be available via shuttle bus."
+  }
+  @exiting_summary_fallback "Visit mbta.com/elevators for more info"
```
robbie-sundstrom, line 56: I'm a bit confused by this naming making these seem like opposites, and it took me a second to figure out what was going on. If I'm reading this correctly, why can't these just be named something like `route_summaries` and `summary_fallback`.

lib/screens/v2/candidate_generator/elevator/closures.ex
```diff
@@ -167,11 +197,36 @@ defmodule Screens.V2.CandidateGenerator.Elevator.Closures do
...
-        summary: backup_route_summary(station_closures, closures)
+        summary: exiting_summary(station_closures, closures)
       }
     end)
   end
 
+  defp build_upcoming_closure([]), do: nil
+
+  defp build_upcoming_closure([closure | _] = closures) do
+    # Note: don't have to account for periods with `nil` start here; those are `happening_now?`
```
robbie-sundstrom, line 208: Maybe log a warning if we do get a case of a nil start_date then?

lib/screens/v2/candidate_generator/elevator/closures.ex
digitalcora: In a way they actually are opposites, so you came away with the correct impression!

* `entering_summaries` is only used for upcoming closures, which show a summary of the elevator's "entering category". It happens that the fallback value here, _"Visit **mbta.com/elevators** for more information"_ is the only one with non-plaintext formatting, so I chose to represent this as `null` and let the client hard-code it instead of the server.

* `exiting_summary_fallback` is only used for active closures, which show a summary of the elevator's "exiting category". It happens that there's exactly one such summary that's shown in a different style than all others, _"Accessible route available"_, so once again I chose to represent this as `null` and let the client hard-code it along with using that as a signal to use the different style.

I'd agree though that it's slightly confusing how in one case the "seek more info" fallback summary is the "special" one and thus represented on the client, and in the other case it isn't and is thus represented on the server. I think there are ways we could resolve this if you think it's worth the lift, though I'm also comfortable letting this sit until the next time we need to change something about it, which might not be for a while.

lib/screens/v2/candidate_generator/elevator/closures.ex
digitalcora: If we do then this will let us know by crashing, on L217 when we attempt to pass `nil` to `Util.service_date/1`. And to be clear, this would be a logic error on our part, not data that's technically possible but we expect to never see — the `closures` argument here ultimately comes from L64 where we split off alerts that are `happening_now?`, which includes alerts with a `nil` start.

assets/src/components/v2/elevator/closures.tsx
```diff
@@ -127,119 +180,126 @@ const CurrentClosures = ({
...
+  useLayoutEffect(() => {
+    if (!listRef.current) return;
+
+    setPageRowCounts(
+      _.chain(listRef.current.children)
+        .toArray()
+        .map((child) => (child as HTMLElement).offsetLeft)
+        .countBy(_.identity)
+        .values()
+        .value(),
```
robbie-sundstrom, line 218: I'd pull this out into a separate function for readability and especially testability.

assets/src/components/v2/elevator/closures.tsx
```diff
@@ -6,15 +6,20 @@ import makePersistent, {
...
-import NormalService from "Images/svgr_bundled/normal-service.svg";
+import CalendarIcon from "Images/svgr_bundled/calendar.svg";
+import CalendarAlertIcon from "Images/svgr_bundled/calendar-alert.svg";
+import NormalServiceIcon from "Images/svgr_bundled/normal-service.svg";
 import AccessibilityAlert from "Images/svgr_bundled/accessibility-alert.svg";
+import { hasOverflowX } from "Util/utils";
 import { CLOSURES_PAGING_INTERVAL_MS } from "./constants";
 
+type UpcomingClosureInfo = {
+  banner: { title: string; trailer: string | null };
```
robbie-sundstrom, line 19: Nit, in part b/c I see how many lines it would require changing -- I found the variable `trailer` confusing at first b/c I haven't seen that used before to refer to trailing words as it's doing here. I'd prefer something like `suffix` or `postfix`, and then when it's used elsewhere it could be `banner_suffix`

lib/screens/v2/candidate_generator/elevator/closures.ex
robbie-sundstrom: Gotcha, thanks for the explanation! This makes more sense now that I've eaten and am coming back to the code.

I agree that having the "seek more info" fallback summary handled on the client is fine for this and not worth dealing with. 

My one suggestion then would be to change `entering_summaries` to something like `upcoming_summaries` and `exiting_summary_fallback` to `active_summary_fallback`, but I don't think it's a huge deal.

assets/src/components/v2/elevator/closures.tsx
robbie-sundstrom: This is part of my reason for [this Slack message](https://mbta.slack.com/archives/C02SXM166N8/p1740416266196749) -- I've seen unexpected issues with Lodash in the past, and like having unit tests for possible edge cases of chaining together so many functions of a library.

lib/screens/v2/widget_instance/elevator_closures.ex
```diff
@@ -32,15 +25,111 @@ defmodule Screens.V2.WidgetInstance.ElevatorClosures do
...
+    @nbsp "\u00A0"
+
+    @spec serialize(t(), DateTime.t()) :: map()
+    def serialize(%__MODULE__{period: {start_date, end_date}, summary: summary}, now) do
+      today = Util.service_date(now)
+      indefinite? = is_nil(end_date)
+      starts_tomorrow? = Date.diff(start_date, today) == 1
+      starts_this_week? = Date.beginning_of_week(start_date) == Date.beginning_of_week(today)
+      relative_start? = starts_tomorrow? or starts_this_week?
+      single_day? = start_date == end_date
```
robbie-sundstrom, line 45: Just to confirm, there's no case where an alert's `period` will have a `start_date` and `end_date` that include times?

assets/src/components/v2/elevator/closures.tsx
digitalcora: Hmm, I like `postfix`, hadn't considered that. I didn't want to call it `suffix` (since that sort of implies something tacked onto the end of running text, which is not the case in the details view) nor `footer` (since that implies a separate block element, which is not the case in the banner). I don't mind changing it since I _also_ didn't love `trailer`.

lib/screens/v2/widget_instance/elevator_closures.ex
digitalcora: Active period endpoints _always_ include times (or rather, are full datetimes). The design seems to assume planned closures are always for the full duration of whatever days they occur on, and doesn't include any presentation of a time-of-day — so far I haven't seen reason to doubt that (the planned closures I've seen have been from `03:00:00` the day they start to `02:59:00` the day they end, or rather the day after), but if we ever do have a "partial-day" elevator closure I think this should degrade somewhat gracefully.

lib/screens/v2/widget_instance/elevator_closures.ex
```diff
@@ -32,15 +25,111 @@ defmodule Screens.V2.WidgetInstance.ElevatorClosures do
...
+
+      same_month? =
+        not is_nil(end_date) and
+          {start_date.year, start_date.month} == {end_date.year, end_date.month}
```
robbie-sundstrom, line 49: Could be simplified b/c you already set `indefinite? = is_nil(end_date)` 
So this could be:
```
same_month? = not indefinite && {start_date.year, start_date.month} == {end_date.year, end_date.month}
```

lib/screens/v2/widget_instance/elevator_closures.ex
```diff
@@ -32,15 +25,111 @@ defmodule Screens.V2.WidgetInstance.ElevatorClosures do
...
+                  ),
+                else: "through #{long(end_date)}"
+
+            {banner, details}
+        end
+
+      %{
+        banner: %{title: banner_title, trailer: banner_trailer},
+        details: %{summary: summary, titles: details_titles, trailer: details_trailer}
+      }
```
robbie-sundstrom, line 98: Is there a reason you don't want to use a struct here? Since it's a nested map, that seems safer to me

lib/screens/v2/widget_instance/elevator_closures.ex
```diff
@@ -32,15 +25,111 @@ defmodule Screens.V2.WidgetInstance.ElevatorClosures do
...
+            {
+              "#{day_name(start_date)}, #{short(start_date)}",
+              [
+                "#{day_name(start_date)}, #{long(start_date)}",
+                "#{day_name(start_date)}, #{short(start_date, _dot = true)}"
+              ]
+            }
+        end
+
+      {banner_trailer, details_trailer} =
```
robbie-sundstrom, line 69: Also here with `banner_postfix` and `details_postfix` instead of `trailer`

lib/screens/v2/widget_instance/elevator_closures.ex
```diff
@@ -32,15 +25,111 @@ defmodule Screens.V2.WidgetInstance.ElevatorClosures do
...
+
+          true ->
+            {
+              "#{day_name(start_date)}, #{short(start_date)}",
+              [
+                "#{day_name(start_date)}, #{long(start_date)}",
+                "#{day_name(start_date)}, #{short(start_date, _dot = true)}"
+              ]
+            }
+        end
```
robbie-sundstrom, line 67: These long blocks make the serialize function a bit long to parse. Could the title and trailer logic be pulled out into their own functions?

```
{banner_title, details_titles} = title_for_start_date(start_date, today)
{banner_postfix, details_postfix} = postfix_for_start_period(start_date, end_date, today)
```

The postfix generation is what uses almost all of the variables you set at the beginning of the serialize method anyway.

digitalcora: I'll give this a shot, definitely.

lib/screens/v2/candidate_generator/elevator/closures.ex
digitalcora: > My one suggestion then would be to change `entering_summaries` to something like `upcoming_summaries` and `exiting_summary_fallback` to `active_summary_fallback`, but I don't think it's a huge deal.

Yeah I think this could be an improvement!

lib/screens/v2/widget_instance/elevator_closures.ex
digitalcora: I think I did it this way to try and clarify that the reason for the short-circuit has nothing to do with however `indefinite?` is defined, but is an explicit `nil` guard for the `end_date.x` field access that comes after it in the conditional. This might change anyway based on [your suggestion here](https://github.com/mbta/screens/pull/2443#discussion_r1968115187) though.

lib/screens/v2/widget_instance/elevator_closures.ex
digitalcora: This is ultimately part of the return value of `serialize/1` on a widget, which is the last step before this data is encoded into JSON and sent to the client, so no further Elixir code would be using the value. In a way the "struct"/authoritative type definition for this data lives in the prop types of the client code.

This is mostly just observation of the conventions we already have in the codebase, so it's not to say we shouldn't change it (except to argue from consistency). I do feel vaguely as though using structs for this would give a _false_ notion of type-safety under the circumstances.

lib/screens/v2/widget_instance/elevator_closures.ex
digitalcora: Committed, let me know what you think!

assets/src/components/v2/elevator/closures.tsx
digitalcora: Thinking it over, this implementation was overcooked, and in [this commit](https://github.com/mbta/screens/pull/2443/commits/cc29f1170d59cbe045d8654032b5ad772266c0da) I replaced it with something simpler that doesn't use chaining. Do you think this is still worth pulling out a function for? I'm less sure now than I was about the original version.

robbie-sundstrom: oh nice, that is much simpler! I don't see a need to pull out a function and test it with your update

lib/screens/v2/widget_instance/elevator_closures.ex
robbie-sundstrom: Gotcha, since this is just getting sent off to the frontend, I see how there's less of a need for type safety from a struct

robbie-sundstrom (approved):
🚀

PR #2450, opened 2025-02-25
Added Sans-Serif fallback on font-family for bus_shelter, lcd_common and pre_fare styles

deanshi:
**Asana task**: [Fix serif font fallback on some screen elements](https://app.asana.com/0/1185117109217413/1209354747944094)

Description
Added Sans-Serif fallback on font-family for bus_shelter, lcd_common and pre_fare styles

- [ ] Tests added? (Unneeded?)

### DIFF
--- a/assets/css/v2/bus_shelter/link_footer.scss
+++ b/assets/css/v2/bus_shelter/link_footer.scss
@@ -23,7 +23,7 @@
 
 .link-footer__message {
   display: inline-block;
-  font-family: Inter;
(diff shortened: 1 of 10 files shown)


digitalcora (approved):
LGTM. At this moment we don't have a great way to live-test this because GDS has already applied their own fix for the font loading issue that resulted in fallback fonts being used, but I think we also don't have any reason to believe this will actively cause issues on these screen types.

PR #2456, opened 2025-03-03
feat: initial support for "solo" Pre-Fares

digitalcora:
**Asana task**: https://app.asana.com/0/1185117109217413/1209255404829828

A Pre-Fare solo is, for most practical purposes, just the right screen of a Pre-Fare duo. This adds a configuration option to formalize this, plus a tweak to the header so the stop name and time are both displayed on the right (only) side.

Widgets otherwise behave identically, except that alerts which would take over both screens of a duo will not appear at all (not egregious compared to the current behavior, where most of the useful information is displayed on the nonexistent left screen, leaving only a vague "Seek alternate route" message on the visible one). Full alert support for solo screens will be implemented in an upcoming task.

Key technical changes:

* `CandidateGenerator.screen_template` now accepts an argument which is the screen configuration, allowing the generated template to vary per individual screen.
* Simplify the `Viewport` component out of existence.

⚠️ Prerequisite: mbta/screens-config-lib#51


lib/screens/v2/widget_instance/audio_only/content_summary.ex
```diff
@@ -43,6 +43,15 @@ defmodule Screens.V2.WidgetInstance.AudioOnly.ContentSummary do
...
   end
 
+  @pre_fare_takeover_slots MapSet.new(~w[
+      full_body_duo
+      full_body_left
+      full_body_right
+      full_duo_screen
+      full_left_screen
+      full_right_screen
+    ]a)
```
robbie-sundstrom, line 53: should `full_body_solo` and `full_solo_screen` be included in this list as well?

assets/css/v2/pre_fare/screen_container.scss
```diff
@@ -1,9 +1,26 @@
...
+  &--left,
+  &--right,
+  &--solo {
+    width: var(--panel-width);
+  }
+
+  &--right > *,
+  &--solo > * {
+    margin-left: calc(var(--panel-width) * -1);
+  }
```
robbie-sundstrom, line 23: nice 😎

lib/screens/v2/widget_instance/audio_only/content_summary.ex
digitalcora: There are no such slots — the full body/screen of a solo is `full_body_right` and `full_right_screen`. For now I chose to keep the names of all slots that are conceptually shared between [solo pre-fares] and [the right side of duo pre-fares] the same, so a solo almost literally _is_ the right side of a duo. This works out well to avoid extensive code changes for the moment, though I'm keeping this in mind as something we should be willing to change if it makes sense for full alert widget support in the subsequent tasks.

robbie-sundstrom: Gotcha, that makes sense to me why the solo generator is using the right side slots now.

robbie-sundstrom (approved):
Awesome! Thought this would be much more complex, but you kept it simple 😄 Comments in `ScreenSide` and the config generator also make this very understandable

mix.exs
```diff
@@ -92,9 +92,7 @@ defmodule Screens.MixProject do
       {:telemetry, "~> 1.2"},
       {:telemetry_poller, "~> 1.0"},
       {:telemetry_metrics, "~> 1.0"},
-      {:screens_config,
-       git: "https://github.com/mbta/screens-config-lib.git",
-       ref: "b2ffefe272d47a4e3173e3eea77cd0e2a84b8179"},
+      {:screens_config, github: "mbta/screens-config-lib"},
```
deanshi, line 95: I guess there'd never be a change to screens-config-lib that we wouldn't see review right?

Out of curiosity was there a reason we were holding onto a specific commit previously?

robbie-sundstrom: oh I missed this last commit before approving! I liked pointing at a specific commit for tracking changes more accurately and in case of the need for a rollback.

digitalcora: My reasoning here is that the lockfile does point to a specific commit, similarly to how e.g. we depend on `{:jason, "~> 1.0"}` and that resolves and is locked to a specific version. There's no way for the underlying version/commit hash to be updated without some explicit action on our part or without a change showing up in `mix.lock`. If we did go through that, merged the change, and it turned out we shouldn't have, the process of rolling it back or pinning it to an older version would be the same as for other dependencies.

Also, this should allow Dependabot to open PRs for updates to this library, to remind us if we've forgotten to bring it up to date.

robbie-sundstrom: ah I didn't realize the lockfile would still point to a specific version. This update makes sense to me in that case!

PR #2457, opened 2025-03-03
refactor: alert parsing and active period handling

digitalcora:
* The `start` of alert active periods is documented as non-nullable in the V3 API, so we can remove some special cases that assumed it could be null. This logic was unreachable anyway, since `Alerts.Parser` did _not_ have such handling and would have crashed if we ever encountered this situation.

* Remove some redundancy in the active period parsing code.

* Add a test case to cover `Alert.fetch/1` and more of the behavior of `Alerts.Parser`, using more realistic test data.

**Asana task**: https://app.asana.com/0/1185117109217413/1209447170673650

### DIFF
--- a/lib/screens/alerts/parser.ex
+++ b/lib/screens/alerts/parser.ex
@@ -47,8 +47,8 @@ defmodule Screens.Alerts.Parser do
           active_period: parse_and_sort_active_periods(active_period),
           lifecycle: lifecycle,
           timeframe: timeframe,
-          created_at: parse_time(created_at),
-          updated_at: parse_time(updated_at),
+          created_at: parse_datetime(created_at),
+          updated_at: parse_datetime(updated_at),
           url: url,
           description: description
         }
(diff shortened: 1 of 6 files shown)


robbie-sundstrom (approved):
nice!

deanshi (approved):
Looks great to me!

Just a simple style question for my own knowledge but otherwise :shipit:

lib/screens/alerts/parser.ex
```diff
@@ -97,32 +97,17 @@ defmodule Screens.Alerts.Parser do
...
-    end_t = parse_time(end_str)
-    {start_t, end_t}
-  end
-
-  defp parse_time(nil), do: nil
-
-  defp parse_time(s) do
-    {:ok, time, _} = DateTime.from_iso8601(s)
+  defp parse_datetime(iso_str) do
+    {:ok, time, _offset} = DateTime.from_iso8601(iso_str)
```
deanshi, line 110: In Elixir, is it just a stylistic preference to having "_" versus "_offset" here? 

I see in a lot of places we use plain ol' underscore, but I really do like the idea of having "_value" instead to be super explicit.

digitalcora: I personally tend towards naming unused variables, and there is a [Credo check](https://github.com/rrrene/credo/blob/965a59c0e71067c7fc6fab26860ac6dc0ea2912f/lib/credo/check/consistency/unused_variable_names.ex#L1) we could enable if we wanted to enforce that, but I like having the option of keeping them anonymous. For example, `[_ | _]` is a common pattern used to match a non-empty list, and to me it wouldn't really make sense to name those variables, as the structure of the pattern is more important than the values themselves.

PR #2458, opened 2025-03-03
[feat] Add logic into pre-fare widget generator to filter elevators with redundancy

robbie-sundstrom:
**Asana task**: [Use identical filtering in pre-fare elevators widget as on elevator screens](https://app.asana.com/0/1185117109217413/1209184999956098)

I tried to consolidate more elevator screen and pre-fare logic, but found it to be too complicated for this task. As it stands, this is largely copied logic from the elevator screen widget generation into the pre-fare widget generation, which itself was very simple.


lib/screens/v2/candidate_generator/widgets/elevator_closures.ex
```diff
@@ -85,4 +121,53 @@ defmodule Screens.V2.CandidateGenerator.Widgets.ElevatorClosures do
...
+        ]
+
+      _multiple ->
+        Report.warning("elevator_closure_affects_multiple", alert_id: alert_id)
+        []
+    end
+  end
+
+  # If we couldn't find alternate/redundancy data for an elevator, assume it's relevant.
+  defp relevant_closure?(%Closure{elevator: nil}, _closures), do: true
```
digitalcora, line 162: This is missing the behavior where a closure is always considered relevant when it's in the same station as the screen displaying it.

lib/screens/v2/candidate_generator/widgets/elevator_closures.ex
```diff
@@ -22,12 +46,24 @@ defmodule Screens.V2.CandidateGenerator.Widgets.ElevatorClosures do
...
+        elevator_closures
+        |> Enum.filter(&relevant_closure?(&1, elevator_closures))
+        |> Enum.map(fn %Closure{alert_id: alert_id} -> alert_id end)
 
       [
         %ElevatorStatusWidget{
-          alerts: elevator_closures,
+          alerts:
+            elevator_alerts
+            |> Enum.filter(fn %Alert{id: id} -> id in elevator_closure_ids end),
```
digitalcora, line 66: `elevator_closure_ids` is a list here, requiring a linear-time traversal for each item in `elevator_alerts`, meaning this takes N² time overall. Admittedly this couldn't possibly matter with the values of N we'd expect here, but in this kind of situation I'd consider it good practice to create a set (`|> MapSet.new()`).

robbie-sundstrom: great callout!

lib/screens/v2/candidate_generator/widgets/elevator_closures.ex
```diff
@@ -1,14 +1,38 @@
...
+    @type t :: %__MODULE__{
+            id: Facility.id(),
+            alert_id: String.t(),
+            name: String.t(),
+            station_id: Stop.id(),
+            periods: [Alert.active_period()],
+            elevator: Elevator.t() | nil
+          }
+
+    @enforce_keys ~w[id alert_id name station_id periods elevator]a
```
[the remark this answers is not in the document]
robbie-sundstrom: I did update this to not be a defined as a separate Closure struct, but I'm still passing struct values around between functions here. That makes more sense to me than a tuple, but not sure if that's still just my brain tending towards OOP rather than more functional.

Unfortunately, I didn't want to mess with the `ElevatorStatusWidget`, which requires a list of alerts. However, to do the relevance filtering based on stations, the easiest way is by converting to Elevator objects to get the redundancies. Because of that, I think there need to be two discrete steps so that we have both a list of alerts and a list of closures used to find any non-relevant closures.

lib/screens/v2/candidate_generator/widgets/elevator_closures.ex
```diff
@@ -1,14 +1,38 @@
...
+  alias Screens.Report
   alias Screens.Routes.Route
   alias Screens.Stops.Stop
   alias Screens.V2.WidgetInstance.ElevatorStatus, as: ElevatorStatusWidget
   alias ScreensConfig.Screen
   alias ScreensConfig.V2.{ElevatorStatus, PreFare}
 
+  import Screens.Inject
+
+  @elevator injected(Elevator)
```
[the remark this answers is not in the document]
robbie-sundstrom: I do want to add test coverage, but have not finished that up yet.

lib/screens/v2/candidate_generator/widgets/elevator_closures.ex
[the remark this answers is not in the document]
digitalcora: > I did update this to not be a defined as a separate Closure struct, but I'm still passing struct values around between functions here.

I guess you mean maps, from the current code — which actually I don't recommend vs. structs, since it's easy to accidentally use the wrong name for a field and not get an obvious compile error/runtime crash because maps don't have a fixed set of keys. Since you also found a use for keeping the `station_id` field though (and I think that's a nice improvement!), this has 4 fields, which is the point where tuple matching starts to get unwieldy and error-prone itself. I think you'd be justified in keeping the struct, if we're still duplicating this much code from elevator screens; the eventual goal is to merge these into one widget anyway. (Don't feel like you have to though, just giving a heads-up about the downsides of maps.)

lib/screens/v2/candidate_generator/widgets/elevator_closures.ex
```diff
@@ -85,4 +92,55 @@ defmodule Screens.V2.CandidateGenerator.Widgets.ElevatorClosures do
...
+
+  # If we couldn't find alternate/redundancy data for an elevator, assume it's relevant.
+  defp relevant_closure?(%{elevator: nil}, _closures, _parent_station_id), do: true
+
+  # If any of a closed elevator's alternates are also closed, it's always relevant.
+  defp relevant_closure?(
+         %{
+           id: _,
+           station_id: station_id,
+           alert_id: _,
```
digitalcora, line 137: If you're not using the value of a struct/map field, leaving it out is fine:

```elixir
%{
  station_id: station_id,
  elevator: %Elevator{alternate_ids: alternate_ids, exiting_redundancy: redundancy}
}
```

lib/screens/v2/candidate_generator/widgets/elevator_closures.ex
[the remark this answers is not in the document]
robbie-sundstrom: I've added tests for the module now!

lib/screens/v2/candidate_generator/widgets/elevator_closures.ex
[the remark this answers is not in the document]
robbie-sundstrom: Converted this back to using a defined `Closure` struct instead of a map. I think that makes it a bit easier to work with since it's 4 fields.

lib/screens/v2/candidate_generator/widgets/elevator_closures.ex
digitalcora: Resurfacing this comment since I think it's still relevant; I don't see this behavior implemented in the latest code.

robbie-sundstrom: I thought that the check for `station_id == parent_station_id` that I added covered this. I'm adding a test case for this now because I did miss this scenario in the tests

robbie-sundstrom: I added a test for this case and believe it's working as intended!

digitalcora: Ahh, that's the part I was missing. I was expecting to see a third function clause here a la `Elevator.Closures` which the rest of this is somewhat copied from:

```ex
# Elevators at the home station ID are always relevant.
defp relevant_closure?(%Closure{station_id: station_id}, station_id, _closures), do: true
```

It doesn't specifically have to be implemented one way or the other but I do like calling out this special case via its own pattern match.

lib/screens/v2/candidate_generator/widgets/elevator_closures.ex
```diff
@@ -20,27 +44,37 @@ defmodule Screens.V2.CandidateGenerator.Widgets.ElevatorClosures do
...
+        |> Enum.filter(&relevant_closure?(&1, elevator_closures, parent_station_id))
+        |> Enum.map(fn %{alert_id: alert_id} -> alert_id end)
+        |> MapSet.new()
 
       [
         %ElevatorStatusWidget{
-          alerts: elevator_closures,
+          alerts:
+            alerts
+            |> Enum.filter(fn %Alert{id: id} -> id in elevator_closure_ids end),
```
digitalcora, line 64: No need for the pipeline here as it's only a single step:

```ex
Enum.filter(alerts, fn %Alert{id: id} -> id in elevator_closure_ids end)
```

lib/screens/v2/candidate_generator/widgets/elevator_closures.ex
```diff
@@ -85,4 +110,53 @@ defmodule Screens.V2.CandidateGenerator.Widgets.ElevatorClosures do
...
+        ]
+
+      _multiple ->
+        Report.warning("elevator_closure_affects_multiple", alert_id: alert_id)
+        []
+    end
+  end
+
+  # If we couldn't find alternate/redundancy data for an elevator, assume it's relevant.
+  defp relevant_closure?(%{elevator: nil}, _closures, _parent_station_id), do: true
```
digitalcora, line 148: nit: Could be `%Closure{elevator: nil}` in the pattern match.

test/screens/v2/candidate_generator/widgets/elevator_closures_test.exs
```diff
@@ -0,0 +1,237 @@
...
+  alias Screens.Alerts.Alert
+  alias Screens.Elevator
+  alias Screens.Routes.Route
+  alias Screens.Stops.Stop
+  alias Screens.V2.CandidateGenerator.Widgets.ElevatorClosures
+  alias Screens.V2.WidgetInstance.ElevatorStatus, as: ElevatorStatusWidget
+  alias ScreensConfig.Screen
+  alias ScreensConfig.V2.{ElevatorStatus, PreFare}
+
+  import Mox
```
digitalcora, line 12: Wherever we use Mox in tests we should also `setup :verify_on_exit!` — this enables the feature of `expect` that if an expected call _wasn't_ received, the test fails. I think we are only using `stub` in these particular tests but it doesn't hurt to include this in case we add something later.

digitalcora (approved):
Some minor comments but LGTM! Thanks so much for adding our first actual tests of this module.

test/screens/v2/candidate_generator/widgets/elevator_closures_test.exs
robbie-sundstrom: that is a very good tip to know!

lib/screens/v2/candidate_generator/widgets/elevator_closures.ex
robbie-sundstrom: much cleaner, thanks!

PR #2467, opened 2025-03-04
feat(admin): inspector shows current screen URL

digitalcora:
* Provide a basic "URL bar" for the inspector iframe, allowing an admin to see and copy the current screen URL.

* Minor refactor to iframe URL building to put all the logic in one place and allow the URL to be used as the key for resetting data controls, fixing potential issues where these might not be reset even when the URL changed.

Screenshot 2025-03-04 at 3 51 41 PM

### DIFF
--- a/assets/css/admin/inspector.scss
+++ b/assets/css/admin/inspector.scss
@@ -65,7 +65,10 @@
 }
 
 .inspector__screen {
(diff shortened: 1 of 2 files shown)


PaulJKim (approved):
Looks good!

PR #2471, opened 2025-03-06
Added support for Pre-fare Solo layout to the Screenplay simulation

deanshi:
**Asana task**: [Screenplay renaming / support for layout](https://app.asana.com/0/1185117109217413/1209255549956022)

Screenshot 2025-03-06 at 4 59 54 PM
Screenshot 2025-03-06 at 5 00 15 PM

Description
- These changes fix the cut off duo layout too
- [ ] Tests added?

### DIFF
--- a/assets/css/pre_fare_v2.scss
+++ b/assets/css/pre_fare_v2.scss
@@ -60,6 +60,7 @@
 
 @import "v2/simulation_common";
 @import "v2/pre_fare/simulation";
+@import "v2/pre_fare/viewport";
 
 @import "v2/pre_fare/disruption_diagram/disruption_diagram";
 
--- a/assets/css/v2/pre_fare/screen_container.scss
+++ b/assets/css/v2/pre_fare/screen_container.scss
@@ -1,26 +1,6 @@
 .screen-container {
   position: relative;
-  height: 1920px;
   margin: 0 auto;
-  overflow: hidden;
-  // width of a single Pre-Fare display unit; "duo" installations are two of
-  // these side-by-side
-  --panel-width: 1080px;
-
-  &--duo {
-    width: calc(var(--panel-width) * 2);
-  }
-
-  &--left,
-  &--right,
-  &--solo {
-    width: var(--panel-width);
-  }
-
-  &--right > *,
-  &--solo > * {
-    margin-left: calc(var(--panel-width) * -1);
-  }
 }
 
 .screen-container-blink {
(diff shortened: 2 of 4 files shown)


assets/css/v2/pre_fare/simulation.scss
```diff
@@ -24,13 +24,40 @@
...
+    &--left,
+    &--right,
+    &--solo {
+      width: var(--panel-width);
+    }
+
+    &--right > *,
+    &--solo > * {
+      margin-left: calc(var(--panel-width) * -1);
+    }
```
digitalcora, line 43: Since all this is identical to the rules for `.screen-container` with only the selector being different, can we extract to a shared selector targeted at `.screen-container, .simulation`? This could be in a separate file, though I'm not sure what a good name would be. `pre_fare/viewport`?

assets/css/v2/pre_fare/simulation.scss
```diff
@@ -24,13 +24,40 @@
...
   .simulation__full-page {
     z-index: 999;
-    max-width: 388px;
     max-height: 407px;
-    margin: 16px;
+
+    --width: 200px;
+
+    &--duo {
+      width: calc(var(--width) * 2);
```
digitalcora, line 53: What's the reason we need to set a width on this, separately from `.simulation`? It seems redundant, and fragile (assuming the width set here needs to line up with the width above multiplied by the transform scale, though this is not _quite_ true of the new value you've set it to).

deanshi: Ah this was leftover code from me trying to fix the issue with the left side and right side pages not showing up as individual items in the simulation.

I was able to get the right side to show up (by changing some filtering and length logic in `simulation_screen_container.tsx`) and this styling was just to handle some of the extra white space due to the solo layout. None of that really for anything more than just me playing around to try and understand the code base.

I eventually took out the code because I wasn't quite sure what the logic of the flex space and the rightScreenPage logic, but I forgot to take this width part out.

digitalcora (approved):
Seems good to me though I haven't personally checked the final appearance in Screenplay — doing that would be a good exercise if you haven't already (either pointing your local copy of Screenplay to your local copy of Screens, or deploying this branch to dev-green and looking at Screenplay dev-green).

deanshi:
> Seems good to me though I haven't personally checked the final appearance in Screenplay — doing that would be a good exercise if you haven't already (either pointing your local copy of Screenplay to your local copy of Screens, or deploying this branch to dev-green and looking at Screenplay dev-green).

Good call - let me push to dev-green real quick and see if it looks good. I looked at things locally but I think it'd make me feel better if it was on a dev environment without my weird local config stuff.

deanshi:
screenplay dev-green is looking good, no more cut off simulation, going to get this merged in!

PR #2476, opened 2025-03-07
refactor: rename ReconstructedAlert field

digitalcora:
In preparation for reconstructed alerts supporting the Pre-Fare solo template, `is_full_screen` is renamed, since this field will no longer directly determine whether an alert is placed in a "full screen" slot. The terminology of a "priority" alert is used instead.


PaulJKim (approved):
look good!

PR #2477, opened 2025-03-07
refactor: more targeted log capturing in tests

digitalcora:
`WidgetInstance.ReconstructedAlert` accepted a logging function as a second argument to `serialize`, to allow suppressing warnings in tests. The more conventional way to do this in ExUnit is `CaptureLog`.

* Switch to using `CaptureLog` in these tests (via the `:capture_log` tag which automatically wraps a test case, discarding the log output).
* Only capture logs for the specific cases we expect to log a warning.
* Use our `Report` module so these warnings are also logged to Sentry, proactively alerting us.


PaulJKim (approved):
Looks good!

PR #2478, opened 2025-03-07
refactor: ReconstructedAlert placement logic

digitalcora:
The goal of this change is to introduce a boundary between determining where a ReconstructedAlert is "placed" (dual screen, single screen, or flex zone) and serializing it. This opens the possibility of the same alert being placed differently depending on whether a screen is using the "duo" or "solo" template, though that is not implemented here.

* Introduce `placement/1` which makes the above-mentioned determination. Use this across various public widget functions to ensure consistency.
* Add missing fields to the "serialized response" typespecs.
* Make some functions private which were public but unused outside of this module.


robbie-sundstrom (approved):
Took me a few minutes to understand everything going on here, but the changes look good to me! This file is a behemoth though.

PR #2483, opened 2025-03-10
Added sync script between screen environment assets and json configs

deanshi:
**Asana task**: [Sync Screens non-prod environments with prod](https://app.asana.com/0/1185117109217413/1209417945686261)

- `sync_s3_env.sh` introduced to sync assets and configs
- allows for -D argument to delete files that exist in the destination env that do not in the source env
- currently uses `aws s3 sync` rather than `aws s3 cp` in order to keep files that are the same across the buckets (but overwrites if the timestamps are different)
- then just `aws s3 cp` the screens.json and pending_screens.json config file from mtd-ctd-config to the other bucket

- [ ] Tests added?


scripts/sync_s3_env.sh
```diff
@@ -0,0 +1,78 @@
...
+PENDING_SCREENS_DEST_JSON="$PENDING_SCREENS_JSON-$DEST_ENV.json"
+
+printf '%s' "Copy *$SOURCE_ENV* screens assets and configs into *$DEST_ENV*? "
+read -r answer
+case $answer in
+[Yy]*)
+  if [ "$DELETE_FLAG" = true ]; then
+    aws s3 sync "$SOURCE_BUCKET" "$DEST_BUCKET" --exact-timestamps --delete --dryrun
+  else
+    aws s3 sync "$SOURCE_BUCKET" "$DEST_BUCKET" --exact-timestamps --dryrun
```
digitalcora, line 72: Does this also set ACLs on each file to the same as the source file? In particular, we need `public-read` on all the media files (but ideally not the `LAST_DEPLOY` file that also exists in this bucket).

scripts/sync_s3_env.sh
```diff
@@ -0,0 +1,78 @@
...
+
+PENDING_SCREENS_SOURCE_JSON="$PENDING_SCREENS_JSON-$SOURCE_ENV.json"
+PENDING_SCREENS_DEST_JSON="$PENDING_SCREENS_JSON-$DEST_ENV.json"
+
+printf '%s' "Copy *$SOURCE_ENV* screens assets and configs into *$DEST_ENV*? "
+read -r answer
+case $answer in
+[Yy]*)
+  if [ "$DELETE_FLAG" = true ]; then
+    aws s3 sync "$SOURCE_BUCKET" "$DEST_BUCKET" --exact-timestamps --delete --dryrun
```
digitalcora, line 70: These all have the `--dryrun` flag, so will not do anything for real. Left over from testing?

scripts/sync_s3_env.sh
```diff
@@ -0,0 +1,78 @@
+#!/bin/sh
+
+# Syncs two environments S3 buckets and their config files.
+# Can only go from prod -> dev <-> dev-green
+
+set -eu
+
+SCREENS_BUCKET="s3://mbta-screens/screens"
+SCREENS_JSON="s3://mbta-ctd-config/screens/screens"
+PENDING_SCREENS_JSON="s3://mbta-ctd-config/screens/pending-screens"
```
digitalcora, line 10: nit: These variable names seem a bit misleading since they're more "prefixes" to an eventual full path. Maybe `SCREENS_BUCKET_PREFIX`, `SCREENS_CONFIG_PREFIX`, and `PENDING_SCREENS_PREFIX`?

scripts/sync_s3_env.sh
deanshi: Let me see if I can do something about the ACLs. 

There are definitely commands to change them to allow public-read, but let me see if I can get some details to see about LAST_DEPLOY or have some options there.

deanshi: okay - seems like a simple `--acl public-read` addition to sync will get us there and we can exclude the `LAST_DEPLOY` file from the initial sync (and sync it over right after with a different ACL).

But for context: What is `LAST_DEPLOY` for and is it something we want to copy over across environments? It's not environment specific right?

digitalcora:
I tried executing this from `prod` to `dev`, but it encountered an error when copying one of the files, and the subsequent commands to copy the configuration were not run:

```
An error occurred (AccessDenied) when calling the GetObjectTagging operation:
User: [redacted] is not authorized to perform: s3:GetObjectTagging
on resource: "arn:aws:s3:::mbta-screens/screens-prod/videos/youth-pass.mp4"
because no identity-based policy allows the s3:GetObjectTagging action
```

I'm not sure why this should require the `GetObjectTagging` permission or why only this particular file causes the problem. Could you investigate @deanshi?

deanshi:
> I tried executing this from `prod` to `dev`, but it encountered an error when copying one of the files, and the subsequent commands to copy the configuration were not run (though all other media files were copied, skipping over the error):
> 
> ```
> An error occurred (AccessDenied) when calling the GetObjectTagging operation:
> User: [redacted] is not authorized to perform: s3:GetObjectTagging
> on resource: "arn:aws:s3:::mbta-screens/screens-prod/videos/youth-pass.mp4"
> because no identity-based policy allows the s3:GetObjectTagging action
> ```
> 
> I'm not sure why this should require the `GetObjectTagging` permission or why only this particular file causes the problem. Could you investigate @deanshi?

Let me take a peek

digitalcora:
Per the discussion in Slack it looks like we'll eventually have permissions to get tagging, but it might be better to keep the `metadata-directive` switch for the sake of being able to start using this right away, and because we currently have no use for tags anyway. If you like we could add a `TODO` saying we should remove the switch once we have the permission.

digitalcora (approved):
Looking good!

PR #2485, opened 2025-03-10
Clean up phrasing of elevator widget audio on pre-fares

jacmrob:
**Asana task**: [Clean up phrasing of elevator widget audio on pre-fares](https://app.asana.com/0/1185117109217413/1209294078164616)

Only renders the additional closures messaging if there are actually additional closures. Added a test suite for this module. First elixir PR so let me know if this is the "elixir-y" way to go about this!  I think some of the functional programming mindset is starting to come back to me (I worked on an app in Clojure a looong time ago) :) 

- [x] Tests added?

### DIFF
--- a/lib/screens_web/views/v2/audio/elevator_status_view.ex
+++ b/lib/screens_web/views/v2/audio/elevator_status_view.ex
@@ -16,7 +16,7 @@ defmodule ScreensWeb.V2.Audio.ElevatorStatusView do
   end
 
   defp render_closures(list_pages, upcoming_at_home_pages, elsewhere_pages) do
-    if not Enum.all?(Enum.concat(list_pages, upcoming_at_home_pages, elsewhere_pages)) do
+    if Enum.empty?(list_pages) and Enum.empty?(upcoming_at_home_pages) and Enum.empty?(elsewhere_pages) do
       "All other MBTA elevators are working or have a backup elevator within 20 feet."
     else
       Enum.map(list_pages, &render_page/1)
--- a/lib/screens_web/views/v2/audio/elevator_status_view.ex
+++ b/lib/screens_web/views/v2/audio/elevator_status_view.ex
@@ -9,14 +9,21 @@ defmodule ScreensWeb.V2.Audio.ElevatorStatusView do
       }) do
     ~E|
     <p>Elevator Closures across the T.</p>
-    <p><%= render_cta() %></p>
     <p><%= render_active_at_home(active_at_home_pages) %></p>
     <p>Other elevator closures:</p>
-    <p><%= Enum.map(list_pages, &render_page/1) %></p>
-    <p><%= Enum.map(upcoming_at_home_pages, &render_page/1) %></p>
-    <p><%= Enum.map(elsewhere_pages, &render_page/1) %></p>
-    <p><%= render_cta() %></p>
+    <p><%= render_closures(list_pages, upcoming_at_home_pages, elsewhere_pages) %></p>
     |
+  end
+
+  defp render_closures(list_pages, upcoming_at_home_pages, elsewhere_pages) do
+    if not Enum.all?(Enum.concat(list_pages, upcoming_at_home_pages, elsewhere_pages)) do
+      "All other MBTA elevators are working or have a backup elevator within 20 feet."
+    else
+      Enum.map(list_pages, &render_page/1)
+      Enum.map(upcoming_at_home_pages, &render_page/1)
+      Enum.map(elsewhere_pages, &render_page/1)
+      render_cta()
+    end
   end
 
   defp render_cta do
--- a/lib/screens_web/views/v2/audio/elevator_status_view.ex
+++ b/lib/screens_web/views/v2/audio/elevator_status_view.ex
@@ -15,16 +15,12 @@ defmodule ScreensWeb.V2.Audio.ElevatorStatusView do
     |
   end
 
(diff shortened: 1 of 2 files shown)


lib/screens_web/views/v2/audio/elevator_status_view.ex
```diff
@@ -9,16 +9,24 @@ defmodule ScreensWeb.V2.Audio.ElevatorStatusView do
...
-    <p><%= Enum.map(upcoming_at_home_pages, &render_page/1) %></p>
-    <p><%= Enum.map(elsewhere_pages, &render_page/1) %></p>
-    <p><%= render_cta() %></p>
+    <p><%= render_closures(list_pages, upcoming_at_home_pages, elsewhere_pages) %></p>
     |
   end
 
+  defp render_closures(list_pages, upcoming_at_home_pages, elsewhere_pages) do
+    if Enum.empty?(list_pages) and Enum.empty?(upcoming_at_home_pages) and
+         Enum.empty?(elsewhere_pages) do
```
robbie-sundstrom, line 20: This isn't really necessary, more of a nit and practice with Elixir, but you could use pattern matching instead of having all these if statements. 
```
defp render_closures([], [], []) do
  "All other MBTA elevators are working or have a backup elevator within 20 feet."
end
```

Then the next function matching could be defined as `defp render_closures(list_pages, upcoming_at_home_pages, elsewhere_pages)` and handle the other logic

test/screens_web/views/v2/audio/elevator_status_view_test.exs
```diff
@@ -0,0 +1,77 @@
...
+  describe "Single elevator alert" do
+    test "renders no additional closures messages" do
+      assigns = %{
+        active_at_home_pages: test_stations(),
+        list_pages: [],
+        upcoming_at_home_pages: [],
+        elsewhere_pages: []
+      }
+
+      assert render(assigns) =~
```
robbie-sundstrom, line 27: I think this and the assert in the above test case should be an exact match check with `==`

test/screens_web/views/v2/audio/elevator_status_view_test.exs
```diff
@@ -0,0 +1,77 @@
+defmodule ScreensWeb.V2.Audio.ElevatorStatusViewTest do
+  use ScreensWeb.ConnCase, async: true
+  alias ScreensWeb.V2.Audio.ElevatorStatusView
+
+  describe "No elevator alerts" do
```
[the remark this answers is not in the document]
digitalcora: I'd agree in this case since each `describe` only contains a single test case right now (so the "grouping" isn't serving any organizational purpose). But will also note this isn't an iron-clad convention. The [documentation for `describe`](https://hexdocs.pm/ex_unit/1.18.3/ExUnit.Case.html#describe/2) contains examples of using it to organize both by the function under test, and by more abstract "scenarios". The latter can make more sense when e.g. a module only has a single public function, but a lot of complex behavior that can be sorted into various categories.

digitalcora:
@jacmrob You'll want to check out those CI failures; I think they point towards a broader issue with how you've implemented this, which isn't being caught by the test cases due to the partial matching @robbie-sundstrom pointed out (technically the test only verifies that the CTA is present). Happy to huddle on this if you'd like to sort it out sync.

Sidebar: Since in this repo we only run CI once a pull request is open (instead of on every push), it's possible to get surprised by a CI failure when you've already put up your PR for review. To avoid this I usually create my PRs as a "Draft" first, which runs CI but doesn't request review yet.

Screenshot 2025-03-11 at 10 54 15 AM

test/screens_web/views/v2/audio/elevator_status_view_test.exs
jacmrob: Because the strings have a lot of garbage characters I opted to just do multiple `=~` asserts on the important substrings for the tests, to make it more readable.

test/screens_web/views/v2/audio/elevator_status_view_test.exs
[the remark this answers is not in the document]
jacmrob: I made two `describe` cases -- one where there are no alerts and one where there are. lmk if this organization works!

test/screens_web/views/v2/audio/elevator_status_view_test.exs
```diff
@@ -0,0 +1,79 @@
...
+    test "with additional closures renders closure message + additional closures message" do
+      assigns = %{
+        active_at_home_pages: test_stations(),
+        list_pages: test_stations(),
+        upcoming_at_home_pages: [],
+        elsewhere_pages: []
+      }
+
+      assert render(assigns) =~ "take a different elevator :)"
+      assert render(assigns) =~ "For a full list of elevator alerts"
```
robbie-sundstrom, line 43: For this test, it would be nice to have at least two different test stations that are being checked for in the `render(assigns)` instead of the same elevator alert twice. But I see that the logic is working now so it's not strictly necessary

test/screens_web/views/v2/audio/elevator_status_view_test.exs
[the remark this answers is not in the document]
robbie-sundstrom: the updated test format make sense to me!

robbie-sundstrom (approved):
Overall looks great! Small suggestion on the unit test, but not necessary to update.

PR #2486, opened 2025-03-11
Fixed admin image upload failing with 400 error

deanshi:
**Asana task**: [Fix Screens admin UI image uploads](https://app.asana.com/0/1185117109217413/1209279595523788)

Description
- Seems like previously we were passing FormData all the way through without serializing it and attaching content-type to the headers, in the new code we attempt to stringify + add the content-type of `application/json`
- Added the `formData` function in `fetch` to allow passing data through without trying to serialize it into JSON and attaching a content-type to it

- [ ] Tests added?

### DIFF
--- a/assets/src/components/admin/admin_image_manager.tsx
+++ b/assets/src/components/admin/admin_image_manager.tsx
@@ -82,7 +82,7 @@ const ImageUpload = (): JSX.Element => {
     formData.append("image", stagedImageUpload, stagedImageUpload.name);
 
     try {
-      const result = await fetch.post("/api/admin/image", formData);
+      const result = await fetch.formData("/api/admin/image", formData);
 
       if (result.success) {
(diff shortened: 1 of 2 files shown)


digitalcora:
Good find, and this looks like a reasonable fix — is it tested on dev-green?

deanshi:
> Good find, and this looks like a reasonable fix — is it tested on dev-green?

Yep, looking good on dev-green. 

One thing to note, it all seems to go into the `psa` file. Is this intended?

digitalcora:
It's intentional that the Image Manager only handles images within the `psa/` prefix, yes. It could definitely be enhanced to work with all objects in this bucket if we wanted to add that as a feature, though maybe for other prefixes there are reasons we _don't_ want to be able to easily delete or overwrite those objects.

digitalcora (approved):
Glad this was an easy fix!

PR #2490, opened 2025-03-13
feat: alert widget support for solo Pre-Fares

digitalcora:
**Asana task**: https://app.asana.com/1/15492006741476/project/1185117109217413/task/1209255549956021

The existing alert widget "hierarchy" of dual screen > single screen > flex zone does not make sense on Pre-Fares that only have one screen to work with. We "adjust" placement for solo screens by moving all alerts down one level in this hierarchy (with some exceptions):

* Alerts that would have been dual-screen are single-screen.
* Alerts that would have been single-screen are (mostly) flex-zone.
* Alerts that would have been in the flex zone stay there.

This reorganizes some of the tests for this widget and adds assertions for solo screens to existing test cases. Initially most of these are placeholders whose main purpose is to ensure we don't crash in any known scenarios.

### DIFF
--- a/lib/screens/v2/widget_instance/reconstructed_alert.ex
+++ b/lib/screens/v2/widget_instance/reconstructed_alert.ex
@@ -366,6 +366,7 @@ defmodule Screens.V2.WidgetInstance.ReconstructedAlert do
   defp get_cause(:unknown), do: nil
   defp get_cause(cause), do: cause
 
+  # Special case for CR Departures; only supported on duo screens.
   defp placement(%__MODULE__{
          is_priority: true,
          screen: %Screen{
@@ -373,7 +374,8 @@ defmodule Screens.V2.WidgetInstance.ReconstructedAlert do
              cr_departures: %CRDepartures{
                enabled: true,
                pair_with_alert_widget: true
-             }
+             },
+             template: :duo
            }
          }
        }),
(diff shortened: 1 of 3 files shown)


lib/screens/v2/widget_instance/reconstructed_alert.ex
```diff
@@ -366,37 +366,58 @@ defmodule Screens.V2.WidgetInstance.ReconstructedAlert do
...
        do: :dual_screen,
        else: :single_screen
   end
 
-  defp placement(%__MODULE__{is_priority: true}), do: :single_screen
-  defp placement(%__MODULE__{}), do: :flex_zone
+  defp duo_placement(%__MODULE__{is_priority: true}, _location), do: :single_screen
+  defp duo_placement(%__MODULE__{}, _location), do: :flex_zone
+
+  # "Downgrade" placement by one level for solo screens. Exception: single-screen alerts that we
```
robbie-sundstrom, line 414: Nit: delete we at end of comment so this makes sense

robbie-sundstrom (approved):
Small comment nit, but overall this looks great! These PRs are being broken up in a way that's very understandable

Also love the updated tests with solo screen asserts in the existing test cases

lib/screens/v2/widget_instance/reconstructed_alert.ex
digitalcora: Huh, I thought this did make sense... the "we" refers to the screen here (the screen is what's "inside" the alert). I can reword it since it was evidently confusing!

robbie-sundstrom: ohhhhh I thought the screens was implied without the we. It's a weird thing to explain in a short comment lol

test/screens/v2/widget_instance/reconstructed_alert_test.exs
```diff
@@ -2157,6 +2258,11 @@ defmodule Screens.V2.WidgetInstance.ReconstructedAlertTest do
       }
 
       assert expected == ReconstructedAlert.serialize(widget)
+
+      # FIXME: flex zone version of this alert for solo screens should not misleadingly say
+      # "[RED LINE] No trains / Use shuttle bus". Don't actually downgrade to the flex zone?
```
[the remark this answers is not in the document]
digitalcora: I think I found the problem with this: We had "fall back to header text" logic for alerts we couldn't come up with a good "reconstructed" form for, but were only applying this in the case where a shuttle/suspension alert affected multiple routes. I expanded this to also apply in the case where we can't find a single set of "endpoints" to describe the extents of the alert, which is the case for multi-branch RL alerts. This is why this case erroneously said "No trains / Use shuttle bus" — it was supposed to say something like "No trains _between JFK/UMass and Ashmont_", but the "endpoints" part was missing.

PR #2491, opened 2025-03-13
refactor: migrate off deprecated Sass `@import`

digitalcora:
https://sass-lang.com/blog/import-is-deprecated/

* Replace `@import` with `@use`.

* This requires files using shared mixins or variables to directly `@use` the file that contains them, rather than relying on them being in a global namespace. Since this could otherwise require a fragile sequence of `../../..` in paths, introduce a Webpack alias that allows referencing a file relative to the root `css` directory.

* Reorganize imports such that it should be less ambiguous how to keep them organized moving forward. Use a small number of "groups" and alphabetize within each group.

### DIFF
--- a/assets/webpack.config.js
+++ b/assets/webpack.config.js
@@ -13,6 +13,7 @@ const common_export_body = {
     extensions: [".ts", ".tsx", ".js", ".jsx"],
     alias: {
       // Please also update the "paths" list in tsconfig.json when you add aliases here!
+      CSS: path.resolve(__dirname, "css"),
       Components: path.resolve(__dirname, "src/components"),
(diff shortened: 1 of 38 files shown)


assets/webpack.config.js
```diff
@@ -13,6 +13,7 @@ const common_export_body = {
     extensions: [".ts", ".tsx", ".js", ".jsx"],
     alias: {
       // Please also update the "paths" list in tsconfig.json when you add aliases here!
+      CSS: path.resolve(__dirname, "css"),
```
deanshi, line 16: 👍

deanshi (approved):
Looks good to me!

PR #2492, opened 2025-03-14
Rs/hanover departures widget

robbie-sundstrom:
**Asana task**: [Generate and populate departures widget](https://app.asana.com/1/15492006741476/project/1185117109217413/task/1209537234084968)

- Creates a new candidate generator for departures for the On Bus screen app
- CSS styling is just placeholder to be able to see the 2 departures that are being passed to the Frontend

- [X] Tests added?

### DIFF
--- a/lib/screens/v2/candidate_generator/on_bus.ex
+++ b/lib/screens/v2/candidate_generator/on_bus.ex
@@ -2,6 +2,7 @@ defmodule Screens.V2.CandidateGenerator.OnBus do
   @moduledoc false
 
   alias Screens.V2.CandidateGenerator
+  alias Screens.V2.CandidateGenerator.Widgets.OnBus
   alias Screens.V2.Template.Builder
   alias Screens.V2.WidgetInstance.Placeholder
 
@@ -26,9 +27,14 @@ defmodule Screens.V2.CandidateGenerator.OnBus do
   end
 
   @impl CandidateGenerator
-  def candidate_instances(_config, query_params) do
+  def candidate_instances(
+        config,
+        query_params,
+        departures_instances_fn \\ &OnBus.Departures.departures_instances/3
+      ) do
     [
-      fn -> body_instances(query_params) end
+      fn -> body_instances(query_params) end,
+      fn -> departures_instances_fn.(config, query_params.route_id, query_params.stop_id) end
     ]
     |> Task.async_stream(& &1.())
     |> Enum.flat_map(fn {:ok, instances} -> instances end)
--- a/lib/screens/v2/candidate_generator/on_bus/departures.ex
+++ b/lib/screens/v2/candidate_generator/on_bus/departures.ex
@@ -1,199 +1,80 @@
 defmodule Screens.V2.CandidateGenerator.Widgets.OnBus.Departures do
   @moduledoc false
 
-  alias Screens.Routes.Route
-  alias Screens.RouteType
-  alias Screens.V2.CandidateGenerator.Widgets.OnBus.Departures
   alias Screens.V2.Departure
   alias Screens.V2.WidgetInstance.Departures, as: DeparturesWidget
   alias Screens.V2.WidgetInstance.Departures.NormalSection
-  alias Screens.V2.WidgetInstance.{DeparturesNoData, DeparturesNoService, OvernightDepartures}
+  alias Screens.V2.WidgetInstance.{DeparturesNoData, DeparturesNoService}
   alias ScreensConfig.Screen
-  alias ScreensConfig.V2.Departures.Filters.RouteDirections
-  alias ScreensConfig.V2.Departures.Filters.RouteDirections.RouteDirection
-  alias ScreensConfig.V2.Departures.{Filters, Query, Section}
 
-  alias ScreensConfig.V2.OnBus
-
-  @type t :: %__MODULE__{
-          stop_id: Stop.id(),
-          route_id: Route.id()
-        }
-
-  @enforce_keys ~w[stop_id route_id]a
-  defstruct @enforce_keys
-
-  @type options :: [
-          departure_fetch_fn: Departure.fetch(),
-          disabled_modes_fn: (-> RouteType.t()),
-          post_process_fn: (Departure.result(), Screen.t() -> Departure.result() | :overnight),
-          route_fetch_fn: (Route.params() -> {:ok, [Route.t()]} | :error),
-          now: DateTime.t()
-        ]
+  @max_departure_results 2
 
   @type widget ::
           DeparturesNoData.t()
           | DeparturesNoService.t()
           | DeparturesWidget.t()
-          | OvernightDepartures.t()
+  @type options :: [
+          departure_fetch_fn: Departure.fetch(),
+          now: DateTime.t()
+        ]
 
-  @spec departures_instances(Screen.t(), options()) :: [widget()]
-  def departures_instances(%Screen{app_params: %app{}} = config, now, options \\ [])
-      when app in [BusEink, BusShelter, Busway, GlEink] do
-    disabled_modes =
-      Keyword.get(options, :disabled_modes_fn, &Screens.Config.Cache.disabled_modes/0).()
-
-    if screen_devops_mode(config) in disabled_modes do
-      [%DeparturesNoData{screen: config, show_alternatives?: false}]
-    else
-      do_departures_instances(
-        config,
-        disabled_modes,
-        Keyword.get(options, :departure_fetch_fn, &Departure.fetch/2),
-        Keyword.get(options, :post_process_fn, fn results, _config -> results end),
-        Keyword.get(options, :route_fetch_fn, &Route.fetch/1),
-        now
+  @spec departures_instances(Screen.t(), String.t(), String.t(), options()) :: [DeparturesWidget]
+  def departures_instances(config, route_id, stop_id, options \\ []) do
+    departures =
+      fetch_departures(
+        route_id,
+        stop_id,
+        Keyword.get(options, :departure_fetch_fn, &Departure.fetch/2)
       )
+
+    case departures do
+      {:error, _} ->
+        [%DeparturesNoData{screen: config, show_alternatives?: true}]
+
+      {:ok, departure_data} ->
+        [
+          %DeparturesWidget{
+            screen: config,
+            sections: [
+              %NormalSection{
+                rows: departure_data,
+                layout: %ScreensConfig.V2.Departures.Layout{
+                  base: nil,
+                  include_later: false,
+                  max: @max_departure_results,
+                  min: 0
+                },
+                header: %ScreensConfig.V2.Departures.Header{
+                  arrow: nil,
+                  read_as: nil,
+                  title: nil
+                }
+              }
+            ],
+            now: DateTime.utc_now(),
+            slot_names: [:main_content]
+          }
+        ]
     end
   end
 
-  defp do_departures_instances(%Departures{
-         route_id: route_id,
-         stop_id: stop_id
-       }) do
-    departures = fetch_departures(stop_id, route_id, departure_fetch_fn)
-    # |> post_process_fn.(config)
-    # |> post_process_no_data(&1, has_multiple_sections, route_fetch_fn)
-
-    departures
-    # departures_instance =
-    #   cond do
-    #     Enum.any?(sections_data, &(&1.result == :error)) ->
-    #       %DeparturesNoData{screen: config, show_alternatives?: true}
-
-    #     match?([%{result: :overnight}], sections_data) ->
-    #       %OvernightDepartures{}
-
-    #     true ->
-    #       sections =
-    #         Enum.map(sections_data, fn
-    #           %{section: %Section{header: header, layout: layout}, result: result} ->
-    #             %NormalSection{
-    #               rows: normal_section_rows(result),
-    #               layout: layout,
-    #               header: header
-    #             }
-    #         end)
-
-    #       %DeparturesWidget{screen: config, sections: sections, now: now}
-    #   end
-
-    # [departures_instance]
-  end
-
-  # When a section has no departures on a screen with multiple sections, populate it with a "no
-  # data" entry. This may include a "representative" route for the section, used to determine an
-  # icon to display alongside the message.
-  #
-  # NOTE: Assumes any given section is configured such that it only displays departures from a
-  # single route-type-or-subway-line. If there would be more than one route-type-or-subway-line,
-  # one is picked arbitrarily.
-  defp post_process_no_data({:ok, []}, section, true = _has_multiple_sections, route_fetch_fn) do
-    %Section{
-      query: %Query{
-        params: %Query.Params{route_ids: route_ids, route_type: route_type, stop_ids: stop_ids}
-      }
-    } = section
-
-    fetch_params =
-      Map.reject(
-        %{
-          limit: 1,
-          ids: route_ids,
-          route_types: if(route_type, do: [route_type], else: []),
-          stop_ids: stop_ids
-        },
-        fn {_key, value} -> value == [] end
-      )
-
-    route =
-      case route_fetch_fn.(fetch_params) do
-        {:ok, [route | _]} -> route
-        _ -> nil
-      end
-
-    {:no_data, route}
-  end
-
-  defp post_process_no_data(fetch_result, _, _, _), do: fetch_result
-
-  defp normal_section_rows({:ok, departures}), do: departures
-
-  # defp normal_section_rows({:no_data, route?}) do
-  #   [
-  #     %FreeTextLine{
-  #       icon: if(route?, do: Route.icon(route?), else: nil),
-  #       text: ["No departures currently available"]
-  #     }
-  #   ]
-  # end
-
-  @spec fetch_departures() :: Departure.result()
-  def fetch_departures(route_id, stop_id, departure_fetch_fn) do
-    fetch_params = %{route_id => route_id, stop_id => stop_id}
-    fetch_opts = %{:include_schedules => true}
+  defp fetch_departures(route_id, stop_id, departure_fetch_fn) do
+    fetch_params = %{:stop_ids => [stop_id]}
+    fetch_opts = [include_schedules: false]
 
     is_bidirectional = true
 
     with {:ok, departures} <- departure_fetch_fn.(fetch_params, fetch_opts) do
       {:ok,
        departures
-       #  |> filter_departures(filters, now) #TODO: Add back filters
+       |> filter_current_route(route_id)
        |> make_bidirectional(is_bidirectional)}
     end
   end
 
-  defp filter_departures(
-         departures,
-         %Filters{max_minutes: max_minutes, route_directions: route_directions},
-         now
-       ) do
-    departures
-    |> filter_by_time(max_minutes, now)
-    |> filter_by_route_direction(route_directions)
-  end
-
-  defp filter_by_time(departures, nil, _now), do: departures
-
-  defp filter_by_time(departures, max_minutes, now) do
-    latest_time = DateTime.add(now, max_minutes, :minute)
-    Enum.reject(departures, &(DateTime.compare(Departure.time(&1), latest_time) == :gt))
-  end
-
-  defp filter_by_route_direction(departures, %RouteDirections{
-         action: :include,
-         targets: targets
-       }) do
-    Enum.filter(departures, &departure_in_route_directions?(&1, targets))
-  end
-
-  defp filter_by_route_direction(departures, %RouteDirections{
-         action: :exclude,
-         targets: targets
-       }) do
-    Enum.reject(departures, &departure_in_route_directions?(&1, targets))
-  end
-
-  defp filter_by_route_direction(departures, nil) do
-    departures
-  end
-
-  defp departure_in_route_directions?(d, route_directions) do
-    route_direction(d) in route_directions
-  end
-
-  defp route_direction(d) do
-    %RouteDirection{route_id: Departure.route(d).id, direction_id: Departure.direction_id(d)}
+  # Only return departures that are not from the bus's current route in either direction
+  defp filter_current_route(departures, route_id) do
+    Enum.filter(departures, &(&1.prediction.trip.route_id != route_id))
   end
 
   # "Bidirectional" mode: take only the first departure, and the next departure in the opposite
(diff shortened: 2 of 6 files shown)


lib/screens/v2/candidate_generator/on_bus.ex
```diff
@@ -26,9 +27,14 @@ defmodule Screens.V2.CandidateGenerator.OnBus do
...
-  def candidate_instances(_config, query_params) do
+  def candidate_instances(
+        config,
+        query_params,
+        departures_instances_fn \\ &OnBus.Departures.departures_instances/3
+      ) do
     [
-      fn -> body_instances(query_params) end
+      fn -> body_instances(query_params) end,
+      fn -> departures_instances_fn.(config, query_params.route_id, query_params.stop_id) end
```
robbie-sundstrom, line 37: Noting that there needs to be a check on the `route_id` and `stop_id` in case they do not exist on `query_params`, but I'll get to that later

robbie-sundstrom: I am updating this to pass in the full `query_params` struct and then to pattern match on if `route_id` and `stop_id`

lib/screens/v2/candidate_generator/on_bus/departures.ex
```diff
@@ -0,0 +1,96 @@
...
+  end
+
+  # Only return departures that are not from the bus's current route in either direction
+  defp filter_current_route(departures, route_id) do
+    Enum.filter(departures, &(&1.prediction.route.id != route_id))
+  end
+
+  @spec process_response({:error, any()}, Screen.t(), DateTime.t()) :: [DeparturesNoData]
+  @spec process_response({:ok, []}, Screen.t(), DateTime.t()) :: [DeparturesNoService]
+  @spec process_response({:ok, [Departure.t()]}, Screen.t(), DateTime.t()) :: [DeparturesWidget]
```
digitalcora, line 58: Technically `[Departure.t()]` includes `[]` — you can use `nonempty_list(Departure.t())` to exclude the empty list.

lib/screens/v2/candidate_generator/on_bus/departures.ex
```diff
@@ -0,0 +1,96 @@
...
+       |> filter_current_route(route_id)}
+    end
+  end
+
+  # Only return departures that are not from the bus's current route in either direction
+  defp filter_current_route(departures, route_id) do
+    Enum.filter(departures, &(&1.prediction.route.id != route_id))
+  end
+
+  @spec process_response({:error, any()}, Screen.t(), DateTime.t()) :: [DeparturesNoData]
```
digitalcora, line 56: The type `[DeparturesNoData]` means this will return a list consisting of the literal atom `DeparturesNoData` — I think you meant `DeparturesNoData.t()` (and etc.)

lib/screens/v2/candidate_generator/on_bus/departures.ex
```diff
@@ -0,0 +1,96 @@
...
+  end
+
+  defp process_response({:ok, departure_data}, config, now) do
+    [
+      %DeparturesWidget{
+        screen: config,
+        sections: [
+          %NormalSection{
+            rows: departure_data,
+            layout: %ScreensConfig.V2.Departures.Layout{
```
digitalcora, line 74: You can just use `%Layout{}` and `%Header{}` here — all fields are optional and have reasonable defaults. This is what the DUP departures candidate generator currently does, for example, since the special DUP version of the frontend Departures widget doesn't support the fields anyway.

lib/screens/v2/candidate_generator/on_bus/departures.ex
```diff
@@ -0,0 +1,96 @@
...
+              base: nil,
+              include_later: false,
+              max: @max_departure_results,
```
digitalcora, line 77: Since the design has a fixed number of departures rather than being able to flex (like the multiple sections on Sectionals, which are the reason the Layout parameters exist), I think this could be better implemented as a limit on the backend, e.g. using `Enum.take`. Then we won't send a potentially large number of departures to the frontend only to throw most of them away.

lib/screens/v2/candidate_generator/on_bus/departures.ex
```diff
@@ -0,0 +1,96 @@
...
+          }
+        ],
+        now: now,
+        slot_names: [:main_content]
+      }
+    ]
+  end
+
+  defp build_no_data_widget(config) do
+    [%DeparturesNoData{screen: config, show_alternatives?: true}]
```
digitalcora, line 94: Including `show_alternatives?` here seems odd since, on the frontend, this is mapped to a generic widget that doesn't understand this prop.

digitalcora (commented):
I don't see any logic here for fetching departures at connecting stops, which we included in the spec for this task. For example, if a bus was heading for a busway at a subway station, subway departures would not be displayed because the trains aren't predicted for the same stopping location as the bus. I think it could be reasonable to bump this out to its own task if you want, due to the non-trivial complexity, but then we should update tasks and point values in Asana accordingly.

lib/screens/v2/candidate_generator/on_bus/departures.ex
robbie-sundstrom: Agreed this should be on the BE! I was thinking that could be done in the follow up filtering story, but it's so quick that I might as well add it now.

robbie-sundstrom:
@digitalcora I don't think it should be too much to add the connecting stops with this PR! That was my mistake as I thought it came in the followup filtering task.

digitalcora:
@robbie-sundstrom OK, cool. [Take a look here](https://groups.google.com/g/massdotdevelopers/c/5caKPkBV0A0/m/nanG4cjeAwAJ) for an explanation of how connecting stops works if you need.

lib/screens/v2/candidate_generator/on_bus/departures.ex
```diff
@@ -0,0 +1,96 @@
...
+  alias Screens.V2.ScreenData.QueryParams
+  alias Screens.V2.WidgetInstance.Departures, as: DeparturesWidget
+  alias Screens.V2.WidgetInstance.Departures.NormalSection
+  alias Screens.V2.WidgetInstance.{DeparturesNoData, DeparturesNoService}
+  alias ScreensConfig.Screen
+
+  @max_departure_results 2
+
+  @type widget ::
+          DeparturesNoData.t() | DeparturesNoService.t() | DeparturesWidget.t()
```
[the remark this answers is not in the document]
robbie-sundstrom: Huh my auto formatter doesn't adjust this and it's been very annoying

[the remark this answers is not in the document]
digitalcora: The official formatter (`mix format`, which ultimately the LSP and I think every other tool would be using) doesn't un-wrap lines that are currently wrapped but could fit on one line. i.o.w. it's not a "parse the code and fully re-emit it in a canonical form" formatter — there are multiple forms it will accept for syntactically identical code. I agree this _is_ very annoying but there's no way around it that I know of. Having your ruler set at 98 characters for Elixir files (which is roughly the breakpoint the formatter uses) helps to be able to eyeball cases where things could be un-wrapped.

lib/screens/v2/candidate_generator/on_bus.ex
```diff
@@ -26,9 +27,14 @@ defmodule Screens.V2.CandidateGenerator.OnBus do
...
   @impl CandidateGenerator
-  def candidate_instances(_config, query_params) do
+  def candidate_instances(
+        config,
+        query_params,
+        departures_instances_fn \\ &OnBus.Departures.departures_candidate/2
+      ) do
     [
-      fn -> body_instances(query_params) end
+      fn -> body_instances(query_params) end,
```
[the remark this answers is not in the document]
robbie-sundstrom: I'm deleting the implementation of displaying query params with the placeholder, but keeping the Placeholder for the No Data state since that ticket is coming later.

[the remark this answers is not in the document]
digitalcora: This change already seems to contain an (unstyled) no-data implementation, though? Is there any way to get this placeholder to display? No matter what comes out of `departures_instances_fn` it would always be a higher priority, I think.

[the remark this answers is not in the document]
robbie-sundstrom: you're totally right! Forgot I initially added that No Data component. Deleting the placeholder in the candidate generator

lib/screens/v2/candidate_generator/on_bus/departures.ex
```diff
@@ -0,0 +1,132 @@
...
+    end
+  end
+
+  @spec fetch_connecting_stops(String.t()) :: nonempty_list(String.t())
+  defp fetch_connecting_stops(stop_id) do
+    case @stop.fetch(%{ids: [stop_id], location_types: [0, 1]}, true) do
+      {:ok, stops} ->
+        Enum.flat_map(stops, &parent_stop_ids/1) ++
+          Enum.flat_map(stops, &connecting_stop_ids/1) ++
+          Enum.flat_map(stops, &child_stop_ids/1) ++
```
digitalcora, line 51: * We expect the stop ID in the query param to only ever be a specific stopping location on a trip (location type 0), so we could simplify away this specific line since "our" stop should never have child stops.
* Though the size of the lists involved here is small, it's good to keep in mind that each `++` has to iterate the contents of the left-hand side, taking time proportional to the number of elements. In a case like `a ++ b ++ c ++ d`, this has to traverse `a`, then all of `a ++ b`, then all of `a ++ b ++ c`, making multiple redundant traversals. Instead we could use `Enum.concat([a, b, c, d])`...
* ...but, since the point is to end up with a flat list of stop IDs, we could simplify even further to:

```elixir
List.flatten([stop_id, Enum.map(stops, &child_stop_ids/1), Enum.map(stops, &connecting_stop_ids/1)])
```

lib/screens/v2/candidate_generator/on_bus/departures.ex
```diff
@@ -0,0 +1,132 @@
...
+
+      :error ->
+        Report.warning("fetch_connecting_stops_error", stop_id: stop_id)
+        [stop_id]
+    end
+  end
+
+  @spec connecting_stop_ids(Stop.t()) :: [String.t()]
+  defp connecting_stop_ids(stop) do
+    stop |> Map.get(:connecting_stops) |> Enum.map(&stop_id(&1))
```
digitalcora, line 62: `stop` here is a struct so it may be confusing (and permit catching fewer errors at compile time) that we're fetching a field out of it as though it's a plain map. We could use pattern matching on the struct instead:

```elixir
defp connecting_stop_ids(%Stop{connecting_stops: stops}), do: Enum.map(stops, & &1.id)
```

This could also be applied to the other `_stop_ids` functions. (This is also a suggestion that we don't need the private function `stop_id` which just extracts the `id` field from stops, since doing so directly within the function capture can be equally or more concise.)

lib/screens/v2/candidate_generator/on_bus/departures.ex
```diff
@@ -0,0 +1,132 @@
...
+  end
+
+  @spec parent_stop_ids(Stop.t()) :: [String.t()]
+  defp parent_stop_ids(stop) do
+    parent_station =
+      stop
+      |> Map.get(:parent_station)
+
+    case parent_station do
+      nil ->
```
digitalcora, line 79: As suggested above we can use pattern-matching to extract the field we want from the stop, but we can go even further here:

```elixir
defp parent_stop_ids(%Stop{parent_station: nil}), do: []

defp parent_stop_ids(%Stop{parent_station: stop}) do
  # ...handle case where parent station exists
end
```

lib/screens/v2/candidate_generator/on_bus/departures.ex
```diff
@@ -0,0 +1,132 @@
...
+        []
+
+      # If a parent_station has any connecting stops, we also get those IDs
+      _ ->
+        parent_station
+        |> Enum.filter(fn stop -> stop.location_type == 0 end)
```
digitalcora, line 85: This doesn't seem like it should work, since `parent_station` will be a singular `Stop` struct, not a list (and in any case since it is a parent station there's no way its location type could be 0). Is there a test that covers this code path?

lib/screens/v2/candidate_generator/on_bus/departures.ex
```diff
@@ -0,0 +1,132 @@
...
+  end
+
+  defp stop_id(stop), do: stop.id
+
+  # Only return departures that are not from the bus's current route in either direction
+  defp filter_current_route(departures, route_id) do
+    Enum.filter(departures, &(&1.prediction.route.id != route_id))
+  end
+
+  @spec process_response(:error, Screen.t(), DateTime.t()) ::
```
digitalcora, line 98: nit: `process_response` is a somewhat generic name — it tells us vaguely what the function is "doing" but not what purpose it serves within the module or what it returns. I'd suggest something like `departures_instances`.

test/screens/stops/stop_test.exs
```diff
@@ -42,7 +42,8 @@ defmodule Screens.Stops.StopTest do
         fn "stops",
            %{
              "filter[id]" => "s1,p1,p2,c3",
-             "include" => "child_stops,parent_station.child_stops"
+             "include" =>
+               "child_stops.connecting_stops,parent_station.child_stops,parent_station.connecting_stops,connecting_stops,parent_station"
```
digitalcora, line 46: Although this updates the value we expect the fetch function to pass through to the API, it doesn't add any test coverage for the new parsing of `connecting_stops` (and this is not covered by any other tests since those will bypass this code entirely by injecting the fetch function). I think we should either augment the fake API response to include some connecting stops and assert that those too are parsed correctly, or add a new test case which does the same.

test/screens/v2/candidate_generator/on_bus/departures_test.exs
```diff
@@ -0,0 +1,192 @@
...
+        trip: %Trip{direction_id: direction_id},
+        arrival_time: arrival_time
+      }
+    }
+  end
+
+  defp build_stop(id, options \\ %{connecting_ids: [], child_ids: [], parent_id: nil}) do
+    [
+      %Stop{
+        :id => id,
```
digitalcora, line 53: I guess the formatter doesn't "fix" this but it shouldn't be necessary to use the long-hand map syntax in this case; like the other structs above, this can be `id: id` and etc. The only time when this should be required is for plain maps where the keys are not atoms (e.g. `%{"key" => "value"}`).

test/screens/v2/candidate_generator/on_bus/departures_test.exs
```diff
@@ -0,0 +1,192 @@
...
+    %Departure{
+      prediction: %Prediction{
+        route: %Route{id: route_id, type: route_type},
+        trip: %Trip{direction_id: direction_id},
+        arrival_time: arrival_time
+      }
+    }
+  end
+
+  defp build_stop(id, options \\ %{connecting_ids: [], child_ids: [], parent_id: nil}) do
```
digitalcora, line 50: This "API" is a bit awkward to use since, if you use any options, you have to pass them all (or else e.g. `options.child_ids` will crash with a nonexistent key error). We could instead have `options \\ %{}` and use `Map.get(options, :child_ids, [])`.

It's also somewhat conventional in Elixir to use keyword lists for these kinds of "options", since there is a special shorthand syntax available when calling a function with a keyword list as the final argument. In this case it would be `options \\ []` and `Keyword.get/3` instead, but otherwise the same.

```elixir
build_stop("100", child_ids: ["200"]) # shorthand keyword-list syntax
```

test/screens/v2/candidate_generator/on_bus/departures_test.exs
```diff
@@ -0,0 +1,192 @@
...
+  alias Screens.V2.WidgetInstance.DeparturesNoService
+  alias ScreensConfig.Screen
+  alias ScreensConfig.V2.OnBus
+
+  import Mox
+  import Screens.Inject
+
+  @stop injected(Stop)
+
+  defp build_config() do
```
digitalcora, line 23: nit: Since this function is argument-less and always returns the same value, it could be a module attribute instead (`@config %Screen{...}`).

lib/screens/stops/stop.ex
```diff
@@ -13,12 +13,15 @@ defmodule Screens.Stops.Stop do
...
     child_stops
+    connecting_stops
     platform_code
     platform_name
     vehicle_type
   ]a
 
   @type id :: String.t()
+  # A location_type of 1 indicates a parent station complex, whereas 0 indicates a distinct boarding location.
+  # A value of2 designates a station entrance/exit, and 3 indicates a generic node within a station, such as the end of a staircase, elevator, or escalator.
```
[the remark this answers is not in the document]
robbie-sundstrom: Agreed this would be great to eventually parse into descriptive atoms!

lib/screens/v2/candidate_generator/on_bus/departures.ex
```diff
@@ -0,0 +1,132 @@
...
+  @stop injected(Stop)
+
+  @max_departure_results 3
+
+  @type widget :: DeparturesNoData.t() | DeparturesNoService.t() | DeparturesWidget.t()
+
+  @spec departures_candidate(Screen.t(), QueryParams.t(), DateTime.t(), Departure.fetch()) :: [
+          widget()
+        ]
+  def departures_candidate(config, %{stop_id: stop_id, route_id: route_id}, now, fetch_fn) do
```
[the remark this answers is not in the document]
robbie-sundstrom: - Great point on the injection of the `Departure.fetch/2` function! That is updated
- I dislike the naming of `[blank]_instances` as the convention we have for candidate generator widgets, including for the plurality you mentioned. I'm leaving that as is in my next commit, but since it is the convention, I could definitely be convinced to change it to `departures_instances`

lib/screens/v2/candidate_generator/on_bus/departures.ex
robbie-sundstrom: Updated this to `departures_widget`, which I like much more than the original name

test/screens/v2/candidate_generator/on_bus/departures_test.exs
robbie-sundstrom: This `options` keyword list feels unnatural to me, but I must admit that it is very clean after converting this function to use it. Thanks for pointing this out, will keep it in mind in the future

lib/screens/v2/candidate_generator/on_bus/departures.ex
```diff
@@ -0,0 +1,132 @@
...
+    with {:ok, departures} <- departure_fetch_fn.(fetch_params, fetch_opts) do
+      {:ok,
+       departures
+       |> filter_current_route(route_id)}
+    end
+  end
+
+  @spec fetch_connecting_stops(String.t()) :: nonempty_list(String.t())
+  defp fetch_connecting_stops(stop_id) do
+    case @stop.fetch(%{ids: [stop_id], location_types: [0, 1]}, true) do
```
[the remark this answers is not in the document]
robbie-sundstrom: thank you for this explanation!! I misunderstood how this part of the API worked. 

To prevent us from calling the departures fetch API with way more stop IDs than we need to, I've added a filter to `child_stop_ids` -- ` |> Enum.filter(fn child -> child.location_type == 0 end)`

lib/screens/v2/candidate_generator/on_bus/departures.ex
robbie-sundstrom: This does not have a test case, but I'm adding one. I've updated the logic here as well, but I don't love it:

lib/screens/v2/candidate_generator/on_bus/departures.ex
robbie-sundstrom: pattern matching here is much cleaner!

test/screens/stops/stop_test.exs
robbie-sundstrom: good point, adding tests for this now!

robbie-sundstrom: updated this test to process connecting_stops!

lib/screens/v2/candidate_generator/on_bus/departures.ex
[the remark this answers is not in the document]
robbie-sundstrom: ah gotcha, that makes sense then! I'll update to `departures_candidates`

lib/screens/v2/candidate_generator/on_bus/departures.ex
digitalcora: Nice. Repeated `Enum.concat/2` does have the same problem as repeated `++` does — we can use `Enum.concat/1` to reduce the number of redundant iterations:

```
Enum.concat([[id], child_stop_ids(stop), connecting_stop_ids(stop)])
```

...though in this case the only actual redundant iteration that would happen would be on `[id]`, which doesn't really count.

lib/screens/v2/candidate_generator/on_bus/departures.ex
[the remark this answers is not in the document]
digitalcora: By definition, stops in `child_stops` have `location_type == 0`, so I don't think this filter will do anything in practice.

lib/screens/v2/candidate_generator/on_bus/departures.ex
robbie-sundstrom: Updated to `Enum.concat/1` here and in the test files! That's handy to know about.

digitalcora (approved):
It's been a journey (...an on-bus journey you might say), but thanks for the revisions, this is looking good! I have one last digression, feel free to respond however you like, including saying "enough of your digressions, this is going in" 😸

test/screens/v2/candidate_generator/on_bus/departures_test.exs
```diff
@@ -0,0 +1,250 @@
...
+      end)
+
+      assert departures_candidate(
+               @config,
+               %QueryParams{route_id: route_id, stop_id: stop_id}
+             ) == [%DeparturesNoService{screen: @config}]
+    end
+  end
+
+  describe "fetch_connecting_stops/1" do
```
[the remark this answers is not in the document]
robbie-sundstrom: I updated the function `fetch_connecting_stops` to be private, and followed your advice in option 1 to make all of these tests happen through the public `departures_candidates` function in their own `describe` block.

PR #2494, opened 2025-03-14
Fixed flex zones being cut off in Pre-fare simulations

deanshi:
**Asana task**: [Pre-fare flex zone is broken in Screenplay](https://app.asana.com/1/15492006741476/project/1185117109217413/task/1209673539034709?focus=true)

(correct me if my CSS knowledge is wrong here)

- Seems like we were transforming the div around the widget instead of just the widget, leading to some cut off widgets.
- Moves the transform into the widget itself rather than the surrounding div
- Adds some scaling logic for the solo/duo panels for simulations to prevent unneeded horizontal scrolling

Screenshot 2025-03-14 at 12 23 30 PM
Screenshot 2025-03-14 at 12 29 37 PM

- [ ] Tests added?

### DIFF
--- a/assets/css/v2/pre_fare/viewport.scss
+++ b/assets/css/v2/pre_fare/viewport.scss
@@ -1,5 +1,5 @@
 .screen-container,
-.simulation {
+.simulation-viewport {
   height: 1920px;
   overflow: hidden;
   // width of a single Pre-Fare display unit; "duo" installations are two of
--- a/assets/src/components/v2/pre_fare/simulation_screen_container.tsx
+++ b/assets/src/components/v2/pre_fare/simulation_screen_container.tsx
@@ -45,7 +45,7 @@ const SimulationScreenLayout: ComponentType<SimulationScreenLayoutProps> = ({
           <div className="simulation__full-page">
             <div className="simulation__title">Live view</div>
             <div
-              className={classWithModifier("simulation", getScreenSide())}
+              className={`simulation ${classWithModifier("simulation-viewport", getScreenSide())}`}
               id="simulation"
             >
               <WidgetTreeErrorBoundary>


assets/css/v2/pre_fare/viewport.scss
```diff
@@ -1,10 +1,20 @@
...
+
+.screen-container {
+  // width of a single Pre-Fare display unit; "duo" installations are two of
+  // these side-by-side
+  --panel-width: #{$panel-width};
+}
+
+.simulation {
+  // width of a single Pre-Fare display unit scaled down by 17.92%
+  --panel-width: #{$panel-width * 0.1792};
```
digitalcora, line 11: This seems like a more fragile solution than what we had before, and I wonder if we need to do it this way (encoding the percentage scale-down in two different places). I'm also confused as to why this specific change fixes the flex zone items, since this doesn't touch anything outside the `.simulation` element, which only includes the "main" screen view. Maybe you can walk me through it at some point today?

assets/css/v2/pre_fare/viewport.scss
```diff
@@ -1,23 +1,23 @@
...
-.simulation {
+#simulation {
   height: 1920px;
   overflow: hidden;
   // width of a single Pre-Fare display unit; "duo" installations are two of
   // these side-by-side
   --panel-width: 1080px;
 
-  &--duo {
+  &.simulation--duo {
```
digitalcora, line 9: After we chatted, I realized this will now no longer work for the _normal_ screen page, because it results in a selector like `.screen-container.simulation--duo` (which will never match anything) instead of the intended `.screen-container--duo`. I think what we need to do is:

1. Keep these modifiers the same (e.g. `&--duo`)
2. Change line 2 to be something like `.simulation-viewport`
3. Change `pre_fare/simulation_screen_container.tsx` to _additionally_ apply this classname, with the modifier, to the element

The result should be:

```
<div id="simulation" class="simulation simulation-viewport simulation-viewport--duo">
```

assets/src/components/v2/pre_fare/simulation_screen_container.tsx
```diff
@@ -45,7 +45,7 @@ const SimulationScreenLayout: ComponentType<SimulationScreenLayoutProps> = ({
           <div className="simulation__full-page">
             <div className="simulation__title">Live view</div>
             <div
-              className={classWithModifier("simulation", getScreenSide())}
+              className={`simulation + ${classWithModifier("simulation-viewport", getScreenSide())}`}
```
digitalcora, line 48: This is mixing string concatenation and interpolation; the result will be

```
<div class="simulation + simulation-viewport simulation-viewport--duo" ...>
```

In addition to the intended classes, the element has a class of `+` (technically valid, but not useful). I think you meant one of these two:

```
className={`simulation ${classWithModifier("simulation-viewport", getScreenSide())}`}
```

```
className={"simulation " + classWithModifier("simulation-viewport", getScreenSide())}
```

digitalcora (approved):
One more small comment, looks good!

PR #2495, opened 2025-03-14
UI for the "all elevators working" state

jacmrob:
**Asana task**: [No elevator alerts](https://app.asana.com/1/15492006741476/project/1185117109217422/task/1209256677260370?focus=true)

Description
When the elevator alerts array is empty, display the "all elevators working screen." If there is an upcoming outage at the elevator, display that in a smaller bar on top.  One issue -- I wasn't sure how to remove the footer from these screens, or if I should. The mock-ups don't have a footer.

 Screenshots added below: 

Screenshot 2025-03-14 at 12 58 25 PM

Screenshot 2025-03-14 at 12 57 49 PM

- [ ] Tests added?

### DIFF
--- a/assets/src/components/v2/elevator/closures.tsx
+++ b/assets/src/components/v2/elevator/closures.tsx
@@ -217,9 +217,11 @@ const sortStations = (
     }
   });
 
+type ClosuresStatus = "no_closures";
+
 interface Props extends WrappedComponentProps {
   id: string;
-  stations_with_closures: StationWithClosures[];
+  stations_with_closures: StationWithClosures[] | ClosuresStatus;
   station_id: string;
   upcoming_closure?: UpcomingClosureInfo;
 }
@@ -230,9 +232,11 @@ const Closures = ({
   upcoming_closure: upcomingClosure,
   updateVisibleData,
 }: Props) => {
-  const numClosuresInStation = stations
-    .filter((s) => s.id === stationId)
-    .flatMap((s) => s.closures).length;
+  const numClosuresInStation =
+    typeof stations === "string"
+      ? 0
+      : stations.filter((s) => s.id === stationId).flatMap((s) => s.closures)
+          .length;
 
   const listRef = useRef<HTMLDivElement>(null);
 
@@ -290,7 +294,9 @@ const Closures = ({
 
   return (
     <div className="elevator-closures">
-      {stations.length > 0 ? (
+      {stations === "no_closures" ? (
+        <NoCurrentClosures closure={upcomingClosure} />
+      ) : (
         <>
           <InStationSummary
             numClosures={numClosuresInStation}
@@ -333,8 +339,6 @@ const Closures = ({
             </div>
           </div>
         </>
-      ) : (
-        <NoCurrentClosures closure={upcomingClosure} />
       )}
       {numPages > 1 && (
         <div className="paging-info-container">
--- a/assets/src/components/v2/elevator/closures.tsx
+++ b/assets/src/components/v2/elevator/closures.tsx
@@ -10,6 +10,7 @@ import { type StationWithClosures } from "Components/v2/elevator/types";
 import useIntervalPaging from "Hooks/v2/use_interval_paging";
 import CalendarIcon from "Images/svgr_bundled/calendar.svg";
 import NormalServiceIcon from "Images/svgr_bundled/normal-service.svg";
+import CalenderAlertIcon from "Images/svgr_bundled/calendar-alert.svg";
 import AccessibilityAlert from "Images/svgr_bundled/accessibility-alert.svg";
 import Logo from "Images/svgr_bundled/logo.svg";
 import { hasOverflowX } from "Util/utils";
@@ -141,7 +142,7 @@ const UpcomingClosure = ({
 
   return (
     <div className="upcoming-closure">
-      <AccessibilityAlert height={249} width={249} />
+      <CalenderAlertIcon height={249} width={249} />
       <div className="upcoming-closure__title" ref={titleRef}>
         {title}:
       </div>
@@ -166,13 +167,12 @@ const NoCurrentClosures = ({ closure }: { closure?: UpcomingClosureInfo }) => {
     <>
       {closure ? (
         <div className="closures-info">
-          <div className="small-info-strip">
+          <div className="in-station-summary">
             <div>All MBTA elevators are working.</div>
             <div>
               <NormalServiceIcon width={72} height={72} fill="#145A06" />
             </div>
           </div>
-          <div className="small-info-strip__divider" />
           <UpcomingClosure closure={closure} />
         </div>
       ) : (
--- a/assets/src/components/v2/elevator/closures.tsx
+++ b/assets/src/components/v2/elevator/closures.tsx
@@ -162,61 +162,21 @@ const UpcomingClosure = ({
   );
 };
 
-const NoCurrentClosures = ({ closure }: { closure?: UpcomingClosureInfo }) => {
-  const getClosureTitle = (titles: string[]) => {
-    const titlesParsed = titles[0].split(" ");
-    if (titlesParsed[0] === "This") {
-      return (
-        <b>
-          {titlesParsed[0]} {titlesParsed[1]}:{" "}
-        </b>
-      );
-    }
-    return;
-  };
-  const getClosureDetail = (titles: string[]) => {
-    const titlesParsed = titles[0].split(" ");
-    if (titlesParsed[0] === "This") {
-      return titlesParsed.slice(1);
-    } else {
-      return titlesParsed.join(" ");
-    }
-  };
-
+const NoCurrentClosures = () => {
   return (
-    <>
-      {closure && (
-        <div className="closures-info">
-          <div className="upcoming-closure-small">
-            <div>
-              {getClosureTitle(closure.details.titles)}
-              This elevator will be closed{" "}
-              {getClosureDetail(closure.details.titles)}{" "}
-              {closure.details.postfix}
-            </div>
-            <div>
-              <CalendarIcon width={72} height={72} />
-            </div>
-          </div>
-          <div className="closure-divider" />
-        </div>
-      )}
-      <div className="no-closures">
-        <NormalServiceIcon height={150} width={150} fill="#145A06" />
-        <div className="no-closures-header">
-          All MBTA elevators are working.
-        </div>
-        <svg height={4} width={936} className="divider">
-          <line x1="0" y1="4" x2={936} y2={4} />
-        </svg>
-        <div className="no-closures-text">
-          For info on elevator outages and alternate paths:{" "}
-          <b>mbta.com/elevators</b> or call the elevator hotline:{" "}
-          <b>617-222-2828</b>
-        </div>
-        {!closure && <Logo height={124} width={124} />}
+    <div className="no-closures">
+      <NormalServiceIcon height={150} width={150} fill="#145A06" />
+      <div className="no-closures-header">All MBTA elevators are working.</div>
+      <svg height={4} width={936} className="divider">
+        <line x1="0" y1="4" x2={936} y2={4} />
+      </svg>
+      <div className="no-closures-text">
+        For info on elevator outages and alternate paths:{" "}
+        <b>mbta.com/elevators</b> or call the elevator hotline:{" "}
+        <b>617-222-2828</b>
       </div>
-    </>
+      <Logo height={124} width={124} />
+    </div>
   );
 };
 
@@ -317,7 +277,7 @@ const Closures = ({
   return (
     <div className="elevator-closures">
       {stations === "no_closures" ? (
-        <NoCurrentClosures closure={upcomingClosure} />
+        <NoCurrentClosures />
       ) : (
         <>
           <InStationSummary
--- a/assets/src/components/v2/elevator/closures.tsx
+++ b/assets/src/components/v2/elevator/closures.tsx
@@ -162,21 +162,66 @@ const UpcomingClosure = ({
   );
 };
 
-const NoCurrentClosures = () => {
+const NoCurrentClosures = ({ closure }: { closure?: UpcomingClosureInfo }) => {
+  const getClosureTitle = (titles: string[]) => {
+    const titlesParsed = titles[0].split(" ");
+    return (
+      <b>
+        {titlesParsed[0]} {titlesParsed[1]}:{" "}
+      </b>
+    );
+  };
+  const getClosureDetail = (titles: string[]) => {
+    const titlesParsed = titles[0].split(" ");
+    if (titlesParsed[0] === "This") {
+      return titlesParsed.slice(1);
+    } else {
+      return titlesParsed.join(" ");
+    }
+  };
+
   return (
-    <div className="no-closures">
-      <NormalServiceIcon height={150} width={150} fill="#145A06" />
-      <div className="no-closures-header">All MBTA elevators are working.</div>
-      <svg height={4} width={936} className="divider">
-        <line x1="0" y1="4" x2={936} y2={4} />
-      </svg>
-      <div className="no-closures-text">
-        For info on elevator outages and alternate paths:{" "}
-        <b>mbta.com/elevators</b> or call the elevator hotline:{" "}
-        <b>617-222-2828</b>
-      </div>
-      <Logo height={124} width={124} />
-    </div>
+    <>
+      {closure ? (
+        <div className="closures-info">
+          <div className="small-info-strip">
+            <div>All MBTA elevators are working.</div>
+            <div>
+              <NormalServiceIcon width={72} height={72} fill="#145A06" />
+            </div>
+          </div>
+          <div className="small-info-strip__divider" />
+          <div className="upcoming-closure">
+            <AccessibilityAlert height={249} width={249} />
+            <div className="maintenance-header">
+              {getClosureTitle(closure.details.titles)}
+            </div>
+            <div className="no-closures__text">
+              This elevator will be closed{" "}
+              {getClosureDetail(closure.details.titles)}{" "}
+              {closure.details.postfix}.
+              <br />
+              <br />
+              {closure.details.summary}
+            </div>
+          </div>
+        </div>
+      ) : (
+        <div className="no-closures">
+          <NormalServiceIcon height={150} width={150} fill="#145A06" />
+          <div className="no-closures__header">
+            All MBTA elevators are working.
+          </div>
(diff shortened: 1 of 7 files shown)


dmaltzan:
@jacmrob for what it's worth, I would prefer if we could remove the footer

jacmrob:
New screenshots without the footer. I also modified the wording of the upcoming closures bar to be more readable/in-line with the mocks, represented below: 
Screenshot 2025-03-17 at 2 49 00 PM
Screenshot 2025-03-17 at 3 07 08 PM
Screenshot 2025-03-17 at 3 09 23 PM

jacmrob:
@digitalcora oops I think I was looking at the wrong mocks in figma. Just removed it.

lib/screens/v2/candidate_generator/elevator/closures.ex
```diff
@@ -65,14 +65,31 @@ defmodule Screens.V2.CandidateGenerator.Elevator.Closures do
     active_closures = Enum.flat_map(active, &elevator_closure/1)
     at_this_elevator? = fn %Closure{id: id} -> id == elevator_id end
 
+    {:ok, %Stop{id: stop_id}} = @facility.fetch_stop_for_facility(elevator_id)
+
+    relevant_closures =
+      active_closures |> Enum.filter(&relevant_closure?(&1, stop_id, active_closures))
```
digitalcora, line 71: Conventional Elixir tends not to use a pipe operator when only a single function is being piped-into:

```elixir
Enum.filter(active_closures, &relevant_closure?(&1, stop_id, active_closures))
```

lib/screens/v2/candidate_generator/elevator/closures.ex
```diff
@@ -65,14 +65,31 @@ defmodule Screens.V2.CandidateGenerator.Elevator.Closures do
...
       nil ->
         upcoming_closures =
           upcoming |> Enum.flat_map(&elevator_closure/1) |> Enum.filter(at_this_elevator?)
 
         [
-          elevator_closures(active_closures, upcoming_closures, app_params, now)
-          | header_footer_instances(config, now)
+          elevator_closures(
+            relevant_closures,
+            active_closures,
```
digitalcora, line 81: I think we should only be passing in `relevant_closures` here, so all logic is consistently using the same set of closures.

lib/screens/v2/candidate_generator/elevator/closures.ex
```diff
@@ -197,7 +261,7 @@ defmodule Screens.V2.CandidateGenerator.Elevator.Closures do
             station_closures,
             fn %Closure{id: id, name: name} -> %WidgetClosure{id: id, name: name} end
           ),
-        summary: active_summary(station_closures, closures)
+        summary: active_summary(station_closures, active_closures)
```
digitalcora, line 264: The value here was previously "relevant closures", but now is _all_ active closures. It should stay `relevant_closures` so the existing logic of this function remains the same.

lib/screens/v2/widget_instance/elevator_closures.ex
```diff
@@ -20,7 +20,7 @@ defmodule Screens.V2.WidgetInstance.ElevatorClosures do
             id: Stop.id(),
             name: String.t(),
             route_icons: list(Route.icon()),
-            closures: list(Closure.t()),
+            closures: list(Closure.t()) | :no_closures,
```
digitalcora, line 23: I don't think it makes sense to also add this value to an individual `Station` — we never generate this, and the frontend would not be able to do anything with it. Due to the way we construct `Station`s, the list of closures is always non-empty. (This does point to the fact that we could be more specific in this type and say `nonempty_list(Closure.t())`, but since that change would be unrelated to this feature, I'd save it for a later time.)

lib/screens/v2/widget_instance/elevator_closures.ex
```diff
@@ -119,7 +119,7 @@ defmodule Screens.V2.WidgetInstance.ElevatorClosures do
   @type t :: %__MODULE__{
           app_params: Elevator.t(),
           now: DateTime.t(),
-          stations_with_closures: list(Station.t()),
+          stations_with_closures: list(Station.t()) | :no_closures,
```
digitalcora, line 122: We could be more specific here now and say `nonempty_list(Station.t())`, since if the list were empty, we would instead use `:no_closures`.

test/screens/v2/candidate_generator/elevator/closures_test.exs
```diff
@@ -71,27 +71,63 @@ defmodule Screens.V2.CandidateGenerator.Elevator.ClosuresTest do
...
+
+        {:ok, alerts}
+      end)
+
+      assert [_current_closed, %NormalHeader{variant: :closed}, %Footer{variant: :closed}] =
+               Generator.elevator_status_instances(@screen, now)
+    end
+  end
+
+  describe "when all elevators are working" do
```
digitalcora, line 90: The method of organization here is getting a little mixed-up: The existing `describe "header and footer"` and `describe "alternate path widget"` and etc. are grouping tests related to specific widgets, but the new `describe "when all elevators are working"` is grouping tests related to a specific data scenario. I think we should try to stick with one or the other, and it feels like the current scheme can accommodate this:

1. The case currently on L91 could go under `header and footer` — this asserts there is no footer at all when there are no relevant closures, and does not assert anything about the closures widget
2. The case currently on L98 could go under `closure list widget` — the point of this is to assert on the `:no_closures` behavior so I think the bits about upcoming closures could be cut (that continues to work exactly as we'd expect from other test cases, there's nothing exceptional about it in this case)

assets/src/components/v2/elevator/closures.tsx
```diff
@@ -180,9 +199,11 @@ const sortStations = (
     }
   });
 
+type ClosuresStatus = "no_closures";
```
digitalcora, line 202: Since this is only used in one place, I don't think we need a separate type definition for it; the prop could be defined as

```ts
stations_with_closures: StationWithClosures[] | "no_closures";
```

This way there is no cross-referencing needed to understand all possible values.

assets/src/components/v2/elevator/closures.tsx
```diff
@@ -161,6 +162,24 @@ const UpcomingClosure = ({
...
 };
 
+const NoCurrentClosures = () => {
+  return (
+    <div className="no-closures">
+      <NormalServiceIcon height={150} width={150} fill="#145A06" />
+      <div className="no-closures-header">All MBTA elevators are working.</div>
+      <svg height={4} width={936} className="divider">
+        <line x1="0" y1="4" x2={936} y2={4} />
+      </svg>
```
digitalcora, line 172: This feels very specific to the dimensions/margins/etc. of the current screen layout, and I don't think it's necessary to do it that way — what if we replaced this entire `svg` element with:

```html
<hr />
```

and used CSS for the color and thickness?

assets/src/components/v2/elevator/closures.tsx
```diff
@@ -161,6 +162,24 @@ const UpcomingClosure = ({
   );
 };
 
+const NoCurrentClosures = () => {
+  return (
+    <div className="no-closures">
+      <NormalServiceIcon height={150} width={150} fill="#145A06" />
+      <div className="no-closures-header">All MBTA elevators are working.</div>
```
digitalcora, line 169: nit: We've been (somewhat haphazardly) following a BEM-style naming convention for classes, where discrete elements of a single component are named as e.g. `no-closures__header` (`no-closures` is the "block", `header` is the "element"). This makes it a little easier to understand the nesting of elements in cases where they have a fixed nesting. If an SCSS selector for an element is nested inside the selector for its block, you can also shorthand to e.g. `&__header`. An example of where this is done already on elevator screens is `.upcoming-closure__title`.

assets/src/components/v2/elevator/closures.tsx
```diff
@@ -253,41 +276,52 @@ const Closures = ({
...
-        {isUpcomingClosurePage ? (
-          <UpcomingClosure closure={upcomingClosure} />
-        ) : (
-          <div className="header-container">
-            <div className="header">
-              <div className="header__title">MBTA Elevator Closures</div>
-              <div>
-                <AccessibilityAlert height={128} width={155} />
+      {stations === "no_closures" ? (
+        <NoCurrentClosures />
```
digitalcora, line 280: This doesn't seem to include the design state mentioned in Asana where all elevators are working but there is an upcoming closure of the current elevator — with this current code, if all elevators are working, we unconditionally display the full-page NoCurrentClosures message, and there is no indication of an upcoming closure if there is one.

lib/screens/v2/candidate_generator/elevator/closures.ex
```diff
@@ -86,14 +103,44 @@ defmodule Screens.V2.CandidateGenerator.Elevator.Closures do
...
       %NormalHeader{text: "Elevator #{elevator_id}", screen: config, time: now, variant: variant},
       %Footer{screen: config, variant: variant}
     ]
   end
 
+  defp header_footer_instances(
+         %Screen{app_params: %ElevatorConfig{elevator_id: elevator_id}} = config,
+         now,
+         variant,
+         closures
```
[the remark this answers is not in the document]
jacmrob: That works for me. I had thought initially that having multiple functions was the "elixir-y" way to do a conditional, even if each was only used once for a particular condition. how do you typically handle situations in elixir that would be an if/else or some other conditional blocking in non-functional code?

lib/screens/v2/candidate_generator/elevator/closures.ex
jacmrob: I had thought so too but this led to some test failures -- see my comment below

lib/screens/v2/candidate_generator/elevator/closures.ex
jacmrob: Actually the value passed here was pre-filtered closures (when `closures` first arrived to the function it was only active closures, and that filtering was done earlier in the pipeline). Using `relevant_closures` here caused some test failures because I guess we weren't able to access the full summary given on the closure. (L409 in `closures_test.exs`)

lib/screens/v2/widget_instance/elevator_closures.ex
jacmrob: I think I was confused on what this data model was for -- you're right it doesn't belong here and I'd vote just to revert this to what it was :)

assets/src/components/v2/elevator/closures.tsx
jacmrob: I made this with the plan in mind of adding the "no relevant closures" case as another keyword (the other half of this ticket). That way the props def could remain `StationWithClosures[] | ClosuresStatus`. But I agree it isn't strictly necessary and maybe just adds boilerplate

assets/src/components/v2/elevator/closures.tsx
jacmrob: yeah I initially thought that the upcoming closure for this elevator was to be displayed in a smaller header above the "all clear" message, but I was looking at the wrong mocks 🤦 . I'll update with the future closure state following that mock.

assets/src/components/v2/elevator/closures.tsx
jacmrob: sounds good. I'll probably just do it as a div.

assets/src/components/v2/elevator/closures.tsx
jacmrob: cool I'll use that convention. thanks for explaining!

jacmrob:
New screen for no closures + upcoming closure at current station case:

Screenshot 2025-03-19 at 1 16 43 PM

lib/screens/v2/candidate_generator/elevator/closures.ex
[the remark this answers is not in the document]
digitalcora: > how do you typically handle situations in elixir that would be an if/else or some other conditional blocking in non-functional code?

Well, Elixir has if/else 😄 But here it's more the duplication I wanted to address rather than the specific approach to the conditional logic — it's easy to imagine a version of this that used `if` or `case` or `cond` all within a single function body, but still had the duplication.

In terms of extracting different branches of a conditional into separate functions, I think it's basically a judgment call on how much it helps or harms the readability of the code, with the operations either co-located or abstracted behind a function call; it's never _necessary_ to do one or the other, given the tools Elixir has for conditional logic. Is the output of the process nearly the same in all conditional branches, with just a couple differing fields? Maybe keep everything together and use local variables or put conditionals directly in the field assignments. Is the output _very_ different depending on which branch is chosen? Maybe put each one behind a descriptively-named function call.

digitalcora:
Looking at the designs for this state, I think some spacing and/or font sizes might have changed since the upcoming-closure component was originally implemented:

Design | This Branch
--- | ---
Screenshot 2025-03-20 at 12 15 40 PM | Screenshot 2025-03-20 at 12 16 13 PM

In particular it's possible for the "accessible route" text to take up 3 lines _and_ for the date description at the bottom to take up 2 lines, and if either of those happened here I think there would be some clipping. Can you take a look at the spacing and sizing in this design and update the upcoming-closure styles (universally) to match?

lib/screens/v2/candidate_generator/elevator/closures.ex
```diff
@@ -65,14 +65,33 @@ defmodule Screens.V2.CandidateGenerator.Elevator.Closures do
...
+    relevant_closures =
+      Enum.filter(active_closures, &relevant_closure?(&1, stop_id, active_closures))
+
     case Enum.find(active_closures, at_this_elevator?) do
       nil ->
         upcoming_closures =
           upcoming |> Enum.flat_map(&elevator_closure/1) |> Enum.filter(at_this_elevator?)
 
+        footer =
+          if relevant_closures ++ upcoming_closures == [], do: [], else: [%Footer{screen: config}]
```
digitalcora, line 79: The `++` operator has to iterate through the entire left-hand side, so using a pattern like this to check for emptiness can be slow when the list is large (relatively speaking; it could probably have thousands of items before you'd really notice, since list iteration itself is extremely fast). The more idiomatic way to handle this would be `relevant_closures == [] and upcoming_closures == []`, or possibly `Enum.empty?(relevant_closures) and Enum.empty?(upcoming_closures)` if that reads better to you.

lib/screens/v2/candidate_generator/elevator/closures.ex
```diff
@@ -132,12 +159,30 @@ defmodule Screens.V2.CandidateGenerator.Elevator.Closures do
   defp elevator_closure(_alert), do: []
 
   defp elevator_closures(
+         [],
+         _active_closures,
+         upcoming_closures,
+         %ElevatorConfig{elevator_id: _elevator_id} = app_params,
```
digitalcora, line 165: The destructuring here is now entirely unused — this line could be replaced with just `app_params,`.

lib/screens/v2/candidate_generator/elevator/closures.ex
digitalcora: Oh, my bad, this does make sense: When displaying accessibility summaries, we need to know if any of a given elevator's backup elevators are also closed, and that's independent of whether or not the backup elevator was considered "relevant" for the purposes of building the overall closure list. I misread this. Thanks!

assets/src/components/v2/elevator/closures.tsx
```diff
@@ -141,7 +141,7 @@ const UpcomingClosure = ({
 
   return (
     <div className="upcoming-closure">
-      <CalendarAlertIcon width={224} />
+      <AccessibilityAlert height={249} width={249} />
```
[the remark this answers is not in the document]
digitalcora: Looks like Snehal decided to change this back to the original "calendar alert" for now to keep things consistent, so we can revert this change.

jacmrob:
Design | This Branch
--- | ---
| Screenshot 2025-03-21 at 1 05 52 PM | Screenshot 2025-03-21 at 1 07 08 PM |

assets/src/components/v2/elevator/closures.tsx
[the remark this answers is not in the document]
jacmrob: yep saw that 👍

assets/src/components/v2/elevator/closures.tsx
```diff
@@ -141,7 +142,7 @@ const UpcomingClosure = ({
 
   return (
     <div className="upcoming-closure">
-      <CalendarAlertIcon width={224} />
+      <CalenderAlertIcon height={249} width={249} />
```
digitalcora, line 145: Should we also revert the size change here? It's 224 in the design and is non-square (though it's actually 224 in _height_ rather than width, which might be a mistake in the original code).

test/screens/v2/candidate_generator/elevator/closures_test.exs
```diff
@@ -71,14 +71,6 @@ defmodule Screens.V2.CandidateGenerator.Elevator.ClosuresTest do
...
 
   describe "header and footer" do
-    test "have no variant when current elevator is not closed", %{now: now} do
-      assert [
-               _elevator_closures,
-               %NormalHeader{screen: @screen, text: "Elevator 111", time: ^now, variant: nil},
-               %Footer{screen: @screen, variant: nil}
-             ] = Generator.elevator_status_instances(@screen, now)
-    end
-
```
digitalcora, line 81: I think this test case still covers a unique and relevant behavior: that the footer is present (with no variant) when the current elevator is not closed but there are active closures elsewhere. Rather than deleting it, we could add some setup that creates an active closure, so the footer is still generated.

test/screens/v2/candidate_generator/elevator/closures_test.exs
```diff
@@ -213,12 +214,6 @@ defmodule Screens.V2.CandidateGenerator.Elevator.ClosuresTest do
     end
 
     test "filters out alerts with no facilities or more than one facility", %{now: now} do
-      expect(@stop, :fetch_parent_station_name_map, fn ->
-        {:ok, %{"place-haecl" => "Haymarket"}}
-      end)
-
-      expect(@route, :fetch, fn _ -> {:ok, [%Route{id: "Red", type: :subway}]} end)
-
```
digitalcora, line 221: Nice that pushing some of the fetch logic further down also simplified the setup of this test.

digitalcora (approved):
Thanks for tracking down that style issue! I had a couple more minor comments but otherwise looks good, pre-approving ✅

assets/src/components/v2/elevator/closures.tsx
jacmrob: I don't think it makes a huge difference.... I'll revert it to 224 height I think.

PR #2499, opened 2025-03-17
Jr/elevators with nearby redundancy

jacmrob:
**Asana task**: [No elevator outages state](https://app.asana.com/1/15492006741476/project/1185117109217422/task/1209256677260370?focus=true)

Adds the case where some elevators are closed across the system but have redundancies. I tested this with some alerts for the Alewife elevators, which seem like they have duplicates. 

- [ ] Tests added?

### DIFF
--- a/lib/screens/v2/candidate_generator/elevator/closures.ex
+++ b/lib/screens/v2/candidate_generator/elevator/closures.ex
@@ -65,14 +65,24 @@ defmodule Screens.V2.CandidateGenerator.Elevator.Closures do
     active_closures = Enum.flat_map(active, &elevator_closure/1)
     at_this_elevator? = fn %Closure{id: id} -> id == elevator_id end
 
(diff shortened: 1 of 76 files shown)


lib/screens/v2/candidate_generator/elevator/closures.ex
```diff
@@ -195,6 +196,24 @@ defmodule Screens.V2.CandidateGenerator.Elevator.Closures do
...
+         upcoming_closures,
+         %ElevatorConfig{elevator_id: _elevator_id} = app_params,
+         now,
+         stop_id
+       ) do
+    %ElevatorClosures{
+      app_params: app_params,
+      now: now,
+      station_id: stop_id,
+      stations_with_closures: :nearby_redundancy,
```
panentheos, line 212: nit: Since this code path is almost the same as the other one, I think it would be cleaner to combine them and use a `case` expression to compute just this one item.

jacmrob: I opted for a ternery-ish statement

PR #2503, opened 2025-03-18
Suppress crowding data for first stop

deanshi:
**Asana task**: [Don't display bus crowding at first stop on trip](https://app.asana.com/0/1185117109217413/1209411176930017/f)

- Removed checks for `current_status` for vehicle when doing crowding_data checks

Description
- [ ] Tests added?

### DIFF
--- a/lib/screens/v2/departure.ex
+++ b/lib/screens/v2/departure.ex
@@ -198,13 +198,12 @@ defmodule Screens.V2.Departure do
   def vehicle_status(_), do: nil
 
   defp crowding_data_relevant?(%Trip{id: trip_trip_id, stops: [first_stop | _]}, %Vehicle{
-         current_status: current_status,
          trip_id: vehicle_trip_id,
          stop_id: next_stop
        })
        when not is_nil(trip_trip_id) and not is_nil(vehicle_trip_id) do
     vehicle_on_prediction_trip? = trip_trip_id == vehicle_trip_id
-    vehicle_started_trip? = not (current_status == :in_transit_to and next_stop == first_stop)
+    vehicle_started_trip? = not (next_stop == first_stop)
     vehicle_on_prediction_trip? and vehicle_started_trip?
   end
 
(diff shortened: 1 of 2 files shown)


lib/screens/v2/departure.ex
```diff
@@ -198,13 +198,12 @@ defmodule Screens.V2.Departure do
...
 
   defp crowding_data_relevant?(%Trip{id: trip_trip_id, stops: [first_stop | _]}, %Vehicle{
-         current_status: current_status,
          trip_id: vehicle_trip_id,
          stop_id: next_stop
        })
        when not is_nil(trip_trip_id) and not is_nil(vehicle_trip_id) do
     vehicle_on_prediction_trip? = trip_trip_id == vehicle_trip_id
-    vehicle_started_trip? = not (current_status == :in_transit_to and next_stop == first_stop)
+    vehicle_started_trip? = not (next_stop == first_stop)
```
digitalcora, line 206: There is an existing `crowding_level/1` test case `"returns nil when the vehicle is in transit to the first stop"`. I don't know if the _content_ of this test needs to change exactly, but maybe the wording could change to remove the reference to a condition that doesn't apply anymore.

digitalcora (approved):
Looks good, some small comments, pre-approved 😄

lib/screens/v2/departure.ex
deanshi: Changed the name just to remove the In Transit part since the test still make sense!

lib/screens/departures/departure.ex
```diff
@@ -294,7 +293,7 @@ defmodule Screens.Departures.Departure do
 
     vehicle_started_trip? =
       not is_nil(first_stop) and
-        not (current_status == :in_transit_to and next_stop == first_stop)
+        not (next_stop == first_stop)
```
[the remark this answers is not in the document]
deanshi: Got it, I was wondering about both places. I reverted this back.

digitalcora (approved):

PR #2505, opened 2025-03-20
[feat] Filter and ordering logic for Hanover screen departures

robbie-sundstrom:
**Asana task**: [Add departure filtering/sorting](https://app.asana.com/0/1185117109217413/1209537234084970/f)

- Add destination filtering to only return the same route and direction twice if there are fewer than 3 unique departures.
- Add mode sorting 

Details of design for filtering is [in Miro](https://miro.com/app/board/uXjVLMQRENQ=/?moveToWidget=3458764612048337852&cot=14)

- [X] Tests added?


lib/screens/v2/candidate_generator/on_bus/departures.ex
```diff
@@ -80,6 +82,47 @@ defmodule Screens.V2.CandidateGenerator.Widgets.OnBus.Departures do
...
+  defp sort_by_mode(departures) do
+    priority = %{
+      :subway => 1,
+      :light_rail => 1,
+      :bus => 2,
+      :commuter_rail => 3,
+      :ferry => 4
+    }
+
+    Enum.sort(departures, fn a, b ->
```
panentheos, line 111: We should be able to use `Enum.sort_by` here, which would remove the need for the explicit comparisons below.

lib/screens/v2/candidate_generator/on_bus/departures.ex
```diff
@@ -80,6 +82,47 @@ defmodule Screens.V2.CandidateGenerator.Widgets.OnBus.Departures do
     Enum.filter(departures, &(&1.prediction.route.id != route_id))
   end
 
+  defp filter_duplicate_routes(departures) do
+    unique_departures =
+      Enum.uniq_by(departures, fn dep ->
+        {dep.prediction.route.id, dep.prediction.trip.direction_id}
```
digitalcora, line 88: This is not quite the uniqueness criteria specified in the task, which is to use the definition of "destination" from the Real-time Destination States proposal. That consists of the Line (so prediction => route => line => ID) and the "representative headsign" of the trip (already implemented as `Screens.Trips.Trip.representative_headsign/1`).

lib/screens/v2/candidate_generator/on_bus/departures.ex
```diff
@@ -80,6 +82,47 @@ defmodule Screens.V2.CandidateGenerator.Widgets.OnBus.Departures do
...
+  end
+
+  defp sort_by_mode(departures) do
+    priority = %{
+      :subway => 1,
+      :light_rail => 1,
+      :bus => 2,
+      :commuter_rail => 3,
+      :ferry => 4
+    }
```
digitalcora, line 109: This could be a module attribute, and I think that would make the function a bit more concise.

lib/screens/v2/candidate_generator/on_bus/departures.ex
```diff
@@ -80,6 +82,47 @@ defmodule Screens.V2.CandidateGenerator.Widgets.OnBus.Departures do
     Enum.filter(departures, &(&1.prediction.route.id != route_id))
   end
 
+  defp filter_duplicate_routes(departures) do
+    unique_departures =
+      Enum.uniq_by(departures, fn dep ->
+        {dep.prediction.route.id, dep.prediction.trip.direction_id}
+      end)
+
+    # Return at least 3 departures if they exist, even if there are not enough unique routes
```
digitalcora, line 91: I don't see this behavior specified in the task. Are we overcomplicating things by including it?

lib/screens/v2/candidate_generator/on_bus/departures.ex
```diff
@@ -80,6 +82,47 @@ defmodule Screens.V2.CandidateGenerator.Widgets.OnBus.Departures do
...
+
+    Enum.sort(departures, fn a, b ->
+      priority_a =
+        Map.get(priority, a.prediction.route.type)
+
+      priority_b =
+        Map.get(priority, b.prediction.route.type)
+
+      if priority_a == priority_b do
+        a.prediction.arrival_time <= b.prediction.arrival_time
```
digitalcora, line 119: Departures should already be sorted the way we want coming from the API, so ideally we wouldn't duplicate that here. Elixir sort functions are stable, so they should leave departures in the correct order within their priority groups after sorting by priority. (It seems like you might have added this to compensate for `filter_duplicate_routes` not necessarily leaving departures in the order it found them, though see my comment there as well. In case we do want to keep that behavior, I think it should be possible for that function to be "stable" with regard to ordering as well.)

lib/screens/v2/candidate_generator/on_bus/departures.ex
robbie-sundstrom: It was not specified in the list of design reqs, but I inferred from [the Figma](https://www.figma.com/design/syAph5uqIe4byahVkgg4yH/Bus-screen-(Hanover)?node-id=685-3617&t=kdFrraFFYFPyAARv-0), which shows multiple destinations to the same route. I didn't think this added too much complication and it seems like a clear improvement if included.

lib/screens/v2/candidate_generator/on_bus/departures.ex
robbie-sundstrom: Good point, I can definitely keep that `filter_duplicate_routes` behavior and maintain ordering by time as well so that this comparison of time isn't necessary.

lib/screens/v2/candidate_generator/on_bus/departures.ex
digitalcora: Hm... I'd be more inclined to view that as a placeholder rather than a representation of an intended feature, given the departure items are exact duplicates of each other, down to the times. It might be worth just calling out to Dave in the task that this is what we're going with, and if he has any suggestions for alternate behavior he can weigh in (maybe taking "extra" departures from higher-priority groups is more desirable? though soonest overall also seems good since this is "imminently relevant" information, i.e. will I catch my connection when I get off at this stop?).

lib/screens/v2/candidate_generator/on_bus/departures.ex
robbie-sundstrom: Great suggestion, this is much cleaner with `sort_by`
```
    Enum.sort_by(departures, fn departure ->
      priority = Map.get(@priority, departure.prediction.route.type)
      {priority, departure.prediction.arrival_time}
    end)
```

lib/screens/v2/candidate_generator/on_bus/departures.ex
robbie-sundstrom: Just logging what we discussed offline -- Dave confirmed that we should go ahead with rounding out the 3 departures shown on the screen with the soonest arrival time when there are fewer than 3 unique routes

lib/screens/v2/candidate_generator/on_bus/departures.ex
```diff
@@ -80,6 +91,29 @@ defmodule Screens.V2.CandidateGenerator.Widgets.OnBus.Departures do
     Enum.filter(departures, &(&1.prediction.route.id != route_id))
   end
 
+  defp filter_duplicate_routes(departures) do
+    unique_departures =
+      Enum.uniq_by(departures, fn dep ->
+        {dep.prediction.route.line.id, dep.prediction.trip.direction_id,
```
digitalcora, line 97: We don't need to include `direction_id` here since headsign should be strictly "more unique" than direction (if the direction is different then the headsign will also be different).

lib/screens/v2/candidate_generator/on_bus/departures.ex
```diff
@@ -80,6 +91,29 @@ defmodule Screens.V2.CandidateGenerator.Widgets.OnBus.Departures do
...
+         Trip.representative_headsign(dep.prediction.trip)}
+      end)
+
+    # If there are fewer than 3 unique connecting departures, then return at least 3 based on soonest departure time.
+    case length(unique_departures) do
+      count when count >= 3 ->
```
digitalcora, line 103: A `case` with only one condition and a fallback might be more naturally expressed as an `if`:

```elixir
if length(unique_departures) >= 3 do
  unique_departures
else
  # ...
end
```

lib/screens/v2/candidate_generator/on_bus/departures.ex
```diff
@@ -80,6 +91,29 @@ defmodule Screens.V2.CandidateGenerator.Widgets.OnBus.Departures do
...
+        unique_departures
+
+      _ ->
+        unique_departures
+        |> Enum.concat(Enum.take(departures -- unique_departures, 3 - length(unique_departures)))
+        |> Enum.sort_by(& &1.prediction.arrival_time)
```
digitalcora, line 109: I think my comment from the previous version of this still applies, namely that we're already sorting departures in a particular way coming from the `fetch` function, and this is re-sorting them (by arrival time, when initially they were sorted by departure time — but even if we fixed that, I think it's still a problem that we're duplicating the sorting, because the multiple places that are doing sorting could fall out of sync).

What we'd want here is basically "sort by position in the original list of departures", which could go something like this:

```elixir
departure_indexes = departures |> Enum.with_index() |> Map.new()
# ...then in the pipeline:
|> Enum.sort_by(&Map.fetch!(departure_indexes, &1))
```

digitalcora (approved):
Pre-approving with some comments, looking good!

PR #2506, opened 2025-03-24
fix: support Fall River/New Bedford line

digitalcora:
The Middleborough/Lakeville line is now the Fall River/New Bedford line, with a new route ID (`CR-NewBedford`).

* Add a route pill abbreviation for this line. Remove the abbreviation for the old route ID.
* Remove a DUP headsign abbreviation that only applied to the old line.
* Clean up some old DUP headsign abbreviations from the Orange Line Super Surge.

**Asana task**: https://app.asana.com/0/1185117109217413/1209770072971451


deanshi (approved):
:shipit:

PR #2507, opened 2025-03-24
ci: add more trigger events to Asana workflow

digitalcora:
* Add the `edited` event to prompt a task sync if a PR description is edited to add/change an Asana URL.

* Add the `review_requested` event, since the workflow only moves tasks to the in-review column when review has been requested (and CODEOWNERS review requests seem to happen after, and separately from, a PR being opened or moved out of draft).


deanshi (approved):
🚢

PR #2508, opened 2025-03-24
feat: "here" station closures for Pre-Fare solo

digitalcora:
The code previously assumed a station closure alert "here" (at the screen's station) would only be presented as a single-screen takeover if it was a closure of a single route at a transfer station; if an alert eliminated all service at the station, it would be a dual-screen takeover. This assumption is no longer valid with the addition of the "solo" template, which uses a single-screen takeover for anything that would have been a dual-screen takeover on a "duo".

This change enables such "full" station closures to be displayed as a single-screen takeover, with a disruption diagram and text matching the dual-screen version.

**Asana task**: https://app.asana.com/0/1185117109217413/1209255549956020

### DIFF
--- a/assets/src/components/v2/pre_fare_single_screen_alert.tsx
+++ b/assets/src/components/v2/pre_fare_single_screen_alert.tsx
@@ -22,8 +22,8 @@ interface PreFareSingleScreenAlertProps {
   routes: EnrichedRoute[];
   unaffected_routes: EnrichedRoute[];
   endpoints: [string, string];
-  effect: string;
-  region: string;
+  effect: "suspension" | "shuttle" | "station_closure" | "delay";
+  region: "here" | "boundary" | "outside";
   updated_at: string;
   disruption_diagram?: DisruptionDiagramData;
 }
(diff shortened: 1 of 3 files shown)


assets/src/components/v2/pre_fare_single_screen_alert.tsx
```diff
@@ -368,6 +374,17 @@ const PreFareSingleScreenAlert: React.ComponentType<
         />
       );
       break;
+    case effect === "station_closure" && region === "here":
```
robbie-sundstrom, line 377: Out of scope for this change, but it does seem like these effect strings, and possibly region as well, should be in some sort of enum for safety.

assets/src/components/v2/pre_fare_single_screen_alert.tsx
```diff
@@ -315,8 +315,14 @@ const MapSection: React.ComponentType<MapSectionProps> = ({
   );
 };
 
-const isMultiLine = (effect: string, region: string) =>
-  effect === "station_closure" && region === "here";
+const isMultiLine = (
```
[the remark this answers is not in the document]
deanshi: Anecdotally - I thought multi-line was dealing with multiple lines on an alert (from a String perspective) thinking it might be dealing with truncating/ellipsis logic rather than dealing with Stations and Lines. I had to look over the logic of it to realize what it actually meant but that just might be my lack of exposure for lines/routes.

assets/src/components/v2/pre_fare_single_screen_alert.tsx
digitalcora: Agreed, and I think it's a pretty small set of values that can appear here, so I will make that improvement.

robbie-sundstrom (approved):

PR #2509, opened 2025-03-24
test: fix unused variable warnings

digitalcora:
Some test assertions using `=` were written with an unpinned variable, causing the variable to be reassigned and therefore unused. We can also cut down on matching some values that are not relevant to the specific test cases in question.

As a separate commit, includes a minor refactor to the no-data logic being tested here:

* Push down an API fetch into the one branch that needs it.
* Push down the no-data conditional logic into the function that generates the only field it affects.


deanshi (approved):
💯 🚀

PR #2511, opened 2025-03-25
deps: update screens-config-lib

digitalcora:
* Update references to the removed `V2` namespace and "root" screen config modules which were moved into the `Screen` namespace.
* Delete a test file that was entirely commented out.


mix.exs
```diff
@@ -92,7 +92,7 @@ defmodule Screens.MixProject do
       {:telemetry, "~> 1.2"},
       {:telemetry_poller, "~> 1.0"},
       {:telemetry_metrics, "~> 1.0"},
-      {:screens_config, github: "mbta/screens-config-lib"},
+      {:screens_config, github: "mbta/screens-config-lib", branch: "cfg-un-v2"},
```
PaulJKim, line 95: Is the thought behind using a branch as the dependency version just to avoid merging in the changes to `screens-config-lib` and testing them out first?

digitalcora: [See comment here](https://github.com/mbta/screenplay/pull/616#discussion_r2016881323)!

PaulJKim (approved):
Looks good!

PR #2512, opened 2025-03-27
feat: use redundancy summaries for all elevators

digitalcora:
Previously elevators with `nearby` or `in_station` exiting redundancy would use the hard-coded "Accessible route available" as an exiting summary. The elevator spreadsheet has now been updated with summaries for all exiting categories, so to simplify further updates to wording and allow elevators with nearby/in-station redundancy to have different summaries, we can now use that in all cases.

Since the value `nil` was previously used to indicate both the use of the hard-coded summary and its special styling, the widget API changes a bit: summaries are now always non-nil, tagged `inside` (the station) or `other`, and styled accordingly.

**Asana task**: https://app.asana.com/0/1185117109217413/1209775829419814


deanshi (approved):
Code looks good to me :shipit:

PR #2513, opened 2025-03-27
feat: "this is the backup" elevator closure text

digitalcora:
When "this" elevator (the one an elevator screen is associated with) is among the alternate elevators for a closure, and the closed elevator has nearby or in-station redundancy (i.e. "this" elevator is at the same station and the alternative path does not involve travel outside the station), display a unique closure summary that identifies "this" elevator as an alternate.

**Asana task**: https://app.asana.com/0/1185117109217413/1209093626932232

### DIFF
--- a/lib/screens/v2/candidate_generator/elevator/closures.ex
+++ b/lib/screens/v2/candidate_generator/elevator/closures.ex
@@ -86,8 +86,8 @@ defmodule Screens.V2.CandidateGenerator.Elevator.Closures do
             relevant_closures,
             upcoming_closures,
             app_params,
-            now,
-            stop_id
+            stop_id,
+            now
           ),
           header_instances(config, now)
           | footer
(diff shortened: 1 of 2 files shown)


lib/screens/v2/candidate_generator/elevator/closures.ex
```diff
@@ -293,29 +293,34 @@ defmodule Screens.V2.CandidateGenerator.Elevator.Closures do
...
+    cond do
+      Enum.any?(closures, fn %Closure{id: id} -> id in alternate_ids end) ->
+        # If any of a closed elevator's alternates are also closed, the normal summary may not be
+        # applicable.
+        @active_summary_fallback
+
+      redundancy in ~w[nearby in_station]a and screen_elevator_id in alternate_ids ->
+        {:inside, "This is the backup elevator"}
+
+      redundancy in ~w[nearby in_station]a ->
```
deanshi, line 308: Self Note: `~w[nearby in_station]a` - I definitely need to read up more about atoms 👀

deanshi (approved):
Just making sure I'm parsing this correctly - is it safe to say that most of this was some name refactors, but then the main bit was the `redundancy` "This is the backup elevator" section?

digitalcora:
> is it safe to say that most of this was some variable name refactors

Mm, I should have been more clear that this includes a refactor to more consistently name variables that hold the screen's station ID or elevator ID. Maybe I'll pull that into a separate commit.

deanshi:
> > is it safe to say that most of this was some variable name refactors
> 
> Mm, I should have been more clear that this includes a refactor to more consistently name variables that hold the screen's station ID or elevator ID. Maybe I'll pull that into a separate commit.

I'm good with either way! My main question stems from being still new to some of the meat of the business logic. I wanted to make sure that I wasn't misinterpreting variables that might be coming from other places or things along that nature.

lib/screens/v2/candidate_generator/elevator/closures.ex
digitalcora: `~w` is a "sigil", specifically a [word list](https://github.com/elixir-lang/elixir/blob/v1.17.2/lib/elixir/pages/getting-started/sigils.md#word-lists). The `a` modifier makes it a list of atoms instead of strings.

digitalcora:
@deanshi I separated the refactor so you can see the behavior change more easily now. 🧹

deanshi (approved):
yee looks great 💯

PR #2514, opened 2025-03-27
Enable offline caching of Sentry events

PaulJKim:
**Asana task**: [Prefare/bus shelter apps collectively making fewer requests](https://app.asana.com/0/1185117109217413/1209235802491489/f)

Description
1. Add sentry logging in cases where we get a failed screen data API request but it hasn't been a minute yet and therefore does not trigger the no-data state
2. Enable offline caching of Sentry events in cases where the screen doesn't have network access when it tries to send the event to sentry. Once the browser is back online, it will retry any cached events before clearing the cache.

Note: Raven doesn't support offline caching out of the box so had to write a custom transport callback and do the caching manually.

New event logged to sentry with `cached` field: https://mbtace.sentry.io/issues/6469468736/?project=6061747&query=&referrer=issue-stream&stream_index=2

### DIFF
--- a/assets/src/hooks/v2/use_api_response.tsx
+++ b/assets/src/hooks/v2/use_api_response.tsx
@@ -92,7 +92,6 @@ const doFailureBuffer = (
   } else {
     const elapsedMs = Date.now() - lastSuccess;
     if (elapsedMs < MINUTE_IN_MS) {
-      SentryLogger.info("Keeping current state.");
       setApiResponse((state) => state);
     }
     if (elapsedMs >= MINUTE_IN_MS) {
@@ -200,6 +199,7 @@ const useBaseApiResponse = ({
       const response = parseRawResponse(json);
 
       if (response.state == "failure") {
+        SentryLogger.info("Request failed.", { json });
         doFailureBuffer(lastSuccess, setApiResponse, response);
       } else {
         setApiResponse((prevApiResponse) => {
@@ -210,7 +210,8 @@ const useBaseApiResponse = ({
         });
         setLastSuccess(now);
       }
-    } catch {
+    } catch (err) {
+      SentryLogger.captureException(err);
       doFailureBuffer(lastSuccess, setApiResponse);
     }
 
--- a/assets/src/hooks/v2/use_api_response.tsx
+++ b/assets/src/hooks/v2/use_api_response.tsx
@@ -91,8 +91,8 @@ const doFailureBuffer = (
     setApiResponse(FAILURE_RESPONSE);
   } else {
     const elapsedMs = Date.now() - lastSuccess;
-
     if (elapsedMs < MINUTE_IN_MS) {
+      SentryLogger.info("Keeping current state.");
       setApiResponse((state) => state);
     }
     if (elapsedMs >= MINUTE_IN_MS) {
--- a/assets/src/hooks/v2/use_api_response.tsx
+++ b/assets/src/hooks/v2/use_api_response.tsx
@@ -91,6 +91,7 @@ const doFailureBuffer = (
     setApiResponse(FAILURE_RESPONSE);
   } else {
     const elapsedMs = Date.now() - lastSuccess;
+
     if (elapsedMs < MINUTE_IN_MS) {
       setApiResponse((state) => state);
     }
--- a/assets/src/util/sentry.tsx
+++ b/assets/src/util/sentry.tsx
@@ -58,7 +58,7 @@ async function resendCachedEvents() {
   const cached = localStorage.getItem(OFFLINE_EVENTS_KEY);
   if (!cached) return;
 
-  let events = JSON.parse(cached);
+  const events = JSON.parse(cached);
 
   for (let i = 0; i < events.length; i++) {
     try {
--- a/assets/src/util/sentry.tsx
+++ b/assets/src/util/sentry.tsx
@@ -54,25 +54,32 @@ function buildSentryAuthHeader(auth) {
   return `Sentry sentry_version=${auth.sentry_version}, sentry_client=${auth.sentry_client}, sentry_key=${auth.sentry_key}`;
 }
 
-function resendCachedEvents() {
+async function resendCachedEvents() {
   const cached = localStorage.getItem(OFFLINE_EVENTS_KEY);
   if (!cached) return;
 
-  // Retry each event in the cache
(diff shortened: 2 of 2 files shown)


assets/src/hooks/v2/use_api_response.tsx
```diff
@@ -91,8 +91,8 @@ const doFailureBuffer = (
     setApiResponse(FAILURE_RESPONSE);
   } else {
     const elapsedMs = Date.now() - lastSuccess;
-
     if (elapsedMs < MINUTE_IN_MS) {
+      SentryLogger.info("Keeping current state.");
```
digitalcora, line 95: I think when we get a request failure we should be logging the actual error, rather than this log which wouldn't tell us much other than "there was an error" (we know there are errors, the question is why). So in the places we _call_ `doFailureBuffer`, we should log the problem that resulted in us making that call. I see two instances of that in `useBaseApiResponse`:

1. If parsing the response fails. We could log the `response` in that case.
2. In the `catch` branch. We should capture the exception we caught.

assets/src/util/sentry.tsx
```diff
@@ -19,6 +19,65 @@ const captureException = (ex: unknown, options?: RavenOptions) => {
   Raven.captureException(ex, options);
 };
 
+const offline_events_key = "sentry_offline_events";
```
digitalcora, line 22: Let's name this `OFFLINE_EVENTS_KEY` so in the places where it's used, it's obvious it's referring to a constant value rather than a local variable.

assets/src/util/sentry.tsx
```diff
@@ -19,6 +19,65 @@ const captureException = (ex: unknown, options?: RavenOptions) => {
...
+      ...options,
+      data: {
+        ...options.data,
+        extra: {
+          ...(options.data.extra || {}),
+          cached: new Date().toISOString(),
+        },
+      },
+    });
+    localStorage.setItem(offline_events_key, JSON.stringify(events));
```
digitalcora, line 39: Should we have a limit on the number of events cached? The official offline transport does this by default and sets the limit at 30 events. If a screen was offline for a long time, it could accumulate a huge number of events, and trying to send them might bog down the connection or use up our event quota.

assets/src/util/sentry.tsx
```diff
@@ -27,8 +86,12 @@ const initSentry = (appString: string) => {
   const { sentry: sentryDsn, environmentName: env } = getDataset();
 
   if (sentryDsn && isRealScreen()) {
+    Raven.setTransport(customTransport);
     Raven.config(sentryDsn, { environment: env }).install();
 
+    // Set listener to retry any cached Sentry events any time the browser transitions from offline to online
+    window.addEventListener("online", resendCachedEvents);
```
digitalcora, line 93: The official transport seems to also have some retry logic built-in, so if the browser never technically goes "offline" by its own definition (and so this event never fires), but there was still a sending error at some point, the queued events would eventually be retried. Maybe a simple timer that ticks every 5 minutes or so?

assets/src/util/sentry.tsx
```diff
@@ -19,6 +19,65 @@ const captureException = (ex: unknown, options?: RavenOptions) => {
...
+      method: "POST",
+      headers,
+      body: JSON.stringify(event.data),
+    });
+  });
+
+  localStorage.removeItem(offline_events_key);
+}
+
+function customTransport(options: RavenTransportOptions) {
```
digitalcora, line 68: It's pretty cool that you were able to replicate this functionality in Raven — but we do only use this because of browser constraints that we hope to soon be rid of (old Bus E-ink), so maybe we can add a comment pointing to the [offline caching docs](https://docs.sentry.io/platforms/javascript/guides/react/best-practices/offline-caching/) in the modern Sentry package, and say we should be using that instead once we upgrade?

assets/src/hooks/v2/use_api_response.tsx
PaulJKim: Ah okay, so are you suggesting we move the sentry logging up a level here to `fetchData()`? And should we keep the existing Sentry log for `Exiting no-data state`?

digitalcora: Pretty much that, yes. And I think it's fine if we leave the existing log.

assets/src/util/sentry.tsx
```diff
@@ -3,22 +3,94 @@ import { isRealScreen } from "Util/utils";
...
+  if (!cached) return;
+
+  // Retry each event in the cache
+  JSON.parse(cached).forEach((event: RavenTransportOptions) => {
+    const headers = {
+      "Content-Type": "application/json",
+      "X-Sentry-Auth": buildSentryAuthHeader(event.auth),
+    };
+
+    fetch(event.url, {
```
digitalcora, line 68: If we try to re-send but it turns out there's still a network issue:

* the error this throws will probably be captured by Sentry, which will then get cached, etc. causing a cascade of error reports.
* if we successfully sent any reports before the one that failed, those ones will get re-sent next time, since the cache is only cleared all-at-once and only after all events have successfully sent.

To address this, I think we should wrap each `fetch` in a `try ... catch {}`, and inside that block, after doing the `fetch`, remove only that specific event from the queue. This way any events that send successfully will not be re-sent, and any events that fail won't interfere with any others and won't cause a cascade of reports.

digitalcora: Looks okay, though I think we could avoid the relative fragility of a `for` loop here by repeatedly pulling off the first item of the array using `shift` until there are no items left.

digitalcora (approved):
Aside from my comment about the `for` loop, looks good — let's test this on dev-green to be sure before merging.

assets/src/util/sentry.tsx
PaulJKim: Agreed that the for loop approach could be fragile, but I went that direction rather than something like using shift and conditioning the loop on the events array being empty since that could cause a case where we remain in the loop for a long time/indefinitely while the screen is offline, assuming we keep re-queueing failed retries.

Maybe we could just do one pass through the events queue (using `shift`) and re-queue any failed attempts in a separate array and then re-cache that array afterwards so that we make sure the next retries happens on the subsequent interval, or when an "online" event fires of course.

digitalcora: Oh, good point. The approach described in your second paragraph sounds over-complex — it's probably fine if we stick with what we have now.

PR #2519, opened 2025-03-31
feat: Stylize Hanover On Bus screen

robbie-sundstrom:
**Asana task**: [Style the widget as per the designs](https://app.asana.com/0/1185117109217413/1209537234084969/f)

- Adds CSS files to style the shared Departures components used by the On Bus screens

image


assets/css/v2/dup/departures/row.scss
```diff
@@ -9,7 +11,7 @@
     width: 1586px;
     height: 4px;
     content: "";
-    background-color: #cccbc8;
```
robbie-sundstrom, line 12: There are a lot of places where the colors constants are not used and they should be. Thinking of doing that in a separate quick PR.

assets/src/components/v2/on_bus/departures.tsx
robbie-sundstrom, line 1: This component is very similar to `Components/v2/dup/departures.tsx`. I chose to break it out into a separate Component for the following reasons, but I'm interested if there are other opinions on doing this differently:
1. This references the DepartureRow component, which the On Bus screen definitely needs to handle differently for the following reasons. 
    1.  It will never have more than two times for the same destination in the same row.
    2. Crowding is not used, and removing that logic makes the Component and styling simpler.
    3. It uses different logic for displaying the headsign that does not involve paging or multiple lines.
2. On Bus component needs a header. This part is very solvable, as it could be pulled out into a separate type->component in `on_bus.tsx` or used conditionally so that dup does not create the header section.
3. It does not need all of the sections that the Dups do. This component removes `HeadwaySection` and `OvernightSection`.

lib/screens/v2/candidate_generator/on_bus.ex
```diff
@@ -32,11 +34,18 @@ defmodule Screens.V2.CandidateGenerator.OnBus do
...
+    [
+      fn -> departures_instances_fn.(config, query_params, now) end,
+      fn -> header_instances(config, now) end
+    ]
     |> Task.async_stream(& &1.())
     |> Enum.flat_map(fn {:ok, instances} -> instances end)
   end
 
+  defp header_instances(config, now) do
+    [%NormalHeader{screen: config, time: now, text: "CONNECTIONS"}]
```
digitalcora, line 46: I think we can make this a section header of the Departures component, rather than a separate Header component, and that would allow simplifying the frontend somewhat (no new "on-bus header" component required, just styling the markup the Departures component already emits for section headers). In `CandidateGenerator.Widgets.OnBus.Departures`, where we currently have the line `header: %ScreensConfig.Departures.Header{}`, we could specify a `title` field for that struct.

robbie-sundstrom: Removed the on bus header component and am now passing the `title` field as you suggested! That definitely does simplify things

assets/css/v2/on_bus/header.scss
```diff
@@ -0,0 +1,16 @@
...
+  align-items: center;
+  height: 108px;
+  margin-left: 26px;
+  font-family: Inter;
+  font-size: 47px;
+  font-style: normal;
+  font-weight: 500;
+  line-height: normal;
+  color: colors.$cool-black-30;
+  text-align: left;
```
digitalcora, line 14:
```css
font-style: normal;
line-height: normal;
text-align: left;
```

Are these all required to "undo" something else being set at a higher level? I actually see a number of instances of `font-style: normal` in this changeset, but no instances of a `font-style` with any other value.

assets/css/v2/on_bus/time.scss
```diff
@@ -0,0 +1,44 @@
+.departure-time {
+  display: inline-block;
+  align-items: center;
+  justify-content: end;
```
digitalcora, line 4: Since this is not a flex or grid container I don't think `align-items` and `justify-content` would have any effect?

assets/css/v2/on_bus/time.scss
```diff
@@ -0,0 +1,44 @@
...
+}
+
+.departure-times-with-crowding,
+.departure-time-with-crowding {
+  height: 100%;
+}
+
+.departure-crowding {
+  width: 0;
+  height: 0;
```
digitalcora, line 20: Can we use `display: none` here? That seems more appropriate for what we want to do.

lib/screens/v2/candidate_generator/on_bus/header.ex
digitalcora, line 1: ✂️ empty file

assets/css/v2/on_bus/time.scss
```diff
@@ -0,0 +1,44 @@
...
+
+.departure-time__minutes-label {
+  display: inline-block;
+  width: 42px;
+  height: 100%;
+  font-weight: 400;
+  text-align: left;
+}
+
+.departure-time__text {
```
digitalcora, line 39: We have styles here for `minutes` and `text` types, but I think we're missing the `timestamp` type, which would be seen for Commuter Rail departures (see `components/v2/departures/departure_time`).

assets/css/v2/on_bus/time.scss
robbie-sundstrom: oh yes those are leftover from when I was using a flex container here!

assets/css/v2/on_bus/header.scss
robbie-sundstrom: Removed the `font-style: normal` from throughout these changes! Also forgot that `text-align: left` is the default so that shouldn't be needed.

assets/css/v2/on_bus/departures_row.scss
```diff
@@ -0,0 +1,55 @@
...
+  position: relative;
+  align-items: center;
+  width: 100%;
+  height: 108px;
+
+  &:not(:first-child)::after {
+    position: absolute;
+    top: 105px;
+    left: 132px;
+    width: 492px;
```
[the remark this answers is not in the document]
robbie-sundstrom: In the design, it is specified as 1px below the bottom, so the height of 2px is needed, and then that 1px difference is captured in the 105px (108-2-1). I do think some of these can be moved into variables, so I'm doing that

assets/css/v2/on_bus/time.scss
robbie-sundstrom: Good catch, added!

assets/css/v2/on_bus/departures_row.scss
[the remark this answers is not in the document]
robbie-sundstrom: Added an `on_bus/vars.scss` with some variables that are reused and used in calculations for other values used throughout this styling!

assets/css/v2/on_bus/time.scss
```diff
@@ -0,0 +1,44 @@
...
+  display: inline-block;
+  width: 100%;
+  font-weight: 800;
+  text-align: right;
+}
+
+.departure-time__timestamp {
+  width: 42px;
```
digitalcora, line 41: Should this be like `__text` and be `width: 100%`, since like a `text` time it is the sole element in its container? Feels like they could share the exact same styles even.

```scss
.departure-time__text,
.departure-time__timestamp {
  width: 100%;
  height: 100%;
  font-weight: 400;
}
```

digitalcora (approved):
One tiny suggestion. Really appreciate the use of variables to reduce "magic numbers" and keep everything in sync! ✨

assets/css/v2/on_bus/time.scss
robbie-sundstrom: True this does make sense to just be combined!

PR #2520, opened 2025-03-31
chore: Use consts defined in colors.scss for all equivalent hex code colors

robbie-sundstrom:
There were a lot of CSS files I noticed that used color definitions where we had already defined constants in the `colors` namespace. Went through all the colors defined in the namespace and replaced any instances where we used the same hex code.


digitalcora (approved):
Incredible 🙌

PR #2523, opened 2025-04-03
refactor: fetch and parse Facility resources

digitalcora:
* Introduce a `Facility` struct and generic fetch functions for it. Allows replacing the narrower `fetch_stop_for_facility` function with fetching the facility and reading its `stop` relationship.

* Parse facilities related to alerts as the complete `Facility` instead of a map of ID and short name, potentially including relationships of the facilities as well. Allows replacing the special-purpose `fetch_elevator_alerts_with_facilities` with the normal `fetch`, since related facilities are now always included.

* Fix a minor bug in Stop parsing where `connecting_stops` being unloaded was represented as `nil`, rather than `:unloaded` as the typespec suggested.

For now only the `excludes-stop` facility property is parsed. This plus the `stop` relationship will ultimately let us determine the child stop IDs affected by a facility alert (e.g. the specific platforms within a station that may be inaccessible if a given elevator is offline), which is a necessary component of determining whether a given elevator outage is "downstream" of a screen location.

### DIFF
--- a/lib/screens/facilities/facility.ex
+++ b/lib/screens/facilities/facility.ex
@@ -3,20 +3,73 @@ defmodule Screens.Facilities.Facility do
   Functions for fetching facility data from the V3 API.
   """
 
-  alias Screens.Stops
+  alias Screens.Facilities.Parser
+  alias Screens.Stops.Stop
+  alias Screens.V3Api
 
   @type id :: String.t()
 
-  @callback fetch_stop_for_facility(id()) :: {:ok, Stops.Stop.t()} | {:error, term()}
-  def fetch_stop_for_facility(facility_id) do
-    case Screens.V3Api.get_json("facilities/#{facility_id}", %{
-           "include" => "stop"
-         }) do
-      {:ok, %{"data" => _data, "included" => [stop_map]}} ->
-        {:ok, Stops.Parser.parse_stop(stop_map, [])}
+  @type type ::
+          :bike_storage
+          | :bridge_plate
+          | :electric_car_chargers
+          | :elevated_subplatform
+          | :elevator
+          | :escalator
+          | :fare_media_assistance_facility
+          | :fare_media_assistant
+          | :fare_vending_machine
+          | :fare_vending_retailer
+          | :fully_elevated_platform
+          | :other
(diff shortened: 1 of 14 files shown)


robbie-sundstrom (approved):
Looks great!!

lib/screens/facilities/facility.ex
```diff
@@ -3,20 +3,73 @@ defmodule Screens.Facilities.Facility do
...
+          | :fare_vending_machine
+          | :fare_vending_retailer
+          | :fully_elevated_platform
+          | :other
+          | :parking_area
+          | :pick_drop
+          | :portable_boarding_lift
+          | :ramp
+          | :taxi_stand
+          | :ticket_window
```
robbie-sundstrom, line 30: Just curious, where did you get this list? I couldn't find this in V3 docs so I'm wondering if it's somewhere else

lib/screens/v2/candidate_generator/elevator/closures.ex
```diff
@@ -60,14 +60,14 @@ defmodule Screens.V2.CandidateGenerator.Elevator.Closures do
         %Screen{app_params: %ElevatorConfig{elevator_id: elevator_id} = app_params} = config,
         now
       ) do
-    {:ok, alerts} = @alert.fetch_elevator_alerts_with_facilities()
+    {:ok, alerts} = @alert.fetch(activity: "USING_WHEELCHAIR")
```
robbie-sundstrom, line 63: Nit, could use a comment explaining that `"USING_WHEELCHAIR"` filters to only alerts that specifically affect customers who use a wheelchair to board or exit.

lib/screens/facilities/facility.ex
digitalcora: It's in the [Swagger docs](https://api-v3.mbta.com/docs/swagger/index.html) — expand `FacilityResource`, then under `attributes` => `type`, expand the "Array [18]".

robbie-sundstrom: Oops that's where I was looking and I just totally missed `attributes` => `type`. Thanks!

PR #2525, opened 2025-04-04
refactor: DRY up V3 API response parsing

digitalcora:
Some patterns were repeated across many of our resource parsing modules, including an implicit "API" for passing around included resources. This centralizes the shared logic in a `V3Api.Parser` module.

### DIFF
--- a/lib/screens/v3_api/parser.ex
+++ b/lib/screens/v3_api/parser.ex
@@ -1,0 +1,90 @@
+defmodule Screens.V3Api.Parser do
+  @moduledoc "Tools for parsing V3 API responses."
+
+  @parsers %{
+    "alert" => Screens.Alerts.Parser,
+    "facility" => Screens.Facilities.Parser,
+    "line" => Screens.Lines.Parser,
+    "prediction" => Screens.Predictions.Parser,
+    "route" => Screens.Routes.Parser,
+    "route_pattern" => Screens.RoutePatterns.Parser,
+    "schedule" => Screens.Schedules.Parser,
+    "stop" => Screens.Stops.Parser,
+    "trip" => Screens.Trips.Parser,
+    "vehicle" => Screens.Vehicles.Parser
+  }
+
+  @typep included :: %{{id :: String.t(), type :: String.t()} => object()}
+  @typep object :: %{String.t() => term()}
+
+  @doc "Parse a complete JSON-decoded V3 API response."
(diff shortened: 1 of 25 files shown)


lib/screens/v3_api/parser.ex
```diff
@@ -0,0 +1,109 @@
...
+  def parse(%{"data" => data} = response) when is_map(data) do
+    [resource] = parse(%{response | "data" => [data]})
+    resource
+  end
+
+  def parse(%{"data" => data} = response) when is_list(data) do
+    included =
+      response
+      |> Map.get("included", [])
+      |> then(&Enum.concat(data, &1))
```
digitalcora, line 30: This added line (versus the previously duplicated copies of this logic in the per-resource parsers) fixes [a crash we've seen](https://mbtace.sentry.io/issues/6492234261) on the test on-bus screen. This occurred when trying to resolve a relationship that referred to a resource in the response's `data` (not `included`).

robbie-sundstrom: great catch and fix idea by moving everything into a shared parser so this logic can be addressed everywhere!

lib/screens/v3_api/parser.ex
```diff
@@ -0,0 +1,109 @@
...
+      long_name: long_name,
+      short_name: short_name,
+      sort_order: sort_order
+    }
+  end
+
+  for {type, module} <- @parsers do
+    def parse_resource(%{"type" => unquote(type)} = data, included),
+      do: unquote(module).parse(data, included)
+  end
```
robbie-sundstrom, line 64: I love the use of `unquote` here! Very cool.

Is there are a reason that `Line` parsing needs to be defined separately instead of defined in its own module like the other resource types? It would make more sense to me to have that defined like every other included resource

digitalcora: It doesn't need to be, I just thought the logic was trivial enough that there was no need for it to live in a separate module. Originally I thought maybe that could be true of other resources (that their logic was minimal enough that it could also live here) but it turned out Line was the only "simple" one. I can put that back in its own module as well.

robbie-sundstrom: Gotcha, I think putting that back in it's own module would make it much cleaner in that case

robbie-sundstrom (approved):
Pre-approving with one small comment. Looks great overall, very understandable Parser module

PR #2526, opened 2025-04-04
refactor: remove dead v1 `Departure` code

digitalcora:
Only a small part of this module was still used anywhere; this moves it to `V2.Departure` and deletes the rest.


robbie-sundstrom (approved):
Nice cleanup!

PR #2527, opened 2025-04-07
feat: Add No Data and No Connections On Bus screen types

robbie-sundstrom:
**Asana task**: [Generate no-data state if predictions are missing](https://app.asana.com/0/1185117109217413/1209537234084971/f)

Designs that were followed are here: https://www.figma.com/design/syAph5uqIe4byahVkgg4yH/Bus-screen-(Hanover)?node-id=1042-7509&t=ztTGYvLcZq32P11i-0

Adds the two "no departures to list" screen types for On Bus screens
- No Data
image

- No available connections (easy to test by putting in any stop id without connections)
image


deanshi:
Is it just me or does "Live connections updates" sound awkward?

assets/src/components/v2/on_bus/no_data.tsx
```diff
@@ -1,8 +1,26 @@
...
-import Placeholder from "../placeholder";
+import { imagePath } from "Util/utils";
 
 const NoData: ComponentType = () => {
-  return <Placeholder color="blue" text="No Data" />;
+  return (
+    <div className="screen-container">
+      <div className="no-data__container">
+        <div className="no-data__header">
+          <div className="no-data__logo-container">
```
digitalcora, line 9: There are several layers of `div`s that seem unnecessary here:

* The SVG is the only element in the `logo-container`
* The `logo-container` is the only element in the `header`
* The `__text` is the only element in the `main-content`

I think we could collapse these into a less-noisy structure:

* `no-data`
  * `no-data__header`
  * `no-data__content`

assets/src/components/v2/on_bus/no_service.tsx
```diff
@@ -0,0 +1,13 @@
+import React, { ComponentType } from "react";
+import Header from "../departures/header";
+
+const DeparturesNoService: ComponentType = () => {
+  return (
+    <div className="departures-container">
+      <Header {...{ title: "CONNECTIONS", arrow: null }} />
```
digitalcora, line 7: I think this should work and would be a more conventional way of expressing this:

```tsx
<Header title="CONNECTIONS" arrow={null} />`
```

assets/src/components/v2/on_bus/no_service.tsx
```diff
@@ -0,0 +1,13 @@
+import React, { ComponentType } from "react";
+import Header from "../departures/header";
```
digitalcora, line 2: Normally when importing something from outside a component's directory we'll use the `Components` alias, which maps to the root `components` directory. This way, reorganizing files/directories is less likely to require noisy changes to the number of `..`s in an import path.

```ts
import Header from "Components/v2/departures/header";
```

assets/css/v2/on_bus/no_data.scss
```diff
@@ -0,0 +1,27 @@
...
+}
+
+.no-data__content {
+  position: relative;
+  top: 28px;
+  font-size: 59px;
+  font-weight: 700;
+  line-height: 66px;
+  color: colors.$cool-black-15;
+  letter-spacing: 0;
```
digitalcora, line 26: This should be equivalent to the default and so not necessary to specify (it does have behavior differences from the actual default value of `normal`, but those should only come into play when text is being justified).

assets/css/v2/on_bus/no_service.scss
```diff
@@ -0,0 +1,12 @@
...
+
+.departures-no-service {
+  position: relative;
+  left: 26px;
+  width: 515px;
+  font-size: 59px;
+  font-weight: 700;
+  line-height: 66px;
+  color: colors.$cool-black-15;
+  letter-spacing: 0;
```
digitalcora, line 11: Same comments in this file re: letter-spacing and `position: relative`.

assets/src/apps/v2/on_bus.tsx
```diff
@@ -26,9 +27,9 @@ import Departures from "Components/v2/departures";
 
 const TYPE_TO_COMPONENT = {
   body_normal: NormalBody,
-  departures: Departures,
+  departures: NoData,
```
digitalcora, line 30: Change left in from testing?

digitalcora (approved):
Couple small things, address them if you like, but looks good!

assets/src/apps/v2/on_bus.tsx
robbie-sundstrom: yes thanks for catching!

assets/css/v2/on_bus/no_data.scss
```diff
@@ -0,0 +1,27 @@
...
+  position: relative;
+  width: 552px;
+  height: 354px;
+  margin-bottom: 23px;
+  margin-left: 36px;
+}
+
+.no-data__header {
+  position: relative;
+  top: 23px;
```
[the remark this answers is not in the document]
robbie-sundstrom: I updated this to use `border`s instead of `margin`, but that does get rid of the `position` property and make it easier to see what's going on

PR #2528, opened 2025-04-07
chore: When running prettier, only log files that have been changed

robbie-sundstrom:
Just a change to the package.json that's been bothering me. Adds the [--list-different option](https://prettier.io/docs/cli#--list-different) so you can easily see what files prettier changed.

- [ ] Tests added?


digitalcora (approved):
Nice find!

PR #2529, opened 2025-04-07
refactor: use facility stop in elevator closures

digitalcora:
Generating elevator screen closures requires, among other things, the ID and name of the elevator's station.

Previously the ID was obtained from the `stop` field of the same `informed_entity` that referenced the elevator (assuming the alerts feed would have to populate this, given "facilities" are a non-standard extension), and the name from a separate API fetch that collected the names of all parent stations in the system.

We can now populate the facilities associated with an alert and the stops associated with those facilities as part of the alert fetch, so this information can be gotten directly from the Stop resources. In particular this removes the need for the separate "station name map" to be fetched, passed around, and stubbed in tests.

Fetching alerts with the full Facility relationship graph populated (`include_all?: true`) will also enable more easily determining which child stop IDs within a station are affected by an elevator closure, which we need to enable "downstream" closure summaries.

### DIFF
--- a/lib/screens/v2/candidate_generator/elevator/closures.ex
+++ b/lib/screens/v2/candidate_generator/elevator/closures.ex
@@ -12,22 +12,19 @@ defmodule Screens.V2.CandidateGenerator.Elevator.Closures do
     alias Screens.Alerts.Alert
     alias Screens.Elevator
     alias Screens.Facilities.Facility
-    alias Screens.Stops.Stop
 
     @type t :: %__MODULE__{
-            id: Facility.id(),
-            name: String.t(),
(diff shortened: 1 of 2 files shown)


robbie-sundstrom (approved):
LGTM! Would be great to also have this logic change shared with pre-fare elevator alerts, but definitely out of scope for this change and that's just another reason to consolidate their logic

digitalcora:
Agreed, I am so far not going to the trouble of "porting" these improvements because it's all in service of downstream elevator summaries, and Pre-Fares currently don't display elevator summaries of any kind. The point when we want them to do so will be a good opportunity to consolidate the logic / make these one unified widget.

PR #2530, opened 2025-04-07
refactor: specify alert activities as atoms

digitalcora:
Alert activities are an enum rather than an arbitrary string. Specify them as atoms and document the possible values as a type.

### DIFF
--- a/lib/screens/alerts/alert.ex
+++ b/lib/screens/alerts/alert.ex
@@ -23,6 +23,16 @@ defmodule Screens.Alerts.Alert do
             url: nil,
             description: nil
 
+  @type activity ::
+          :board
+          | :bringing_bike
+          | :exit
+          | :park_car
+          | :ride
+          | :store_bike
+          | :using_escalator
+          | :using_wheelchair
+
   @type cause ::
           :accident
           | :amtrak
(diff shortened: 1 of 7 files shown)


digitalcora:
> does it ultimately boil down to performance/type safety?

There isn't a _whole_ lot of either that this change gets us — it's mostly down to aligning with the data model and improving clarity for those reading and changing the code. In Elixir, "enums" tend to be represented using atoms, whereas if a field is a string the expectation is that it could be "any" string (possible values are not known ahead of time). In support of this view, atoms are a "divisible" type in typespecs, i.e. `:ok | :error` is a different type from `:ok | :ignore` and Dialyzer can sometimes use this extra information to call out mistakes — the only string type is `String.t()`.

deanshi:
> > does it ultimately boil down to performance/type safety?
> 
> There isn't a _whole_ lot of either that this change gets us — it's mostly down to aligning with the data model and improving clarity for those reading and changing the code. In Elixir, "enums" tend to be represented using atoms, whereas if a field is a string the expectation is that it could be "any" string (possible values are not known ahead of time). In support of this view, atoms are a "divisible" type in typespecs, i.e. `:ok | :error` is a different type from `:ok | :ignore` and Dialyzer can sometimes use this extra information to call out mistakes — the only string type is `String.t()`.

Very cool, thank you!

PR #2531, opened 2025-04-08
fix: alert activity atoms not existing

digitalcora:
Atoms used only in typespecs seem to not reliably "exist" at runtime. Unfortunately this means we have to duplicate the list of atoms in the typespec for the sake of parsing.


panentheos (approved):
:atom:

PR #2538, opened 2025-04-11
ci: fix caches never being populated on main

digitalcora:
Because we only had CI running on `pull_request`, it was never actually running on the main branch, which is the only branch whose caches other branches are allowed to use. This resulted in builds on feature branches slowing down as the main cache became progressively less usable due to dependency changes.

Also includes a small improvement to the caching logic as a separate commit; see commit message.


robbie-sundstrom (approved):
nice fix!

PR #2546, opened 2025-04-24
refactor: improve Webpack configuration

digitalcora:
The overall goal here is to simplify the configuration, and therefore the upgrade to Webpack 5, by removing unused or redundant components.

* Remove duplication between the configuration of the "packaged" DUP entrypoint and the others.
* Instead of always building the "packaged" DUP entrypoint alongside all the others, add a `deploy:dup` script which only builds this.
* Remove `useBuiltIns: "usage"` configuration from the "packaged" DUP entrypoint. This was redundant since we also included the full set of polyfills in this build.
* `copy-webpack-plugin` was redundantly copying files that were already copied by a `file-loader` rule. Move fonts and images out of the `static` directory so this is no longer the case; `static` is now reserved for files that are not copied/transformed by another rule (i.e. they are not "referenced" in a way that a Webpack plugin can detect and incorporate into the build automatically).
* Remove `plugin-proposal-do-expressions` which appears to be unused.
* Remove unused configuration that would have allowed importing non-SVG image files into JS code. If needed later, we can restore this.
* Update `core-js` to the latest version for improved polyfill support.

---
  - https://app.asana.com/0/0/1210072126243999


robbie-sundstrom (approved):
Changes look good! I've only done some basic testing of the different screen types, but everything seems to work as expected. All moved/re-directoried files seem to be in the right places. Webpack config also looks right to me, and actually seems much more understandable now!

PR #2549, opened 2025-04-24
fix: elevator screen no-closures cosmetic issues

digitalcora:
* The flexbox container for the no-closures component used the `gap` property, which is not supported by the old browser version on elevator screens. Replace this with a `margin`-based solution.

* Fix the "true" no-closures state (no closures at all, including nearby-redundancy elevators) having an unwanted space between the text "All MBTA elevators are working" and the period.

* Fix the alerts URL and phone number not necessarily having a line of text to themselves as in the designs, resulting in some oddly-placed line breaks (e.g. in the middle of the phone number).

Unrelated to the no-closures state:

* Fix that the upcoming closure component had every word of its trailer capitalized ("Until Further Notice" rather than the intended "Until further notice").

* Remove some duplication when invoking the "normal service" SVG icon.

**Asana task**: [Elevator screen no-closures style issues](https://app.asana.com/1/15492006741476/project/1185117109217413/task/1210023570906845)


robbie-sundstrom (approved):
LGTM!

PR #2554, opened 2025-04-29
MVP Destination Grouping for Departures

deanshi:
**Asana task**: [MVP destination ordering logic to be used on pre-fares](https://app.asana.com/1/15492006741476/project/1185117109217413/task/1209857808687042?focus=true)

Description
- Added Destination Grouping logic for Departures widgets
- Backend now takes in a new screen-config field (:destination), and will sort the departures/rows by `direction_id` as well as filtering them using `uniq_by` to grab only one instance of each headsign/direction_id. Using these rows, it will calculate extra fields that are passed to the frontend.
- Frontend has separate trimming logic based on destination grouping. It will use a hardcoded `Layout` param (base: number of unique destinations, min: 2, max: 0, include_later: false) and a `direction_trim_priority` both passed from the backend. If the departures only has one direction, it will continue to treat it as a timed grouping and just trim the furthest departure. If it is bidirectional, it will first check which direction_id group is larger and trim rows from the largest group. If they are equal, it will use direction_trim_priority to trim rows (calculated based on the original departure list to see which direction_id group has more to begin with). `direction_trim_priority` will default to 1 based on the logic if the direction groups are equal.
- Non-destination grouping will remain the same, defaulting to :time grouping.

Screenshot 2025-04-29 at 8 45 54 AM
Screenshot 2025-04-29 at 8 46 26 AM
Screenshot 2025-04-29 at 8 46 59 AM

- [x] Tests added?


assets/src/components/v2/departures/section.ts
```diff
@@ -106,37 +111,90 @@ const trimOneBy = (
...
+    aboveFold.some((row) => isDepartureRow(row) && row.direction_id == 1);
+
+  // If there's only one direction, treat it as a timed list and trim the furthest time
+  const trimmedRow = isBidirectional
+    ? trimBidirectionDestinationGrouping(aboveFold, directionTrimPriority)
+    : _.last(aboveFold);
+
+  if (isDepartureRow(trimmedRow)) {
+    aboveFold.splice(aboveFold.indexOf(trimmedRow), 1);
+    belowFold.unshift({
```
digitalcora, line 165: Since sections with destination-based grouping never use "later departures", I think there is some simplification we could do here. We don't technically need to shift trimmed departures below the fold, we just need to discard them.

assets/tests/components/v2/departures/factories.ts
```diff
@@ -1,22 +1,28 @@
...
   layout: { min: 1, base: null, max: null, include_later: false },
   rows: departureRow.buildList(1),
   header: { title: null, arrow: null },
+  grouping_type: "time",
+  direction_trim_priority: 0,
 }));
 
-export const departureRow = Factory.define<Row>(({ sequence }) => ({
+export const departureRow = Factory.define<
+  DepartureRow & { type: "departure_row" }
```
digitalcora, line 17: Why was the change to this type needed?

lib/screens/v2/widget_instance/departures.ex
```diff
@@ -154,8 +167,52 @@ defmodule Screens.V2.WidgetInstance.Departures do
...
+      layout:
+        Layout.to_json(%Layout{
+          max: nil,
+          base: length(serialized_rows),
+          min: 2,
+          include_later: false
+        }),
+      header: Header.to_json(header),
+      grouping_type: :destination,
+      direction_trim_priority:
```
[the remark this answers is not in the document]
deanshi: I think that makes sense, and now that you say it I'm definitely over complicating here.

In my mind, I was thinking which direction had more departures from the _entire_ departure list that comes back, and not just from the list of unique destinations, but that's definitely too much. I'll switch over that logic to the front end and clean up some of this extra data and tests.

deanshi:
Also - @digitalcora I was seeing this in some testing, where the route-id seems to be different but the headsign is the same towards Government Center and I think this is what I was seeing before. Would this be reason to switch the `uniq_by` check? Or am I completely misunderstanding how the routes/headsigns work. 

Screenshot 2025-04-29 at 6 30 12 PM

digitalcora:
@deanshi This is fine and how we expect it to work for now. The task description alludes to this:

> For the sake of this task, a "destination" is whatever the existing grouping logic uses, not what we eventually want a "destination" to be — we will address this discrepancy in a separate task.

assets/src/components/v2/departures/section.ts
```diff
@@ -106,37 +123,88 @@ const trimOneBy = (
...
+
+const trimSectionForDestinationGrouping = (
+  section: FoldedDepartureGroupedSection,
+): boolean => {
+  const {
+    rows: { aboveFold },
+    direction_to_trim,
+  } = section;
+
+  // If there's only one direction, treat it as a timed list and trim the furthest time
```
digitalcora, line 168: The comment here is now out-of-sync with the code, which no longer has this conditional. I think it can just be removed.

assets/src/components/v2/departures/section.ts
```diff
@@ -106,37 +123,88 @@ const trimOneBy = (
...
+  const trimmedRow = trimRowForDestinationGrouping(
+    aboveFold,
+    direction_to_trim,
+  );
+
+  if (isDepartureRow(trimmedRow)) {
+    aboveFold.splice(aboveFold.indexOf(trimmedRow), 1);
```
digitalcora, line 175: The way this works is slightly indirect — first we call a function that returns the row itself that should be trimmed, but doesn't mutate the section, then we go back and find where that row is in the section, and trim it. Can we simplify this so it works more like `trimSection`, where "finding the row to trim" and "trimming it" are a single operation?

assets/src/components/v2/departures/section.ts
```diff
@@ -106,37 +123,88 @@ const trimOneBy = (
...
+  if (directionZeroRows.length > directionOneRows.length) {
+    return _.last(directionZeroRows);
+  }
+
+  return directionToTrim == 1
+    ? _.last(directionOneRows)
+    : _.last(directionZeroRows);
+};
+
+const calculateDirectionToTrim = (rows: Row[]) => {
```
digitalcora, line 205: nit: "calculate" is a word that doesn't add much meaning to a function name; it could be omitted here without losing anything, I think.

assets/src/components/v2/departures/section.ts
```diff
@@ -106,37 +123,88 @@ const trimOneBy = (
...
+  const counts = _.countBy(rows, "direction_id");
+  return (counts[1] || 0) > (counts[0] || 0) ? 0 : 1;
+};
+
+const isDepartureRow = (row: Row | undefined) => row?.type == "departure_row";
```
digitalcora, line 210: By introducing this function we've loosened the conditional in `trimSection` so that it allows `undefined` when previously it didn't. The logic here is so concise already that I'm not sure it's worth extracting — can we keep the condition in `trimSection` as it was, and inline this logic in `trimSectionForDestinationGrouping`?

assets/src/components/v2/departures/departure_row.tsx
```diff
@@ -9,15 +9,20 @@ type DepartureRow = {
   route: Pill;
   headsign: Destination;
   times_with_crowding: TimeWithCrowding[];
+  direction_id: 0 | 1;
+  showBoldDivider: boolean;
```
digitalcora, line 13: nit: something like `isBeforeDirectionSplit` I think would better express what this indicates, without being as prescriptive about how it should be presented.

test/screens/v2/candidate_generator/widgets/departures_test.exs
```diff
@@ -376,6 +376,7 @@ defmodule Screens.V2.CandidateGenerator.Widgets.DeparturesTest do
     test "rejects departures with excluded route-directions" do
       section = %Section{
         query: %Query{params: %Query.Params{stop_ids: ["S"]}},
+        grouping_type: :time,
```
digitalcora, line 379: This should be the default value for this field — do we need to provide it here?

test/screens/v2/widget_instance/departures_test.exs
```diff
@@ -44,6 +44,21 @@ defmodule Screens.V2.WidgetInstance.DeparturesTest do
                Departures.serialize_section(section, bus_shelter_screen, now)
     end
 
+    test "returns serialized normal_section with destination grouping", %{
```
digitalcora, line 47: Since it doesn't have any rows, this test case doesn't seem like it's doing much other than making sure this code path doesn't crash outright, which the full test case on L288 also does. Could we remove it without losing anything?

test/screens/v2/widget_instance/departures_test.exs
```diff
@@ -260,6 +284,80 @@ defmodule Screens.V2.WidgetInstance.DeparturesTest do
...
+      section = %NormalSection{
+        rows: rows,
+        layout: %Layout{},
+        header: %Header{title: "Section Header"},
+        grouping_type: :destination
+      }
+
+      assert %{
+               type: :normal_section,
+               header: %{title: "Section Header", arrow: nil, read_as: nil},
```
digitalcora, line 344: The header isn't really relevant to this test case; I think we could remove this from the pattern match.

assets/css/v2/lcd_common/departures/section.scss
```diff
@@ -11,4 +11,9 @@
     background-color: colors.$warm-neutral-80;
     border-radius: 2px;
   }
+
+  & > .departure-row.direction-split::after {
+    left: 0;
+    width: 100%;
```
digitalcora, line 17: This should also be using the darker-gray color that's used to divide sections, rather than the lighter gray used between departure rows normally.

digitalcora (approved):
Some minor things I noticed that could be cleaned up, but looks good!

PR #2556, opened 2025-05-01
feat: Handle wayfinding_only config option in departures widget

robbie-sundstrom:
**Asana task**: [Wayfinding-only sections in departures widget](https://app.asana.com/1/15492006741476/project/1185117109217413/task/1209972434938245?focus=true)

- Builds off this `screens-config-lib` PR: https://github.com/mbta/screens-config-lib/pull/56
- This ended up being pretty simple! Just checks if the `wayfinding_only` option is set to true, and if so, does not fetch any departures. I have not found any way this effects things downstream on the FE, including with autosizing.  
 
- [X] Tests added?


mix.exs
```diff
@@ -80,7 +80,8 @@ defmodule Screens.MixProject do
       {:recon, "~> 2.5.6"},
       {:remote_ip, "~> 1.2"},
       {:retry, "~> 0.18.0"},
-      {:screens_config, github: "mbta/screens-config-lib"},
+      {:screens_config,
+       github: "mbta/screens-config-lib", branch: "rs/wayfinding-only-departures-section"},
```
robbie-sundstrom, line 84: Noting that this will of course need to be modified back before merging, once the `screens-config-lib` PR is merged. This is needed for testing currently though.

lib/screens/v2/candidate_generator/widgets/departures.ex
```diff
@@ -176,6 +186,11 @@ defmodule Screens.V2.CandidateGenerator.Widgets.Departures do
     end
   end
 
+  def fetch_section_departures(_, _, _, _) do
+    # Only get to this pattern match if wayfinding_only_text is not nil, in which case, we fetch no departures
+    :wayfinding_only
```
digitalcora, line 191: suggestion: This could be `{:ok, []}` and then we would not need the extra clause of `normal_section_rows`.

lib/screens/v2/candidate_generator/widgets/departures.ex
```diff
@@ -176,6 +186,11 @@ defmodule Screens.V2.CandidateGenerator.Widgets.Departures do
     end
   end
 
+  def fetch_section_departures(_, _, _, _) do
+    # Only get to this pattern match if wayfinding_only_text is not nil, in which case, we fetch no departures
```
digitalcora, line 190: suggestion: Rather than explaining this in a comment we could "explain" it via explicitly pattern-matching on `wayfinding_only: true`. In fact I think it might make more sense to reverse the order of these clauses so the "exceptional" behavior (wayfinding-only) comes first.

```elixir
def fetch_section_departures(%Section{wayfinding_only: true}, _, _, _), do: {:ok, []}

def fetch_section_departures(
  %Section{
    query: %Query{opts: opts, params: params},
    filters: filters,
    bidirectional: is_bidirectional
  },
  disabled_route_types,
  departure_fetch_fn,
  now
) do
  # ...
end
```

lib/screens/v2/candidate_generator/widgets/departures.ex
```diff
@@ -139,6 +139,8 @@ defmodule Screens.V2.CandidateGenerator.Widgets.Departures do
 
   defp normal_section_rows({:ok, departures}), do: departures
 
+  defp normal_section_rows(:wayfinding_only), do: []
```
digitalcora, line 142: suggestion: I think this extra clause would not be needed if we returned `{:ok, []}` instead of using the separate value `:wayfinding_only` for this purpose.

lib/screens/v2/candidate_generator/widgets/departures.ex
robbie-sundstrom: good call, that is cleaner!

test/screens/v2/candidate_generator/widgets/departures_test.exs
```diff
@@ -282,6 +282,61 @@ defmodule Screens.V2.CandidateGenerator.Widgets.DeparturesTest do
...
+                arrow: nil,
+                read_as: nil,
+                title: nil,
+                subtitle: nil
+              },
+              layout: %ScreensConfig.Departures.Layout{
+                base: nil,
+                include_later: false,
+                max: nil,
+                min: 1
```
digitalcora, line 327: I don't think any of the fields of `Header` or `Layout` are required — we should be able to leave them at defaults and shorten this irrelevant data significantly.

digitalcora (approved):
Approving pending the merge of the config changes, plus one small suggestion.

PR #2558, opened 2025-05-06
feat: Process subtitle field in Departures widget on the FE

robbie-sundstrom:
**Asana task**: [Wayfinding sub-headers in departures sections](https://app.asana.com/1/15492006741476/project/1185117109217413/task/1209972434938244?focus=true)

Adds processing of subtitles to the departures widget on the Frontend. 
image

- [X] Tests added?


digitalcora:
What's the rationale for not implementing this end-to-end in this PR? (for Sectionals only, since Pre-Fare departures is not supported yet, but it would presumably be the same frontend code) In particular I was thinking that if we wanted to do this without `dangerouslySetInnerHTML` we'd probably have to do the "bold formatting" on the frontend, and doing the entire implementation within this change would be a good way to confirm or refute that.

robbie-sundstrom:
@digitalcora I wasn't aware that we were also including subtitles for sectionals, and can't find any details related to that in the ticket or Figma. I was going to do the Frontend piece as a followup once the Pre-Fares departures section is created.

That's an interesting point I didn't think about for the Frontend display piece. I'll confirm that this sort of implementation could work by building out the FE

digitalcora:
I don't think there is presently any intent to use subtitles on Sectionals, but the frontend components are common to all screen types that use Departures, and Pre-Fares are also going to reuse the existing `lcd_common` styles that apply to Sectionals and Bus Shelters, so from a technical standpoint very little (none?) of this will actually be specific to Pre-Fares.

assets/src/components/v2/departures/header.tsx
```diff
@@ -23,13 +24,31 @@ const DirectionArrow = ({ arrow }: { arrow: CardinalDirection }) => (
...
-      {(title || arrow) && <span>{title}</span>}
-      {arrow && <DirectionArrow arrow={arrow} />}
-    </header>
+    <div>
+      <header className="departures-header">
+        {(title || arrow) && <span>{title}</span>}
+        {arrow && <DirectionArrow arrow={arrow} />}
+      </header>
+      {formatted_subtitle && (
+        <div className="departures-header__subtitle">{formatted_subtitle}</div>
```
digitalcora, line 37: We're sort of doing this "null check" twice, once here and once in `formatSubtitle` itself. Rather than passing a null through and requiring `formatSubtitle` to handle it, I think we could simplify this if we did:

```tsx
{subtitle && <div ...>{formatSubtitle(subtitle)}</div>}
```

digitalcora (approved):
Nice, this was very straightforward! Some minor comments, pre-approving however you decide to address those.

assets/src/components/v2/departures/header.tsx
robbie-sundstrom: nice simplification!

assets/src/components/v2/departures/header.tsx
```diff
@@ -23,13 +24,31 @@ const DirectionArrow = ({ arrow }: { arrow: CardinalDirection }) => (
   />
 );
 
-const Header = ({ title, arrow }: Header) => {
+const Header = ({ title, arrow, subtitle }: Header) => {
+  const formatted_subtitle = format_subtitle(subtitle);
```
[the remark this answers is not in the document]
robbie-sundstrom: ugh still not great at switching between Elixir and TS I guess 🤦  thanks for pointing out

PR #2559, opened 2025-05-06
feat: admin image manager improvements

digitalcora:
A recent motivation for [making files in the assets bucket public by default](https://app.asana.com/1/15492006741476/project/1185117109217413/task/1209486399971517) was that we often forgot to add the `public-read` ACL when manually uploading line map images. However, the admin UI already includes a file manager that allows uploading images to S3 and sets the correct ACL automatically; it's just limited to a specific prefix. This PR is a quick diversion into the idea of removing that limitation, and sprucing up the experience a bit while we're here.

👉 [Try it out on dev-green!](https://screens-dev-green.mbtace.com/admin/image-manager)

* The Image Manager can now manage all objects under the environment's `images/` prefix, rather than only in `images/psa/`. When uploading an image, admins can enter the prefix to use within `images/` (and the filename, removing the limitation that the uploaded name had to be the same as the local one).

* Uploading and deleting images no longer forces a full page reload, only a refresh of the image list. Uploading an image also selects the just-uploaded image, confirming its presence.

* The prefix of an upload defaults to the prefix of whichever file is selected for preview. This allows easily uploading images to an existing prefix without having to type it in, and combined with the above behavior, uploading multiple images to the same prefix without typing/selecting it each time.

* Limit the dimensions of the preview for the selected image to reasonable values. Provide a link to view the image at full size.

* Add a button to copy the path of the selected image to the clipboard, suitable for pasting directly into screen configuration.

* Fix an issue where deleting images was not possible due to a missing CSRF token.

* Remove an unneeded S3 "pass-through" route that was only used by the Image Manager. Instead, send the full URLs of images to the client along with their keys.

### DIFF
--- a/assets/css/admin.scss
+++ b/assets/css/admin.scss
@@ -115,39 +115,10 @@
     margin-left: 8px;
   }
 
-  &__thumbnail-container {
-    display: flex;
-    flex-flow: row wrap;
-    margin-top: 16px;
-  }
-
   &__thumbnail {
-    box-sizing: border-box;
-    display: inline-flex;
-    width: 100px;
-    height: 100px;
-    padding: 4px;
-    margin-right: 8px;
+    max-width: 640px;
+    max-height: 480px;
     margin-bottom: 8px;
-    border: 1px solid #eaeaea;
-    border-radius: 2px;
-
-    & &--full-size {
-      width: unset;
-      height: unset;
-    }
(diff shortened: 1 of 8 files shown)


robbie-sundstrom (approved):
Looks great, and worked really smoothly in testing! Awesome improvements

PR #2560, opened 2025-05-06
Support Departures Widget on PreFare Screens

deanshi:
**Asana task**: [Implement departures widget on Pre-Fares](https://app.asana.com/1/15492006741476/project/1185117109217413/task/1209972434938243?focus=true)

Description
- Departures takes priority over the Line Map as long as it is
  configured
- Added departures fields to admin page
- Renamed some of the classes in cr_departures to prevent overlapping
- Moved alerts to flex zone when departures are enabled
- Modified audio_sort_keys for departures on PreFares and updated other
  widgets to have them after the departures widget
- Renamed duo_placement to base_placement and added an adjust_placement
  function specifically for moving single screen alerts into the flex
  zone when departures are being shown

- [x] Tests added? 

Screenshot 2025-05-06 at 11 45 46 AM
Screenshot 2025-05-06 at 11 46 04 AM
Screenshot 2025-05-07 at 8 50 23 AM


lib/screens/v2/candidate_generator/widgets/reconstructed_alert.ex
```diff
@@ -386,4 +388,12 @@ defmodule Screens.V2.CandidateGenerator.Widgets.ReconstructedAlert do
...
     end
   end
+
+  defp showing_departures?(%Screen{
+         app_params: %PreFare{departures: %Departures{sections: [_ | _]}, template: :duo}
+       }),
+       do: true
+
+  defp showing_departures?(_),
+    do: false
```
robbie-sundstrom, line 398: This logic is exactly the same as in the candidate generator `pre_fare.ex`. First thought is that this should be consolidated/shared in case it changes (i.e. not just duos in the future), but I'm not sure the best place for that

deanshi: Not sure if `screens/util.ex` might be a good place to put it? Otherwise I could even create a separate util under `screens/v2/` somewhere?

assets/css/busway_v2.scss
```diff
@@ -4,6 +4,7 @@
 
 @use "v2/lcd_common/departures" as *;
 @use "v2/lcd_common/departures_no_data" as *;
+@use "v2/lcd_common/departures/header.scss" as *;
```
digitalcora, line 7: In general (for here and the lines changed in `pre_fare_v2`):

* `as *` is only required when there is another module of the same name being imported.
* The `.scss` extension is not needed.

lib/screens/v2/candidate_generator/pre_fare.ex
```diff
@@ -78,18 +78,23 @@ defmodule Screens.V2.CandidateGenerator.PreFare do
...
+       fn -> elevator_status_instance_fn.(config, now) end,
+       fn -> evergreen_content_instances_fn.(config) end,
+       fn -> commuter_rail_departures_instance_fn.(config, now) end,
+       fn -> shuttle_bus_info_instances(config, now) end
+     ] ++
+       if should_show_departures?(config) do
+         [fn -> departures_instances_fn.(config, now) end]
+       else
+         [fn -> full_line_map_instances_fn.(config) end]
+       end)
```
digitalcora, line 97: We may not actually need this conditional at all, which could accidentally resolve the question of how to share the conditional logic. Consider:

* Generating the line map widget is basically free. If given a lower priority than the departures widget, it just won't be chosen for the slot, if the departures widget is also generated.
* With a small tweak to the departures widget generator, we could generate _no_ widget at all (return `[]`) when there are no sections configured. This feels like a reasonable behavior and I don't think it would break any existing uses of the widget — we'd never have configured one with zero sections before.
* So, when no sections are configured, the line map would be the only widget wanting to be in its slot, and would appear as normal.

lib/screens/v2/candidate_generator/widgets/departures.ex
```diff
@@ -99,7 +99,18 @@ defmodule Screens.V2.CandidateGenerator.Widgets.Departures do
                 }
             end)
 
-          %DeparturesWidget{screen: config, sections: sections, now: now}
+          slot_names =
+            case config do
+              %Screen{
+                app_params: %PreFare{departures: %Departures{sections: [_ | _]}, template: :duo}
```
digitalcora, line 105: This feels overly "defensive" — we can just make this conditional on the app being `PreFare`, since for now we will never want it to go into a different slot on Pre-Fares. Either way you slice it, configuring departures on a solo won't do anything useful. (Note this would only work as intended if we go with my suggestion above to not generate the widget at all when zero sections are configured.)

lib/screens/v2/candidate_generator/widgets/reconstructed_alert.ex
```diff
@@ -174,7 +175,8 @@ defmodule Screens.V2.CandidateGenerator.Widgets.ReconstructedAlert do
         informed_stations: get_stations(alert, fetch_stop_name_fn),
         is_terminal_station: is_terminal_station,
         is_priority: is_priority,
-        partial_closure_platform_names: all_platforms_names_at_informed_station
+        partial_closure_platform_names: all_platforms_names_at_informed_station,
+        showing_departures: showing_departures?(config)
```
digitalcora, line 179: The widget instance already has access to the screen configuration, so I don't think we need to define this here just to pass it to the instance — the instance can itself do this pattern-match on the configuration.

lib/screens/v2/widget_instance/audio_only/content_summary.ex
```diff
@@ -27,8 +28,22 @@ defmodule Screens.V2.WidgetInstance.AudioOnly.ContentSummary do
...
+        %__MODULE__{
+          screen: %Screen{
+            app_id: :pre_fare_v2,
+            app_params: %PreFare{departures: %ScreensConfig.Departures{sections: [_ | _]}}
+          }
+        } = t
+      ) do
+    %{
+      lines_at_station: t.lines_at_station,
+      showing_departures: true
```
digitalcora, line 41: "Showing" is a slightly odd word to use in the context of audio, where we're not "showing" anything. How about `has_departures`?

lib/screens/v2/widget_instance/departures.ex
```diff
@@ -95,6 +92,16 @@ defmodule Screens.V2.WidgetInstance.Departures do
       %{sections: Enum.map(sections, &Departures.audio_serialize_section(&1, screen, now))}
     end
 
+    def audio_sort_key(%Departures{
+          screen: %Screen{
+            app_params: %PreFare{
+              departures: %ScreensConfig.Departures{sections: [_ | _]},
+              template: :duo
```
digitalcora, line 99: As previously commented, I think we can just condition this on the screen type being `PreFare` and ignore the rest — there is only a single way we want departures to behave on pre-fares.

lib/screens/v2/widget_instance/reconstructed_alert.ex
```diff
@@ -391,6 +393,8 @@ defmodule Screens.V2.WidgetInstance.ReconstructedAlert do
     t |> duo_placement(location) |> adjust_placement(location, template)
   end
 
+  defp duo_placement(%__MODULE__{showing_departures: true}, _location), do: :flex_zone
```
digitalcora, line 396: This isn't the logic we want — if an alert would normally take over the full dual screens, we still want it to do that. This puts _every_ alert in the flex zone.

assets/css/pre_fare_v2.scss
```diff
@@ -7,6 +7,9 @@
 @use "v2/lcd_common/no_data" as *;
 @use "v2/lcd_common/page_load_no_data";
 @use "v2/lcd_common/route_pill";
+@use "v2/lcd_common/departures" as *;
+@use "v2/lcd_common/departures_no_data" as *;
+@use "v2/lcd_common/departures/header.scss";
```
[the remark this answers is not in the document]
deanshi: 🤦 oof that's what I get for rushing and not looking at my other files as well

digitalcora (approved):
Looking good! Let's get the config change in, then update this and merge ✨

PR #2562, opened 2025-05-09
feat: Include subtitle in departures header audio readout

robbie-sundstrom:
**Asana task**: [Wayfinding sub-headers in departures sections](https://app.asana.com/1/15492006741476/project/1185117109217413/task/1209972434938244?focus=true)

- Adds the subheader to the departures widget audio readout value when present

- [X] Tests added?


digitalcora (approved):
This seems like a reasonable handling of the readouts for now! If we end up wanting to frequently "customize" the readout of titles or subtitles separately from each other, we could revisit this.

PR #2565, opened 2025-05-12
bugfix: Add nil stop_id check into DUP alerts fetcher

robbie-sundstrom:
**Asana task**: [Investigate Sentry report volume](https://app.asana.com/1/15492006741476/project/1185117109217413/task/1210072157534313?focus=true)

When the `stop_id` for Alerts is set to nil, the `LocationContext.fetch` call throws an error. We want to return an empty list from the `Alerts` fetcher before that method is ever called. 

Notes on implementation:
- This nil stop_id case is only relevant for the DUPs alert fetcher at present. 
- I included in the function instead of breaking out into a separate pattern matched function header b/c that ended up being cleaner and more concise with the default values being set.


digitalcora (approved):
🧹

PR #2567, opened 2025-05-12
chore: improve departures widget audio readouts

digitalcora:
* Previously, each section could read out up to 15 departures, without regard for its visual layout. This could result in impractically long readouts due to e.g. only 2 departures fitting on the screen visually, but the audio including 10. To address this, make audio serialization respect the configured maximum/base size of the section visually.

* Ensure audio readouts of sections using destination-based grouping only include one departure per "destination", in line with the visual presentation, instead of two.

---
  - https://app.asana.com/0/0/1210225474039152


panentheos (approved):
Looks good!

PR #2569, opened 2025-05-13
Changed mbta.com/alerts to mbta.com/status for SubwayStatus widgets

deanshi:
**Asana task**: [Modify Subway Status logic to match website widget](https://app.asana.com/1/15492006741476/project/1185117109217413/task/1209950740101100?focus=true)

Description
Changed mbta.com/alerts to mbta.com/status for SubwayStatus widgets
Let me know if I missed any places, I know there a few places where we might concat the actual URL but I couldn't 100% find them.

- [X] Tests modified

### DIFF
--- a/lib/screens/v2/widget_instance/subway_status.ex
+++ b/lib/screens/v2/widget_instance/subway_status.ex
@@ -86,6 +86,8 @@ defmodule Screens.V2.WidgetInstance.SubwayStatus do
 
   @green_line_branches ["Green-B", "Green-C", "Green-D", "Green-E"]
 
+  @mbta_status_url "mbta.com/status"
+
   defimpl Screens.V2.WidgetInstance do
     alias ScreensConfig.Audio
     alias ScreensConfig.Screen.BusShelter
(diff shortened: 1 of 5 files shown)


lib/screens/v2/widget_instance/subway_status.ex
```diff
@@ -426,7 +426,7 @@ defmodule Screens.V2.WidgetInstance.SubwayStatus do
                                 other {# stops}}",
             num_informed_platforms: length(informed_platforms)
           ),
-        location: %{full: "mbta.com/alerts", abbrev: "mbta.com/alerts"}
+        location: %{full: "mbta.com/status", abbrev: "mbta.com/status"}
```
digitalcora, line 429: At least within this module, we should put this in a module attribute so it's not repeated 4 times.

lib/screens_web/views/v2/audio/subway_status_view.ex
```diff
@@ -229,8 +229,8 @@ defmodule ScreensWeb.V2.Audio.SubwayStatusView do
   defp get_location_string(nil), do: ""
 
   # We never read out the alerts URL.
-  defp get_location_string(%{full: "mbta.com/alerts"}), do: ""
-  defp get_location_string("mbta.com/alerts"), do: ""
+  defp get_location_string(%{full: "mbta.com/status"}), do: ""
+  defp get_location_string("mbta.com/status"), do: ""
```
digitalcora, line 233: We could use a pattern like `"mbta.com" <> _` here so it works no matter what URL we happen to be using — the intent feels closer to just "don't read out URLs as part of the individual line statuses" rather than "don't read out this specific URL".

digitalcora (approved):
Pre-approving with some small comments, looks good.

PR #2570, opened 2025-05-13
fix: Pre-Fare audio consistency with takeovers

digitalcora:
The `ContentSummary` and `AlertsIntro` audio-only widgets both had logic that prevented their inclusion when the screen had widgets in "takeover" slots, but these checks were duplicated and had drifted out of sync with regard to slot names. This moves the logic into the main candidate generator so it can be done in one place.

---
  - https://app.asana.com/0/0/1210237684564486


robbie-sundstrom (approved):
nice 🚀

PR #2573, opened 2025-05-14
Style Fixes for Pre-Fare Departures

deanshi:
**Asana task**: [[extra] style fixes for Pre-Fare departures](https://app.asana.com/1/15492006741476/project/1185117109217413/task/1210237684564480?focus=true)

Description
- Padding for subtitle, divider for subtitles with departures (only on prefares)
- FreeText styling fixed specifically for prefare departures notice_row to prevent any rippling style changes

- [ ] Tests added?

Screenshot 2025-05-14 at 9 24 08 AM


assets/css/v2/pre_fare/free_text.scss
```diff
@@ -59,3 +59,61 @@
     line-height: unset;
   }
 }
+
+.departures__notice-row {
+  // Some of the same properties as, just copied specifically for departures.
```
deanshi, line 64: Maybe worth pulling into a mixin for some of these properties? 

Or maybe there's a better way of handling this specific issue with free-text overlap here and above?

robbie-sundstrom: I think a `mixin` makes sense here! Could even just be defined earlier in this .scss file

robbie-sundstrom (approved):
Looks great!

assets/css/v2/pre_fare/departures/header.scss
```diff
@@ -0,0 +1,9 @@
+@use "CSS/colors";
+
+.departures-header__subtitle {
+  padding-bottom: 44px;
+
+  &.divider {
+    border-bottom: 4px solid colors.$true-grey-70;
+  }
```
digitalcora, line 8: This should be part of the existing `lcd_common` styles for the `subtitle` — presumably if we configure a subtitle on a non-Pre-Fare screen, we also want it to appear as specified in the designs, and not have the missing padding/line.

assets/src/components/v2/departures/normal_section.tsx
```diff
@@ -70,7 +70,7 @@ const NormalSection: ComponentType<FoldedSection> = ({
 
   return (
     <div className="departures-section">
-      <Header {...header} />
+      <Header {...header} withDivider={aboveFold.length > 0} />
```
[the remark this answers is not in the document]
deanshi: I initially had a pure-CSS version with `has()` but was worried about browser versions. Swapped it with the adjacent selector and definitely prefer this to the JS version.

assets/css/v2/lcd_common/departures/header.scss
```diff
@@ -35,8 +35,17 @@
...
   font-size: 40px;
   line-height: 54px;
+  padding-bottom: 44px;
 
   strong {
     @include fonts.inter--bold;
   }
 }
+
+.departures-header-container + .departure-row::before {
```
digitalcora, line 45: This seems like it would add the divider even when there is no subtitle (and in that case would result in some unwanted doubling-up on borders, since there is also a bottom border on the header). Do the header and subtitle actually need to be in a common container element? If not, i.e. if the element tree could look like:

* header
* subtitle
* departure
* departure
* ...

Then we could use a selector like `.departures-subtitle + .departure-row::before`.

deanshi: Sounds good - removed the <div> for <> to keep it flat, now it's looking like header, subtitle, departure, departure, etc.

PR #2576, opened 2025-05-15
Fix bus berth pronounciation

PaulJKim:
**Asana task**: [Fix pronunciation of bus berths on screen audio](https://app.asana.com/1/15492006741476/project/1185117109217413/task/1210077353559487?focus=true)

Description
Unfortunately, wrapping the berth identifier in a single SSML `say-as` tag and indicating to pronounce it as an address or any other option still results in the pronounciation "ah-two" or "ay tee double u oh".

The best option I could find was adding a slight break between "Berth" and the identifier to improve clarity, and then only specifying to "spell-out" the letter.

### DIFF
--- a/lib/screens_web/views/v2/audio/departures_view.ex
+++ b/lib/screens_web/views/v2/audio/departures_view.ex
@@ -103,6 +103,12 @@ defmodule ScreensWeb.V2.Audio.DeparturesView do
   end
 
   defp render_track_number(nil, _), do: ~E""
+
+  defp render_track_number(<<letter, digit>>, :bus)
+       when letter in ?A..?Z and digit in ?0..?9 do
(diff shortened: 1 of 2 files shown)


digitalcora:
Well, that's annoying. Thanks for doing the research on it!

I think we could implement this a bit more "generically" so it will work for more platform codes in case we ever encounter them. Something like:

```elixir
String.replace("A12", ~r/\D+/, ~S(<say-as interpret-as="spell-out">\0</say-as>))
# => "<say-as interpret-as=\"spell-out\">A</say-as>12"
```

This ensures any non-digits anywhere in the platform code are "spelled out" regardless of the placement and number of digits or non-digits, and also works fine when the code is only digits (leaving it as-is).

digitalcora (approved):

PR #2577, opened 2025-05-15
feat: support footers with no stop ID

digitalcora:
See mbta/screens-config-lib#60.

All screen types that support `Footer` configuration now support having no stop ID specified, for screens that are not associated with a single stop. The "more info" URL displayed in this case is simply `mbta.com`.

Includes minor refactors to remove unneeded "variability" of values that were always the same.

**Asana task**: https://app.asana.com/1/15492006741476/project/1185117109217413/task/1209206882712033


PaulJKim (approved):
Looks good!

PR #2580, opened 2025-05-16
chore: remove ShuttleBusInfo and old "surge" logic

digitalcora:
* Remove the `ShuttleBusInfo` widget and related logic/components. This was developed specifically for the 2022 Orange Line "Super Surge" and has not been used since then.

* Remove some special-case logic related to specific past disruptions other than the Super Surge.


deanshi (approved):
🟥

PR #2581, opened 2025-05-16
feat(admin): don't reload page on config update

digitalcora:
The Config Editor and the "single-screen" editor used in the Inspector no longer reload the entire page when the config is updated. This was mostly unnecessary and even counter-productive in the Inspector, since the reloaded screen would most likely be built from the pre-update cached copy of the config, requiring a second manual reload after a few seconds to allow the cache to update.

Also add a reload button to the Inspector to allow manually reloading only the screen iframe, so the current zoom level is retained.


deanshi (approved):
The persistent zoom level is so nice. Appreciate these changes :shipit:

PR #2582, opened 2025-05-20
bugfix: Only show overnight screen on DUP if every route in the section is overnight

robbie-sundstrom:
**Asana task**: [Investigate Kendall DUP evening bus issue](https://app.asana.com/1/15492006741476/project/1185117109217413/task/1210196927622064?focus=true)

In the example of the bug shown in the ticket, this occurred because only 2/3 routes shown on this section of the Kendall DUP was in overnight mode. However, there was still one bus that had trips scheduled for later in the night.

To solve this, I added a check that there are as many overnight trips as there are total routes available.

This also changes, and I believe improves, the following case
- When there are no trips scheduled for the next day, as in the case of the Kendall buses on Friday, then we will show a duplicate of the `primary_departures` screen instead of an Overnight screen

- [ ] Tests added?

### DIFF
--- a/lib/screens/v2/candidate_generator/dup/departures.ex
+++ b/lib/screens/v2/candidate_generator/dup/departures.ex
@@ -175,9 +175,9 @@ defmodule Screens.V2.CandidateGenerator.Dup.Departures do
       |> Enum.uniq()
 
     # Check if there is any room for overnight rows before running the logic.
-    overnight_schedules_for_section =
+    {section_contains_active_route, overnight_schedules_for_section} =
       if (is_only_section and length(departures) >= 4) or length(departures) >= 2 do
-        []
+        {false, []}
       else
         get_overnight_schedules_for_section(
           routes_with_live_departures,
@@ -193,8 +193,10 @@ defmodule Screens.V2.CandidateGenerator.Dup.Departures do
     headway_mode = get_headway_mode(stop_ids, routes, alert_informed_entities, now)
 
     cond do
-      # All routes in section are overnight
-      departures == [] and overnight_schedules_for_section != [] ->
+      # No remaining departures or active routes, but we do have routes with overnight schedules
+      # Show a takeover Overnight section for the given route types
+      departures == [] and !section_contains_active_route and
+          overnight_schedules_for_section != [] ->
         %OvernightSection{routes: routes}
 
       # Headway mode
@@ -458,8 +460,8 @@ defmodule Screens.V2.CandidateGenerator.Dup.Departures do
       MapSet.disjoint?(not_informed, informed_stop_ids)
   end
 
-  # For sections configured only to show predictions and not schedules,
-  # we will override by returning a list of scheduled Departures from `get_overnight_schedules_for_section`
+  # For sections configured only to show predictions and not schedules, we may override
+  # by returning a list of scheduled Departures from `get_overnight_schedules_for_section`
   # if one of the following is true for a given route_id/direction_id combo:
   # 1. Service for the route is done for the day, so we may display the first departure of tomorrow.
   # 2. Service for the route has not started for the day, so we may display the first departure of today.
@@ -467,8 +469,9 @@ defmodule Screens.V2.CandidateGenerator.Dup.Departures do
   #    (possible for CR/buses/routes with interruptions tomorrow), so return a Departure
   #    with nil departure_time and arrival_time to be handled by the serializer.
   #
-  # If any routes still have scheduled trips left today, return an empty list, as we are not 'overnight'
-  # and do not want to display scheduled times for tomorrow when there are future scheduled trips today.
+  # Returns a tuple where:
+  #  - First value is a boolean to track if there are any remaining scheduled trips for the section.
+  #  - Second value is a list of Departure Schedules to potentially show on screen given enough space.
   defp get_overnight_schedules_for_section(
          routes_with_live_departures,
          stop_ids,
@@ -498,7 +501,7 @@ defmodule Screens.V2.CandidateGenerator.Dup.Departures do
       )
 
     # Get schedules for each route_id in config
-    results =
+    overnight_schedules =
       today_schedules
       |> Enum.map(&{&1.route.id, &1.direction_id})
       |> Enum.uniq()
@@ -533,15 +536,14 @@ defmodule Screens.V2.CandidateGenerator.Dup.Departures do
         )
       end)
 
-    # If any routes have a :departure_scheduled_today, then do not return tomorrow's schedules for any route
-    if Enum.any?(results, &(&1 == :departure_scheduled_today)) do
-      []
-    else
-      results
-      # Routes not in overnight mode will be nil. Can ignore those.
-      |> Enum.reject(&is_nil/1)
-      |> Enum.sort_by(fn %Departure{schedule: %Schedule{departure_time: dt}} -> dt end)
-    end
+    # We flag if any routes have a :route_service_ongoing, so the section is not fully overnight
+    is_sections_service_active = Enum.any?(overnight_schedules, &(&1 == :route_service_ongoing))
+
+    {is_sections_service_active,
+     overnight_schedules
+     # Routes not in overnight mode will be nil. Can ignore those.
+     |> Enum.reject(fn elem -> is_nil(elem) or elem == :route_service_ongoing end)
+     |> Enum.sort_by(fn %Departure{schedule: %Schedule{departure_time: dt}} -> dt end)}
   end
 
   defp get_overnight_schedules_for_section(
@@ -567,11 +569,11 @@ defmodule Screens.V2.CandidateGenerator.Dup.Departures do
         fetch_vehicles_fn
       )
     else
-      []
+      {false, []}
     end
   end
 
-  defp get_overnight_schedules_for_section(_, _, _, _, _, _, _), do: []
+  defp get_overnight_schedules_for_section(_, _, _, _, _, _, _), do: {false, []}
 
   # Verifies we are meeting the timeframe conditions for overnight mode and generates the departure widget
   defp get_overnight_departure_for_route(
@@ -609,7 +611,7 @@ defmodule Screens.V2.CandidateGenerator.Dup.Departures do
       }
     else
       # Return an atom so we can track that there is still a departure for this route during today's service.
-      :departure_scheduled_today
+      :route_service_ongoing
     end
   end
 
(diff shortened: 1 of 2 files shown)


lib/screens/v2/candidate_generator/dup/departures.ex
```diff
@@ -194,7 +194,7 @@ defmodule Screens.V2.CandidateGenerator.Dup.Departures do
 
     cond do
       # All routes in section are overnight
-      departures == [] and overnight_schedules_for_section != [] ->
+      departures == [] and Enum.count(overnight_schedules_for_section) == Enum.count(routes) ->
```
digitalcora, line 197: I'm not entirely convinced this new condition accomplishes what we want. The `routes` value here comes from a Route fetch for the screen's configured stop IDs, so will include each route that has any scheduled service at the stops. The `overnight_schedules_for_section` on the other hand consists of A) one "overnight departure" per _route+direction combination_, B) that has scheduled service at the stop _only_ on the current day, and C) some of those can also be filtered out depending on other conditions. So it seems very easy (typical, even, because of point A) for these counts to be different even in scenarios where we want to show the overnight message — and for them to be equal in scenarios where we don't!

Conceptually I think it's straightforward to describe what we want here: If there are any scheduled departures remaining today (i.e. in the future, but still within the current service day) that match the section's parameters, we should fail this condition and not generate an OvernightSection. The existing comment on L598 even sort of gets at this: _"If now is before any of today's schedules ... we do not display overnight mode."_ The problem is, that particular code is only deciding whether to generate an "overnight departure" for an _individual_ route+direction — so if that condition occurs for some route+directions but not for others, we run into the problem that prompted this task.

Spitballing: Some branches of `get_overnight_departure_for_route` currently return `nil` when we are in the condition mentioned above (there are scheduled departures remaining for today). Those `nil`s get lumped in with other `nil`s that are returned in other conditions, so we're "throwing away" the fact that we're in that scenario, and higher-level logic has no way to know or act on it. It feels like these branches should instead be returning some kind of sentinel value in this specific scenario, which doesn't get filtered out. If any such value is present, we can pick it up and avoid generating an OvernightSection. Does that make sense?

robbie-sundstrom: Yes that makes sense to me! I considered an approach like that, but then thought the solution I went with would be the simplest approach. I see how the issues you brought up could cause my solution to fail though.

lib/screens/v2/candidate_generator/dup/departures.ex
```diff
@@ -624,6 +633,11 @@ defmodule Screens.V2.CandidateGenerator.Dup.Departures do
           DateTime.compare(now, first_schedule_today.departure_time) == :lt ->
         %Departure{schedule: first_schedule_today}
 
+      # Before 3am and before the last scheduled trip of the day.
+      # Return an atom so we can track that there is still a departure for this route during today's service.
+      DateTime.compare(now, last_schedule_today.departure_time) == :lt ->
+        :departure_scheduled_today
```
digitalcora, line 639: I think this is not the only place this can happen. We only enter this clause of `get_overnight_departure_for_route` when there is a scheduled departure tomorrow — there's another clause (the second, currently on L588) that applies when there is no scheduled departure tomorrow, and that produces a `nil` in at least some cases when there is a scheduled departure remaining today. (tbh, not sure how to make sense of the extra "`or` now is before today's first scheduled departure" there... it feels like either a mistake or a hack. might be worth digging into the blame on that one, and/or removing it and seeing if any tests fail)

lib/screens/v2/candidate_generator/dup/departures.ex
```diff
@@ -175,11 +175,11 @@ defmodule Screens.V2.CandidateGenerator.Dup.Departures do
       |> Enum.uniq()
 
     # Check if there is any room for overnight rows before running the logic.
-    overnight_schedules_for_section =
+    {todays_schedules_for_section, overnight_schedules_for_section} =
```
robbie-sundstrom, line 178: @digitalcora just wanted to draw your attention to a fundamental update to this logic in this revision.

Changed how this logic works, b/c `overnight_schedules_for_section` could include schedules that were not actually overnight. Splitting this up into `{todays_schedules_for_section, overnight_schedules_for_section}` makes more sense to me, and makes the logic below easier to track.

 Doesn't need to be done exactly like this -- i.e. `get_next_scheduled_trips_for_section` could return a struct, splitting up the departures into today and tomorrow could happen elsewhere. However, I did realize while drawing up the different logic branches that differentiating between the scheduled departures for today and tomorrow is how we actually process these cases differently and should be made clearer in the code.

digitalcora: I wonder if we should call this second value `tomorrow_schedules_for_section` or similar. Previously the word "overnight" was basically used as shorthand for "scheduled departures relevant to the overnight logic" which could include today's and/or tomorrow's. Now that it's split out, we could be more specific.

lib/screens/v2/candidate_generator/dup/departures.ex
```diff
@@ -460,7 +461,7 @@ defmodule Screens.V2.CandidateGenerator.Dup.Departures do
 
   # If we are currently overnight, returns the first schedule of the day for each route_id and direction for each stop.
```
digitalcora, line 462: This summary of the logic could use an update, as could the comments on the various clauses of `get_next_schedule_for_route`.

lib/screens/v2/candidate_generator/dup/departures.ex
```diff
@@ -525,9 +526,24 @@ defmodule Screens.V2.CandidateGenerator.Dup.Departures do
     # Routes not in overnight mode will be nil. Can ignore those.
     |> Enum.reject(&is_nil/1)
     |> Enum.sort_by(fn %Departure{schedule: schedule} -> schedule.departure_time end)
+    |> Enum.split_with(fn
+      # Split up the upcoming departures for each route into a tuple of lists, {todays_schedules, overnight_schedules}
+      %{schedule: %{departure_time: nil}} ->
```
digitalcora, line 531: * We should use the structs for pattern-matching to enable better compile-time checks and code analysis, e.g. `%Departure{schedule: %Schedule{departure_time: nil}}`.
* The flag for producing the overnight icon is _both_ departure and arrival time being `nil`. One or the other could potentially be `nil` for an ordinary, "real" departure; we might filter out "arrival-only" departures at an earlier stage of things anyway, but best to align things as closely as possible in case that ever changes.

lib/screens/v2/candidate_generator/dup/departures.ex
```diff
@@ -525,9 +526,24 @@ defmodule Screens.V2.CandidateGenerator.Dup.Departures do
...
+        # Return as an overnight schedule so it can be displayed with a moon 'overnight' icon if applicable
+        false
+
+      %{schedule: %{departure_time: departure_time}} ->
+        Util.service_date(departure_time) == Util.service_date(now)
+
+      departure ->
+        # Defensive fallback that we should never reach.
+        Report.warning("dup_overnight_schedule_without_departure_time", departure: departure)
+        false
```
digitalcora, line 542: This should be entirely impossible/unreachable — having it here is just noise, I think. These are structs, not plain maps, so they cannot be missing a field.

lib/screens/v2/candidate_generator/dup/departures.ex
```diff
@@ -624,8 +640,9 @@ defmodule Screens.V2.CandidateGenerator.Dup.Departures do
           DateTime.compare(now, first_schedule_today.departure_time) == :lt ->
         %Departure{schedule: first_schedule_today}
 
+      # Before 3am and before the last scheduled trip of the day.
       true ->
-        nil
+        %Departure{schedule: last_schedule_today}
```
digitalcora, line 645: I'm not sure this is how we want to handle this, and I'll explain my thinking:

Previously, assuming a section was configured not to display scheduled departures, we would only "override" that (displaying a scheduled departure anyway) when it was the _first_ trip of either today (call this the "early morning" scenario) or tomorrow ("late night"). This changes that behavior so we may also display the _last_ scheduled departure of today — which in some cases could be confusing since there may be other scheduled departures today before that one. Additionally, the way this is implemented using the `split_with`, this "last schedule today" is bundled together with the potential "first schedule today", so we can't distinguish these cases and display only the latter as we'd want.

digitalcora: IMO (2) is an undesired change to the current display behavior, and feels like it could generate some odd edge cases. When displaying a scheduled departure is specifically tied to the "early morning" and "late night" scenarios, it's easier to reason about. If we make this change, it's then possible for a scheduled departure to pop up at any time of day, whenever there's a sufficiently long gap in predictions. I don't know if I could confidently say this is an improvement, and it is certainly outside the scope of fixing the bug at hand. (It would also be a point of further divergence from PA/ESS and other touchpoints.)

However I don't think (1) is workable given the current logic either, because that gets us right back to still having the bug: if we don't "know" somehow that there _is_ a scheduled departure later today, we'll enter overnight mode when we shouldn't.

I'm a little confused as to how we ended up going down this path vs. the "signal atom" approach in the previous revision, which I thought worked fine and just needed to be extended slightly to cover an additional case where we were returning `nil` before (the clause where there is no scheduled departure tomorrow).

robbie-sundstrom: Thanks for breaking that down! To keep parity with other screens, I see how we should avoid showing these schedules during the day then.

As for why I changed the approach today, I wanted to specifically handle the case where:
- there are multiple routes in the section
- multiple routes have no predictions
- some of the routes have scheduled trips remaining today, but others have no schedules upcoming today

In this case, ideally we could show the overnight rows for the routes that are done for the day. This case isn't handled by the `:departure_scheduled_today` atom, as it will not show the overnight row for the completed routes. 

However, for simplicity's sake to resolve this bug, I've reverted to the atom based fix and updated it to handle when there are no scheduled trips tomorrow. The case I laid out is such an edge case, that I think it's ok to not show the overnight rows if it happens.

lib/screens/v2/candidate_generator/dup/departures.ex
```diff
@@ -458,8 +460,18 @@ defmodule Screens.V2.CandidateGenerator.Dup.Departures do
       MapSet.disjoint?(not_informed, informed_stop_ids)
   end
 
-  # If we are currently overnight, returns the first schedule of the day for each route_id and direction for each stop.
-  # Otherwise, return an empty list.
+  # For sections configured only to show predictions and not schedules, we may override
```
digitalcora, line 463: Technically the behavior here does not depend on whether the section is configured to show schedules. Even if we are showing scheduled departures, "overnight departures" may still be generated and displayed (in particular for departures outside the current service day, or the "service ended" rows with the moon icon). In general though, appreciating the added detail in this comment!

digitalcora (approved):
Whew, looks good to me!

PR #2588, opened 2025-05-29
feat: support multi-stop alerts configuration

digitalcora:
Minimal changes required to support Bus E-ink screens (only) being configured to display alerts for multiple stops. Preceded by some separate refactor commits, probably best reviewed per-commit.

Presentation may not be ideal due to conditions involving the screen's "home stop" always being considered false (as there is no single "home stop"), but since we have exactly one use case for this feature and don't expect that to change soon, the amount of effort applied here is appropriate to the impact.

**Asana task**: https://app.asana.com/1/15492006741476/project/1185117109217413/task/1209206882712035?focus=true


lib/screens/v2/location_context.ex
```diff
@@ -35,100 +33,192 @@ defmodule Screens.LocationContext do
...
+      {:ok, []} -> :error
+      {:ok, patterns} -> {:ok, get_tagged_parent_station_sequences(patterns)}
+      _ -> :error
+    end
+  end
+
+  # NOTE: only public due to use in tests. Should be treated as private.
+  @spec untag_stop_sequences(%{Route.id() => [[Stop.id()]]}) :: [[Stop.id()]]
+  def untag_stop_sequences(tagged_stop_sequences),
+    do: Enum.flat_map(tagged_stop_sequences, &elem(&1, 1))
```
robbie-sundstrom, line 192: This use of `elem/2` on a tuple seems brittle. Especially since this is a map, should we do something like: `Enum.flat_map(tagged_stop_sequences, fn {_route_id, sequences} -> sequences end)`

mix.exs
```diff
@@ -80,7 +80,7 @@ defmodule Screens.MixProject do
       {:recon, "~> 2.5.6"},
       {:remote_ip, "~> 1.2"},
       {:retry, "~> 0.18.0"},
-      {:screens_config, github: "mbta/screens-config-lib"},
+      {:screens_config, github: "mbta/screens-config-lib", branch: "cfg-alerts-stop-ids"},
```
robbie-sundstrom, line 83: Just commenting so this is updated before merge

lib/screens/v2/location_context.ex
digitalcora: I'd agree your version makes it easier to understand what's going on. This code was copied as-is from elsewhere, but no harm in improving it.

PR #2592, opened 2025-06-02
refactor: remove custom logic around GL E-ink LineMap generation

digitalcora:
Generating the LineMap widget for GL E-ink invokes a good bit of special-purpose fetching and parsing logic that isn't used anywhere else and has no test coverage. At this point we have a more flexible and better-tested `RoutePattern.fetch/1`, so let's use it.

This is incidental to investigating the linked Asana task / Sentry report. I'm not sure if it will resolve it outright, but I also don't think it could hurt.

**Asana task**: https://app.asana.com/1/15492006741476/project/1185117109217413/task/1210441924659979


robbie-sundstrom (approved):
I wasn't familiar with this GL map generation until looking through this code, but the changes make sense to me and seems correct in testing it!

PR #2598, opened 2025-06-05
fix: Pre-Fare flex zone crash with departures

digitalcora:
At transfer stations, having departures enabled could force an "inside" shuttle or suspension alert into the flex zone, which we did not have a relevant function clause for. To resolve this crash in the short term, reuse the presentation of "boundary" shuttle/suspension alerts, which does not actually have any elements specific to being on the boundary.

**Asana task**: https://app.asana.com/1/15492006741476/project/1185117109217413/task/1210474553813649

- [x] Tests added?

### DIFF
--- a/lib/screens/v2/widget_instance/reconstructed_alert.ex
+++ b/lib/screens/v2/widget_instance/reconstructed_alert.ex
@@ -796,18 +796,25 @@ defmodule Screens.V2.WidgetInstance.ReconstructedAlert do
     }
   end
 
-  defp boundary_flex_fields(%__MODULE__{alert: %Alert{effect: :suspension}} = t, location) do
+  # Boundary presentation of shuttle/suspension alerts is also valid when inside them.
+  defp inside_flex_fields(%__MODULE__{alert: %Alert{effect: effect}} = t)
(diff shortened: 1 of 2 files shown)


lib/screens/v2/widget_instance/reconstructed_alert.ex
```diff
@@ -796,77 +796,44 @@ defmodule Screens.V2.WidgetInstance.ReconstructedAlert do
...
-        effect: :shuttle,
+        effect: effect,
         urgent: true
       }
     end
   end
 
-  defp boundary_flex_fields(%__MODULE__{alert: %Alert{effect: :station_closure}}, _location),
-    do: nil
-
```
robbie-sundstrom, line 869: It doesn't look like we have a case to handle `%Alert{effect: :station_closure}` anymore. Is that intentional?

robbie-sundstrom: gotcha, makes sense! I could see us having a comment explaining that or Log a warning if we do end up in that case, but I don't think that's needed before merging this fix

PR #2602, opened 2025-06-09
feat: Omit no-data section if an alert is present

robbie-sundstrom:
**Asana task**: [DUP screens: Omit no-data section if an alert is present](https://app.asana.com/1/15492006741476/project/1185117109217413/task/1210227886153331?focus=true)

This essentially adds logic to not show a `NoDataSection` for routes with closure alerts for all routes in section
Some pieces to note: 
- This is not explicitly tied to the alerts widget, since those candidate generators run in parallel. However, if we have no departures for a given section and alerts tied to each, it's a fair assumption that we don't want to show a `NoDataSection` regardless.
- In this alerts scenario, behavior on the FE depends on the number of routes in a section, but returning an empty list of departures works for both:
  - If a page contains only one section of routes and they are closed, the alerts widget will override the entire screen with a full page alert. Returning an empty list of departures here works b/c it will be overridden anyway.
  - If a page contains multiple sections and only one has closed routes, the alerts widget will place a partial alert on the bottom of the screen. In this case, returning an empty list of departures works b/c it will take up no screen space.
- Testing this manually beyond unit tests is a bit messy and required overriding the V3 API and stubbing a few different scenarios. Namely returning no departures for a specific route and different variants of alerts. Happy to run through that with a reviewer if desired!

- [X] Tests added?

### DIFF
--- a/lib/screens/alerts/informed_entity.ex
+++ b/lib/screens/alerts/informed_entity.ex
@@ -3,9 +3,9 @@ defmodule Screens.Alerts.InformedEntity do
   Functions to query alert informed entities.
   """
 
-  alias Screens.Trips.Trip
-  alias Screens.Routes.Route
   alias Screens.Alerts.Alert
+  alias Screens.Routes.Route
+  alias Screens.Trips.Trip
 
   @type t :: Alert.informed_entity()
 
--- a/lib/screens/alerts/informed_entity.ex
+++ b/lib/screens/alerts/informed_entity.ex
@@ -32,21 +32,13 @@ defmodule Screens.Alerts.InformedEntity do
     match?(%{stop: "place-" <> _}, ie)
   end
 
-  @spec all_routes_represented?([Route.id()], Trip.direction() | :both | nil, [t()]) :: boolean()
-  def all_routes_represented?(route_ids, direction_id, informed_entities) do
-    Enum.all?(route_ids, fn route_id ->
-      Enum.any?(informed_entities, fn entity ->
-        entity_matches?(entity, route_id, direction_id)
-      end)
-    end)
-  end
-
-  defp entity_matches?(
-         %{route: entity_route, direction_id: entity_direction},
-         route_id,
-         direction_id
-       )
-       when entity_route.id == route_id do
+  @spec present_alert_for_route?(t(), Route.id(), Trip.direction() | nil) :: boolean()
+  def present_alert_for_route?(
+        %{route: entity_route, direction_id: entity_direction},
+        route_id,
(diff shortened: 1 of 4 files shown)


lib/screens/alerts/informed_entity.ex
```diff
@@ -29,4 +31,20 @@ defmodule Screens.Alerts.InformedEntity do
   def parent_station?(ie) do
     match?(%{stop: "place-" <> _}, ie)
   end
+
+  @spec present_alert_for_route?(t(), Route.id(), Trip.direction() | nil) :: boolean()
+  def present_alert_for_route?(
```
robbie-sundstrom, line 36: I kind of hate this function name if there are any suggestions 😅 

Wanted to make this updatable with the alert changes coming down the pipeline. We can either update this to also accept a list in that case and process the direction_ids of multiple entities.

lib/screens/v2/candidate_generator/dup/departures.ex
```diff
@@ -209,6 +210,11 @@ defmodule Screens.V2.CandidateGenerator.Dup.Departures do
           headsign: headsign
         }
 
+      # No departures, but no routes in the section are running, so no departures are expected.
+      # In this case, the alerts widget will display info on the closure, so we return an empty row.
```
digitalcora, line 213: typo? `empty row` => `empty section`

lib/screens/v2/candidate_generator/dup/departures.ex
```diff
@@ -231,6 +237,21 @@ defmodule Screens.V2.CandidateGenerator.Dup.Departures do
...
 
+  # Determines if the active alerts for a section apply to all routes that are enabled in the section
+  defp section_routes_disabled?(params, alert_informed_entities) do
+    case alert_informed_entities do
+      [] ->
+        false
+
+      _ ->
+        InformedEntity.all_routes_represented?(
+          params.route_ids,
```
digitalcora: This won't necessarily work, since configuring sections with a specific list of routes is not required (and, really, we'd _prefer_ not doing so if it's possible to get the same departures with only a stop ID and perhaps a route type filter, since that's more resilient to changes in service). I think what we want here is to pass in the `routes` field of the "section data" passed as the first argument to `get_section_instance`, which is populated via some data fetching performed in `get_routes_serving_section`.

robbie-sundstrom: Thanks for the callout! Updated to use the fetched routes instead of the routes from the configuration.

PR #2612, opened 2025-06-26
Normalize informed entities for new alert direction_id logic

deanshi:
**Asana task**: [Screens: support direction_id changes to alerts feed](https://app.asana.com/1/15492006741476/project/1185117109217413/task/1210571516808215?focus=true)

- Allow for handling informed entities with direction_ids going both ways by transforming the alert list to modify two informed entities into to one with a nil direction_id which will allow for the old nil handling logic to work

- [x] Tests added?

### DIFF
--- a/lib/screens/alerts/alert.ex
+++ b/lib/screens/alerts/alert.ex
@@ -167,8 +167,14 @@ defmodule Screens.Alerts.Alert do
         |> Map.put("include", Enum.join(includes, ","))
 
       case get_json_fn.("alerts", params) do
-        {:ok, response} -> {:ok, V3Api.Parser.parse(response)}
-        _ -> :error
+        {:ok, response} ->
+          {:ok,
+           response
+           |> V3Api.Parser.parse()
+           |> normalize_informed_entities_for_direction_id()}
+
+        _ ->
+          :error
       end
     end)
   end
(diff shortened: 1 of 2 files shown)


lib/screens/alerts/alert.ex
```diff
@@ -377,4 +384,36 @@ defmodule Screens.Alerts.Alert do
   def informs_stop_id?(%__MODULE__{informed_entities: informed_entities}, stop_id) do
     Enum.any?(informed_entities, &(&1.stop == stop_id))
   end
+
+  @spec normalize_informed_entities_for_direction_id([Alert.t()]) :: [Alert.t()]
+  def normalize_informed_entities_for_direction_id(alerts) do
+    Enum.map(alerts, fn alert ->
+      %{
```
digitalcora, line 391: Being explicit here and using `%__MODULE__{` gets us better compile-time checks and LSP functionality.

lib/screens/alerts/alert.ex
```diff
@@ -377,4 +384,36 @@ defmodule Screens.Alerts.Alert do
...
+            alert.informed_entities
+            |> Enum.reduce(%{}, fn entity, acc ->
+              key = Map.drop(entity, [:direction_id])
+
+              if entity.direction_id == nil do
+                Map.put(acc, key, entity)
+              else
+                acc
+                |> Map.fetch(key)
+                |> case do
```
digitalcora, line 403: This seems slightly overcooked — could be simplified to

```elixir
case Map.fetch(acc, key) do
```

test/screens/alerts/alert_test.exs
```diff
@@ -16,7 +16,27 @@ defmodule Screens.Alerts.AlertTest do
     "header" => "Route 1 experiencing delays up to 20 minutes due to an accident.",
     "image" => nil,
     "image_alternative_text" => nil,
-    "informed_entity" => [%{"activities" => ~w[BOARD EXIT RIDE], "route" => "1"}],
+    "informed_entity" => [
```
digitalcora, line 19: This is intended to be the _minimal_ valid attributes as per the naming and comment above. Rather than adding data for testing specific functionality to the "base" attributes that apply to all test cases, we should add the relevant attributes only within the test case, as in e.g. the `parses related facilities` test. (This could be a new `combines informed entities by direction_id` test.)

lib/screens/alerts/alert.ex
```diff
@@ -1,6 +1,7 @@
 defmodule Screens.Alerts.Alert do
   @moduledoc false
 
+  alias Screens.Alerts.Alert
```
digitalcora, line 4: This is the module we're in, so conventionally we'd refer to it using `__MODULE__` rather than aliasing it. However it looks like this was only added to allow putting `Alert.t()` in the typespec for the new function, which can also be expressed as just `t()`.

PR #2575, opened 2025-05-15
feat: standardize support for header types

digitalcora:
See mbta/screens-config-lib#59.

All screen types that previously supported one or both of the "stop name" and "stop ID" header types (Bus E-ink, Bus Shelter, Busway, DUP, Pre-Fare) now support both. The header widget generation is refactored into a shared module to reduce duplication.

**Asana task**: https://app.asana.com/1/15492006741476/project/1185117109217413/task/1209206882712031

PR #2615, opened 2025-07-03
chore: standardize headers across "LCD" screens

digitalcora:
Fold Pre-Fare header styles into the "LCD" styles already shared between Bus Shelter and Sectional screens. When these conflict, prefer the style of Pre-Fare headers.

**Asana task**: https://app.asana.com/1/15492006741476/project/1185117109217413/task/1210440988220696


assets/src/components/v2/normal_header.tsx
```diff
@@ -28,14 +28,6 @@ const ICON_TO_SRC: Record<Icon, string> = {
   logo_negative: "logo-black.svg",
 };
 
-const abbreviateText = (text: string) => {
-  if (text === "Government Center") {
-    return "Government Ctr";
```
digitalcora, line 33: Note: This one-off bit of abbreviation logic was only used on Pre-Fares. Since these [now support](https://github.com/mbta/screens/pull/2575) configuring arbitrary text as the header, we don't need this anymore.

robbie-sundstrom (approved):
Nice work!

PR #2616, opened 2025-07-03
fix: DUP special section heights

digitalcora:
In developer browsers, but not on real-world DUPs, no-data and headway sections would always take up two rows of height, even in the "reduced" layout where they were intended to only take up one. This was a CSS specificity issue which may not have affected real DUPs due to their very different (older) browser versions.

**Asana task**: https://app.asana.com/1/15492006741476/project/1185117109217413/task/1210675381138616

Screen Shot 2025-07-03 at 16 38 03

### DIFF
--- a/assets/css/v2/dup/departures/overnight.scss
+++ b/assets/css/v2/dup/departures/overnight.scss
@@ -1,5 +1,5 @@
 .overnight-section__row {
-  height: 408px;
+  height: 416px;
 }
 
 .overnight-departures__container {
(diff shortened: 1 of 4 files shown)


assets/css/v2/dup/departures/overnight.scss
```diff
@@ -1,5 +1,5 @@
 .overnight-section__row {
-  height: 408px;
```
robbie-sundstrom, line 2: Looks like 408px was just a mistake b/c it should be double 208, correct?

PR #2618, opened 2025-07-09
Added `hide_logo` configuration to Headers

deanshi:
**Asana task**: [Make T logo configurable in Sectional header](https://app.asana.com/1/15492006741476/project/1185117109217413/task/1210440988220701?focus=true)

Description
- Allows for hiding/showing the T Logo on headers
- Still only setup for Busway/Dup to match previous behavior and not having to configure every single one
- Currently defaults to always showing (maybe flip this so that we can allow for PreFares and other headers to have access to this field?)
- Waiting on https://github.com/mbta/screens-config-lib/pull/67

- [x] Tests added?


digitalcora:
Looking good so far, but we should add admin UI to the Busway table to be able to change this new field easily. As-is, the only way to change it would be using the "raw" config editors.

assets/src/components/admin/admin_tables.tsx
```diff
@@ -186,6 +186,19 @@ const alertsColumn = {
   FormCell: FormTextarea,
 };
 
+const includeLogoColumn = {
+  Header: "Include Logo?",
```
digitalcora, line 190: To economize on horizontal space even more, this could just be `Logo?`

assets/src/components/admin/admin_tables.tsx
```diff
@@ -186,6 +186,19 @@ const alertsColumn = {
...
+  accessor: (row) => row.app_params.include_logo_in_header,
+  mutator: (row, value) => {
```
digitalcora, line 192: I think we should be able to simplify this using `buildAppParamAccessor` and `buildAppParamMutator` here.

digitalcora (approved):
Couple comments on the admin UI, seems good pending the merge of the config-lib PR and updating the dep so it's not pointing to the branch.

PR #2620, opened 2025-07-14
feat: move screen_checker into app

digitalcora:
**Asana task**: https://app.asana.com/1/15492006741476/project/1185117109217413/task/1210487957046633

Incorporate the functions of `screen_checker` into this app, now named `DeviceMonitor`.

Since this app has multiple instances in production, some way to avoid duplicate logging is needed. We use the existing Memcached server to store the timestamp of the last logging run (see `DeviceMonitor.Store`), which then allows using Memcached CAS features to avoid race conditions.


deanshi (approved):
Just a small elixir question in a comment, but otherwise :shipit: 

With regards to multiple applications running in production, how does that work? Are we doing that for redudancy? or is there something else I'm missing with how `screen_checker` is working.

PR #2625, opened 2025-07-18
chore: use new Mercury device monitoring APIs

digitalcora:
Mercury introduced a `verbose` parameter for their API which allows fetching the details of all devices at once, instead of requiring N+1 requests. Using this also allows us to shorten the logging interval since we no longer have to wait multiple minutes for the N+1 requests to complete.

Also removes a `timeout` option from `Task.Supervisor.async_nolink/3` which does not actually exist.

**Asana task**: https://app.asana.com/1/15492006741476/project/1185117109217413/task/1210825464414329

### DIFF
--- a/lib/screens/device_monitor.ex
+++ b/lib/screens/device_monitor.ex
@@ -22,10 +22,6 @@ defmodule Screens.DeviceMonitor do
 
   alias __MODULE__.Store
 
-  @log_interval_minutes 5
-  # ensure logging functions don't run for longer than the log interval
-  @vendor_mod_timeout_ms 4 * 60 * 1000
-
   def start_link(opts \\ []) do
     with {:ok, store} <- Store.start_link(),
          do: GenServer.start_link(__MODULE__, %State{store: store}, opts)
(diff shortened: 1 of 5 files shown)


lib/screens/device_monitor.ex
```diff
@@ -92,19 +84,13 @@ defmodule Screens.DeviceMonitor do
...
-      |> truncate_to_interval(1)
+      |> truncate_seconds()
       |> DateTime.add(1, :minute)
       |> DateTime.add(1, :second)
       |> DateTime.diff(now, :millisecond)
 
     Process.send_after(self(), :run, send_after)
   end
 
-  # Truncate a datetime to the given interval size in minutes. For example, when the size is 5,
```
deanshi, line 103: Does this rollback some of the interval stuff that was just added?

digitalcora: Sort of, yeah 😅 That extra complexity was needed because we were using an interval longer than 1 minute, so just truncating off the seconds wasn't sufficient. Now that checking every minute is fine, it can be simplified back to whence it came.

deanshi (approved):

PR #2628, opened 2025-07-22
Single-tracking messaging for Subway Status

digitalcora:
Most widgets technically already support single-tracking alerts since these have an effect of `DELAY`. However, where possible we want to:

1. call out that these are "due to single tracking"
2. show them regardless of severity, allowing "informational" alerts

This change adds these aspects to the Subway Status widget.

---

Includes several distinct but related commits, best reviewed per-commit.

**Asana task**: https://app.asana.com/1/15492006741476/project/1185117109217413/task/1210764886971057


robbie-sundstrom (approved):
Looks great! Thanks for simplifying the tests so much as well with this change

PR #2629, opened 2025-07-25
chore: Add basic logging in case of overnight departures appended in dups

robbie-sundstrom:
**Asana task**: [Investigate: DUPs showing "overnight" departures incorrectly](https://app.asana.com/1/15492006741476/project/1208877354406742/task/1210559658847355?focus=true)

Adds a  basic log in the bug we had raised in the ticket. When there are still departures being shown, but not enough departures to fill up the entire section, we include the first of the sorted overnight departures. Log the overnight departures and the available departures we've found.

- [ ] Tests added?

### DIFF
--- a/lib/screens/v2/candidate_generator/dup/departures.ex
+++ b/lib/screens/v2/candidate_generator/dup/departures.ex
@@ -4,6 +4,7 @@ defmodule Screens.V2.CandidateGenerator.Dup.Departures do
   require Logger
 
   alias Screens.Alerts.{Alert, InformedEntity}
+  alias Screens.LogScreenData
   alias Screens.Report
   alias Screens.Routes.Route
   alias Screens.Schedules.Schedule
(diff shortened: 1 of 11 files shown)


lib/screens/v2/candidate_generator/dup/departures.ex
```diff
@@ -224,15 +237,18 @@ defmodule Screens.V2.CandidateGenerator.Dup.Departures do
...
-        departures = departures ++ overnight_schedules_for_section
-
-        visible_departures =
-          if is_only_section do
-            Enum.take(departures, 4)
+        departures =
+          if length(departures) < max_visible_departures and overnight_schedules_for_section != [] do
+            # Temporary logs for insight into a bug with incorrect departure for next day being shown
+            # https://app.asana.com/1/15492006741476/project/1185117109217413/task/1210559658847355
+            LogScreenData.log_dup_data(screen_name, departures, overnight_schedules_for_section)
```
robbie-sundstrom, line 244: I also considered adding the raw schedule data that we are pulling from the API in `get_overnight_schedules_for_section` by adding an additional field to be returned by this function. I think that could be helpful for us to have logged, but will definitely make this messier just for the sake of debugging

deanshi:
Is this something you could deploy to dev-green/dev and collect logs from there? Are the bugs specific to production?

robbie-sundstrom:
@deanshi I haven't been able to reproduce this bug, so this logging is to hopefully give us some insight into what happens if it's ever reported again in the future. I did actually deploy this to dev-green, and did some brief refreshing of different DUP screens in dev-green throughout the day. However, I wasn't even able to properly trigger this specific path where the bug occurred and logging happens, let alone reproduce the bug down that path in dev-green.

I also see that I broke a unit test and am looking at that now.

robbie-sundstrom:
Closing because I figured out the root cause of this bug, which is addressed here: https://github.com/mbta/screens/pull/2631

PR #2630, opened 2025-07-29
Single-tracking messaging for DUP Alerts

digitalcora:
Most widgets technically already support single-tracking alerts since these have an effect of `DELAY`. However, when the cause of a delay is single-tracking specifically, we want to make this info more prominent, including showing the alert at severity levels we otherwise wouldn't. This updates the handling of DUP alerts accordingly.

---

Includes several distinct but related commits, best reviewed per-commit.

**Asana task**: https://app.asana.com/1/15492006741476/project/1185117109217413/task/1210603551669549

### DIFF
--- a/config/test.exs
+++ b/config/test.exs
@@ -92,6 +92,14 @@ config :screens,
     ]
   },
   dup_alert_headsign_matchers: %{
+    "place-r1" => [
+      %{
+        informed: "place-r2",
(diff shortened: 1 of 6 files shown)


panentheos (approved):
Looks good! Thanks for splitting this out into multiple commits, it made it much easier to review.

PR #2631, opened 2025-07-31
Sort overnight schedules in DUPs departures

robbie-sundstrom:
**Asana task**: [Investigate: DUPs showing "overnight" departures incorrectly](https://app.asana.com/1/15492006741476/project/1208877354406742/task/1210559658847355?focus=true)

To fix the incorrect overnight times being shown on DUPs, this passes the param to the v3 API to sort departures by ascending order time. This ensures that the first/last schedule of today and first schedule of tomorrow that are found are all correct.


digitalcora (commented):
Great find!! And I'm sorry fixing this caused so much pain — I think this was sort of my fault, due to some recent refactoring of departure fetching logic. It looks like everywhere _else_ we fetch schedules, we already sort them either by using the API param or on the client side.

PR #2633, opened 2025-08-05
Added Silver Line headway mapping to stop_ids

deanshi:
**Asana task**: [SL Headway Key Mapping Update](https://app.asana.com/1/15492006741476/project/1185117109217413/task/1210911817494840?focus=true)

Description
- Also added headway logic for parsing the stop_id and route_id into the headway key
- Added tests for the new silver line headway logic

- [x] Tests added?


lib/screens/headways.ex
```diff
@@ -203,4 +219,9 @@ defmodule Screens.Headways do
       {_, false} -> :off_peak
     end
   end
+
+  defp silver_line_route_headway_key("741"), do: "sl_one"
+  defp silver_line_route_headway_key("742"), do: "sl_two"
+  defp silver_line_route_headway_key("743"), do: "sl_three"
```
robbie-sundstrom, line 225: There's also the 746, SLW, which I think you should include. See here: https://api-v3.mbta.com/routes/746

lib/screens/headways.ex
```diff
@@ -173,8 +184,9 @@ defmodule Screens.Headways do
     defp headway_key(unquote(to_string(stop_id)), _route_id), do: unquote(to_string(key))
   end
 
-  for {key, stations} <- @stations, station <- stations do
-    defp headway_key("place-" <> unquote(station), _route_id), do: unquote(to_string(key))
+  for {_, ranges} <- @sl_stops, range <- ranges, stop_id <- range do
+    defp headway_key(unquote(to_string(stop_id)), route_id),
+      do: silver_line_route_headway_key(route_id)
```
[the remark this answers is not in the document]
robbie-sundstrom: I think using `sl_one`, etc. as the keys makes this a bit clearer for people who are newer to the codebase, and `silver_line_route_headway_key` is a nice little helper function. Even though at this point, the SL `route_ids` are forever embedded in my head.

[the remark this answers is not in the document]
digitalcora: I don't think this is implemented correctly — it only works for the case where a route ID is provided (ultimately, only for `get_with_route`, but not `get`). Try testing e.g. `Headways.get("30250")`.

It seems like this _would_ work correctly if we omitted this extra macro block and made the contents of `@sl_stops` part of `@stops` instead of a separate thing. Any reason that wouldn't work? (I'm not sure I understand the point about "using 741/etc. as the key"; we _are_ doing that already in this version of the code, in `@multi_stations`, and that is the only place expressing a route ID is required)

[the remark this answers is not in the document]
deanshi: I initially had a version of this where I had the common stop IDs in the `@stops`, which was causing picking the correct route. When I moved those overs to the @multi_stations section, I failed to move the @sl_stops back into the @stops.

lib/screens/headways.ex
```diff
@@ -173,16 +184,22 @@ defmodule Screens.Headways do
...
   end
 
+  for {key, {route_ids, stations}} <- @sl_multi_stations,
+      route_id <- route_ids,
+      station <- stations do
+    defp headway_key(unquote(station), unquote(route_id)), do: unquote(to_string(key))
+  end
+
+  for {key, stations} <- @stations, station <- stations do
+    defp headway_key("place-" <> unquote(station), _route_id), do: unquote(to_string(key))
```
digitalcora, line 200: This probably shouldn't be moved, as it has been here — the order of definitions matters for the pattern-matching.

lib/screens/headways.ex
```diff
@@ -124,16 +126,25 @@ defmodule Screens.Headways do
...
-    red_trunk: {~w[Red], ~w[pktrm dwnxg]}
+    red_trunk: {~w[Red], ~w[pktrm dwnxg]},
+    silver_seaport: {~w[741 742 746], ~w[conrd wtcst crtst sstat]},
+    silver_chelsea: {~w[743], ~w[conrd wtcst crtst sstat aport]}
+  }
+
+  @sl_multi_stations %{
+    # congress_st_at_wtc 17_096
+    silver_seaport: {~w[741 742 746], ~w[17096]},
+    silver_chelsea: {~w[743], ~w[17096]}
```
digitalcora, line 147: It seems the only reason this needs to be a separate module attribute is that `17096` is a "standalone" stop and not a parent station, so it doesn't have the `place-` prefix which is otherwise assumed. To be clear about this we could call it `@multi_stops` (or something, it just shouldn't have "station" in the name IMO).

lib/screens/headways.ex
```diff
@@ -12,34 +12,36 @@ defmodule Screens.Headways do
   # Compact mapping of stop IDs to headway keys, leaning on the fact that subway stop IDs happen
   # to be numeric and often contiguous as we "traverse" the line in a given direction.
   @stops %{
-    blue_trunk: [70038..70060],
-    glx_medford: [70505..70514],
-    glx_union: [70503..70504],
+    blue_trunk: [70_038..70_060],
+    glx_medford: [70_505..70_514],
+    glx_union: [70_503..70_504],
```
digitalcora, line 17: There is a lot of re-formatting noise here that doesn't seem necessary (the old version passed format checking as well).

lib/screens/headways.ex
```diff
@@ -124,16 +126,27 @@ defmodule Screens.Headways do
...
     green_trunk: {~w[Green-B Green-C Green-D Green-E], ~w[north haecl gover pktrm]},
     orange_trunk: {~w[Orange], ~w[north haecl state dwnxg]},
-    red_trunk: {~w[Red], ~w[pktrm dwnxg]}
+    red_trunk: {~w[Red], ~w[pktrm dwnxg sstat]},
+    silver_seaport: {~w[741 742 746], ~w[conrd wtcst crtst sstat]},
+    silver_chelsea: {~w[743], ~w[conrd wtcst crtst sstat aport]}
+  }
+
+  # Mapping of parent stops and route IDs to headway keys for the Silver Line,
+  # for parent stops which serve more than one route.
```
digitalcora, line 145: We shouldn't use "parent" here as these are specifically _not_ parent stops (they are "standalone stops", as we sometimes call stopping locations with no parent).

digitalcora: Follow-up: "parent" stops has been removed from this line, but not the line above ("Mapping of parent stops and...")

deanshi:
So I wanted to bring up - I tried testing it with some alerts but it seems like I ran into some issues dealing with the `dup_alert_headsign_matchers`. It looks like there's a script that needs to be run `add_dup_headsign_config` in order to generate the headsigns for the appropriate routes. 

Should I run through this and make sure that it's appropriately setup in cases for Alerts? The alerts should be made obsolete by the new headway work right?

digitalcora:
Good point, "alert headsign matchers" do represent a unique piece of functionality in the existing logic and I'm not sure whether we want to carry that forward (and only use "xbound" messages when there is no alert) or just replace it with the "xbound" messages entirely.

The advantages of the former are that it might be a more targeted, less invasive code change (basically adding a "non-alert headway mode" separate from the existing "alert headway mode"), and it maintains parity with alert messages, which will still be using headsigns instead of direction names. This is especially relevant since in the "boundary" case, full-screen headways combine with the alert banner to create what looks like a single message: "[Bowdoin trains every 6-8 minutes // No Wonderland trains](https://github.com/mbta/screens/pull/2646)". It would be strange if this said instead "**Southbound** trains every 6-8 minutes // No Wonderland trains", with a direction name in one half of the message and a destination in the other.

The advantage of the latter would be that we consistently use the same presentation for headway messages regardless of whether an alert is involved, and the overall complexity of the resulting code might be lower.

I'm leaning on the side of the former approach due to the rider-facing implications, which maybe we should loop in Sarah about, but let me know what you think.

deanshi:
I think I'm in favor of the former approach too. I think it just makes sense to have it be consistent for riders instead of the `xBound` + `No ____ trains` messaging in cases of alerts, especially if we could have that information anyway. 

In terms of complexity/maintenance - I think I need to look at the script but I think from my skim over it looked like it was pretty well documented and I don't _think_ we're adding any more DUPs soon right? So it might be a one off for a little while (famous last words). 

I'll message Sarah to see if it should go in now, or if we want it as a fast follow or even something that isn't too much of a priority

PR #2636, opened 2025-08-06
feat: Add CapeFlyer route pill color

robbie-sundstrom:
**Asana task**: [Fix incorrect route pill for Cape Flyer](https://app.asana.com/1/15492006741476/project/1185117109217413/task/1210794505820587?focus=true)

Add the route pill color for the CapeFlyer. According to the V3 API, this route's color is #006595. See [route data here](https://api-v3.mbta.com/routes/CapeFlyer).
image

- [X] Tests added?

### DIFF
--- a/assets/css/v2/dup/departures/route_pill.scss
+++ b/assets/css/v2/dup/departures/route_pill.scss
@@ -38,7 +38,7 @@
   &--ocean_blue {
     background-color: colors.$line-capeflyer;
   }
-  
+
   &--yellow {
     width: 280px;
     height: 168px;
--- a/assets/css/v2/dup/departures/route_pill.scss
+++ b/assets/css/v2/dup/departures/route_pill.scss
@@ -35,7 +35,7 @@
     background-color: colors.$line-ferry;
   }
 
-  &--cape_blue {
+  &--ocean_blue {
     background-color: colors.$line-capeflyer;
   }
   
--- a/assets/css/v2/dup/departures/route_pill.scss
+++ b/assets/css/v2/dup/departures/route_pill.scss
@@ -35,6 +35,10 @@
     background-color: colors.$line-ferry;
   }
 
+  &--cape_blue {
+    background-color: colors.$line-capeflyer;
+  }
+  
   &--yellow {
     width: 280px;
     height: 168px;
--- a/lib/screens/routes/route.ex
+++ b/lib/screens/routes/route.ex
@@ -36,8 +36,8 @@ defmodule Screens.Routes.Route do
         }
 
   @typep name_colors :: :blue | :green | :orange | :red | :silver
-  @type color :: name_colors() | :purple | :teal | :yellow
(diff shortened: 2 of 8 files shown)


lib/screens/routes/route.ex
```diff
@@ -128,6 +129,7 @@ defmodule Screens.Routes.Route do
   def icon(%{id: "Mattapan"}), do: :mattapan
   def icon(%{id: "Orange"}), do: :orange
   def icon(%{id: "Red"}), do: :red
+  def icon(%{id: "CapeFlyer"}), do: :capeflyer
```
digitalcora, line 132: I think this may not be sufficient to support this pill type in `FreeText` as the doc comment references — we might need updates to `screens-config-lib` where the `FreeText` type is defined, the FreeText frontend component, and the CSS files which support that (of which there are some different ones for different screen types).

assets/css/v2/dup/departures/route_pill.scss
```diff
@@ -35,6 +35,10 @@
     background-color: colors.$line-ferry;
   }
 
+  &--ocean_blue {
+    background-color: colors.$line-capeflyer;
+  }
```
digitalcora, line 40: We should make corresponding changes to other CSS "implementations" of route pills; this file is only relevant to DUPs.

robbie-sundstrom: I was intentionally only doing this for DUPs since it's only relevant at South Station and Braintree, but thinking about it more, this does seem safer to future proof for any additional screens we add at these stations.

lib/screens/routes/route.ex
digitalcora: I think the only impact on `screens-config-lib` itself is indeed the typespec, which doesn't need to block this. However there are also the changes to the actual implementation of FreeText. If you trace where this `icon` function is used you'll find it's put into `FreeTextLine` structs which end up as `FreeText` components on the frontend. (Annoyingly this is a second, completely divergent implementation of the concept of "route pills".)

assets/css/v2/dup/departures/route_pill.scss
digitalcora: Yeah, I think it'd be best not to leave any unnecessary surprises here for our future selves. The handling of route pills in general is already reasonably "complete" and consistent across various contexts and screen types, so it would be unexpected for this one route to only work on DUPs.

digitalcora:
It occurs to me that since this changes DUP frontend code, it would also require shipping a new build to Outfront. We may want to "simulate" what would happen with current DUPs by locally undoing the frontend changes but leaving the backend ones. If the result is a display "more broken" than what we currently have, we should think about doing an intermediate quick fix that maybe uses the existing CR purple pill for the CapeFlyer, and then a separate more-comprehensive fix that adds the proper color (and whose merging will be blocked on getting that new build shipped and tested).

PR #2638, opened 2025-08-08
Rs/capeflyer shows cr pill

robbie-sundstrom:
**Asana task**: [Fix incorrect route pill for Cape Flyer](https://app.asana.com/1/15492006741476/project/1185117109217413/task/1210794505820587?focus=true)

Updates the CapeFlyer to use the purple commuter rail color with a rail icon on DUPs.

image

- [X] Tests added?

### DIFF
--- a/lib/screens/routes/route.ex
+++ b/lib/screens/routes/route.ex
@@ -123,6 +123,7 @@ defmodule Screens.Routes.Route do
   @spec icon(t()) :: icon()
   def icon(%{id: "Blue"}), do: :blue
   def icon(%{id: "Boat-" <> _}), do: :ferry
+  def icon(%{id: "CapeFlyer"}), do: :capeflyer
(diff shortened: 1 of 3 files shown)


lib/screens/routes/route.ex
```diff
@@ -123,6 +123,7 @@ defmodule Screens.Routes.Route do
   @spec icon(t()) :: icon()
   def icon(%{id: "Blue"}), do: :blue
   def icon(%{id: "Boat-" <> _}), do: :ferry
+  def icon(%{id: "CapeFlyer"}), do: :capeflyer
```
digitalcora, line 126: I think for this fix we should just use `:cr` here to be minimally disruptive. As I mentioned in the other PR, this also applies to `FreeText` which is used in some places on DUPs, and that component will not understand this new "icon type" even if departure rows do.

PR #2647, opened 2025-08-13
fix: tolerate invalid config in no-data message

digitalcora:
The logic for directional no-data messages added in b2b1ca40 encountered [a crash](https://mbtace.sentry.io/issues/6807063531/?alert_rule_id=9523677&alert_timestamp=1755060442227&alert_type=email&environment=screens-prod&notification_uuid=076671d8-80e3-47c4-b4d5-eacc60bc4d6a&project=6061747&referrer=alert_email) when screens were mistakenly configured with a `direction_id` of `null`, when the correct value to use for no direction filtering is `"both"`. This restructures the logic to check specifically for a valid direction ID, and otherwise show the generic message.


robbie-sundstrom (approved):
nice reordering to make this more "fault tolerant"

PR #2653, opened 2025-08-21
chore: target frontend build to supported browsers

digitalcora:
Previously we configured Babel using a browserslist query of `> 0.25%`, which happened to include all the compatibility we needed but also some we didn't (for example, IE 11). Earlier attempts to improve on this were reverted when they broke GDS E-ink units. Now that we no longer have to support this hardware, we can try again.

* Add a `.browserslistrc` enumerating all the hardware we support and their browser versions. When not configured with an explicit `targets` option, Babel uses this file automatically.

* Remove `regenerator-runtime` and `whatwg-fetch` as dependencies. The former is no longer needed to transform generator syntax as of Babel 7.18, and the latter should no longer be necessary since all of our supported browsers have native Fetch.

* Configure Babel with `useBuiltIns: "usage"`. This removes the need to explicitly import `core-js`, instead automatically inserting imports to polyfill only the features our code uses and only if they are needed by our target browsers (and since this was the only import left in our `polyfills` entrypoint, that can be entirely removed).

* Update Babel packages to their latest versions.

All together this shaves ~130KB off our total shipped code size for each "screen app".

**Asana task**: https://app.asana.com/1/15492006741476/project/1185117109217413/task/1211080013682284


robbie-sundstrom (approved):
LGTM 🚢

PR #2657, opened 2025-08-22
chore: disable unneeded telemetry logging

digitalcora:
This method of sending telemetry was regularly overloading the Splunk log handler and causing more important logs to be dropped (in particular DeviceMonitor, which we rely on for alerting us to screen outages). We have not actually used these log spans to investigate anything in a long time, so we'll remove them until we may need them again.

---
  - https://app.asana.com/0/0/1211123747782268

### DIFF
--- a/config/config.exs
+++ b/config/config.exs
@@ -38,7 +38,7 @@ config :ex_aws, json_codec: Jason
 config :elixir, :time_zone_database, Tzdata.TimeZoneDatabase
 
 config :hackney, mod_metrics: :hackney_telemetry
-config :hackney_telemetry, report_interval: 5_000
+config :hackney_telemetry, report_interval: 10_000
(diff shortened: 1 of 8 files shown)


deanshi (approved):
:shipit: What made you suspect that we were overloading the Splunk log handler?

Also if you think this covers the Prefare/Bus Shelter ticket, I can reassign that to you if you want.

digitalcora:
When a log handler is overloaded a message like this appears in the logs:

```
[notice] Handler :default switched from :sync to :drop mode
```

The mode can also be `:async`, but in any case `:drop` mode means that log messages are just dropped. I found that when our DeviceMonitor logging was "missing", it was often had a "switching to drop mode" message just before the time where the log should have appeared.

PR #2659, opened 2025-08-25
build: use automatic JSX transform

digitalcora:
Babel 7.9.0 and up support an alternate method of transforming JSX (to become the default in Babel 8) which does not require using `import React from "react"` for JSX element syntax to be "available". Instead, it automatically inserts a more targeted import statement. See:

https://legacy.reactjs.org/blog/2020/09/22/introducing-the-new-jsx-transform.html

This switches to the new method and eliminates now-unnecessary import statements.


deanshi (approved):
🧹 :shipit:

PR #2662, opened 2025-08-27
Reintroduce OvernightSection for DUPs

deanshi:
- We previously disabled OvernightSections due to a bug dealing with it during the day for stations that had predictions suppressed.
- This was due to the previous logic handling, which checked if we have any active routes (which would be false due to prediction suppressions) and any scheduled trains for tomorrow (which would be true) and we would default to showing the overnight section.
- This new logic adds a check to see if our current time is past the last scheduled trip for today OR before the first scheduled trip for today (which would indicate that we're inbetween schedules and overnight). It also adds the overnight check for alerts, so we're not showing alert headways in cases where it's overnight.
- As a note, this is very similar to some logic we're already using today to calculate for overnight_departures, and also is similar to some of the overnight logic we're using in RTS with the exception of the is_last_trip checks (which is omitted here)

**Asana task**: [Follow-up on temporary disabling of DUP overnight mode](https://app.asana.com/1/15492006741476/project/1185117109217413/task/1210998350896155?focus=true)

Description

- [X] Tests (re)added?

### DIFF
--- a/lib/screens/v2/candidate_generator/dup/departures.ex
+++ b/lib/screens/v2/candidate_generator/dup/departures.ex
@@ -176,20 +176,29 @@ defmodule Screens.V2.CandidateGenerator.Dup.Departures do
 
     max_visible_departures = if is_only_section, do: 4, else: 2
     # Check if there is any room for overnight rows before running the logic.
+
+    route_and_direction_schedules =
+      get_route_direction_schedules(params, routes, now, fetch_schedules_fn)
+
     {section_contains_active_route, overnight_schedules_for_section} =
       if length(departures) >= max_visible_departures do
         {false, []}
       else
         get_overnight_schedules_for_section(
+          route_and_direction_schedules,
           routes_with_live_departures,
           params,
           routes,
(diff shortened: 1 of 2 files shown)


lib/screens/v2/candidate_generator/dup/departures.ex
```diff
@@ -473,6 +482,47 @@ defmodule Screens.V2.CandidateGenerator.Dup.Departures do
...
+          Enum.find(
+            todays_schedules,
+            &(&1.route.id == route_id and &1.direction_id == direction_id)
+          )
+
+        last_schedule_today =
+          Enum.find(
+            Enum.reverse(todays_schedules),
+            &(&1.route.id == route_id and &1.direction_id == direction_id)
+          )
```
digitalcora, line 501: There seems to be a lot of overlap between this function and `get_overnight_schedules_for_section`, with some code duplicated outright. Could these be consolidated? Maybe `overnight_time_period?` could take as an input the values returned from that function?

PR #2663, opened 2025-08-27
fix: increase timeout for Mercury devices endpoint

digitalcora:
[Slack thread for context](https://mbta.slack.com/archives/C059D8A3J1G/p1756215540250559). This should resolve some false positive alerts about Mercury devices being offline.


deanshi (dismissed):

digitalcora:
@deanshi I'll take the approval — the only reason this was still draft is because I wanted to let it sit on dev-green for a bit and check the logs.

deanshi (approved):

PR #2664, opened 2025-08-27
Decreased font size for DUP route pills text length > 3

deanshi:
- Added the small modifier for route-pill__text
- Had to change the size modifiers logic to check explicitly for TR text

**Asana task**: [Adjust font size for 2-digit CR Track Pills](https://app.asana.com/1/15492006741476/project/1185117109217422/task/1211138992434689?focus=true)

- [ ] Tests added?

Screenshot 2025-08-27 at 2 21 43 PM
Screenshot 2025-08-27 at 2 22 00 PM

### DIFF
--- a/assets/src/components/v2/departures/route_pill.tsx
+++ b/assets/src/components/v2/departures/route_pill.tsx
@@ -56,7 +56,11 @@ const TextRoutePill: ComponentType<TextPill & { outline?: boolean }> = ({
     modifiers.push(size);
   } else {
     const routeNum = Number(text);
-    modifiers.push(isNaN(routeNum) || routeNum > 199 ? "small" : "large");
+    if (isNaN(routeNum)) {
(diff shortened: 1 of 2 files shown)


assets/src/components/v2/departures/route_pill.tsx
```diff
@@ -54,6 +54,8 @@ const TextRoutePill: ComponentType<TextPill & { outline?: boolean }> = ({
 
   if (size) {
     modifiers.push(size);
+  } else if (text.startsWith("TR")) {
+    modifiers.push(text.length > 3 ? "small" : "large");
```
deanshi, line 58: A note - switched the logic here because `Number(text)` was returning NaN for all `TR` text. I can modify the logic there, but I'm not sure what other text can show up within this flow that might actually want to be NaN.

digitalcora (approved):
✨

PR #2670, opened 2025-08-29
Improve audio readout admin tools

digitalcora:
Two small Inspector features to assist with gathering info for the audio readouts epic:

* In the "Show SSML" dialog, show the count of billable and total characters in the SSML, and the limits for each.
* Show controls for the audio readout, enabling pausing and seeking the audio as well as seeing the length of the readout.

---
  - https://app.asana.com/0/0/1211176661292117


panentheos (approved):
Nice improvements!

PR #2672, opened 2025-09-02
Added boundaries as part of severe single tracking alerts

deanshi:
**Asana task**: [Update Single Tracking on DUPs Boundary Treatment](https://app.asana.com/1/15492006741476/project/1185117109217413/task/1211138992434694?focus=true)

- [x] Tests added?

### DIFF
--- a/test/screens/v2/widget_instance/dup_alert_test.exs
+++ b/test/screens/v2/widget_instance/dup_alert_test.exs
@@ -209,9 +209,10 @@ defmodule Screens.V2.WidgetInstance.DupAlertTest do
       assert serialized(screen, context, alert) == [banner, banner, banner]
     end
 
-    test "inside severe single-tracking" do
+    test "inside or on the boundary of severe single-tracking" do
       screen = build_screen()
-      context = build_location_context("place-r2")
+      inside_context = build_location_context("place-r2")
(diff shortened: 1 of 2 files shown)


test/screens/v2/widget_instance/dup_alert_test.exs
```diff
@@ -244,6 +244,41 @@ defmodule Screens.V2.WidgetInstance.DupAlertTest do
       assert serialized(screen, context, alert) == [banner, takeover, banner]
     end
 
+    test "on boundary of severe single-tracking" do
```
digitalcora, line 247: Since this entire test is identical to the one above except for the `build_location_context` line, and this is ultimately testing the same code path, the signal-to-noise ratio seems less than ideal. What about folding this into the same scenario and just testing two different contexts?

```elixir
screen = build_screen()
inside_context = build_location_context("place-r2")
boundary_context = build_location_context("place-r3")

# ...

assert serialized(screen, inside_context, alert) == [banner, takeover, banner]
assert serialized(screen, boundary_context, alert) == [banner, takeover, banner]
```

digitalcora (approved):
One suggestion on the testing setup but functionally looks fine!

PR #2683, opened 2025-09-04
deps(dev): remove `mix_test_watch`

digitalcora:
We don't seem to use this tool on any other Screens team projects, and I personally have never used it in this project. Propose removing it on the basis that it's one less dependency to install/compile/update. Similar functionality can be had by passing `mix test --stale` into a generic "watch for file changes in a directory and run a command" tool, such as [fswatch](https://github.com/emcrisostomo/fswatch?tab=readme-ov-file#usage) or [hwatch](https://github.com/blacknon/hwatch).


panentheos (approved):
Makes sense to me!

PR #2707, opened 2025-09-17
DUP removed banner for single tracking information alert

deanshi:
**Asana task**: [DUP Info Level Single Tracking Alerts](https://app.asana.com/1/15492006741476/project/1185117109217413/task/1211229715108431?focus=true)

Description
- For informational severity alerts for single tracking, we now omit the alert banner

- [ ] Tests added?

### DIFF
--- a/lib/screens/v2/candidate_generator/dup/alerts.ex
+++ b/lib/screens/v2/candidate_generator/dup/alerts.ex
@@ -112,8 +112,12 @@ defmodule Screens.V2.CandidateGenerator.Dup.Alerts do
     end
   end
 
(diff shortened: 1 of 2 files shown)


deanshi:
Side question - does this change warrant a unit test? Is there a DUP Alert Candidate Generator test file somewhere I would add this to?

digitalcora (approved):
Looks good. To your question, there seems to be no unit test suite for this module yet. Since this is a very minor and "safe" change I'm comfortable pushing it out without digging into the larger task of creating that test suite, but if you have time, it would certainly be appreciated!

PR #2708, opened 2025-09-18
fix: rework text auto-sizing hook

digitalcora:
`useTextResizer` had a known bug where the "sizing step" could stop one step short of the correct one, leaving the element visibly overflowing. This was most likely due to its use of `useEffect`, which runs async with DOM updates and is not guaranteed to "see" the DOM in a particular state. In React 17 it happened that this bug only occurred rarely, but in React 18, presumably due to internal changes or optimizations in the render cycle, it now occurs more often.

The main change necessary here is to switch to `useLayoutEffect`, which *is* guaranteed to see the DOM after it has been updated but before the browser repaints. However, while we're here, we can make some further improvements:

* Check for overflow directly rather than requiring the hook user to specify a maximum pixel height. This allows size constraints to be expressed in CSS where they belong.

* Remove the `isDone` return value. No users of the hook actually needed this other than to know when the final resizing step had been reached, which can also be checked more directly.

**Asana task**: https://app.asana.com/1/15492006741476/project/1185117109217413/task/1208206770696512?focus=true


robbie-sundstrom (approved):
I haven't tested this thoroughly, but the changes make sense to me and look good!

digitalcora:
If you're interested, the main method I used to test this end-to-end was to change widget generation on the server to randomly return either a short or long piece of text rather than the text it would have normally returned. Then tested that the text was sized correctly A) when initially loaded as either short or long, and B) when a data refresh changed it from long to short or vice-versa.

PR #2725, opened 2025-09-26
chore: disable "auto" behaviors in admin forms

digitalcora:
We don't want things like spell check being applied to JSON editors, where they are more annoying than helpful.


panentheos (approved):
Looks good!

PR #3003, opened 2026-04-10
chore: update typespec for serialize_time/3

rwaskiewicz:
**Asana task**: [Handle edgecase with crowding data for timestamp departures](https://app.asana.com/1/15492006741476/project/1213833239363217/task/1213225188829790)

Description

- [X] Tests added?

Notes for Reviewers:

This PR updates the typespec for `serialize_time/3`

**Important/Historical Knowledge**

In https://github.com/mbta/screens/pull/3003#issuecomment-4226584975, Cora called out that this work is unnecessary following the completion of https://github.com/mbta/screens/pull/2948 (3d11f35). There was a typespec change from my PR that we will propagate forward, but the original task/PR is a no-op/unneeded at this time 

<details>
<summary>Expand to see original PR summary</summary>

Original Title: fix: do not show crowding when departures are more than an hour out 

Original Summary

- Robbie and I discussed this task in a huddle. I originally asked how we handle the acceptance criteria/eng note:
  > Rather than not show crowding for timestamp, instead do not show crowding when the time display is too wide for the UI to account for it

    Talking it through, we think this can be accomplished on the backend (which runs slightly counter to what's in the ticket
- I'm still getting used to the domain + Elixir - I'm sure at least one thing is wrong here. I'm going to try to call out things I'm unsure of, but let me know what I'm missing/got wrong
</details>

### DIFF
--- a/lib/screens/v2/departure.ex
+++ b/lib/screens/v2/departure.ex
@@ -12,6 +12,8 @@ defmodule Screens.V2.Departure do
   alias Screens.V3Api
   alias Screens.Vehicles.Vehicle
 
+  @hour_in_seconds 3600
+
   @type t :: %__MODULE__{
           prediction: Screens.Predictions.Prediction.t() | nil,
           schedule: Screens.Schedules.Schedule.t() | nil
@@ -110,10 +112,12 @@ defmodule Screens.V2.Departure do
 
   def cancelled?(%__MODULE__{}), do: false
 
-  def crowding_level(%__MODULE__{prediction: p}) when not is_nil(p) do
+  @spec crowding_level(t, DateTime.t()) :: non_neg_integer() | nil
+  def crowding_level(%__MODULE__{prediction: p}, now) when not is_nil(p) do
     case p do
       %Prediction{trip: %Trip{} = trip, vehicle: %Vehicle{occupancy_status: status} = vehicle} ->
-        if crowding_data_relevant?(trip, vehicle) do
+        if crowding_data_relevant?(trip, vehicle) and
+             not departure_too_far_in_future?(p.departure_time, now) do
           crowding_level_from_occupancy_status(status)
         else
           nil
@@ -124,7 +128,7 @@ defmodule Screens.V2.Departure do
     end
   end
 
-  def crowding_level(%__MODULE__{prediction: nil, schedule: _}), do: nil
+  def crowding_level(%__MODULE__{prediction: nil, schedule: _}, _now), do: nil
 
   def direction_id(%__MODULE__{prediction: %Prediction{trip: %Trip{direction_id: direction_id}}}) do
     direction_id
@@ -242,10 +246,14 @@ defmodule Screens.V2.Departure do
 
   def vehicle_status(_), do: nil
 
-  defp crowding_data_relevant?(%Trip{id: trip_trip_id, stops: [first_stop | _]}, %Vehicle{
-         trip_id: vehicle_trip_id,
-         stop_id: next_stop
-       })
+  @spec crowding_data_relevant?(Trip.t(), Vehicle.t()) :: boolean()
+  defp crowding_data_relevant?(
+         %Trip{id: trip_trip_id, stops: [first_stop | _]},
+         %Vehicle{
+           trip_id: vehicle_trip_id,
+           stop_id: next_stop
+         }
+       )
        when not is_nil(trip_trip_id) and not is_nil(vehicle_trip_id) do
     vehicle_on_prediction_trip? = trip_trip_id == vehicle_trip_id
     vehicle_started_trip? = not (next_stop == first_stop)
@@ -253,6 +261,14 @@ defmodule Screens.V2.Departure do
   end
 
   defp crowding_data_relevant?(_trip, _vehicle), do: false
+
+  @spec departure_too_far_in_future?(DateTime.t() | nil, DateTime.t()) :: boolean()
+  defp departure_too_far_in_future?(departure_time, now)
+       when not is_nil(departure_time) do
+    DateTime.diff(Util.to_eastern(departure_time), Util.to_eastern(now)) > @hour_in_seconds
+  end
+
+  defp departure_too_far_in_future?(_departure_time, _now), do: false
 
   defp crowding_level_from_occupancy_status(:many_seats_available), do: 1
   defp crowding_level_from_occupancy_status(:few_seats_available), do: 2
--- a/lib/screens/v2/departure.ex
+++ b/lib/screens/v2/departure.ex
@@ -12,8 +12,6 @@ defmodule Screens.V2.Departure do
   alias Screens.V3Api
   alias Screens.Vehicles.Vehicle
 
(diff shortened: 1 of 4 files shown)


lib/screens/v2/widget_instance/departures.ex
```diff
@@ -528,21 +528,24 @@ defmodule Screens.V2.WidgetInstance.Departures do
           serialized_time_with_crowding()
   defp serialize_time_with_crowding(departure, screen, now) do
     serialize_time(departure, screen, now)
-    |> Map.merge(%{id: Departure.id(departure), crowding: serialize_crowding(departure, screen)})
+    |> Map.merge(%{
+      id: Departure.id(departure),
+      crowding: serialize_crowding(departure, screen, now)
```
rwaskiewicz: It _felt_ like determining if we're >60 minutes from the departure was a concern for `departure#serialize_crowding/3` once I started looking at that code - but it could be hoisted into this file pretty easily. LMK if anyone has a preference here

lib/screens/v2/widget_instance/departures.ex
```diff
@@ -528,21 +528,24 @@ defmodule Screens.V2.WidgetInstance.Departures do
...
-  defp serialize_crowding(_departure, %Screen{app_id: :dup_v2}), do: nil
+  defp serialize_crowding(_departure, %Screen{app_id: :dup_v2}, _now), do: nil
   # Otherwise, include crowding information for any valid departure.
   # Will return nil for schedules or predictions without crowding data.
-  defp serialize_crowding(departure, _screen), do: Departure.crowding_level(departure)
+  defp serialize_crowding(departure, _screen, now), do: Departure.crowding_level(departure, now)
 
   @spec serialize_time(Departure.t(), Screen.t(), DateTime.t()) ::
-          %{time: serialized_time(), time_in_epoch: integer()}
+          %{time: serialized_time(), time_in_epoch: integer(), is_live: boolean()}
```
rwaskiewicz, line 545: I think this field in the typespec was just missing

test/screens/v2/widget_instance/departures_test.exs
```diff
@@ -1208,6 +1208,82 @@ defmodule Screens.V2.WidgetInstance.DeparturesTest do
       assert [%{is_live: true}] =
                Departures.serialize_times_with_crowding([departure], screen, now)
     end
+
+    test "hides crowding information for departures greater than an hour in the future", %{
```
rwaskiewicz: I wrote these tests as a starting point (my original implementation was solely in `departures.ex`) - where I wanted to get tests that failed, then make them green, etc. 

However, these tests feel like they'll get a little hard to maintain long term - there are a _lot_ of fields here to keep track of on the various structs, which lead to specific code paths. I'm not opposed to:
1. Wrapping these in a `describe` block and `setup` callback to minimize duplication and draw attention to the values that change between tests
2. Deleting them altogether
3. Anything else (thoughts welcome)

lib/screens/v2/departure.ex
```diff
@@ -254,6 +262,14 @@ defmodule Screens.V2.Departure do
 
   defp crowding_data_relevant?(_trip, _vehicle), do: false
 
+  @spec departure_too_far_in_future?(DateTime.t() | nil, DateTime.t()) :: boolean()
+  defp departure_too_far_in_future?(departure_time, now)
+       when not is_nil(departure_time) do
+    DateTime.diff(Util.to_eastern(departure_time), Util.to_eastern(now)) > @hour_in_seconds
```
rwaskiewicz: Callout: The ticket calls out "over an hour", which means we'd show crowding info for exactly an hour. I'm not sure if that takes more space or not IRL (and is potentially screen dependent?).

digitalcora:
Does #2948 actually obviate this? That work introduced a new style/positioning for the crowding icons, which I think was intended to allow for displaying crowding even alongside the "widest" possible countdown types (at the expense of headsign space). cc @deanshi

deanshi:
> Does #2948 actually obviate this? That work introduced a new style/positioning for the crowding icons, which I think was intended to allow for displaying crowding even alongside the "widest" possible countdown types (at the expense of headsign space). cc @deanshi

Yep, so sorry I missed this Asana task when I was doing the work for the Live Icon. Like Cora said, my PR _should_ handle the case with timestamps, crowding information and the live icon but at the expense of the Headsign width. 

Screenshot 2026-04-13 at 8 13 42 AM

digitalcora:
Well, I think I missed it too, given I saw this task get pulled in and didn't remember at the time that we'd already made these other changes. Apologies for the runaround @rwaskiewicz, but it seems like the problem described in this task isn't actually a problem anymore. The typespec update we should still merge, though!

rwaskiewicz:
All good! I'm probably going to spend some time with this area of the codebase this morning to strengthen my knowledge of this part of the codebase while it's still on top of mind. I'll probably take this PR and basically revert the first commit for GitHub's history, then squash it all down once approved to just get the typespec changes in

lib/screens/v2/widget_instance/departures.ex
digitalcora: Oops, actually, I think the existing typespec is correct here. The E-ink-specific code path that produces a `time_in_epoch` field also never produces an `is_live` field.

rwaskiewicz: @digitalcora Can you point me to which code you're referring to? I thought https://github.com/mbta/screens/blob/95974c2cf3c3555caf806f6a1997857e0584fcf5/lib/screens/v2/widget_instance/departures.ex#L551-L567 was the eink-specific code, but I'm likely missing something/not seeing a codepath here.

rwaskiewicz: Or is it that `is_live` is always `false` and we don't document for that reason?

digitalcora: Double oops, that code does in fact produce an `is_live` field which is always `false`. I was going solely off my memory of how we'd implemented this and it turns out memories are faulty!

digitalcora (approved):

test/screens/v2/widget_instance/departures_test.exs
digitalcora: > However, these tests feel like they'll get a little hard to maintain long term - there are a _lot_ of fields here to keep track of on the various structs, which lead to specific code paths.

The best solution to this in my mind is factories — functions that build a valid data structure when given a subset of its fields, initializing the other fields with "realistic" values. We have some of these scattered in various test modules already, as either private `build_` functions or just module attributes defining a base struct you can modify with the `|` update syntax, though have yet to adopt a full-featured factory library, [ExMachina](https://github.com/beam-community/ex_machina) being the main one in Elixir.

The advantage of factories is that all data relevant to a given test case — and _only_ that data — is defined explicitly within it. You avoid the "mystery guest" problem where test data is defined in a separate part of the file or a separate file entirely, and has to be cross-referenced to understand what's going on; but also avoid the verbosity of defining complete valid data for every test case, including fields that "don't matter".

PR #3004, opened 2026-04-10
chore: switch from HTTPoison to Req

digitalcora:
[Req](https://hexdocs.pm/req/readme.html), already adopted by some other MBTA projects, seems to be emerging as the preferred HTTP client for the Elixir ecosystem.

Req is a "batteries-included" HTTP client, built on [Finch](https://hexdocs.pm/finch/Finch.html), which is built on [Mint](https://github.com/elixir-mint/mint), all of which are "native" Elixir projects (as compared to other popular clients which either are, or are built on, Erlang libraries).

During the recent performance investigation, when I noticed large discrepancies between memory usage reported by the Erlang VM vs. the OS, I suspected something to do with Hackney since HTTP connections are one of our biggest uses of memory. Looking into it, I found that we'd been stuck on an old version of Hackney for a while. This was because `hackney_telemetry`, the third-party library we used to get connection pool metrics, was not updated to be compatible with a new major version of Hackney, because it (Hackney) now has built-in telemetry. Dependabot didn't alert us that we had a major update available that it could not resolve automatically.

Faced with the requirement to do some work anyway to change how we report pool metrics, and given Finch's method of doing so was much better documented than Hackney's, I decided this was a good excuse to make the switch — although in the end, I did not find evidence that Hackney was to blame for the memory accounting discrepancy.

* Use the Req adapter for ExAws.
* Use the Req adapter for Sentry (this is the default).
* Replace direct uses of HTTPoison with Req in the V3 API client and Device Monitor.
* Remove separate Jason decoding steps; Req decodes JSON automatically using Jason by default.
* Remove `retry` package we were only using in the V3 API client; Req retries with exponential backoff on relevant status codes by default.
* Replace `hackney_telemetry` with `Finch.get_pool_status/2` by way of `telemetry_poller`.


lib/screens/device_monitor/fetch.ex
```diff
@@ -3,26 +3,25 @@ defmodule Screens.DeviceMonitor.Fetch do
...
 
-      {:response_success, %{status_code: status_code}} ->
-        log_fetch_error(vendor_name, :bad_response_code, url, status_code: status_code)
+      {:response, {:error, e}} ->
+        log_fetch_error(vendor_name, :fetch_error, url, exception: inspect(e))
 
-      {:parse, _} ->
-        log_fetch_error(vendor_name, :parse_error, url)
+      {:decode, _} ->
+        log_fetch_error(vendor_name, :decode_error, url)
```
panentheos, line 24: Just calling out that this will change the log signature, so we should make sure any downstream reports and dashboards continue to work.

panentheos (approved):
Nice improvements!

lib/screens/device_monitor/fetch.ex
digitalcora: I don't believe we have any reports or alerts that relied on the specific "error type" logged here, but I'll keep an eye out!

PR #3005, opened 2026-04-13
feat: FE for CR Shuttle Pill on LCDs and DUPs

robbie-sundstrom:
**Asana task**: [Departures: CR Shuttle Pill Icon](https://app.asana.com/1/15492006741476/project/1185117109217413/task/1209370311249958?focus=true)

### Summary
Adds logic to our Frontend to add the split CR Shuttle pill to our LCD and DUP screens. Though we have done similar types of pills as SVGs for our e-ink screens, I made this a reusable React component that uses the `clip-path` from the associated Figma files to create the proper route pill shape.
The CSS for this is handled differently on DUPs and our other LCD screens. This frontend change will need to be deployed to DUPs before we make the relevant BE change.

### Breakdown of Changes
- Added `@mixin` for route_pill background and text colors, since these are re-used multiple places
- Adds a new type of `dual` route pill (open to naming changes here). This is currently set up to just take text for the primary pill and an icon for the secondary circle pill, but this would be easily modifiable in the future.
- LCD specific Changes
  - Modifies the CSS within `departures/row.scss` to use a grid layout. Shortens the departures row when there is a dual route pill
- DUP Specific Changes
  -  Since grid layouts aren't supported until Chrome 57 and DUPs are stuck on 51, we could not use the same approach with these. We also already have a way to shorten headsigns for an entire section when there is an extended time column. However, in this case, we only want to shorten headsigns for a specific row rather than the entire section. 

### Examples
Note for the examples below, I had the BE `Enum.random`ly return some route pills as dual pills with the color of their actual route, which is why their colors differ. 

#### Sectional
WASHINGTON ST @ ESSEX ST

#### DUP
image

### DIFF
--- a/assets/src/components/departures/route_pill.tsx
+++ b/assets/src/components/departures/route_pill.tsx
@@ -101,14 +101,15 @@ const DualRouteIconPill: ComponentType<DualPill> = ({
   secondary_color: secondaryColor,
 }) => {
   const iconSrc = imagePath(pathForIcon[icon]);
-  const textModifiers: string[] = [color];
+
   const routeNum = Number(text);
+  const textModifiers: string[] = [color];
   if (isNaN(routeNum)) {
     textModifiers.push(text.length > 3 ? "small" : "large");
   } else {
     textModifiers.push("large");
   }
-  classWithModifiers("route-pill__dual-primary-pill", textModifiers);
+
   return (
     <div className="route-pill__dual-container">
       <div
--- a/assets/src/components/departures/route_pill.tsx
+++ b/assets/src/components/departures/route_pill.tsx
@@ -108,7 +108,7 @@ const DualRouteIconPill: ComponentType<DualPill> = ({
   } else {
     textModifiers.push("large");
   }
-  classWithModifiers("route-pill__dual-text-pill", textModifiers);
+  classWithModifiers("route-pill__dual-primary-pill", textModifiers);
   return (
     <div className="route-pill__dual-container">
       <div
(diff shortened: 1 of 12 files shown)


assets/src/components/departures/route_pill.tsx
```diff
@@ -87,6 +94,39 @@ const IconRoutePill: ComponentType<IconPill> = ({ icon }) => {
...
+}) => {
+  const iconSrc = imagePath(pathForIcon[icon]);
+  const textModifiers: string[] = [color];
+  const routeNum = Number(text);
+  if (isNaN(routeNum)) {
+    textModifiers.push(text.length > 3 ? "small" : "large");
+  } else {
+    textModifiers.push("large");
+  }
+  classWithModifiers("route-pill__dual-primary-pill", textModifiers);
```
rwaskiewicz: question(non-blocking): Is the result of this statement used anywhere? It doesn't look like this utility function modifies either of its inputs...but perhaps I'm missing something?

assets/src/components/departures/route_pill.tsx
```diff
@@ -87,6 +94,39 @@ const IconRoutePill: ComponentType<IconPill> = ({ icon }) => {
...
+        className={classWithModifiers("route-pill__dual-text", textModifiers)}
+      >
+        {text}
+      </div>
+      <div
+        className={classWithModifiers("route-pill__dual-secondary", [
+          secondaryColor,
+        ])}
+      >
+        <img className="route-pill__dual-secondary-image" src={iconSrc} />
```
rwaskiewicz, line 125: question(non-blocking): Does alt text ever make sense across any of our screens here? My _guess_ is no, but wanted to double check

rwaskiewicz (approved):

assets/src/components/departures/route_pill.tsx
robbie-sundstrom: I don't think there's any case where we'd want/need alt-text. We serialize audio text separately for each of our widgets, and we don't need to consider things like screen readers with any of our screens.

assets/src/components/departures/route_pill.tsx
robbie-sundstrom: thanks for catching! This was just a leftover from an earlier arrangement of this component that I didn't notice during cleanup.

PR #3009, opened 2026-04-16
feat: RDS Last Trip tracking

deanshi:
**Asana task**: [RDS Core: Last Trip tracking](https://app.asana.com/1/15492006741476/project/1185117109217413/task/1212463750135887?focus=true)

Description
- Added Nebulex to cache last trip departure times when building Departures
- Last Trips have a `TTL` till the end of the service day
- RDS consumes the last trip departure time and if the current time is 3 seconds greater than the departure time, we consider the branch to be service ended regardless of the last scheduled departure time
- Red Trunk stops to Alewife require two last trips in order to declare service ended

- [x] Tests added?

### DIFF
--- a/lib/screens/last_trip/last_trip.ex
+++ b/lib/screens/last_trip/last_trip.ex
@@ -1,0 +1,50 @@
+defmodule Screens.LastTrip.LastTrip do
+  @moduledoc """
+  Last Trip of the Day cache interface
+  """
+  alias Screens.LastTrip.Cache
+  alias Screens.Predictions.Prediction
+  alias Screens.V2.Departure
+  alias Screens.V2.RDS
+
+  @callback update_last_trip_cache([Departure.t()], DateTime.t()) :: :ok
+  def update_last_trip_cache(departures, now) do
+    departures
+    |> Enum.filter(&Departure.last_trip?(&1))
+    |> Enum.group_by(
+      &{Departure.stop(&1).id, Departure.route(&1).line.id,
+       Departure.representative_headsign(&1)},
+      fn %Departure{prediction: %Prediction{departure_time: departure_time}} -> departure_time end
+    )
+    |> Enum.each(fn {destination_key, new_departure_times} ->
+      if Cache.get(destination_key, nil) != new_departure_times do
+        Cache.put(destination_key, new_departure_times, ttl: time_to_service_end(now))
+      end
+    end)
+
+    :ok
+  end
+
+  @callback last_trip_departure_times(RDS.destination_key()) :: DateTime.t() | nil
+  def last_trip_departure_times(destination_key) do
+    case Cache.get(destination_key) do
+      {:ok, nil} -> nil
+      {:ok, departure_times} -> departure_times
+    end
+  end
+
+  defp time_to_service_end(%DateTime{time_zone: time_zone} = now) do
+    service_end_time =
+      now
+      |> DateTime.to_date()
+      |> DateTime.new!(~T[04:00:00], time_zone)
+
+    if DateTime.compare(now, service_end_time) == :gt do
+      service_end_time
+      |> DateTime.add(1, :day)
+      |> DateTime.diff(now, :millisecond)
+    else
+      DateTime.diff(service_end_time, now, :millisecond)
+    end
+  end
+end
--- a/lib/screens/last_trip/last_trip.ex
+++ b/lib/screens/last_trip/last_trip.ex
@@ -1,0 +1,70 @@
+defmodule Screens.LastTrip.LastTrip do
+  @moduledoc """
+  Last Trip of the Day cache interface
+  """
+  alias Screens.LastTrip.Cache
+  alias Screens.Predictions.Prediction
+  alias Screens.V2.Departure
+  alias Screens.V2.RDS
+
+  @callback update_last_trip_cache([Departure.t()], DateTime.t()) :: :ok
+  def update_last_trip_cache(departures, now) do
+    departures
+    |> Enum.filter(&Departure.last_trip?(&1))
+    |> Enum.group_by(
+      &{Departure.stop(&1).id, Departure.route(&1).line.id,
+       Departure.representative_headsign(&1)},
+      fn %Departure{prediction: %Prediction{departure_time: departure_time}} = departure ->
+        {Departure.trip_id(departure), departure_time}
+      end
+    )
+    |> Enum.each(fn {destination_key, last_trip_departure_times} ->
+      last_trip_departure_map = Map.new(last_trip_departure_times)
+
+      Cache.transaction(
+        fn ->
+          Cache.get_and_update(
+            destination_key,
+            fn current_value ->
+              updated_value =
+                if is_nil(current_value) do
+                  last_trip_departure_map
+                else
+                  Map.merge(current_value, last_trip_departure_map)
+                end
+
+              {current_value, updated_value}
+            end,
+            ttl: time_to_service_end(now)
+          )
+        end,
+        keys: [destination_key]
+      )
+    end)
+
+    :ok
+  end
+
+  @callback last_trip_departure_times(RDS.destination_key()) :: [DateTime.t()]
+  def last_trip_departure_times(destination_key) do
+    case Cache.get(destination_key) do
+      {:ok, nil} -> []
+      {:ok, departure_times_by_trip_id} -> Map.values(departure_times_by_trip_id)
+    end
+  end
+
+  defp time_to_service_end(%DateTime{time_zone: time_zone} = now) do
+    service_end_time =
+      now
+      |> DateTime.to_date()
+      |> DateTime.new!(~T[04:00:00], time_zone)
+
+    if DateTime.compare(now, service_end_time) == :gt do
+      service_end_time
+      |> DateTime.add(1, :day)
+    else
+      service_end_time
+    end
+    |> DateTime.diff(now, :millisecond)
+  end
+end
--- a/lib/screens/v2/rds.ex
+++ b/lib/screens/v2/rds.ex
@@ -15,6 +15,7 @@ defmodule Screens.V2.RDS do
   alias Screens.Alerts.InformedEntity
   alias Screens.Config.Cache
   alias Screens.Headways, as: Headway
+  alias Screens.LastTrip.LastTrip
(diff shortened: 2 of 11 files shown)


lib/screens/last_trip/last_trip.ex
```diff
@@ -0,0 +1,50 @@
...
+      if Cache.get(destination_key, nil) != new_departure_times do
+        Cache.put(destination_key, new_departure_times, ttl: time_to_service_end(now))
+      end
+    end)
+
+    :ok
+  end
+
+  @callback last_trip_departure_times(RDS.destination_key()) :: DateTime.t() | nil
+  def last_trip_departure_times(destination_key) do
```
deanshi, line 49: I know the AC says that we shouldn't tie Last Trip API to RDS concepts. Is there an alternative that we can use for describing a last trip for a destination? My mental model has it so that destinations (or at least some modified definition of "destinations") _are_ the best way to lookup last trips regardless of RDS.

lib/screens/last_trip/last_trip.ex
```diff
@@ -0,0 +1,50 @@
...
+  defp time_to_service_end(%DateTime{time_zone: time_zone} = now) do
+    service_end_time =
+      now
+      |> DateTime.to_date()
+      |> DateTime.new!(~T[04:00:00], time_zone)
+
+    if DateTime.compare(now, service_end_time) == :gt do
+      service_end_time
+      |> DateTime.add(1, :day)
+      |> DateTime.diff(now, :millisecond)
```
panentheos: nit: We could factor this out after the `if`, since we want to do the same `diff` for both cases.

lib/screens/v2/rds.ex
```diff
@@ -67,6 +68,11 @@ defmodule Screens.V2.RDS do
     :suspension
   ]
 
+  @red_trunk [70_061..70_061, 70_063..70_084]
```
panentheos, line 71: Is `70_061..70_061` a typo?

lib/screens/v2/rds.ex
```diff
@@ -392,6 +401,31 @@ defmodule Screens.V2.RDS do
     %Countdowns{departures: departures_for_headsign}
   end
 
+  @spec after_last_trip?(destination_key(), DateTime.t()) :: boolean()
+  # Red Trunk stops need two last trip departures in order to classify as Service Ended
+  defp after_last_trip?({stop_id, _line_id, "Alewife"} = destination_key, now)
```
panentheos: Just checking my understanding: Is this logic limited to Alewife headsigns because the Ashmont and Braintree statuses are calculated and reported separately on the trunk?

lib/screens/last_trip/last_trip.ex
```diff
@@ -0,0 +1,50 @@
...
+    departures
+    |> Enum.filter(&Departure.last_trip?(&1))
+    |> Enum.group_by(
+      &{Departure.stop(&1).id, Departure.route(&1).line.id,
+       Departure.representative_headsign(&1)},
+      fn %Departure{prediction: %Prediction{departure_time: departure_time}} -> departure_time end
+    )
+    |> Enum.each(fn {destination_key, new_departure_times} ->
+      if Cache.get(destination_key, nil) != new_departure_times do
+        Cache.put(destination_key, new_departure_times, ttl: time_to_service_end(now))
```
panentheos: Unless I'm misunderstanding something, it doesn't look like this will behave the way we want. The `Departures.Builder` returns a list of _current_ departures, but the Ashmont/Braintree last trips may very well come at different times, so the later one would overwrite the earlier one in the cache, and the RL trunk logic in `after_last_trip?` would never trigger. We probably need to aggregate these over time instead.

lib/screens/v2/rds.ex
```diff
@@ -392,6 +401,31 @@ defmodule Screens.V2.RDS do
...
+        departure_times
+        |> Enum.max()
+        |> after_last_trip_with_buffer?(now)
+
+      _ ->
+        false
+    end
+  end
+
+  defp after_last_trip?(destination_key, now) do
```
panentheos: nit: These two function clauses seem more similar than different, we could consider just factoring out the bit of RL trunk logic into a case or private helper.

lib/screens/v2/rds.ex
deanshi: This was taken from `headways.ex` where it was defined as `red_trunk: [70061..70061, 70063..70084]`. Could that have been a typo to begin with? I can check.

lib/screens/v2/rds.ex
deanshi: Yep - RDS will return two different destinations based on the headsign (Ashmont/Braintree) while Alewife is just a singular destination and so this will be the singular destination that we need to check for two last trips on at each station.

lib/screens/v2/rds.ex
digitalcora: It is likely not a typo there; the shorthand format requires that all items in the list are ranges, for simplicity.

lib/screens/last_trip/last_trip.ex
deanshi: Updated this so that we now combine last trips. Now if they come at different times, we'll appropriately get two last trips instead of throwing the old one away.

lib/screens/last_trip/last_trip.ex
```diff
@@ -0,0 +1,60 @@
...
+      last_trip_times =
+        case Cache.get(destination_key) do
+          {:ok, %MapSet{} = departure_times} -> departure_times
+          _ -> MapSet.new()
+        end
+
+      new_last_trip_times = MapSet.new(new_departure_times)
+
+      unless MapSet.subset?(new_last_trip_times, last_trip_times) do
+        combined_last_trip_times = MapSet.union(new_last_trip_times, last_trip_times)
```
panentheos: I don't think this will work the way we want. Because it's just tracking timestamps, if a prediction changes while it's active, we'll end up accumulating multiple timestamps, so the downstream logic won't work correctly. We'll probably need to track thing by trip_id in order to deduplicate properly. RTS uses that approach, although its model is a little different, but you might find inspiration there.

lib/screens/last_trip/last_trip.ex
```diff
@@ -0,0 +1,60 @@
...
+
+        Cache.put(destination_key, combined_last_trip_times, ttl: time_to_service_end(now))
```
panentheos: Not sure if this is a potential issue, but we should think about race conditions between the read/write of the cache, if multiple signs might be computing the same RDS simultaneously. Nebulex seems to have transaction support if we need it, though I haven't used it before.

lib/screens/last_trip/last_trip.ex
deanshi: Okay - so I've rewritten the logic so that we take into consideration the `trip_id` as well when we're saving cached last trips. 

Just verifying to make sure my understanding is correct though: 
When predictions change, `trip_id` remains the same even if the prediction itself is different (presumably because it's the same trip). Does the prediction id change as well since it's technically a new prediction? And so we use `trip_id` because it's the thing that remains constant in the data that comes back?

lib/screens/last_trip/last_trip.ex
deanshi: 👍 I'll wrap this code around a `transaction` according to the Nebulex docs, but also let me double check about handling race conditions and the most recent transactions.

lib/screens/last_trip/last_trip.ex
```diff
@@ -0,0 +1,72 @@
...
+              {:ok, nil} -> %{}
+              {:ok, current_last_trips} -> current_last_trips
+            end
+            |> Map.merge(departure_times_by_trip_id)
+
+          Cache.put(destination_key, updated_last_trips, ttl: time_to_service_end(now))
+        end,
+        keys: [destination_key]
+      )
+    end)
```
panentheos, line 43: I _think_ this logic is correct, though it took me a couple minutes of reading it to understand it. Maybe we could tighten things up a bit? I can think of a couple approaches:
1. If we're willing to wrap the whole thing in one `transaction`, we could just do a single `filter`+`each` over the departures, and `Cache.update`+`Map.put` each value directly.
2. If it's important to group updates or use multiple transactions, then it might be clearer to generate `{key,trip,time}` tuples and then `group_by` key, rather than constructing an intermediate Map. (Although I doubt there's a need to do this instead of option 1).

In any case, look for opportunities to use higher level utilities like `update_in`, `put_in`, `Cache.update`, etc where possible, since those can shrink the code and make the intention clearer at a glance.

lib/screens/last_trip/last_trip.ex
panentheos: I'm not sure about the stability of prediction IDs since I haven't used them before, but yes, the trip_id should be a stable identifier for this case.

lib/screens/last_trip/last_trip.ex
deanshi: Let me take a look at optimization and I can bring back a code snippet to see if it's better.

deanshi:
```
departures
    |> Enum.filter(&Departure.last_trip?(&1))
    |> Enum.group_by(
      &{Departure.stop(&1).id, Departure.route(&1).line.id,
       Departure.representative_headsign(&1)},
      fn %Departure{prediction: %Prediction{departure_time: departure_time}} = departure ->
        {Departure.trip_id(departure), departure_time}
      end
    )
    |> Enum.each(fn {destination_key, last_trip_departure_times} ->
      last_trip_departure_map = Map.new(last_trip_departure_times)

      Cache.transaction(
        fn ->
          Cache.get_and_update(
            destination_key,
            fn current_value ->
              updated_value =
                if is_nil(current_value) do
                  last_trip_departure_map
                else
                  Map.merge(current_value, last_trip_departure_map)
                end

              {current_value, updated_value}
            end,
            ttl: time_to_service_end(now)
          )
        end,
        keys: [destination_key]
      )
    end)
 ```
 
I ended up going with the separate `transactions` instead of one big one wrapping everything just so we're not unnecessarily locking the whole cache. I also ended up using `get_and_update` with the cache to make things slightly more clear on what we're doing with the Maps.

panentheos (approved):
Looks good, thanks for iterating on this!

PR #3010, opened 2026-04-16
fix: Fixed Bus Route Pill spacing on DUPS

deanshi:
Description
Removed some spacing/width requirements for specifically yellow (bus) route pills within the row CSS but added some margin on the right for the pill itself.

Screenshot 2026-04-16 at 10 08 25 AM
Screenshot 2026-04-16 at 10 19 46 AM

### DIFF
--- a/assets/css/dup/departures/row.scss
+++ b/assets/css/dup/departures/row.scss
@@ -17,13 +17,16 @@
 
 .departure-row__route {
   display: inline-block;
-  width: 248px;
-  height: 144px;
-  margin: 32px;
   vertical-align: middle;
 
   .departure-row--extended-route_pill & {
     width: 351px;
+  }
+
+  &:not(.departure-row__route--yellow) {
+    width: 248px;
+    height: 144px;
+    margin: 32px;
   }
 }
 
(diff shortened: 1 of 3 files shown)


assets/css/dup/departures/row.scss
```diff
@@ -17,14 +17,17 @@
...
-  width: 248px;
-  height: 144px;
-  margin: 32px;
   vertical-align: middle;
 
   .departure-row--extended-route_pill & {
     width: 351px;
   }
+
+  &:not(:has(.route-pill--yellow)) {
```
digitalcora: `:has` is only supported on Chrome 105 and up; we need something that will work on Chrome 51 for the DUPs.

digitalcora (changes requested):

deanshi:
Converting it back to draft because I'm noticing some spacing differences between the pills on the top screenshot versus the bottom. Going to look into it real quick.

deanshi:
Realized that the spacing I'm seeing has been intentionally added as margin. I'm going to get feedback from design, but reopening this PR for now.

robbie-sundstrom (approved):
This looks right to me with only the 32px margin on the right side for the bus route pills!

digitalcora (approved):

PR #3011, opened 2026-04-16
fix: Kenmore special case handling for inside alerts

deanshi:
**Asana task**: [DUP behavior when Kenmore is in middle of diversion](https://app.asana.com/1/15492006741476/project/1185117109217413/task/1213877855438317?focus=true)

Description
- When we are now inside an alert at Kenmore, we will disregard the special case and allow the full screen presentation for handling the alert
- This assumes that we are not overlapping alerts, which I think we agreed we wouldn't do in Slack. If we unfortunately do overlap alerts, it will show one of the alerts associated.

- [x] Tests added?

### DIFF
--- a/lib/screens/v2/candidate_generator/dup/alerts.ex
+++ b/lib/screens/v2/candidate_generator/dup/alerts.ex
@@ -55,7 +55,7 @@ defmodule Screens.V2.CandidateGenerator.Dup.Alerts do
            {:ok, alerts} <- fetch_alerts_fn.(route_ids: route_ids) do
         alerts
         |> relevant_alerts(config, location_context, now)
-        |> alert_special_cases(config)
+        |> alert_special_cases(config, location_context)
         |> create_alert_widgets(config, location_context, stop_name)
       else
         :error -> []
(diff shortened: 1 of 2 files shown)


lib/screens/v2/candidate_generator/dup/alerts.ex
```diff
@@ -139,6 +139,11 @@ defmodule Screens.V2.CandidateGenerator.Dup.Alerts do
     ]
   end
 
+  @spec inside_alert?(Alert.t(), LocationContext.t()) :: boolean()
+  defp inside_alert?(alert, location_context) do
+    LocalizedAlert.location(%{alert: alert, location_context: location_context}) == :inside
+  end
+
```
robbie-sundstrom: Extreme nit, but should we move this function below `alert_special_cases` since that's where it's called? I believe that's the general standard we follow.

lib/screens/v2/candidate_generator/dup/alerts.ex
```diff
@@ -157,25 +162,39 @@ defmodule Screens.V2.CandidateGenerator.Dup.Alerts do
...
 
   # In the case where Kenmore has 2 or more boundary shuttles / suspensions to the west,
   # don't only select 1 alert; instead look at all alerts and make custom text
-  @spec kenmore_special_case(list(Alert.t())) :: {:special, list(Screens.V2.WidgetInstance.t())}
-  defp kenmore_special_case(alerts) do
+  #
+  # In cases where we are inside any of the alerts, assume that we will just use the
+  # full screen presentation of the alert and bypass this special case altogether
+  @spec kenmore_special_case(list(Alert.t()), boolean()) ::
+          {:special, list(Screens.V2.WidgetInstance.t())}
```
robbie-sundstrom: This spec should be updated to return `{:normal, list(Alert.t())} | {:special, list(Screens.V2.WidgetInstance.t())}`. You could even make this its own type since its used elsewhere.

lib/screens/v2/candidate_generator/dup/alerts.ex
```diff
@@ -157,25 +162,39 @@ defmodule Screens.V2.CandidateGenerator.Dup.Alerts do
...
          alerts,
-         %Screen{app_params: %Dup{alerts: %AlertsConfig{stop_id: stop_id}}}
+         %Screen{app_params: %Dup{alerts: %AlertsConfig{stop_id: stop_id}}},
+         location_context
        ) do
+    inside_alert = Enum.any?(alerts, &inside_alert?(&1, location_context))
+
     case stop_id do
-      "place-kencl" -> kenmore_special_case(alerts)
+      "place-kencl" -> kenmore_special_case(alerts, inside_alert)
```
robbie-sundstrom: The `inside_alert?` call could be moved inside the "place-kencl" case since that's the only time we actually want to use it. Something like 
```
    case stop_id do
      "place-kencl" ->
        inside_alert = Enum.any?(alerts, &inside_alert?(&1, location_context))
        kenmore_special_case(alerts, inside_alert)

      "place-wtcst" ->
        wtc_special_case(alerts)
      
       _ ->
        {:normal, alerts}
    end
```

robbie-sundstrom (approved):
Small comments, but overall this looks right to me!

lib/screens/v2/candidate_generator/dup/alerts.ex
deanshi: Nah, I was looking at the code this morning and thought the same thing. I'll move it down.

PR #3012, opened 2026-04-17
Add rotation index to DUP header in version number

robbie-sundstrom:
Just a little change to include the rotation index with the version displayed in the DUP header that I found helpful when debugging. Figure this could be helpful for potential production debugging as well, since this would allow us to easily see the rotation being shown in any image of a DUP screen.

Note that this doesn't show up when looking at the DUPs on Screens Admin or Screenplay, since those don't technically have a rotation, just a wider viewport to show all three rotations. 

image


rwaskiewicz (approved):

digitalcora:
Good idea!

PR #3013, opened 2026-04-17
fix: Live Icon spacing issue when displaying am/pm times

deanshi:
Description
Some late night testing uncovered some spacing issues dealing with specifically the AM/PM timestamps. This adds in some shotened headsign spacing to allow for the full timestamp to appear. 

Opted to make this one a class modifier instead of making the spacing default since it feels that showing am/pm happens less often than other cases (Let me know if this is a false assumption, I can just make it the default spacing in that case). 

Screenshot 2026-04-17 at 12 24 48 PM
Screenshot 2026-04-17 at 12 23 49 PM

### DIFF
--- a/assets/src/components/dup/departures/departure_row.tsx
+++ b/assets/src/components/dup/departures/departure_row.tsx
@@ -15,15 +15,21 @@ const DepartureRow: ComponentType<DepartureRowBase> = ({
   const parentClassModifier =
     route.type === "dual" ? "extended-route_pill" : "";
 
+  const classModifier = timesWithCrowding.some(
+    (time) => time.time?.type === "timestamp" && time?.time.am_pm,
+  )
+    ? "shortened-headsign"
+    : "";
+
   return (
     <div className={classWithModifier("departure-row", parentClassModifier)}>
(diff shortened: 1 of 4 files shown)


assets/src/components/dup/departures/departure_row.tsx
```diff
@@ -15,15 +15,21 @@ const DepartureRow: ComponentType<DepartureRowBase> = ({
   const parentClassModifier =
     route.type === "dual" ? "extended-route_pill" : "";
 
+  const classModifier = timesWithCrowding.find(
```
rwaskiewicz: Nit: `some` ([mdn](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/some)) might be slightly more semnatically correct instead of `find`, but both work equally well

rwaskiewicz (approved):
LGTM!

assets/src/components/dup/departures/departure_row.tsx
deanshi: I agree I think that reads better. Let me swap it.

PR #3014, opened 2026-04-22
feat: Emergency Takeover on Sectionals, PreFares, and Dups

deanshi:
**Asana task**: [Screens: implement emergency takeover feature](https://app.asana.com/1/15492006741476/project/1185117109217413/task/1213370222596877?focus=true)

Description
- Configuring `emergency_takeover` within the config will force an emergency visual and audio takeover for the screen
- The `EmergencyTakeover` widget serializes into an EvergreenContent that takes over the full screen for the respective screen type
- `visual_asset_path` can be configured to be an uploaded asset to display
- If `text_for_audio` is provided, we use the EvergreenContent audio view in order to get the ssml and use Polly to speak
- If `audio_asset_path` is used, a new redirect has been added in order to use the uploaded asset for audio equivalence

Screenshot 2026-04-23 at 4 06 25 PM
Screenshot 2026-04-23 at 4 07 53 PM
Screenshot 2026-04-23 at 4 09 10 PM

- [x] Tests added?


lib/screens/v2/widget_instance/emergency_takeover.ex
```diff
@@ -0,0 +1,46 @@
...
+            text_for_audio: nil,
+            asset_path_for_audio: nil
+
+  @type t :: %__MODULE__{
+          screen: Screen.t(),
+          slot_names: [atom()],
+          asset_path_for_visual: String.t(),
+          text_for_audio: String.t(),
+          asset_path_for_audio: String.t()
+        }
```
rwaskiewicz: Similar [to a question I had yesterday](https://github.com/mbta/screens-config-lib/pull/87/changes#r3123957818) (apologies if this is just rehashing the same-ish thing!) - looking at `screen`, `asset_path_for_visual`, `text_for_audio` and `asset_path_for_audio`:
1. I see the first two there are a part of `@enforce_keys`, they're typed with a concrete type (`Screen.t()`, `String.t()`, respectively). However, we default `screen` to `nil` and `asset_path_for_visual` as an empty string. Is this valid because those two keys are a part of the `@enforce_keys` above? I.E. Does having `screen` in `@enforce_keys` make it unnecessary to type it as `Screen.t() | nil`? 
2. `text_for_audio` and `asset_path_for_audio` _aren't_ in `@enforce_keys`, but we default to `nil` for those, but they're both typed as `String.t()`. From an elixir standpoint, why don't we need to make those the union of `String.t()` and `nil`?

test/screens/v2/widget_instance/emergency_takeover_test.exs
```diff
@@ -0,0 +1,66 @@
...
+        slot_names: [:test_slot_name],
+        asset_path_for_visual: "test_path_name",
+        text_for_audio: "Text for audio",
+        asset_path_for_audio: "test_audio_path_name"
+      }
+    }
+  end
+
+  describe "priority/1" do
+    test "returns 2", %{instance: instance} do
```
rwaskiewicz: Should this be 'returns [0]'? 
```suggestion
    test "returns [0]", %{instance: instance} do
```
Or maybe something like "the highest priority"?

```suggestion
    test "returns the highest priority possible", %{instance: instance} do
```

lib/screens/v2/candidate_generator/widgets/emergency_takeover.ex
```diff
@@ -0,0 +1,52 @@
...
+  def emergency_takeover_instances(
+        %Screen{app_params: %{emergency_takeover: emergency_takeover} = app} =
+          config,
+        _now
+      ) do
+    case app do
+      %PreFare{template: :duo} -> [:full_left_screen, :full_right_screen]
+      %PreFare{template: :solo} -> [:full_right_screen]
+      %Busway{} -> [:full_screen]
+      %Dup{} -> [:full_rotation_zero, :full_rotation_one, :full_rotation_two]
```
rwaskiewicz, line 28: Is it possible for `app` to not match one of these types? Does it make sense to add a catch-all that returns an empty list?

```suggestion
      %Dup{} -> [:full_rotation_zero, :full_rotation_one, :full_rotation_two]
      _ -> []
```

which I suppose would require an additional `emergency_takeover` definition:
```elixir
defp emergency_takeover(_emergency_takeover, _config, _slot_name), do: nil
```

But that widens the return type of `emergency_takeover_instances` 😅 Maybe I'm off here and `app` will always match one of these types. Let me know if I'm barking up the wrong tree here.

test/screens/v2/candidate_generator/widgets/emergency_takeover_test.exs
```diff
@@ -0,0 +1,111 @@
...
+  describe "emergency_takeover_instance/2" do
+    test "returns empty array when emergency_takeover is nil" do
+      screen = build_screen(Screen.PreFare, nil)
+
+      assert Generator.emergency_takeover_instances(screen, @now) == []
+    end
+
+    test "returns emergency takeover with PreFare Duo full screen slots" do
+      screen =
+        struct(build_screen(Screen.PreFare))
```
rwaskiewicz: Non-blocking question (for my own understanding): Do we need to wrap this call to `build_screen` in `struct`? If so, what does that do? My current understanding is that this would effectively become:
```elixir
struct(struct(Screen, app_params: struct(Screen.PreFareen, %{emergency_takeover: emergency_takeover})))
```

rwaskiewicz (approved):
All my questions are non-blocking/mostly for my own understanding

lib/screens/v2/widget_instance/emergency_takeover.ex
deanshi: To both of these, I think I just missed making them non-nil values for enforced keys, and unioned with nil for the unenforced keys. 

Long story short, I messed up and need to change this hahaha, thanks for the catch.

test/screens/v2/widget_instance/emergency_takeover_test.exs
deanshi: Ah good catch, that's what I get for copy pasting old code and missing this. Let me swap.

lib/screens/v2/candidate_generator/widgets/emergency_takeover.ex
deanshi: Based on the `screens_config`, we only allow `emergency_takeover` to be configured for PreFares, Busways (Sectionals) and Dups. I believe if you tried to configure an `emergency_takeover` for another screen type, it would just remove it. 

That being said, thinking about this here, I think that if for whatever reason we are able to get to here on a screen type that we're not supporting we should crash fast and get a Sentry error informing us of what's happening, especially if it's for Emergency Takeover where we would want something to show up rather than have it safely fail. I'm pretty swayable to this change though if you feel strongly hahaha.

test/screens/v2/candidate_generator/widgets/emergency_takeover_test.exs
deanshi: Ah, good catch. 

No this is just something I leftover from a previous implementation I was trying. Let me remove the extra `struct` and just keep `build_screen` like I do elsewhere.

lib/screens/v2/widget_instance/emergency_takeover.ex
deanshi: Ah sorry - but to your other question about `@enforce_keys` and nil, I think we see a few different versions of initializing the defstruct around the code base. I think if you have `@enforce_keys`, when you create the type you will be required to set the value to something non-nil.

I do feel like explicitly setting it to `nil` breaks the typing though, so I'm going to change this to: 
```
@enforce_keys ~w[screen asset_path_for_visual]a
defstruct @enforce_keys ++ [:text_for_audio, :asset_path_for_audio, slot_names: []]
```

like we do in a few other places.

lib/screens/v2/candidate_generator/widgets/emergency_takeover.ex
rwaskiewicz: > I'm pretty swayable to this change though if you feel strongly hahaha.

Nah, I probably should reframe my mindset from "does this value/type combination make sense" to "how should work in the event of an emergency" (particularly for this PR). To your point:

> I think that if for whatever reason we are able to get to here on a screen type that we're not supporting we should crash fast and get a Sentry error informing us of what's happening, especially if it's for Emergency Takeover where we would want something to show up rather than have it safely fail

If that's the fastest feedback loop from an operations/engineering standpoint, I'm onboard with it. Like you said, silently failing is likely _not_ in our best interest here.

lib/screens/v2/widget_instance/emergency_takeover.ex
```diff
@@ -0,0 +1,42 @@
+defmodule Screens.V2.WidgetInstance.EmergencyTakeover do
```
digitalcora: I think we shouldn't need this separate widget type on the backend — as suggested in the Asana task, it can be generated as an evergreen content widget. In other words rather than:

1. emergency message candidate generator
2. ...generates emergency message backend widget
3. ...serializes to evergreen content frontend component

We could have:

1. emergency message candidate generator
2. ...generates **evergreen content backend widget**
3. ...serializes to evergreen content frontend component

deanshi:
@rwaskiewicz re-requesting a review because of a few changes I made in the latest commits. No crazy changes, but it feels like it maybe should still get a review just in case. 🙇

lib/screens/v2/candidate_generator/widgets/emergency_takeover.ex
```diff
@@ -0,0 +1,56 @@
...
+         slot_name
+       ) do
+    %EvergreenContent{
+      screen: config,
+      slot_names: [slot_name],
+      asset_url: visual_asset_path,
+      priority: [0],
+      now: now,
+      text_for_audio: text_for_audio,
+      audio_priority: [0]
```
rwaskiewicz, line 53: For my own understanding - we define the `audio_priority` here, but don't attach the audio itself to the `EvergreenContent` - is that to make ensure the audio from other widgets is deprioritized over this widget's audio?

rwaskiewicz (approved):
Looks good!

lib/screens/v2/candidate_generator/widgets/emergency_takeover.ex
digitalcora: There are two possible audio code paths for emergency content:

1. The content uses text-to-speech audio. In this case we _are_ attaching the audio content to the evergreen widget (see `text_for_audio` above) and it will go through the normal audio synthesis path.
2. The content uses pre-recorded audio. In this case the chosen `audio_priority` for the widget is irrelevant and will never be used/accessed, because when audio is requested for the screen we should instead respond with a redirect straight to the pre-recorded asset, and will not go through the widget generation at all.

digitalcora: (to clarify _why_ this is the case: the widget system has no existing support for pre-recorded audio clips. we decided to use this "bypass" approach just for emergency content because we have no other foreseeable need for the widget system to support pre-recorded audio, and we know emergency content always takes over the entire screen, so pre-recorded emergency audio never needs to coexist with anything else.)

rwaskiewicz: Ah, I see - I had though the synthesis took place before this step, hence my confusion. Thanks!

PR #3016, opened 2026-04-22
fix: Filter to only subway platforms when determining closure type

robbie-sundstrom:
**Asana task**: [[title](regression: "N platforms closed" when entire station is closed)](https://app.asana.com/1/15492006741476/project/1185117109217413/task/1214148649236060?focus=true)

With the change to use platform_names from the `InformedEntity`'s within an `Alert`, we did not filter them to only include stops with `vehicle_types` for `:subway` and `:light_rail`. This adds a function within the Stop module that does this filtering so that its done consistently everywhere these alerts are handled

GREEN LINE

### DIFF
--- a/lib/screens/alerts/alert.ex
+++ b/lib/screens/alerts/alert.ex
@@ -420,16 +420,13 @@ defmodule Screens.Alerts.Alert do
     # Typically, these partial closures affecting child stops will only affect a single station.
     # However, we do want to consider the case in which multiple stations have closures,
     # but not every child stop at those parent stations are closed.
-
-    informed_parent_stations = informed_parent_stations(alert)
-
     platforms_affected_by_alert =
       informed_platforms_from_entities(informed_entities, platforms_at_informed_stations)
 
-    case informed_parent_stations do
+    case informed_parent_stations(alert) do
       [_single_parent_station] ->
         # Compare number of platforms in alert to total number of child platforms at station
-        if length(platforms_affected_by_alert) != length(platforms_at_informed_stations) do
+        if length(platforms_affected_by_alert) < length(platforms_at_informed_stations) do
           :partial_closure
         else
           :full_station_closure
--- a/lib/screens/v2/widget_instance/subway_status/serialize.ex
+++ b/lib/screens/v2/widget_instance/subway_status/serialize.ex
@@ -5,6 +5,7 @@ defmodule Screens.V2.WidgetInstance.SubwayStatus.Serialize do
 
   alias Screens.Alerts.Alert
   alias Screens.Routes.Route
+  alias Screens.Stops.Stop
   alias Screens.V2.WidgetInstance.SubwayStatus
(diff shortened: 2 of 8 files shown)


lib/screens/alerts/alert.ex
```diff
@@ -420,16 +420,13 @@ defmodule Screens.Alerts.Alert do
...
-
     platforms_affected_by_alert =
       informed_platforms_from_entities(informed_entities, platforms_at_informed_stations)
 
-    case informed_parent_stations do
+    case informed_parent_stations(alert) do
       [_single_parent_station] ->
         # Compare number of platforms in alert to total number of child platforms at station
-        if length(platforms_affected_by_alert) != length(platforms_at_informed_stations) do
+        if length(platforms_affected_by_alert) < length(platforms_at_informed_stations) do
```
deanshi, line 429: Just for my own knowledge, we technically can't have more platforms affected by an alert than actually exist at a station right? What other platforms exist for stations in these cases?

lib/screens/v2/widget_instance/subway_status/serialize.ex
```diff
@@ -256,7 +257,7 @@ defmodule Screens.V2.WidgetInstance.SubwayStatus.Serialize do
           child_stops -> child_stops
         end
       )
-      |> Enum.filter(&(&1.location_type == 0))
+      |> Stop.filter_subway_platforms()
```
deanshi, line 260: So looking at the typespec for `Stop`, it looks like `vehicle_type` can be `nil`. Are we good with that change here expecting `vehicle_type` to always be populated here? I would expect it to exist (especially for subway), but I'm not 100% certain.

deanshi (approved):
Makes sense to me! A few questions for my own knowledge, but otherwise looks good.

lib/screens/v2/widget_instance/subway_status/serialize.ex
digitalcora: I believe the reason `vehicle_type` is nullable is because it only applies to stopping locations (location type 0). Other types of GTFS "stops", like parent stations or pathway nodes, don't have this field, but stopping locations always should; it would not make sense to have a "stopping location" that no type of vehicle actually stops at.

lib/screens/alerts/alert.ex
robbie-sundstrom: Yes that's true! In this case, `!=` was essentially equivalent to `<` since it could never be greater. I just changed this to less than to make it more explicit.

lib/screens/v2/widget_instance/subway_status/serialize.ex
robbie-sundstrom: Missed this notification, but good question and something I also looked at! What Cora said is exactly it -- we can always expect there to be a `vehicle_type` on a `Stop` with `location_type` 0

PR #3021, opened 2026-04-24
feat: CR and Subway Shuttle Pill Icons (BE piece)

robbie-sundstrom:
**Asana task**: [Departures: CR Shuttle Pill Icon](https://app.asana.com/1/15492006741476/project/1208877354406742/task/1209370311249958?focus=true)

Note that the FE piece has already been deployed to DUP hardware, so this is good to deploy .

- We identify a shuttle route based on its ID containing the phrase shuttle - I checked and is the case for every shuttle pattern at this point in time
- Adds a list of `shuttle_pill_enabled_screens` that determine if a headsign and route pill is modified when it is a `shuttle_route?`
- Replaces the shuttle pill and shortens the headsign for all subway (BL, OL, RL) and CR lines

### DIFF
--- a/lib/screens/lines/line.ex
+++ b/lib/screens/lines/line.ex
@@ -11,4 +11,12 @@ defmodule Screens.Lines.Line do
           short_name: String.t(),
           sort_order: integer()
         }
+
+  def color("line-Red"), do: :red
+  def color("line-Mattapan"), do: :red
+  def color("line-Orange"), do: :orange
+  def color("line-Green" <> _), do: :green
+  def color("line-Blue"), do: :blue
+  def color("line-CR-" <> _), do: :purple
+  def color(_), do: :yellow
 end
--- a/lib/screens/lines/line.ex
+++ b/lib/screens/lines/line.ex
@@ -11,12 +11,4 @@ defmodule Screens.Lines.Line do
           short_name: String.t(),
           sort_order: integer()
         }
-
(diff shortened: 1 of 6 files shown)


lib/screens/lines/line.ex
```diff
@@ -11,4 +11,12 @@ defmodule Screens.Lines.Line do
...
           sort_order: integer()
         }
+
+  def color("line-Red"), do: :red
+  def color("line-Mattapan"), do: :red
+  def color("line-Orange"), do: :orange
+  def color("line-Green" <> _), do: :green
+  def color("line-Blue"), do: :blue
+  def color("line-CR-" <> _), do: :purple
+  def color(_), do: :yellow
```
rwaskiewicz: For my own understanding, where do these functions get used? I see we're using `Route.color()` in the route pill serializer in this PR, but not these `color` functions. I commented them out locally, ran the tests, and they all passed 🤔 However maybe they're not under test directly or indirectly, and/or I'm missing something in the grand scheme of things?

rwaskiewicz (approved):

lib/screens/lines/line.ex
robbie-sundstrom: Thanks for pointing out, you're totally right! I changed around the logic here so that these were no longer used, and then totally missed that when cleaning up the code

PR #3023, opened 2026-04-24
fix: Dependabot not updating composite actions

digitalcora:
Using `/` as the directory of a `github-actions` ecosystem is actually an "alias" for `.github/workflows`, and so ignores our composite actions which live in subdirectories of `.github/actions`.

By switching to the `directories` field, we can specify multiple directories and they can include wildcards. Use this to include `actions/*` and also make the `workflows` "alias" explicit.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,7 +1,9 @@
 version: 2
 updates:
   - package-ecosystem: github-actions
-    directory: "/"
+    directories:
+      - "/.github/actions/*"
+      - "/.github/workflows"
     schedule:
       interval: weekly
       time: "10:00"


rwaskiewicz (approved):
Nice! What I wouldn't give for a 1st party validation tool for GitHub configuration files that could call things like this out 😆

digitalcora:
It is somewhat bizarre no built-in validation exists for Dependabot config files, considering this general concept/UX _does_ exist on GitHub and is currently used for `CODEOWNERS` files — in a PR that changes such a file, you have a nice little indicator telling you whether it's valid, before you merge it.

Though to be fair our current configuration is entirely _valid_, it simply doesn't include some directories that could have been included (which, for all GitHub knows, is an intentional decision). The "hole" in the experience here is created by there not being a designated directory for in-repo composite actions. Unlike `workflows`, which is "special" to GitHub, the `.github/actions` directory is just a convention. Your composite actions _could_ be anywhere in your repo.

rwaskiewicz:
Absolutely agreed - IMO the ways these config files/DSLs (particularly looking at you, GH Actions) can/should be validated would probably require various checks that may require some manual inspection - e.g. I'd love it if GH had something like ESLint's more verbose logging would tell us what files are being linted/being picked up by the globs defined in the ESLint config file (and in this case, by `directory`/`directories`)

PR #3030, opened 2026-04-28
deps(dev): bump eslint-plugin-react-hooks

digitalcora:
This is holding up #3026 due to failures with the new version of `eslint-plugin-react-hooks`.

* Improvements to `set-state-in-effect` catch more occurrences. Fix or ignore these as appropriate.

* Improvements to `immutability` catch an unconventional use of variable reassignment in `DisruptionDiagram`. Ignore this for now to avoid a disruptive refactoring of the component that might affect behavior.


robbie-sundstrom (approved):
Looks good, and the ignored cases seem fine to me!

PR #3031, opened 2026-04-29
chore(repo): add initial zizmor configuration file

rwaskiewicz:
**Asana task**: [Try Zizmor](https://app.asana.com/1/15492006741476/project/1213833239363217/task/1214349478400567)

Description

This commit introduces zizmor via a new GitHub action to the project.
zizmor is a static analysis tool for GitHub workflows, actions, and
Dependabot configuration files, meant to sniff out less than ideal
practices and code smells, and is being eyed by the organization for
adoption into all MBTA repositories.

The new GitHub action, `zizmor.yml`, runs using zizmor's strictest
persona, the "auditor". This setting runs more audit checks than other
personas in zizmor, driving more change in this commit than other
personas would.

The changes to existing workflows are driven from running zizmor, fixing
audits that failed:
- [Anonymous definition](https://docs.zizmor.sh/audits/#anonymous-definition)
- [Artipacked](https://docs.zizmor.sh/audits/#artipacked)
- [Concurrency limits](https://docs.zizmor.sh/audits/#concurrency-limits)
- [Excessive permissions](https://docs.zizmor.sh/audits/#excessive-permissions)
- [Template injection](https://docs.zizmor.sh/audits/#template-injection)
- [Undocumented permissions](https://docs.zizmor.sh/audits/#undocumented-permissions)
- [Unpinned uses](https://docs.zizmor.sh/audits/#unpinned-uses)

The only audit that has been ignored is `secrets-outside-env`, which
requires moving secrets to GitHub environments. As this requires
escalated permissions, its recommended that we do that as a separate
task, where we can easily test secrets in GitHub environments and
potentially rollback a smaller set of changes if necessary


skyqrose (commented):
I see you added zizmor as a new workflow. Remember to add this workflow to the list of required checks in the repo settings.

Alternatively, you could add zizmor as another step to the existing CI workflow, but it's up to you which way you wanna do it.

docs/github-actions/zizmor.md
```diff
@@ -0,0 +1,37 @@
+# Zizmor
```
rwaskiewicz: Question(s) for the screens team - is this useful?

If so, is it discoverable in this directory? 
If not, is there somewhere else y'all think we should put this?

.github/workflows/zizmor.yml
```diff
@@ -0,0 +1,29 @@
...
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+        with:
+          persist-credentials: false
+
+      - name: Run zizmor
+        uses: zizmorcore/zizmor-action@b1d7e1fb5de872772f31590499237e7cce841e8e # v0.5.3
+        with:
+          advanced-security: false
```
rwaskiewicz, line 29: Unsure if we have GH Advanced Security as a part of our plan. If we don't, [this is required](https://github.com/zizmorcore/zizmor-action#usage-with-github-advanced-security-recommended). I'll circle back to this

.github/actions/setup/action.yml
```diff
@@ -4,8 +4,8 @@ description: Installs tools and dependencies
 runs:
   using: composite
   steps:
-    - uses: asdf-vm/actions/setup@v4
-    - uses: actions/cache@v5
+    - uses: asdf-vm/actions/setup@b7bcd026f18772e44fe1026d729e1611cc435d47 # v4.0.1
```
rwaskiewicz, line 7: Note: Dependabot updates the version comment automatically when updating SHAs - https://github.blog/changelog/2022-10-31-dependabot-now-updates-comments-in-github-actions-workflows-referencing-action-versions/

This does mean that in the event of manual upgrade, folks will need to bump the version manually, but hopefully those events are few and far between

docs/github-actions/zizmor.md
digitalcora: > is this useful?

I'm not convinced that it is — it's largely a duplication of the relevant parts of the official Zizmor docs, except this copy won't be updated automatically when the tool is. All the information here can be quickly and easily discovered or reconstructed from more authoritative sources.

.github/workflows/zizmor.yml
rwaskiewicz: Actually, I don't know if I can check this with my current permissions - @skyqrose do you know if we pay for Advanced Security (or have the ability to check)?

docs/github-actions/zizmor.md
rwaskiewicz: That's kind of where I was as well. The impetus for this doc was "as a developer whose PR is failing the zizmor check, how do I fix that?". But if the docs aren't discoverable/are a regurgitation of existing docs.... 🤷 Perhaps the output from a failing audit is enough - we can start there and always resurrect this doc if needed.

.github/workflows/zizmor.yml
skyqrose: It's something we've considered, but aren't currently paying for. ([Asana task](https://app.asana.com/1/15492006741476/project/1202111961403803/task/1205800604717300))

Public repos (including this one) get some fraction of Advanced Security for free ([docs](https://docs.github.com/en/get-started/learning-about-github/about-github-advanced-security#github-code-security)).

But I don't really know how it works. I think as we're exploring this new tool, we should do the simple thing that's more likely to work, which is treating it like a normal CI check that could fail, rather than trying to plug into GitHub's special features.

rwaskiewicz: Sounds good to me - we'll keep this setting as-is for now in that case, and can always revisit in the future. Thanks!

.github/workflows/deploy-ecs.yml
```diff
@@ -10,20 +10,29 @@ on:
   push:
     branches: ["main"]
 
+permissions: {}
+
+concurrency:
+  group: ${{ github.workflow }}
```
digitalcora: I don't think this is the concurrency behavior we want, since e.g. if you start a prod deploy, and then I start a dev-green deploy, the latter will clobber the former even though there's no "conflict" between them. Having the concurrency scoped to the environment name would make more sense to me.

.github/workflows/zizmor.yml
```diff
@@ -0,0 +1,29 @@
+name: GitHub Actions Security Check
+
+on:
+  push:
+    branches: ['main']
+  pull_request:
+    branches: ['**']
```
digitalcora: * Is this different behavior from not specifying `branches` at all?
* Should we have this only run on PRs that change files within `.github/actions` and `.github/workflows`?

.github/workflows/deploy-ecs.yml
```diff
@@ -10,20 +10,29 @@ on:
...
+
 jobs:
   sentry-upload:
+    name: "Upload sourcemaps to Sentry"
     runs-on: ubuntu-latest
     env:
-      SENTRY_AUTH_TOKEN: ${{ secrets.SENTRY_AUTH_TOKEN }}
-      SENTRY_ORG: ${{ secrets.SENTRY_ORG }}
-      SENTRY_PROJECT: ${{ secrets.SENTRY_PROJECT }}
+      SENTRY_AUTH_TOKEN: ${{ secrets.SENTRY_AUTH_TOKEN }} # zizmor: ignore[secrets-outside-env] - TODO: Move to a dedicated environment
```
skyqrose, line 24: What's your opinion on this rule? Is it worth having all of our repos follow it?

Link to the zizmor docs about it: https://docs.zizmor.sh/audits/#secrets-outside-env (but I haven't thought through it yet)

When you make a decision about it, you should document the rule in our [Notion doc](https://www.notion.so/mbta-downtown-crossing/Zizmor-GHA-audit-350f5d8d11ea8021bc7bc0c3fd6de6a4) so other teams can follow your advice.

skyqrose (commented):
LGTM, thanks for spearheading this investigation!

digitalcora (approved):
🔒✨

.github/workflows/deploy-ecs.yml
rwaskiewicz: I don't know where my comments on this went (if GH ate them or otherwise), but I think you're right - as you've seen, updated in https://github.com/mbta/screens/pull/3031/commits/f8f19a88460ba2bb8a457938a5ff948e43a593c0

.github/workflows/zizmor.yml
rwaskiewicz: I don't know where my comments on this went (if GH ate them or otherwise), but for posteriy the tl;dr is/was:

> Is this different behavior from not specifying branches at all?

It's not - I personally prefer the explicitness here, but that's not something I'm deeply attached to.

> Should we have this only run on PRs that change files within .github/actions and .github/workflows?

👍  Updated to look at `.github` as a whole, since this also handles dependabot config

.github/workflows/deploy-ecs.yml
rwaskiewicz: My current opinion is that it is worth following, especially if there's a chance that a workflow has used an event that triggers an action like `pull_request_target` in the past that could be used to extract secrets (speaking from personal experience 😭). But I'm sure GH has added some additional protections since then - I'll spend some time on Monday verifying my current knowledge/using what I learned to have a better informed opinion & write it up

skyqrose (approved):

.github/workflows/deploy-ecs.yml
rwaskiewicz: I've added a section to the Notion doc. I'm going to merge this as-is, and we can go from there - either move this repo to env secrets (and remove the TODOs I've added), or suppress the rule (and remove the TODOs I've added).

PR #3032, opened 2026-04-29
feat: Use new Predictions trip_headsign field for headsign displays

robbie-sundstrom:
**Asana task**: [Screens: Use new `trip_headsign` field for departure headsigns](https://app.asana.com/1/15492006741476/project/1185117109217413/task/1214064512691055?focus=true)

- Adds new `trip_headsign` field to be used as the highest priority option in `Departure.headsign`
- Removes the `stop_headsign` field from the `Prediction` struct, since this field can only exist on a `Schedule` at this time

### DIFF
--- a/lib/screens/v2/departure.ex
+++ b/lib/screens/v2/departure.ex
@@ -137,27 +137,22 @@ defmodule Screens.V2.Departure do
     end
   end
 
-  def headsign(%__MODULE__{prediction: p, schedule: s}) when not is_nil(p) do
-    %Prediction{trip: %Trip{headsign: headsign}} = p
+  # The headsign to display in order of preference.
+  # May be updated in the schedule/prediction data w/ the trip's or stop's final destination.
+  @spec headsign(t()) :: String.t()
+  def headsign(%__MODULE__{prediction: %Prediction{trip_headsign: trip_headsign}})
+      when not is_nil(trip_headsign),
+      do: trip_headsign
 
-    case s do
-      %Schedule{stop_headsign: stop_headsign} when not is_nil(stop_headsign) ->
-        stop_headsign
+  def headsign(%__MODULE__{schedule: %Schedule{stop_headsign: stop_headsign}})
+      when not is_nil(stop_headsign),
+      do: stop_headsign
 
-      _ ->
-        headsign
-    end
-  end
+  def headsign(%__MODULE__{prediction: %Prediction{trip: %Trip{headsign: headsign}}}),
+    do: headsign
 
-  def headsign(%__MODULE__{prediction: nil, schedule: s}) do
-    case s do
-      %Schedule{stop_headsign: stop_headsign} when not is_nil(stop_headsign) ->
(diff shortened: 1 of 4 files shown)


lib/screens/v2/departure.ex
```diff
@@ -137,27 +137,22 @@ defmodule Screens.V2.Departure do
...
+  # The headsign to display in order of preference.
+  # May be updated in the schedule/prediction data w/ the trip's or stop's final destination.
+  @spec headsign(t()) :: String.t()
+  def headsign(%__MODULE__{prediction: %Prediction{trip_headsign: trip_headsign}})
+      when not is_nil(trip_headsign),
+      do: trip_headsign
+
+  def headsign(%__MODULE__{schedule: %Schedule{stop_headsign: stop_headsign}})
+      when not is_nil(stop_headsign),
+      do: stop_headsign
```
rwaskiewicz, line 149: For my own understanding, why do we prefer `trip_headsign` coming from a `%Prediction{}` rather than the `stop_headsign` from a `%Schedule{}`? It looks like previously we [were preferring the latter](https://github.com/mbta/screens/pull/3032/changes#diff-8e82f72a4752d353df1dfa394d9d3c47cb3354cccedc52652a661441edfd8df4L144-L148) over the `Prediction#headsign` field. Is it that predictions are expected to be more up to date/dynmaic, whereas the schedule is more fixed?

I'm not sure my logic is right there, because I see tests where we prefer the `%Schedule#stop_headsign` over the `%Prediction#headsign`:
https://github.com/mbta/screens/blob/f517cdb697c173f3e190f090a54ee9929f1761e6/test/screens/v2/departure_test.exs#L122-L128
but maybe that's more about the differentiation between a `stop_headsign` and a `headsign`?

rwaskiewicz (approved):

lib/screens/v2/departure.ex
robbie-sundstrom: Totally understandable to be confused here. It took me a bunch of reading through documentation and [this Slack thread ](https://mbta.slack.com/archives/C03K6NLKKD1/p1776275705539589)to understand. A `Schedule` and `Prediction` that come from the V3 API each have a `headsign` value that comes from the schedule in GTFS. As you alluded to in your comment, this can be a different value than the `Schedule.stop_headsign` or `Prediction.trip_headsign`. 

For instance, a train on the that is stopping at Park Street but is scheduled to stop at Government Center will have [TripProperties.trip_headsign ](https://gtfs.org/documentation/realtime/reference/#message-tripproperties)set to "Park Street" in the TripUpdates that we are sending for that trip, which will now be reflected in the /predictions endpoint of the V3 API. The `headsign` value of this `Prediction` would still be set to the scheduled destination of "Government Center". If `trip_headsign` is present then, we want to override the `headsign` that comes from the initial scheduled value. This logic is the same for overriding with `stop_headsign`, but we're giving `trip_headsign` priority.

PR #3036, opened 2026-04-30
chore: add ueberauth_fake_oidcc in dev and test

rwaskiewicz:
**Asana task**: [Adopt ueberauth_fake_oidcc in Screens](https://app.asana.com/1/15492006741476/project/1185117109217413/task/1213899774844024)

Description
Introduce the `ueberauth_fake_oidcc` library to the project to be used in the dev and test environments. Moving to this library allows us to delete the `fake.ex` Ueberauth.Strategy (and its corresponding test).

The configuration for `dev` remains essentially unchanged, other than the `ttl` option being provided now in order to maintain parity with the now-deleted fake strategy.

The configuration for `test` introduces `ttl` for the same reason, as well as an `auto-redirect` option so that the login form is skipped during test time.


test/screens_web/controllers/auth_controller_test.exs
```diff
@@ -5,37 +5,18 @@ defmodule ScreensWeb.Controllers.AuthControllerTest do
...
-        }
-      }
-
       conn =
-        conn
-        |> assign(:ueberauth_auth, auth)
-        |> get(ScreensWeb.Router.Helpers.auth_path(conn, :callback, "keycloak"))
+        get(
+          conn,
+          ScreensWeb.Router.Helpers.auth_path(conn, :callback, "keycloak",
```
rwaskiewicz: I left this as using the generated helper function, as opposed to writing out the path like [we did in Screenplay](https://github.com/mbta/screenplay/pull/708/changes#diff-1822bf39952440ae24a045382b32ee22a9c3b49f840041a53b35f7044d19c48cR14). It looks like helpers are [now deprecated](https://hexdocs.pm/phoenix/Phoenix.Router.html#module-helpers-deprecated) - LMK if you'd prefer we write the path out for this (or another) reason

digitalcora: I actually missed in my review of that PR that we changed the use of the [~p sigil](https://hexdocs.pm/phoenix/Phoenix.VerifiedRoutes.html) (which is what replaced router helpers) with a plain un-verified string. We should use that sigil here if possible.

digitalcora (approved):
Looks good aside from my one comment reply!

PR #3039, opened 2026-05-04
refactor: grouped the states in RDS tests

deanshi:
Description
- No functionality changes in the tests themselves, but added multiple `describes` instead of a single `get/1` to group the multiple states


robbie-sundstrom (approved):
thanks for cleaning up!

PR #3040, opened 2026-05-04
fix: add Alerts/Emergency Takeover Candidate Generator to DupNew

deanshi:
Description
- Also had to refactor for injecting Stop/LocationContext so we could probably stub them out in the `DupNewTest` without having to add params to the candidate generators. 

- [x] Tests modified?

### DIFF
--- a/lib/screens/v2/candidate_generator/dup_new.ex
+++ b/lib/screens/v2/candidate_generator/dup_new.ex
@@ -2,6 +2,7 @@ defmodule Screens.V2.CandidateGenerator.DupNew do
   @moduledoc false
 
   alias Screens.V2.CandidateGenerator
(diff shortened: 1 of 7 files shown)


lib/screens/v2/candidate_generator/dup_new.ex
```diff
@@ -2,6 +2,7 @@ defmodule Screens.V2.CandidateGenerator.DupNew do
   @moduledoc false
 
   alias Screens.V2.CandidateGenerator
+  alias Screens.V2.CandidateGenerator.Dup.Alerts, as: AlertsInstances
```
digitalcora: nit: `AlertsInstances` seems like a confusing name for this module. `AlertsGenerator`?

digitalcora (approved):

PR #3041, opened 2026-05-04
feat: use async_stream to fetch data in RDS

deanshi:
Description
- timeouts were causing errors with RDS as opposed to the default variant
- pulled out the data fetching using `async_stream` as opposed to having it do it synchronously within the with clause
- added some typing for readability of the data results


robbie-sundstrom (approved):
Great idea to do all of these fetches async!

PR #3045, opened 2026-05-05
feat: remove am/pm from timestamps

deanshi:
**Asana task**: [[Bug] First Trip states are on two lines instead of one](https://app.asana.com/1/15492006741476/project/1212581278818608/task/1214442254145703?focus=true)

Description
- currently AM/PM seems to be only used on DUPs and with RDS we want to remove AM/PM completely
- this removes instances and usages of am/pm within timestamps, serialization and on the frontend
- fixes an RDS issue for the First Trip state + am/pm timestamps causing the times to split onto two lines

- [x] Tests modified?

### DIFF
--- a/assets/src/components/departures/departure_time.tsx
+++ b/assets/src/components/departures/departure_time.tsx
@@ -8,7 +8,7 @@ import { classWithModifier, classWithModifiers } from "Util/utils";
 type DepartureTime =
   | { type: "text"; text: string }
   | { type: "minutes"; minutes: number }
-  | { type: "timestamp"; hour: number; minute: number; am_pm: string | null }
+  | { type: "timestamp"; hour: number; minute: number }
   | { type: "status"; pages: string[] }
   | { type: "overnight" };
 
@@ -37,14 +37,7 @@ const DepartureTimePart: ComponentType<DepartureTimePartProps> = ({
       const paddedMinute = time.minute < 10 ? "0" + time.minute : time.minute;
       const timestamp = `${time.hour}:${paddedMinute}`;
 
-      return (
-        <div className="departure-time__timestamp">
-          <span className="departure-time__time">{timestamp}</span>
-          {time.am_pm && (
-            <span className="departure-time__ampm">{time.am_pm}</span>
-          )}
-        </div>
-      );
+      return <span className="departure-time__timestamp">{timestamp}</span>;
     }
 
     case "status": {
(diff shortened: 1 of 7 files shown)


assets/src/components/departures/departure_time.tsx
```diff
@@ -40,9 +40,6 @@ const DepartureTimePart: ComponentType<DepartureTimePartProps> = ({
       return (
         <div className="departure-time__timestamp">
           <span className="departure-time__time">{timestamp}</span>
```
digitalcora: Can we collapse the nesting of `__time` inside `__timestamp` since it is now the only child?

assets/src/components/dup/departures/departure_row.tsx
```diff
@@ -16,7 +16,7 @@ const DepartureRow: ComponentType<DepartureRowBase> = ({
     route.type === "dual" ? "extended-route_pill" : "";
 
   const classModifier = timesWithCrowding.some(
-    (time) => time.time?.type === "timestamp" && time?.time.am_pm,
```
digitalcora, line 19: This changes the condition — now it's true if any times are of the `timestamp` type, rather than if any times have an AM/PM indicator. Is this intentional or can we remove the entire `shortened-headsign` logic from existence since the condition it was previously checking is now always false?

deanshi: It was intentional for the sake of simplicity, but extremely fair point. Let me check to see if anything completely breaks and if we can remove `shortened-headsign` or maybe at least adjust the logic for the special cases.

deanshi: I took a look and it seems like we do still want `shortened-headsigns` for the First Trip case with the `First` text. I added that extra check here.

deanshi: Waiting on Design about a question for this particular screen: 
Screenshot 2026-05-06 at 2 32 36 PM

Seeing if we think it's too close and we want to shorten the headsign.

digitalcora: The new condition seems redundant — if a time-with-crowding is a First Trip, it will always have exactly one time and that time will always be a timestamp.

deanshi: okay - looks like the above headsign is just a bit too tight to the timestamps according to design.  

I'm going to change it again so that we'll just have `shortened-headsign` for timestamp times, if we're okay with that.

assets/src/components/departures/departure_time.tsx
digitalcora: @deanshi I see you 👍'd this a bit ago, but it's unchanged.

deanshi: Yep apologies, I'm working on that as we speak just combing the CSS. I'll have it up shortly.

digitalcora (approved):

PR #3047, opened 2026-05-05
fix: fixed route pill for Green Line headways on multiple branches

deanshi:
**Asana task**: [[Bug] GL-Branch pill showing for Eastbound headways](https://app.asana.com/1/15492006741476/project/1212581278818608/task/1214550997506231?focus=true)

Description
- in the previous code, we were only taking the first departure into account when we were generating route pills
- we now calculate to see if there are multiple route_ids from destinations that have headways. If we do, we will manually create a GL pill rather than going through the regular route pill creation
- also added tests to widget_instance/departures_test.exs for serializing first trip, last trip and headways

Screenshot 2026-05-05 at 3 50 29 PM

- [x] Tests added?

### DIFF
--- a/lib/screens/v2/candidate_generator/dup_new/departures.ex
+++ b/lib/screens/v2/candidate_generator/dup_new/departures.ex
@@ -367,12 +367,15 @@ defmodule Screens.V2.CandidateGenerator.DupNew.Departures do
     end)
     |> Enum.flat_map(fn
       {{_line, _direction_name}, [%RDS{state: %Headways{departure: departure, range: range}}]} ->
-        [{departure, range, nil, :headways}]
+        [{[departure], range, nil, :headways}]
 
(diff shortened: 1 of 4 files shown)


digitalcora:
Note: I don't think this is specific to the Green Line, as the Silver Line also consists of multiple routes with distinct pills that should roll up to a "common" pill (note though that the "Silver Line" is two Lines, SL Waterfront and SL Washington). I'd rather we not add even more "Green Line special cases" — part of the point of basing the RDS logic on Lines was that these directly represent the relevant concept without us having to hard-code it. Probably the only thing we're missing is a way to translate a Line into a pill.

deanshi:
Closing this PR for a slightly larger one that encompasses the Green Line grouping but also the omission of headways for lines/direction_id destinations that are still active.

PR #3048, opened 2026-05-06
fix: shuttle bus message overlapping

rwaskiewicz:
**Asana task**: [Screens bug: shuttle bus message overlapping](https://app.asana.com/1/15492006741476/project/1185117109217413/task/1214364461704789?focus=true)

**Description**

Extend the `max-height` that the text sections on prefare screens may
take. In https://github.com/mbta/screens/commit/de699b640147a9ae73e8e7a2607da636104f3888, we updated the styling for "Use shuttle buses",
amongst other bits of text. In the case of alerts where station name(s)
are long (such as "Braintree & Ashmont"), the height of the section was
exceeding the previous `max-height`. The value of `850px` was arbitrarily
chosen to fit the content associated with the bug.

Remove class name assignments of `.alert-card__content-block__text`. In
https://github.com/mbta/screens/commit/de699b640147a9ae73e8e7a2607da636104f3888, we removed the css class from
`prefare_single_screen_alert.scss`, but did not remove it from the
components consuming it.

Update reference to disruption diagram 

**Screenshots**

Firefox:
Screenshot 2026-05-06 at 13 47 43

Chrome:
Screenshot 2026-05-06 at 13 50 20

**Notes**

You'll see that in the screenshots above (included Firefox and Chrome to demonstrate, even though we're target Chrome IRL) that Braintree's last 'E' gets cut off. I believe this to be a pre-existing bug and not strictly related to flexbox - it seems like the `viewbox` on the SVG is setting the minY value incorrectly. However, this might be an artifact of simulating the screen - open to discussion, but my vote is to spin that bug out into a separate task

### DIFF
--- a/assets/css/pre_fare/prefare_single_screen_alert.scss
+++ b/assets/css/pre_fare/prefare_single_screen_alert.scss
@@ -114,7 +114,7 @@
   }
 
   .alert-card__content-block__text-sections {
-    max-height: 772px;
+    max-height: 850px;
   }
 
   .alert-card__fallback__icon {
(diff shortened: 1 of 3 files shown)


deanshi (approved):
Thanks for the extra cleanup/help 🙏

PR #3054, opened 2026-05-07
fix: fixed RDS headway grouping and route pills for grouped headways

deanshi:
**Asana task**: [[Bug] GL-Branch pill showing for Eastbound headways](https://app.asana.com/1/15492006741476/project/1212581278818608/task/1214550997506231?focus=true)

Description
- Fixed showing headways when there were still other states for similar line/direction_id destinations. We now omit headways when we find other destinations in other states other than headway
- Fixed pill grouping when we have headways, added a `serialize_line` within `RoutePill` and we now use the line to create the Route Pill rather than using individual departures
- Cleaned up some parameters being passed through in order to simplify serialization
- Added test for the headway omission case and also `serialize_line` within RoutePills, and a missing First Trip/Service Ended serialization

- [x] Tests added?

### DIFF
--- a/lib/screens/v2/candidate_generator/dup_new/departures.ex
+++ b/lib/screens/v2/candidate_generator/dup_new/departures.ex
@@ -3,6 +3,7 @@ defmodule Screens.V2.CandidateGenerator.DupNew.Departures do
 
   import Screens.Inject
 
+  alias Screens.Lines.Line
   alias Screens.Routes.Route
 
   alias Screens.V2.Departure
@@ -12,6 +13,7 @@ defmodule Screens.V2.CandidateGenerator.DupNew.Departures do
   alias Screens.V2.WidgetInstance.Departures, as: DeparturesWidget
 
   alias Screens.V2.WidgetInstance.Departures.{
+    HeadwayRow,
     HeadwaySection,
     NoDataSection,
     NormalSection,
@@ -171,10 +173,10 @@ defmodule Screens.V2.CandidateGenerator.DupNew.Departures do
 
   # Bidirectional -> Use no headsign for the trains message
   defp create_headway_section([
-         %RDS{state: %{route_id: route_id, direction_name: direction_name_one, range: range}}
-         | [%RDS{state: %{route_id: route_id, direction_name: direction_name_two, range: range}}]
+         %RDS{state: %{route_id: route_id, direction_id: direction_id_one, range: range}}
+         | [%RDS{state: %{route_id: route_id, direction_id: direction_id_two, range: range}}]
        ])
-       when direction_name_one != direction_name_two do
+       when direction_id_one != direction_id_two do
     %HeadwaySection{
       route: route_id,
       time_range: range,
@@ -188,35 +190,38 @@ defmodule Screens.V2.CandidateGenerator.DupNew.Departures do
   defp create_headway_section(
          [
            %RDS{
-             headsign: first_headsign,
+             headsign: headsign,
+             line: %Line{id: first_line_id},
              state: %Headways{
-               route_id: first_route_id,
-               direction_name: first_direction_name,
-               range: first_range
+               route_id: route_id,
+               direction_name: direction_name,
+               direction_id: direction_id,
+               range: range
              }
            }
            | _
          ] = destinations
        ) do
     %HeadwaySection{
-      route: first_route_id,
-      time_range: first_range,
+      route: route_id,
+      time_range: range,
       headsign:
         cond do
-          Enum.all?(destinations, fn %RDS{headsign: headsign} ->
-            headsign == first_headsign
+          Enum.all?(destinations, fn %RDS{headsign: other_headsign} ->
+            headsign == other_headsign
           end) ->
-            first_headsign
+            headsign
 
           Enum.all?(
             destinations,
             fn %RDS{
-                 state: %Headways{route_id: other_route_id, direction_name: other_direction_name}
+                 line: %Line{id: other_line_id},
+                 state: %Headways{direction_id: other_direction_id}
                } ->
-              other_direction_name == first_direction_name and other_route_id == first_route_id
+              first_line_id == other_line_id and direction_id == other_direction_id
             end
           ) ->
-            first_direction_name
+            direction_name
 
           true ->
             nil
@@ -314,7 +319,7 @@ defmodule Screens.V2.CandidateGenerator.DupNew.Departures do
     rds = Map.get(grouped_rds, :other, [])
 
     sorted_departures_from_rds(rds) ++
-      headways_from_rds(headway_rds) ++
+      headways_from_rds(rds, headway_rds) ++
       sorted_departures_from_rds(service_ended_rds, true)
   end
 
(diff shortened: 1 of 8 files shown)


lib/screens/v2/widget_instance/departures.ex
```diff
@@ -23,7 +25,8 @@ defmodule Screens.V2.WidgetInstance.Departures do
 
     @type special_trip_type :: :first_trip | :last_trip
 
-    @type headway_row :: {Departure.t(), Headways.range(), String.t() | nil, :headways}
+    @type headway_row ::
+            {String.t(), Line.id(), Trip.direction(), Headways.range(), String.t(), :headways}
```
robbie-sundstrom: This tuple feels like it's become large, but that all the info is definitely needed. Seems like this should become a struct for some STRUCTure instead

lib/screens/v2/candidate_generator/dup_new/departures.ex
```diff
@@ -171,10 +172,10 @@ defmodule Screens.V2.CandidateGenerator.DupNew.Departures do
 
   # Bidirectional -> Use no headsign for the trains message
   defp create_headway_section([
-         %RDS{state: %{route_id: route_id, direction_name: direction_name_one, range: range}}
-         | [%RDS{state: %{route_id: route_id, direction_name: direction_name_two, range: range}}]
+         %RDS{state: %{route_id: route_id, direction_id: direction_id_one, range: range}}
+         | [%RDS{state: %{route_id: route_id, direction_id: direction_id_two, range: range}}]
        ])
-       when direction_name_one != direction_name_two do
+       when direction_id_one != direction_id_two do
```
robbie-sundstrom, line 179: Since this is just checking if the headways are for both `direction_id`s 0 and 1, this could be simplified.

```suggestion
         %RDS{state: %{route_id: route_id, direction_id: 0, range: range}}
         | [%RDS{state: %{route_id: route_id, direction_id: 1, range: range}}]
]) do
```

lib/screens/v2/candidate_generator/dup_new/departures.ex
```diff
@@ -351,28 +355,72 @@ defmodule Screens.V2.CandidateGenerator.DupNew.Departures do
...
+
+          %Countdowns{departures: [first_departure | _]} ->
+            [{line_id, Departure.direction_id(first_departure)}]
+
+          %FirstTrip{first_scheduled_departure: departure} ->
+            [{line_id, Departure.direction_id(departure)}]
+
+          %ServiceEnded{last_scheduled_departure: departure} ->
+            [{line_id, Departure.direction_id(departure)}]
+        end
```
robbie-sundstrom: I think this makes sense to pull out into a separate function -- I'm not sure if it would be used elsewhere, but it seems like a separate piece of logic. Especially when it's involving all these different cases for RDS states, I think we want it to be easy to modify in the future. Maybe something like: 
```
defp extract_line_direction_pairs(%RDS{line: %Line{id: line_id}, state: state}) do
  case state do
    %NoService{} ->
      [{line_id, 0}, {line_id, 1}]

    %Countdowns{departures: [first_departure | _]} ->
      [{line_id, Departure.direction_id(first_departure)}]

    %FirstTrip{first_scheduled_departure: departure} ->
      [{line_id, Departure.direction_id(departure)}]

    %ServiceEnded{last_scheduled_departure: departure} ->
      [{line_id, Departure.direction_id(departure)}]
  end
end
```

This would make it easier to call as part of a `flat_map` as well, which I think would be more efficient than this `Enum.reduce` that creates a `MapSet` on each iteration

```
  lines_in_other_states =
    rds
    |> Enum.flat_map(&extract_line_direction_pairs/1)
    |> MapSet.new()
```

lib/screens/v2/candidate_generator/dup_new/departures.ex
```diff
@@ -351,28 +355,72 @@ defmodule Screens.V2.CandidateGenerator.DupNew.Departures do
...
+           state: %Headways{
+             departure_id: departure_id,
+             direction_name: direction_name,
+             direction_id: direction_id,
+             range: range
+           }
+         }
+         | _
+       ]} ->
+        [
```
robbie-sundstrom: Could also simplify this pattern matching  and make it clearer what's happening, or pull it out into its own function: 

```
    |> Enum.flat_map(fn {{line_id, direction_id}, rds_list  ->
        %RDS{
          headsign: headsign,
          state: %Headways{
            departure_id: departure_id,
            direction_name: direction_name,
            range: range
          }
        } = hd(rds_list)
      
        displayed_headsign = if length(rds_list) == 1, do: headsign, else: direction_name
     
```

lib/screens/v2/widget_instance/departures.ex
```diff
@@ -491,33 +494,25 @@ defmodule Screens.V2.WidgetInstance.Departures do
...
       id: hash_and_encode(departure_id),
       type: :departure_row,
-      route: serialize_route(departures, route_pill_serializer, screen),
-      headsign:
-        if headsign do
-          %{headsign: headsign}
-        else
-          serialize_headsign(departures, screen)
-        end,
+      route: RoutePill.serialize_line(line_id),
```
robbie-sundstrom: I think we should try to stick to the standard and testable `route_pill_serializer` flow like all other departures. A few different ways I could see this happening, like adding an additional spec option to pass either the full `Line.t()` or just the `line_id`, and then call that directly with `route_pill_serializer.(line, track_number, screen)`

And then an additional spec and logic handling bundled underneath that version like:

`@spec serialize_for_departure(Line.t(), pos_integer() | nil, Screen.t()) :: t()`

lib/screens/v2/candidate_generator/dup_new/departures.ex
deanshi: Kept this the same since we discussed it - down to do whatever suggestion we end up wanting!

lib/screens/v2/candidate_generator/dup_new/departures.ex
deanshi: Pulled it out into a function - Realized the `MapSet.new()` already does the union that I wanted so this definitely is much better.

lib/screens/v2/candidate_generator/dup_new/departures.ex
deanshi: Yep! I think this makes sense. Pulled it out like you suggested and it's working great.

lib/screens/v2/widget_instance/departures.ex
deanshi: Implemented it using the full line and refactored the HeadwayRow to use the full Line instead of `Line.id()`. I figured it was better than having `Line.id()` or `String.t()` as the identifier in case it was confused to be used for something else. 

In the Line case, I also opted to keep the `track_number` always nil and the `screen` param isn't used in this case. Let me know if you'd prefer I keep everything nil or just not use the variables in each case.

lib/screens/v2/candidate_generator/dup_new/departures.ex
robbie-sundstrom: I think leaving the the way you originally had it for the purpose of different ordering is totally fine!

lib/screens/v2/widget_instance/serializer/route_pill.ex
```diff
@@ -109,6 +109,24 @@ defmodule Screens.V2.WidgetInstance.Serializer.RoutePill do
     end
   end
 
+  @spec serialize_for_departure(Line.t(), nil, Screen.t()) :: t()
```
robbie-sundstrom: I like this new implementation, but I think the new `@spec` should go directly below the existing one so it's easy to see that there are two possibilities for the function. This seems to be the accepted standard convention.

robbie-sundstrom (approved):
One small comment about `serialize_for_departure`, but overall looks great!

PR #3055, opened 2026-05-07
chore: restore permissions to deploy-ecs

rwaskiewicz:
**Asana task**: None

Description

In 9502e1b we removed the permissions for deploy-ecs in order to appease zizmor. This resulted in the following error:
```
The workflow is not valid. .github/workflows/deploy-ecs.yml (Line: 34,
Col: 3): Error calling workflow.  The nested job 'deploy' is requesting
'contents: read, id-token: write', but is only allowed 'contents: none,
id-token: none'.
```
This commit restores the necessary permissions

- [ ] Tests added?

### DIFF
--- a/.github/workflows/deploy-ecs.yml
+++ b/.github/workflows/deploy-ecs.yml
@@ -37,8 +37,8 @@ jobs:
       app-name: screens
       environment: ${{ github.event.inputs.environment || 'dev' }}
     permissions: {
-      contents: read,
-      id-token: write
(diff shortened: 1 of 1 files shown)


digitalcora (approved):
Looks good assuming you've tested this with e.g. a deploy to dev-green from this branch.

rwaskiewicz:
> Looks good assuming you've tested this with e.g. a deploy to dev-green from this branch.

Yup! Won't make that mistake again 😓

PR #3057, opened 2026-05-08
chore: remove legacy screen request logging

digitalcora:
**Asana task**: https://app.asana.com/1/15492006741476/project/1185117109217413/task/1214548587982633?focus=true

All active Splunk alerts have been updated to match the new request logging format, so we should be able to remove this (effectively duplicate) legacy logging code.


rwaskiewicz (approved):
For my own edification, when we introduced the `ScreenRequest` plug, we didn't hardcode the search indices that were cordoned off into the `LegacyLogging` plug in 8fad02c, correct? And everything going to splunk now is a result of [this Logster call](https://github.com/mbta/screens/blob/858fa9204af98811241455e83288f184bfd9b00f/lib/screens_web/plug/screen_request.ex#L82), in addition to its configuration in `config/[ENV].exs`? 

I.E. what I'm calling "search indices" are really the request type, which is a part of the [`meta-base` keyword list](https://github.com/mbta/screens/blob/858fa9204af98811241455e83288f184bfd9b00f/lib/screens_web/plug/screen_request.ex#L69) what we add to the logger metadata?

digitalcora:
I'm not quite sure how to interpret your questions so I might have to ask for a rephrase... but, after this change, the only log entry that appears for each screen request is indirectly produced by [this line](https://github.com/mbta/screens/blob/5ca50cf/lib/screens/application.ex#L37), which also logs all HTTP requests in general (but only "screen requests", as defined by `Plug.ScreenRequest`, would have the full set of metadata). The change in Splunk alerts was largely to replace searches for `[screen data request]` with `state=sent request_type=data is_real_screen=true`.

rwaskiewicz:
Oof, yeah - not my best worded comment/question 😅 But that makes sense & ties my questions up, thank you!

PR #3063, opened 2026-05-12
fix: subway status pills overlapping with status text

deanshi:
**Asana task**: [Screens bug: overlapping elements in Subway Status](https://app.asana.com/1/15492006741476/project/1185117109217413/task/1214355771717040?focus=true)

Description
Screenshot 2026-05-12 at 3 14 04 PM

- seems like having min-width was causing it so that the divs weren't respecting the full size of the pills
- alongside that, we needed to add Shuttle as a specific alert type to show the alert link
- removes min-width from the container for the RoutePill and also pulled out statuses into a boolean check for showing the alert URL

### DIFF
--- a/assets/src/components/subway_status/subway_status_common.tsx
+++ b/assets/src/components/subway_status/subway_status_common.tsx
@@ -163,30 +163,18 @@ export const useSubwayStatusTextResizer = (alert: Alert, type: SectionType) => {
       : [FittingStep.FullSize, FittingStep.Abbrev];
   const { ref, step: fittingStep } = useAutoSize(steps, id);
 
-  const isStopsSkipped = /Stops? Skipped/.test(alert.status);
-  const isSuspension = firstWord(alert.status) === "Suspension";
   const isDelays = firstWord(alert.status) === "Delays";
 
-  const location = (() => {
-    if (
-      fittingStep === FittingStep.PerAlertEffect &&
-      (isStopsSkipped || isSuspension)
-    ) {
-      return "mbta.com/alerts";
-    } else if (isAlertLocationMap(alert.location)) {
-      return fittingStep === FittingStep.Abbrev ||
-        fittingStep === FittingStep.PerAlertEffect
-        ? alert.location.abbrev
-        : alert.location.full;
-    } else {
-      return alert.location;
-    }
-  })();
+  let location = getAlertLocationString(alert.location, fittingStep);
+  let status = alert.status;
 
-  const status =
-    fittingStep === FittingStep.PerAlertEffect && isDelays
-      ? "Delays"
-      : alert.status;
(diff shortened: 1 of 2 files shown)


assets/src/components/subway_status/subway_status_common.tsx
```diff
@@ -163,14 +163,13 @@ export const useSubwayStatusTextResizer = (alert: Alert, type: SectionType) => {
...
-  const isSuspension = firstWord(alert.status) === "Suspension";
-  const isDelays = firstWord(alert.status) === "Delays";
+  const alertStatus = alert.status;
+  const isDelays = firstWord(alertStatus) === "Delays";
 
   const location = (() => {
     if (
       fittingStep === FittingStep.PerAlertEffect &&
-      (isStopsSkipped || isSuspension)
+      shouldShowAlertLink(alertStatus)
```
digitalcora: I'm a little confused as to why we do this for specific statuses rather than "not delays", which is the one status that has its own shortening logic for the `PerAlertEffect` step. This is the last step — if we've reached this point, the text is still overflowing and there's nothing else left to try, so we should try _something_ regardless of status.

Basically right now the logic for this `PerAlertEffect` step is:

1. If stops-skipped, shuttle, or suspension: replace location with URL
2. If delay: shorten status string to just "Delays"
3. If (something that's not one of the above): do nothing?

I would propose this makes more sense as:

1. If delay: shorten status string to just "Delays"
2. Otherwise: replace location with URL

These two are sort of equivalent right now since Shuttle/Suspension/Skipped/Delay covers all the alert types that currently go into Subway Status, but I think simpler logic that covers all possibilities would be an improvement. (it actually doesn't currently account for "Single Tracking", which is status text that can result from a Delay alert)

deanshi: Yep, those changes make sense. I'll add those changes and try to clean-up the code a little bit as well. 

Is there a world where we want to also replace the location with the URL when we're displaying "Delays" as well? Does shortening the `Delays` guarantee there will be enough space for the location? 

It seems like it might, but wanted to check in case there were some locations that were particularly long for the GL. 

Screenshot 2026-05-13 at 9 47 24 AM

digitalcora: > Is there a world where we want to also replace the location with the URL when we're displaying "Delays" as well? Does shortening the `Delays` guarantee there will be enough space for the location?

I think Yes and No, respectively, but delays actually having a location (as opposed to being whole-line, which leaves the location field blank) is unusual enough that maybe we could punt on that. The alternative would be creating an extra resizing step, which would be unused by any alert status other than delays, so delays could do something like full location => abbrev location => abbrev to "Delays" => replace location with URL (if that's even the right sequence, I'm not sure).

digitalcora (approved):

PR #3064, opened 2026-05-13
fix: headway text style/alignment and headways for no service

deanshi:
Description
- added a new FreeTextLine layout that provides spacing in between spans which allows for the Headsign to the on the left and the headway values to be on the right. Currently, this is only used for Partial Headway Sections with a headway defined in order to match the new Headway Rows. Full Headway Sections and Headway Sections with no headsign will continue to be the same
- also fixed an issue where No Service destinations were causing Headways not to show up.

Screenshot 2026-05-13 at 12 43 22 PM

- [x] Tests modified?

### DIFF
--- a/assets/css/dup/free_text.scss
+++ b/assets/css/dup/free_text.scss
@@ -1,4 +1,5 @@
 @use "CSS/colors";
+@use "font";
 
 // General styles for free text / route pills
 .free-text {
@@ -135,7 +136,8 @@
 
 // Partial alert and one-line headway have similar styles
 .partial-alert,
-.headway-section--row {
+.headway-section--row,
+.headway-section--row_with_headsign {
   .free-text__line {
     width: 1608px;
     white-space: nowrap;
@@ -150,6 +152,33 @@
     height: 128px;
     padding-top: 40px;
     padding-bottom: 40px;
+  }
+}
+
+.headway-section--row_with_headsign {
+  .free-text__line {
+    display: flex;
+    align-items: center;
+  }
+
+  .free-text__line-container {
+    display: flex;
+  }
+
+  .free-text__line {
+    display: flex;
+    justify-content: space-between;
+    margin-right: 32px;
+  }
+
+  .free-text__string--bold {
+    font-size: 138px;
+    font-weight: 600;
+  }
+
+  .free-text__string--small {
+    font-size: font.$font-size-small-label;
+    font-weight: 400;
   }
 }
 
--- a/lib/screens/v2/widget_instance/departures.ex
+++ b/lib/screens/v2/widget_instance/departures.ex
@@ -209,7 +209,13 @@ defmodule Screens.V2.WidgetInstance.Departures do
         is_only_section
       ) do
     pill_color = Route.color(route)
-    layout = if is_only_section, do: :full_screen, else: :row
+
+    layout =
+      cond do
+        is_only_section -> :full_screen
+        headsign == nil -> :row
+        true -> :row_with_headsign
+      end
 
     text = get_headway_text(headsign, time_range, pill_color, route, is_only_section)
 
--- a/assets/css/dup/departures/time.scss
+++ b/assets/css/dup/departures/time.scss
@@ -1,4 +1,5 @@
 @use "CSS/colors";
+@use "../font";
 
 .departure-time {
   font-size: 138px;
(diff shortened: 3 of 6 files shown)


lib/screens/v2/widget_instance/departures.ex
```diff
@@ -209,7 +209,13 @@ defmodule Screens.V2.WidgetInstance.Departures do
...
       ) do
     pill_color = Route.color(route)
-    layout = if is_only_section, do: :full_screen, else: :row
+
+    layout =
+      cond do
+        is_only_section -> :full_screen
+        headsign == nil -> :row
+        true -> :row_spaced
+      end
```
robbie-sundstrom, line 218: Maybe a name like `row_with_headsign` vs. `row` would be more descriptive than the `row_spaced` naming is. Otherwise it's not really clear why we're differentiating and adding spacing, in which case I think a comment would be helpful for the future

assets/css/dup/free_text.scss
```diff
@@ -153,6 +154,33 @@
...
+    margin-right: 32px;
+  }
+
+  .free-text__string--bold {
+    font-size: 138px;
+    font-weight: 600;
+  }
+
+  .free-text__string--small {
+    font-size: 0.729em;
```
robbie-sundstrom: I see that the existing `departure-time__minutes-label` label uses essentially the same styling as `.free-text__string--small`.  Almost definitely not necessary, but I'm wondering about having a constant in a shared file for DUPs like `$font-size-small-label: 0.729em`. Since this is the only one I see though, this is just a musing and not a blocking suggestion by any means

robbie-sundstrom (approved):
Just small comments, but looks good!

lib/screens/v2/widget_instance/departures.ex
deanshi: Sounds good - my thought here was that since `FreeTextLine` can theoretically be used in other places, I would keep it function agnostic, but realistically, the only place this is being used right now is for headsigns on Headway Sections. If we need to change it in the future, we can maybe do that then.

assets/css/dup/free_text.scss
deanshi: Sounds good - this is something we might also eventually run by design so we can break out everything out in general but I think just adding the small minutes label for now is good.

digitalcora:
> added a new FreeTextLine layout that provides spacing in between spans which allows for the Headsign to the on the left and the headway values to be on the right.

Is there a particular reason we shouldn't switch to implementing this as a normal section with a single "headway row"? Conceptually it seems much closer to that, now, than to a free text component. Maybe free text made sense for the old implementation of this concept a long time ago, but less so today.

deanshi:
> > added a new FreeTextLine layout that provides spacing in between spans which allows for the Headsign to the on the left and the headway values to be on the right.
> 
> Is there a particular reason we shouldn't switch to implementing this as a normal section with a single "headway row"? Conceptually it seems much closer to that, now, than to a free text component. Maybe free text made sense for the old implementation of this concept a long time ago, but less so today.

Yeah, I think the thought here was that I didn't want to necessarily combine `HeadwaySection` serialization and `NormalSection` usage since in my mind they contained different data. If we want, I can switch my thinking and go that route. 

Another path I was thinking was pulling out a common `HeadwayRow` and using that in `NormalSection` and `HeadwaySection`. More than happy to do either thing if we think that draws cleaner lines.

digitalcora:
> Yeah, I think the thought here was that I didn't want to necessarily combine `HeadwaySection` serialization and `NormalSection` usage since in my mind they contained different data.

The relevant "data" at this level is more about the presentation than anything, so in my mind it makes sense for things that are presented similarly to be modeled similarly — and we already have the concept of a "headway row" for multi-direction sections, so in a sense it's odd to have two different ways ("headway section" and "headway row") to present something that behaves the same and serves the same purpose. I think we do have some special logic to "collapse" headway sections to one row-height instead of two, which may not apply to normal sections, so that could be an argument for keeping this the way it is given it's already approved. Up to you and @robbie-sundstrom, I am just driving by 😎

deanshi:
I spent some time on Friday refactoring some of the serializing/HeadwaySection usage to try and conform to making a NormalSection with a single headway row and it ended up being a little bit more painful than I expected. I'm going to merge this, but put up a follow-up PR with those changes when I get to finishing it.

PR #3070, opened 2026-05-15
chore: enable non-Outfront DUPs

digitalcora:
This incorporates several improvements to the client packaging process and a general separation of the concept of a "DUP screen" from the concept of a "packaged client" (deployed to Outfront DUPs). The overall goal is to enable using the "DUP app" on non-Outfront devices that have a similar form factor.

Key changes to enable non-Outfront DUPs:

* Reduce refresh rate from 30 to 20 seconds, in line with other "LCD" screen types. In practice this was never used (and continues not to be used) on Outfront DUPs, since they load the page fresh every time the content appears; it would have only applied to e.g. Screenplay.

* Display the package version information in the DUP header only in packaged copies of the client, as it is only relevant there; "live" clients should always be using the currently deployed code version.

* Refactor the mechanism that checks for outdated client code: Instead of bypassing this for screens whose IDs start with `DUP-`, bypass it when the client sends `last_refresh=packaged`, a sentinel value only present in the packaged client template. This allows non-Outfront DUPs to still have IDs starting with `DUP-` if we'd like them to be that way for organization.

Other changes to improve client packaging:

* Rename `isDup` to `isOutfront`, since this condition will be `false` on "DUP screens" if they are not the packaged client for Outfront devices. Make this check more reliable by reading a data value only present in the packaged client's HTML, instead of looking at the URL.

* Remove the need for three separate Webpack builds by moving the definition of the rotation index from the client JS into the template HTML. Packaged DUP builds are consequently 3x faster. 🚀

* Simplify a few additional instances of `isOutfront` conditional logic by using existing or new data attributes in the template HTML instead.

* Make the packaged client's "version" string solely a property of the built packages, instead of a committed value in the repo. Populate it automatically with the current Git commit short hash, since this is likely to be a more useful value for knowing "what version the client is" than the date the package was built. For special cases, the build script accepts an argument to override the version string.

* File some rough edges off the build script. Client packages are now built into a separate (gitignored) `priv/packaged` directory and the `priv/static` directory is kept clean of packaged DUP files; package filenames now include the version; the script clears the output directory on each run to prevent old versions from piling up.

**Asana task**: https://app.asana.com/1/15492006741476/project/1185117109217413/task/1213739174252146?focus=true

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -33,6 +33,9 @@ npm-debug.log
 # this depending on your deployment strategy.
 /priv/static/
 
+# packaged client assets
+/priv/packaged/
+
 # Local config files
 /priv/local.json
 /priv/local_pending.json
--- a/assets/src/components/dup/README.md
+++ b/assets/src/components/dup/README.md
@@ -1,24 +1,21 @@
 # DUP app packaging
 
-- Ensure [Corsica](https://hexdocs.pm/corsica/Corsica.html) is used on the server to allow CORS requests (ideally limited to just the DUP-relevant routes). It should already be configured in [the API controller](/lib/screens_web/controllers/v2/screen_api_controller.ex) -- if it is, you don't need to do anything for this step.
-- Double check that any behavior specific to the DUP screen environment happens inside of an `isDup()` check. This includes:
-  - `useApiPath` in [`use_api_response.ts`][use_api_response.ts] should return a full URL for the API path: the URL base should be "https://screens.mbta.com".
-  - `imagePath` in [`util.ts`][util.ts] should return relative paths (no leading `/`).
-
-- Set the version string in [`version.ts`][version.ts]. Refer to this file's JSDoc for instructions on how to set the version string value.
+- Double check that any behavior specific to the packaged environment happens inside of an `isOutfront()` condition.
 - If you've renamed / removed image assets, you might want to delete the corresponding folder in `/priv/static`. The folder accumulates assets without clearing old ones out, and these will be included in the built bundle!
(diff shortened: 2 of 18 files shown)


deanshi (approved):
🚢 

Wondering if @robbie-sundstrom might have some thoughts on any of this since he went through the deploy process, but otherwise looks good to me.

robbie-sundstrom:
@digitalcora I really like the changes to the packaging process! I have one clarification about the removal of the `version.tsx` file. I like having the easily-human-parsable date format for the version, but it's also important to be able to track which commit a deployed package is tied to in our Git history. Having the commit hash solves the second problem, but I'm wondering if we want to keep a version date in some way tracked in Git. This could live within the shell script instead of TS, but the commit hash alone obfuscates information about how long it's been since a deployment and could make marking package changes more prone to human error when collaborating with OFM for scheduling these packages.

digitalcora:
> I'm wondering if we want to keep a version date in some way tracked in Git. This could live within the shell script instead of TS, but the commit hash alone obfuscates information about how long it's been since a deployment and could make marking package changes more prone to human error when collaborating with OFM for scheduling these packages.

I agree this would be a useful function to have but I think it might make more sense to track this as a Git tag or something else outside of the code. If we think of DUPs as another "environment" that we deploy code to, our other environments (prod, dev, etc.) do not require a code change and a PR to deploy a new version of the app. Or we could even go further and make an `outfront` GitHub environment like prod and the rest, and have a "deploy" task for it that does nothing (but running it would update GitHub's understanding of what commit is "deployed") — maybe that would be too much of a false consistency with the other environments though. We could discuss options tomorrow at review, I'll put a note in.

PR #3072, opened 2026-05-18
fix: multiple GL alerts causing the pill branch to repeat

deanshi:
Asana Task: [Screens: duplicate branch letters in Subway Status](https://app.asana.com/1/15492006741476/project/1185117109217413/task/1214901633382571?focus=true)

In `serialize_gl_pill_with_branches` it looks like there's a possibility where we can have multiple repeating `route_ids` cause repeating branches when we serialize to the Subway Status route pill (caused by multiple relevant alerts targeting the same route). 

This just adds a unique check for the branches that we create when we use those `route_ids` in `serialize_gl_pill_with_branches`. 

- [x] Tests added?

### DIFF
--- a/test/screens/v2/widget_instance/subway_status_test.exs
+++ b/test/screens/v2/widget_instance/subway_status_test.exs
@@ -573,6 +573,51 @@ defmodule Screens.V2.WidgetInstance.SubwayStatusTest do
       assert expected == WidgetInstance.serialize(instance)
     end
 
+    test "handles 3+ alerts of the same GL branch" do
+      instance = %SubwayStatus{
+        subway_alerts:
+          subway_alerts([
+            %Alert{
+              effect: :delay,
+              severity: 9,
+              informed_entities: [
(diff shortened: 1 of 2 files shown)


test/screens/v2/widget_instance/subway_status_test.exs
```diff
@@ -573,6 +573,51 @@ defmodule Screens.V2.WidgetInstance.SubwayStatusTest do
...
+              severity: 5,
+              informed_entities: [
+                ie(route: "Green-C", stop_id: "second_stop")
+              ]
+            },
+            %Alert{
+              effect: :station_closure,
+              informed_entities: [
+                ie(route: "Green-C", stop_id: "third_stop"),
+                ie(route: "Green-C", stop_id: "fourt_stop")
```
rwaskiewicz: Non-blocking typo
```suggestion
                ie(route: "Green-C", stop_id: "fourth_stop")
```

rwaskiewicz (approved):

PR #3073, opened 2026-05-19
feat: handle June 13th late night service

deanshi:
**Asana task**: [Support Very Late Service in Screens](https://app.asana.com/1/15492006741476/project/1185117109217413/task/1214365239209201?focus=true)

Description
- added hard-coded logic in `service_date/1` to specifically handle the service date for the extended late night
- reworked GL E-Ink to use `service_date/1` in order to use the late night logic instead of using the default service date cutoff times

Using the [late night schedule knowledge](https://www.notion.so/mbta-downtown-crossing/Overnight-service-June-13th-14th-352f5d8d11ea80f9a3fec8bc8ed8e0d0) this will have a two effects:
1. For GL E-Ink, according to the written Green-D final schedule it will just have one minute of potential overlap with the new schedule. In this case, it will be in the Overnight mode for just a minute more. 
2. All other lines will be scheduled to begin after 5 o clock, so it should match our current logic

### DIFF
--- a/lib/screens/v2/candidate_generator/gl_eink/headways.ex
+++ b/lib/screens/v2/candidate_generator/gl_eink/headways.ex
@@ -106,41 +106,45 @@ defmodule Screens.V2.CandidateGenerator.GlEink.Headways do
       Enum.find(
         @dayparts,
         {:overnight, nil, nil},
-        &match(&1, local_time, {stop_id, direction_id})
+        &match(&1, local_time, {stop_id, direction_id}, datetime)
       )
 
     daypart
   end
 
-  defp match({daypart, :open, t_end}, local_time, {stop_id, direction_id}) do
-    t_start = service_start(stop_id, direction_id)
+  defp match({daypart, :open, t_end}, local_time, {stop_id, direction_id}, now) do
+    t_start = service_start(stop_id, direction_id, now)
 
     case t_start do
-      %Time{} -> match({daypart, t_start, t_end}, local_time, nil)
+      %Time{} -> match({daypart, t_start, t_end}, local_time, nil, now)
       nil -> false
     end
   end
 
-  defp match({daypart, t_start, :close}, local_time, {stop_id, direction_id}) do
-    t_end = service_end(stop_id, direction_id)
+  defp match({daypart, t_start, :close}, local_time, {stop_id, direction_id}, now) do
+    t_end = service_end(stop_id, direction_id, now)
 
     case t_end do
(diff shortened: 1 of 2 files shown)


lib/screens/v2/candidate_generator/gl_eink/headways.ex
```diff
@@ -139,8 +139,12 @@ defmodule Screens.V2.CandidateGenerator.GlEink.Headways do
       Time.compare(local_time, t_end) == :lt
   end
 
-  defp service_start_or_end(stop_id, direction_id, min_or_max_fn) do
-    with {:ok, schedules} <- Schedule.fetch(%{stop_ids: [stop_id], direction_id: direction_id}),
+  defp service_start_or_end(stop_id, direction_id, min_or_max_fn, now \\ DateTime.utc_now()) do
```
rwaskiewicz: For my own knowledge, why do we increase the arity of this function? It doesn't look like any of its callers take advantage of `now`. Is the plan to eventually use this argument? Or does it make sense to hardcode `DateTime.utc_now()` in the call to `service_date/1`?

rwaskiewicz (approved):
General question - Do we have an Asana task for removing the hardcoded logic once June 13th/14th has passed?

deanshi:
> General question - Do we have an Asana task for removing the hardcoded logic once June 13th/14th has passed?

Good call, let me create a task for that now.

lib/screens/v2/candidate_generator/gl_eink/headways.ex
deanshi: That's a good point. I think I did it this way literally for style in my eyes. I don't think we'll use it for testing purposes here, so I can move it into the function.

deanshi: Right I remember now - we get a nice little Credo warning for something like this: 

```
To prevent flaky unit tests, a `now` value must be passed as an argument (with a default value if the function is public), not created in the function body.
```

we do have `now` at the top level public function, so let me take a peek to see if there's something I can do to massage this to be a little more clean.

rwaskiewicz: Ah okay, that makes sense - I should have called out this was non-blocking, so if that doesn't shake out, no worries

deanshi: Definitely good call out. I mindlessly just put it in the function since it was a one off hard coded change but this Schedule fetching with the Date portion of it might remain so something I can fix.

PR #3074, opened 2026-05-20
fix: Emergency Takeover should always take precedence over other Evergreen Content

robbie-sundstrom:
**Asana task**: [Bug: Emergency Takeovers should always override Evergreen Content](https://app.asana.com/1/15492006741476/project/1185117109217413/task/1214950231271848?focus=true)

### Summary
Emergency Takeovers should always take precedence over other types of Evergreen Content, even those with priority 0. 

I implemented a new Protocol to solve this, that will default to 1 for all Widget Instances except when defined by setting `@fallback_to_any true`. It's only defined for `EvergreenContent` at the moment to specially return 0 when `is_emergency_takeover` is `true`. This seems like the cleanest way to implement it without working precedence into all existing `WidgetInstance` modules, which absolutely seems like overkill, but I'm 

### Other Approaches Considered
* Reordering the instance functions so that emergency_takeover_instances comes before evergreen_content_instances. This seemed like a brittle solution and could be easily messed up in the future for a new screen type.
* Setting the priority of the emergency takeover to -1 and modifying priority logic so it can accept a negative number. I wanted to avoid adding a new priority type.


robbie-sundstrom:
With this change, I'm also thinking of moving this and a similar protocol, `Screens.V2.AlertsWidget`, into a shared directory named something like `Screens.V2.WidgetProtocols`. I think this organization would be nice, but am holding off on doing until I get feedback on this.

deanshi:
Out of curiosity, what other evergreen content has priority 0? 

In my mind, being a 0 priority is the Emergency priority but maybe I've configured that incorrectly in the past.

robbie-sundstrom:
@deanshi This happened in dev-green on PRE-108, where an updated line map for the world cup is set with priority 0.

deanshi:
I think this change makes sense anyway so no one configures something to be above the Emergency Takeover by accident. Let me take a look at the full PR!

digitalcora:
> In my mind, being a 0 priority is the Emergency priority

I don't think this is really correct to how we use it — it's more a priority level reserved for custom content that we know no "built-in" widgets are going to use. _Most_ custom content we configure is priority `[0]` when we want it to just unconditionally override whatever's normally in the slot.

digitalcora:
@robbie-sundstrom Another, possibly simpler approach to consider: Have `EvergreenContent` implement `priority` by appending an extra `0` to the configured priority value when the emergency content flag is _not_ set. So if the user configures the priority as `[0]`, it is actually `[0, 0]`, and emergency takeovers are the only way to get a "true" `[0]`, which is considered less than (higher priority than) `[0, 0]`. This would eliminate the need for the new protocol, and have minimal interactions with built-in widgets since we never use priority level 0 for them (it would slightly change the behavior of configuring custom content with exactly the same priority as a built-in widget, but we already shouldn't be doing that since the behavior would have been fragile depending on the order of widget generation).

robbie-sundstrom:
@digitalcora I like that suggestion and will look at modifying the implementation to work that way!

robbie-sundstrom:
@digitalcora thanks for the suggestion! This way of doing it is indeed much more straightforward.

lib/screens/v2/candidate_generator/widgets/evergreen.ex
```diff
@@ -49,7 +49,8 @@ defmodule Screens.V2.CandidateGenerator.Widgets.Evergreen do
       schedule: schedule,
       now: now,
       text_for_audio: text_for_audio,
-      audio_priority: audio_priority
+      audio_priority: audio_priority,
+      is_emergency_takeover: false
```
digitalcora: suggestion: this line shouldn't be necessary since the struct field is optional and defaults to `false`.

lib/screens/v2/widget_instance/evergreen_content.ex
```diff
@@ -131,6 +135,8 @@ defmodule Screens.V2.WidgetInstance.EvergreenContent do
   def audio_serialize(%__MODULE__{text_for_audio: text_for_audio}),
     do: %{text_for_audio: text_for_audio}
 
+  # Ensure emergency takeover content is always prioritized above other evergreen content
+  def audio_sort_key(%__MODULE__{is_emergency_takeover: false, audio_priority: [0]}), do: [0, 0]
```
digitalcora: suggestion: I don't think we really need this logic for the audio sort key — by definition, an emergency takeover completely "takes over" the screen, so there should be no other widgets present to be sorted with (at the very least, there should be no other evergreen content instances, which is what this would affect).

test/screens/v2/screen_data/layout_test.exs
```diff
@@ -260,6 +261,39 @@ defmodule Screens.V2.ScreenData.LayoutTest do
 
       assert %{{0, :flex} => %MockWidget{content: "1"}} = actual_instance_placement
     end
+
+    test "prefers emergency takeover evergreen content over configured evergreen content at equal priority" do
```
digitalcora: suggestion: It feels a bit odd for this to be part of `LayoutTest` when it is not actually testing any logic that exists in `Layout` (anymore) — same goes for `ScreenAudioDataTest`. I could see value in a test for the non-trivial behavior of `priority/1` within `WidgetInstance.EvergreenContentTest`.

digitalcora (approved):
Some suggestions, take or leave!

lib/screens/v2/widget_instance/evergreen_content.ex
robbie-sundstrom: true, this shouldn't be necessary for the audio sorting!

test/screens/v2/screen_data/layout_test.exs
robbie-sundstrom: I agree, this is just where the test already was from my original pass. I can move a version of this  EvergreenContentTest

PR #3087, opened 2026-05-28
fix: headway row crash for RDS

deanshi:
**Asana task**: [Bug: RDS DUP Headways Crash](https://app.asana.com/1/15492006741476/project/1185117109217413/task/1215197336744751?focus=true)

Description
- it seems like we were somehow getting into a situation where we were getting schedules for a destination, but had zero departures (with a nil headway range) forcing the state into a Headway state. This should be impossible in cases where have no headsign, due to how we filter out departures for only scheduled ones. This was causing a crash down the line when we tried to serialize it because of the nil headway
- this modifies how `Schedule.fetch` works so we're always passing in the service date when we fetch instead of relying on the backend to choose what service day we're in. This will keep the Departures schedules and the RDS schedules in sync so there should never be a case where we have scheduled departures, but no departures for departures without headways

Note: This makes it so that #3073 can be fully reverted

- [ ] Tests added?

### DIFF
--- a/lib/screens/v2/rds.ex
+++ b/lib/screens/v2/rds.ex
@@ -63,6 +63,7 @@ defmodule Screens.V2.RDS do
           | :active_period
           | :service_impacted
           | :no_service
+          | :error
 
   # These alert types eliminate service to a destination.
   @relevant_alert_effects [
@@ -286,6 +287,7 @@ defmodule Screens.V2.RDS do
           now
         )
       end)
+      |> Enum.reject(&is_nil(&1.state))
 
     {:ok, section_rds}
   end
@@ -418,6 +420,17 @@ defmodule Screens.V2.RDS do
           direction_id: direction_id,
           range: headway_for_stop
         }
+
+      :error ->
+        Logster.warning([
+          "rds_state_creation_failed",
+          destination_key: destination_key,
+          now: now,
+          first_scheduled_departure: first_scheduled_departure.schedule.id,
+          last_scheduled_departure: last_scheduled_departure.schedule.id
+        ])
+
+        nil
     end
   end
 
@@ -573,8 +586,11 @@ defmodule Screens.V2.RDS do
           Util.service_date(now) == Util.service_date(Departure.time(last_departure)) ->
         :after_scheduled_end
 
-      true ->
+      headway_for_stop != nil ->
         :active_period
+
+      true ->
+        :error
     end
   end
 
--- a/lib/screens/schedules/schedule.ex
+++ b/lib/screens/schedules/schedule.ex
@@ -2,6 +2,7 @@ defmodule Screens.Schedules.Schedule do
   @moduledoc false
 
   alias Screens.Trips.Trip
+  alias Screens.Util
   alias Screens.V2.Departure
 
   defstruct id: nil,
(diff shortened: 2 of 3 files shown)


digitalcora:
It seems like if we ever got into a situation where we were generating a `HeadwayRow` but the headway range was `nil` (i.e. the stop does not support headways), there must be something wrong with the logic of `RDS` itself, not just about how the data is fetched. Per the spec, to produce a Headway item, [all destinations in a given "headway group" must be "headway-eligible"](https://www.notion.so/mbta-downtown-crossing/Real-time-destination-states-114f5d8d11ea8077b79cef05225e5725?source=copy_link#326f5d8d11ea80b396d3d123d9bfcb0c), and in order for a destination to be headway-eligible, the stop we are generating destinations for (the specific stopping location the destination is reachable "from") [must have headway values](https://www.notion.so/mbta-downtown-crossing/Real-time-destination-states-114f5d8d11ea8077b79cef05225e5725?source=copy_link#2e7f5d8d11ea80508636dc84cf353792). So if we were following the spec, we should have never produced a Headway item at all if the headway range was `nil`, regardless of the behavior of schedule data fetching. Does this make sense with how the code is currently implemented? Is there still a possible bug here in `RDS`, which fixing the schedule fetching is essentially "covering up" but may not really fix?

deanshi:
> So if we were following the spec, we should have never produced a Headway item at all if the headway range was `nil`, regardless of the behavior of schedule data fetching. Does this make sense with how the code is currently implemented? Is there still a possible bug here in `RDS`, which fixing the schedule fetching is essentially "covering up" but may not really fix?

My thought process here is that since we have scheduled start and end departure, we should have at least created one scheduled Countdown item. 

I spent some time looking at Schedule data from the V3 API for one of the particular days we were seeing a crash (Malden on 5-23/5-24). It seems like in the crash, we were still in the active period before the end of the schedule (`now` being 04:54:48.463682Z while the last scheduled trip seemed to be 2026-05-24T00:56:00-04:00) and so we should have shown the last scheduled trip as a Departure in this particular case but instead we passed through to Headways. I'm going to spend more time tomorrow digging into this, looking in particular at the `Builder` filtering and see if there's something we need to do to our schedules to match that as well.

digitalcora:
> we should have shown the last scheduled trip as a Departure in this particular case but instead we passed through to Headways

To reiterate, I think "passing through to Headways" is incorrect behavior — generating Headways should _not_ be a fallback, it should be conditional on a headway range existing for the relevant stopping location, as per the spec. The fallback should be nothing at all, i.e. not generating any state/item. Having now looked at the actual `RDS` code it is clear there is a problem there where a "service state" of `:active_period` _always_ results in a Headways state and does not check whether a headway range actually exists. I don't discount that there could also be a problem elsewhere.

deanshi:
> Having now looked at the actual `RDS` code it is clear there is a problem there where a "service state" of `:active_period` _always_ results in a Headways state and does not check whether a headway range actually exists.

Got it. Reading the spec, I took the [last bullet point](https://app.notion.com/p/mbta-downtown-crossing/Real-time-destination-states-114f5d8d11ea8077b79cef05225e5725?source=copy_link#325f5d8d11ea807c80bff145e3ae847a) as basically saying that we can safely assume that Headways will always be the last state that we're in based on the order of operations of creating states. 

I'll change that in the code here to return no state/item for this particular value (and maybe add some kind of logging to indicate that we've filtered out a destination?)

digitalcora:
I see. That extra bullet was intended to be descriptive, not prescriptive; i.e. "if the logic has been implemented exactly as described, then we _should_ always have a headway range if we make it to this point". It wasn't really intended as "it's safe to ignore the previously-stated condition on headway values". I'll see if I can rewrite to clarify.

lib/screens/v2/rds.ex
```diff
@@ -418,6 +420,17 @@ defmodule Screens.V2.RDS do
           direction_id: direction_id,
           range: headway_for_stop
         }
+
+      :error ->
+        Logster.warning([
+          "[rds_state_creation_failed]",
```
digitalcora: suggestion: Conventionally we don't put brackets around log messages; I don't think there's really a need to.

lib/screens/v2/rds.ex
```diff
@@ -573,8 +586,11 @@ defmodule Screens.V2.RDS do
           Util.service_date(now) == Util.service_date(Departure.time(last_departure)) ->
         :after_scheduled_end
 
-      true ->
+      headway_for_stop != nil ->
         :active_period
+
+      true ->
+        :error
```
digitalcora, line 593: If feasible, we should add a test corresponding to this branch, and maybe assert that it logs.

digitalcora (approved):

PR #3088, opened 2026-05-28
feat: GL branch handling for DUP alerts

deanshi:
**Asana task**: [Partial Suspensions on GL Trunk DUPs](https://app.asana.com/1/15492006741476/project/1185117109217413/task/1214815711763938?focus=true)

Description
- handle single and multiple branch alerts when displaying the alert banner/full screen alerts for Green Line branches
- this changes it so that we will display the branches that are under the alert instead of a broad Green Line text in cases where we have service for other branches for both the banner text and the full screen text
- also modify the DUP alert logic to check that all green line routes are included in the alert in order to display full screen alerts across all rotations

Screenshot 2026-05-28 at 9 52 03 AM
Screenshot 2026-05-28 at 11 42 27 AM
Screenshot 2026-05-28 at 12 13 16 PM
Screenshot 2026-05-28 at 12 14 49 PM
Screenshot 2026-05-28 at 12 42 22 PM

- [x] Tests added?

### DIFF
--- a/assets/css/dup/free_text.scss
+++ b/assets/css/dup/free_text.scss
@@ -52,7 +52,9 @@
 }
 
 .free-text__route-pill,
-.free-text__text-pill {
+.free-text__text-pill,
+.free-text__branch-route-pill__main-line,
+.free-text__branch-route-pill__branch {
   display: inline-block;
   text-align: center;
 
@@ -103,12 +105,42 @@
 }
 
 .free-text__route-pill__text,
-.free-text__text-pill__text {
+.free-text__text-pill__text,
+.free-text__branch-route-pill {
   font-family: Helvetica, sans-serif;
   font-weight: 700;
 
   &--branch {
     font-size: 84px;
+  }
+}
+
+.free-text__branch-route-pill {
+  display: inline-flex;
+  align-items: center;
+  overflow: hidden;
+  font-size: 96px;
+  line-height: 144px;
+
+  &__main-line,
+  &__branch {
+    box-sizing: border-box;
+    display: flex;
+    align-items: center;
+    justify-content: center;
+    border-radius: 80px;
+  }
+
+  &__main-line {
+    width: 248px;
+    height: 144px;
+  }
+
+  &__branch {
+    min-width: 158px;
+    height: 158px;
+    margin: 0 0 0 -28px;
+    border: 6px solid colors.$cool-black-15;
   }
 }
 
--- a/lib/screens/v2/widget_instance/dup_alert.ex
+++ b/lib/screens/v2/widget_instance/dup_alert.ex
@@ -106,7 +106,8 @@ defmodule Screens.V2.WidgetInstance.DupAlert do
     # some is (either we're at a boundary and there's still service in one direction, or we're at
     # a transfer station and the alert only affects one line).
     if LocalizedAlert.location(t) == :inside and not partial_station_closure?(t) and
-         length(get_affected_lines(t)) == length(primary_sections),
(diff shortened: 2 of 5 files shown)


assets/css/dup/free_text.scss
```diff
@@ -112,6 +115,36 @@
   }
 }
 
+.free-text__branch-route-pill {
```
deanshi, line 118: Note that this code is similar to the Subway Status implementation of branched route pills. We can probably consolidate if we need.

lib/screens/v2/widget_instance/dup_alert/serialize.ex
```diff
@@ -125,6 +132,38 @@ defmodule Screens.V2.WidgetInstance.DupAlert.Serialize do
...
+  defp get_banner_text_for_single_line(
+         %DupAlert{location_context: location_context} = t,
+         "Green" <> _ = line
+       ) do
+    affected_routes = LocalizedAlert.informed_routes_at_home_stop(t)
+
+    all_green_line_routes =
+      LocationContext.route_ids(location_context) |> Enum.filter(&String.contains?(&1, "Green"))
+
+    if affected_routes == all_green_line_routes do
```
robbie-sundstrom: This comparison assumes both lists of routes are in a sorted order. Would that always be the case here?

lib/screens/v2/widget_instance/dup_alert/serialize.ex
```diff
@@ -125,6 +132,38 @@ defmodule Screens.V2.WidgetInstance.DupAlert.Serialize do
...
+      ["No", bold(line), "trains"]
```
robbie-sundstrom: Also pointing out that this logic is exactly the same as the beginning of `get_full_screen_single_line_no_trains_text`, so that might b

lib/screens/v2/widget_instance/dup_alert.ex
```diff
@@ -127,6 +128,17 @@ defmodule Screens.V2.WidgetInstance.DupAlert do
 
   defp eliminated_service_type(_other), do: :none
 
+  defp all_green_line_routes_affected?(lines, %{location_context: location_context} = t) do
```
robbie-sundstrom: Is this supposed to be checking the same logic I commented on in `serialize.ex`? The function name implies yes, but the comparison is not filtering down to only GL routes.

deanshi: Yep, minus the wrap around checking to see if there's a Green Line in the list or not. 

I'll pull this function out and put it into `DupAlerts` for reuse.

robbie-sundstrom: gotcha, that makes sense to me now!

lib/screens/v2/widget_instance/dup_alert/serialize.ex
deanshi: In the process of pulling this out into a function, I converted these into MapSets in order to compare them directly. I also replaced here (and below) with the new function that I created.

robbie-sundstrom (approved):

lib/screens/v2/widget_instance/dup_alert/serialize.ex
digitalcora: If these are MapSets here, I don't think `==` is guaranteed to work with those — it's an opaque struct and it's possible the internal representation could mean (either now or later) that MapSets containing equal elements are not themselves "equal" in the `==` sense. That is why `MapSet.equal?/2` exists, I think.

PR #3090, opened 2026-06-01
fix: fixed crash mixing no service and headways states

deanshi:
Description
- originally, the code was added in aea810d8593fe37ff5bc570a8fb64dace210a248 so that we would show headways for all destinations even when a single one of them was in the NoService state, but this is [causing crashes](https://mbtace.sentry.io/issues/7518132467/?environment=screens-prod&project=6061747&query=is%3Aunresolved&referrer=issue-stream) because of how we parameter match for creating headway sections
- this backs out the extra `||` check so that we only create headway sections when every state is exactly Headway mode. In cases where we have mixed states now, we'll default to showing headway rows with individual headsigns rather than grouped (which makes more sense in representing the data properly)

### DIFF
--- a/test/screens/v2/candidate_generator/dup_new/departures_test.exs
+++ b/test/screens/v2/candidate_generator/dup_new/departures_test.exs
@@ -60,7 +60,7 @@ defmodule Screens.V2.CandidateGenerator.DupNew.DeparturesTest do
       stop: %Stop{id: stop_id},
       line: %Line{id: line_id},
       headsign: headsign,
-      state: %RDS.NoService{routes: routes}
+      state: %RDS.NoService{routes: routes, direction_id: 0}
     }
   end
 
@@ -1169,7 +1169,7 @@ defmodule Screens.V2.CandidateGenerator.DupNew.DeparturesTest do
       assert actual_instances == expected_instances
     end
 
-    test "creates NormalSections but removes headways when similar line/direction_id destinations are not in headway mode" do
+    test "creates NormalSections but removes headways when similar line/direction_id destinations in Countdown state" do
       expected_route_id = "r1"
       expected_time_range = {6, 10}
       expected_direction_name = "Northbound"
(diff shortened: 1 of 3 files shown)


digitalcora:
> In cases where we have mixed states now, we'll default to showing headway rows with individual headsigns rather than grouped

Just to confirm, in general we should not be showing headway rows with specific destination headsigns when there is more than one possible destination for a line+direction — for example we should never show "Braintree every X-Y min" because the X and Y are set according to the expected frequency for _both_ Ashmont and Braintree trains combined ([relevant section of the spec](https://app.notion.com/p/mbta-downtown-crossing/Real-time-destination-states-114f5d8d11ea8077b79cef05225e5725?source=copy_link#326f5d8d11ea80b396d3d123d9bfcb0c)). Was this the case with the existing code and is it still the case with this change? I don't see any changes to tests here, though there is a behavior that's changing, and it seems like one worth testing.

deanshi:
> Was this the case with the existing code and is it still the case with this change?
 
Yep, so we handle headsigns and combining headsigns within `headways_for_rds/3` where we have logic to handle combining headsigns into a direction or using a specific headsign instead. There could be some tests to handle some of these specific cases, so I can add them. 
 
The one question I had, in the case of Ashmont/Braintree, if we were to get into a state where say, `Ashmont` is in a Headways state but the `Braintree` has NoService, should we filter out headways in this particular edge case since not all destinations techniclly are Headway eligible? Can this case ever technically happen?

digitalcora:
> The one question I had, in the case of Ashmont/Braintree, if we were to get into a state where say, `Ashmont` is in a Headways state but the `Braintree` has NoService, should we filter out headways in this particular edge case since not all destinations techniclly are Headway eligible? Can this case ever technically happen?

I would expect the specific case you mention would never happen, because if a subway route ever doesn't have scheduled service, it will be because there's a disruption, and there would be an alert for that. However, we _should_ be strictly following the logic that all destinations in a "headway group" _must_ be headway-eligible for a headway message to be generated. Ideally this is just enforced in one place, when we decide whether to generate a headway item; we shouldn't have to think about edge cases, it should just be straightforwardly impossible to violate the rule.

digitalcora:
@robbie-sundstrom I'd appreciate if you could take a look at this, since you're more familiar with the code being modified here?

test/screens/v2/candidate_generator/dup_new/departures_test.exs
```diff
@@ -1243,5 +1243,96 @@ defmodule Screens.V2.CandidateGenerator.DupNew.DeparturesTest do
 
       assert actual_instances == expected_instances
     end
+
+    test "creates NormalSections but removes headways when similar line/direction_id destinations are in Service Ended state" do
```
robbie-sundstrom: Should this test also include destinations in the `NoService` state since we want the same behavior as `ServiceEnded` states being removed?

deanshi: So sorry, this is a typo actually. If you look at L1299, we actually do use the `NoService` state, not the Service Ended state. Let me change that.

robbie-sundstrom (approved):

PR #3091, opened 2026-06-02
dev: auto-tag releases in DUP client build script

digitalcora:
Update the build script to, by default, automatically generate and push a release tag as per the below task. The generated version number is also used (once again) as the version string embedded in the package instead of the Git short hash, since with the addition of the tagging, this should point back to a specific commit.

...I might have had a little fun with the formatting and emojis. ❌ ⚠️ ℹ️ ✨ 

**Asana task**: https://app.asana.com/1/15492006741476/project/1185117109217413/task/1214901633382581?focus=true

### DIFF
--- a/assets/src/components/dup/README.md
+++ b/assets/src/components/dup/README.md
@@ -3,19 +3,19 @@
 - Double check that any behavior specific to the packaged environment happens inside of an `isOutfront()` condition.
 - If you've renamed / removed image assets, you might want to delete the corresponding folder in `/priv/static`. The folder accumulates assets without clearing old ones out, and these will be included in the built bundle!
 - **Only if you are packaging for local testing:** To test against your local screens backend instead of production, then replace the value of `data-api-origin` in [`priv/dup-app.html`][dup-app.html] with `http://localhost:4000`.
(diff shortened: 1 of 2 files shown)


robbie-sundstrom (approved):
🏷️ This looks really great to me!!!!! Huge improvement to the process. 

I'm wondering if we should default to having `is_tagged = false` instead such that you specify a flag when you do want to create a tag. I don't think this matters too much though since we'll be able to identify the proper releases even if there are extras in there.

digitalcora:
> I'm wondering if we should default to having `is_tagged = false` instead such that you specify a flag when you do want to create a tag.

I did want to reduce the likelihood of forgetting to ever create the tag — it feels like the most common flow is going to be that you create a package, it (hopefully) works fine in the lab as-is, and you subsequently deploy that package. In that case you want to have created the tag already, instead of having to go back and re-run the script (in which the actual Webpack build would be pointless, you'd just be running it for the tagging "side effect").

That said, I think I'm talking myself around to making the release tag opt-in, especially since it also pushes to origin. When run without the option, the output can make it clear what you're doing and what the alternative is (i.e. "Not tagging a release, use `-t` for that").

PR #3092, opened 2026-06-03
fix: fixed DUP cross origin access within Screenplay

deanshi:
**Asana task**: [Screenplay bug: DUP previews don't display](https://app.asana.com/1/15492006741476/project/1185117109217413/task/1215329063361089?focus=true)

Description
- trying to access `parent?.parent?.mraid` was causing cross origin access problems within Screenplay
- this just wraps the access in a try catch, resulting in MRAID being false and forgoes some of the metadata, header titles and extras that we normally add to the Screens page (which I think might be fine for screenplay)

### DIFF
--- a/assets/src/util/outfront.tsx
+++ b/assets/src/util/outfront.tsx
@@ -92,7 +92,13 @@ interface OFMTag {
 }
 
 export const getMRAID = (): MRAID | false => {
-  return (parent?.parent as OFMWindow)?.mraid ?? false;
+  // We wrap this in a try catch because Screenplay shows
+  // Screens via an iframe and tries to get this value for DUPs
+  try {
+    return (parent?.parent as OFMWindow)?.mraid ?? false;
+  } catch {
+    return false;
+  }
 };
 
 /**


assets/src/util/outfront.tsx
```diff
@@ -92,7 +92,11 @@ interface OFMTag {
 }
 
 export const getMRAID = (): MRAID | false => {
-  return (parent?.parent as OFMWindow)?.mraid ?? false;
+  try {
```
robbie-sundstrom, line 97: I would maybe just add a comment explaining that this logic is here specifically for Screenplay

robbie-sundstrom (approved):

assets/src/util/outfront.tsx
deanshi: Just noting that previously we were using
```
export const isDup = () => /.*dup-app.*/.test(location.href);
```
which was forcing the `getMRAID` to return false. Without this, we would have had the same issue with the old version.

PR #3097, opened 2026-06-04
dev: remove unneeded `concurrently` option

digitalcora:
`--prefix-colors auto` is the default in `concurrently` v10 (version bumped in 2ce5678f).


rwaskiewicz (approved):
Ah, missed this while reviewing the version bump. Good catch!

PR #3098, opened 2026-06-04
feat: ship RDS departures on DUPs

digitalcora:
**Asana task**: https://app.asana.com/1/15492006741476/project/1185117109217413/task/1212463750135892?focus=true

See commit messages for details.


lib/screens/v2/rds.ex
```diff
@@ -220,7 +220,9 @@ defmodule Screens.V2.RDS do
         fetch_relevant_alerts(stop_ids)
       end,
       departures: fn params, _now ->
-        params |> Map.from_struct() |> @departure.fetch(now: now)
+        params
+        |> Map.from_struct()
+        |> @departure.fetch(include_scheduled_cancelled?: true, now: now)
```
deanshi: With this change, I was seeing some crashes on dev-green caused by `nil` times (both `arrival_time` and `departure_time` being `nil` and being passed all the way to the front end). I saw on our old DUP code we handled these departures by going through `Widgets.Departures.fetch_section_departures/3` where we include `schedule_route_type_filter: [:ferry, :rail]`. 

Should this also include the `schedule_route_type_filter` by default? And with `[:ferry, :rail, :bus]`?

digitalcora: Good catch — could you point me to those crashes? I don't see any error logs on dev-green over the past 24 hours in Splunk, and nothing is coming up for me in Sentry either. I agree we should be providing a value for this filter in any case, because it defaults to all route types, and we don't normally want schedules for subway.

digitalcora: I'm also curious about how this could happen:

> both `arrival_time` and `departure_time` being `nil` and being passed all the way to the front end

...because I _think_ the `Builder` should have filtered them out if that were the case.

deanshi: > Good catch — could you point me to those crashes?

I was just actually just poking around on dev-green since it was deployed and came across it at Back Bay and other DUP stations quite randomly. I'm not sure if it actually was sent to Sentry since we serialize everything properly but then crash in Typescript when we try to show `DepartureTimes`. Do we have Sentry set up for that? 

> I'm also curious about how this could happen...

That's what I thought as well. Going through the `relevant_departures/3` flow it seems like we reject cancelled unscheduled trips by default, but these seem to be scheduled trips that were cancelled and because we're passing `include_scheduled_cancelled?: true` we include them in our departures list (bypassing `maybe_reject_scheduled_cancelled`).

digitalcora: > Do we have Sentry set up for that?

Frontend crashes _should_ be reported to Sentry, and we see them somewhat frequently (mostly fetch failures from transient network conditions). If you could capture one of these crashes that would be helpful, I'll also poke around for a bit locally.

> these seem to be scheduled trips that were cancelled

In that case, we shouldn't be sending a `nil` time to the client, we should be sending the originally scheduled time so it knows to show a struck-through presentation. This worked on non-RDS so I'll look into why it might not work here.

deanshi: > In that case, we shouldn't be sending a `nil` time to the client, we should be sending the originally scheduled time so it knows to show a struck-through presentation.

For that piece, I was looking at `serialize_time/3` in `Screens.V2.WidgetInstance.Departures` and there we have hard coded logic to only include scheduled times for rail and ferry (although I didn't dive too deep into this portion of the code so it might be somewhere else).

Appreciate the help looking into this!

digitalcora: Interesting! We might be able to remove that, not sure.

lib/screens/v2/rds.ex
robbie-sundstrom, line 172: Not your change, but something that's occurred to me when doing an initial pass of this PR. In this case statement, we'll just return an `:error` if we don't have any `routes` for a section. If this is the case, I think we want to log a warning if we had a way to know that the request was coming from a real screen, since this indicates that a section is misconfigured. Just thinking out loud in this comment as I'm trying to think if there's a way to easily track when a section configuration ends up with no valid routes at all.

digitalcora: I agree it seems relevant to monitor if we ever hit this case and will add a warning report, since I don't expect it would happen often. We can re-evaluate if it turns out to be noisy. (I think it may not even be possible to encounter unless a section is configured _only_ with stops that have no service at all)

deanshi (approved):
Also tested by poking around on screen-dev-green and everything is looking good to me.

@robbie-sundstrom maybe since you've done an initial pass, if there are any other comments you wanted to add so I'm not just overlooking what you might have seen - but otherwise we can 🚢 🚢 🚢

robbie-sundstrom:
@deanshi nothing else I really noticed yesterday that needed addressing before getting this out! With your testing, I'm ready to give it a 👍 🚢  as well

robbie-sundstrom (approved):
mostly looked through the refactoring commit in depth, but I really like the improvements!

digitalcora:
Thanks very much @deanshi @robbie-sundstrom, in addition to the reporting you suggested I've pushed two more smaller refactor commits. Would appreciate eyes on those if you have time.

lib/screens/v2/departure.ex
```diff
@@ -63,15 +63,19 @@ defmodule Screens.V2.Departure do
...
-    all_types = RouteType.all()
+    all_types = RouteType.all() |> Enum.sort()
     opt_route_types = Keyword.get(opts, :schedule_route_type_filter, all_types)
     param_route_types = List.wrap(params[:route_type] || all_types)
 
-    # An empty list here, which we'd intend as "no route types", would encode as an empty string
-    # in the fetch params, which means "no route type *filter*" a.k.a. all route types. Fetching
-    # schedules for "no route types" is just not fetching any schedules at all, so we can skip it.
-    case Enum.filter(opt_route_types, &(&1 in param_route_types)) do
+    case opt_route_types |> Enum.filter(&(&1 in param_route_types)) |> Enum.sort() do
```
robbie-sundstrom: Nitpick suggestion for cleaning this up so it's a bit more understandable: 
```suggestion
    all_types = RouteType.all() |> Enum.sort()
    param_route_types = List.wrap(Map.get(params, :route_type, all_types))
    opt_route_types =
      opts
      |> Keyword.get(:schedule_route_type_filter, all_types)
      |> Enum.filter(&(&1 in param_route_types))
      |> Enum.sort()

    # An empty list here, which we'd intend as "no route types", would encode as an empty string
    # in the fetch params, which means "no route type *filter*" a.k.a. all route types. Fetching
    # schedules for "no route types" is just not fetching any schedules at all, so we can skip it.
    case opt_route_types do
```

robbie-sundstrom:
@digitalcora the commits you've added look good to me! One small readability comment

lib/screens/v2/departure.ex
digitalcora: I think I managed to improve this in a slightly different way using MapSets.

PR #3099, opened 2026-06-04
fix: Add takeover styles for Sectional solos

robbie-sundstrom:
These styles are needed on Sectional solos for full screen takeovers to be the correct width and height, otherwise, the width and height is set to that of two panels.


digitalcora (approved):
Thanks for the find! These CSS updates for Sectional Duos got a little tangled at one point and I must have lost track of this piece.

PR #3100, opened 2026-06-05
Script to get destination list for all screen configs

robbie-sundstrom:
**Asana task**: [Enumerate route destinations for screen configs](https://app.asana.com/1/15492006741476/project/1185117109217413/task/1214711732390209?focus=true)

See the Asana task or the CSV I accidentally added in my first commit for what this data looks like after running the script.

Could definitely be cleaned up a bit, but it's just a script that I don't see us running much more often.


scripts/screen_destinations.exs
```diff
@@ -0,0 +1,212 @@
...
+            case filtered_screens do
+              [] -> true
+              filtered_screens -> screen_id in filtered_screens
+            end
+          end)
+          |> Enum.map(fn {screen_id, screen_data} ->
+            {screen_id, screen_data["name"], screen_data["app_params"]["departures"]["sections"]}
+          end)
+          |> Enum.filter(fn {_screen_id, _name, sections} -> sections != nil end)
+          |> Enum.map(fn {screen_id, name, sections} ->
```
deanshi, line 54: super nit kind of like you said in your comment: I don't know if it's intentional in terms of how the filtering/map is done, but it does seem like it could be collapsed. That being said, this is how my brain works in terms of mapping/filtering step by step, so I'd be okay with keeping it since it's a one off script.

scripts/screen_destinations.exs
```diff
@@ -0,0 +1,212 @@
+# Script to get all of the destinations we could show on a given Screen
+# Uses the local.json file to find each relevant screen's departure query params,
```
deanshi, line 2: Should this pull from the production S3 (or be given the option?) instead of using the `local.json` file? I'm just thinking if there's an off chance we accidentally use a non-prod `local.json` for the script.

scripts/screen_destinations.csv
```diff
@@ -0,0 +1,122 @@
+Screen ID,Name,Destination Count,Destinations
```
deanshi: I know it's removed but - 

Is there a sort order on this? Just curious if there's an easy way for us to see which ones have the most destinations off the bat or something like that, or have it sorted alphabetically or by screen type or something.

scripts/screen_destinations.csv
```diff
@@ -0,0 +1,122 @@
+Screen ID,Name,Destination Count,Destinations
+TPS-103,,8,"Bowdoin, Wonderland, Forest Hills, Oak Grove, Riverside, Union Square, Heath Street, Medford/Tufts"
+TPS-129,N/A,10,"Beth Israel (Limited Stops), Andrew (Limited Stops), Kenmore via South Bay Center, Kenmore, Central Square, Assembly, Harbor Point via South Bay Center, Fields Corner, Broadway, Ruggles"
```
deanshi: I see that some of the screens just don't have a name associated or the name is just `N/A` with them (but mostly TPS screens). Do we need to worry about that at all? It feels like it might be okay but wanted to ask just in case.

deanshi (commented):
I think looks good, but just a few clarifying questions from me!

scripts/screen_destinations.csv
robbie-sundstrom: I set it up to sort by screen type, and design didn't have a preference for how this is sorted. Should be easy to sort after the fact in Excel.

scripts/screen_destinations.csv
robbie-sundstrom: This CSV was generated from a WIP script, so for some reason it didn't catch all of the missing names. The finalized script gets all of the screen's without names and properly labels them as having N/A name

scripts/screen_destinations.exs
robbie-sundstrom: That's a good suggestion! I think I can make that the default and have it optionally pull from the local config

scripts/screen_destinations.exs
robbie-sundstrom: This could totally be condensed! This is definitely a case of "non-production code" YAGNI, where i just didn't want to sink time into optimizing this code. This could probably be updated pretty quickly though so I might while I'm implementing the config changes you mentioned

scripts/screen_destinations.exs
robbie-sundstrom: I was having some issues with getting the S3 permissions to work within this Elixir script, and I don't want to sink too much time into this. I added some clarifying comments to the instructions for running this script if you think that will be effective enough!

deanshi: Yep, I think that's reasonable enough. Script runners beware. 🚢

deanshi (approved):

PR #3101, opened 2026-06-09
Exclude special case routes in connections script

robbie-sundstrom:
**Asana task**: [Run connection change script for new rating](https://app.asana.com/1/15492006741476/project/1185117109217413/task/1215226273421683?focus=true)

This was just a small improvement I made while running the connections script for June's rating change to automatically remove routes that we don't want to display to riders.

### DIFF
--- a/scripts/connecting_routes.exs
+++ b/scripts/connecting_routes.exs
@@ -35,16 +35,9 @@ route_patterns = [
 ]
 
 defmodule RouteConnections do
-  # Routes that we want to omit from the list of connections to show to riders
+  # Routes that we want to omit from the list of connections to show to riders (eg SLW & combined routes)
   # See the Notion for On-Train Screen Rating Update for details
-  @excluded_routes %{
-    "place-sstat": ["746"],
-    "place-alfcl": ["627"],
-    "place-forhl": ["4050"],
-    "place-haecl": ["8993"],
-    "place-sull": ["8993"],
-    "place-balsq": ["8993"]
-  }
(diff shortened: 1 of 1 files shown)


rwaskiewicz (approved):

scripts/connecting_routes.exs
```diff
@@ -35,6 +35,17 @@ route_patterns = [
 ]
 
 defmodule RouteConnections do
+  # Routes that we want to omit from the list of connections to show to riders
+  # See the Notion for On-Train Screen Rating Update for details
+  @excluded_routes %{
+    "place-sstat": ["746"],
+    "place-alfcl": ["627"],
+    "place-forhl": ["4050"],
```
digitalcora: Routes `627` and `4050` no longer exist; we don't need to suppress them.

scripts/connecting_routes.exs
```diff
@@ -35,6 +35,17 @@ route_patterns = [
...
+    "place-haecl": ["8993"],
+    "place-sull": ["8993"],
+    "place-balsq": ["8993"]
```
digitalcora: I think when we don't display a route as a connection, we want that to be universal (i.e. if the SLW is never shown then it's never shown, the answer doesn't depend on where you are). It might be less fragile to just have a list of route IDs to not output.

robbie-sundstrom: I think this is likely true and would make it easier to maintain in the future!

PR #3102, opened 2026-06-09
fix: Full closure at Symphony showing as partial in reconstructed_alert

robbie-sundstrom:
**Asana task**: [Bug: "skipping n platforms" alert message](https://app.asana.com/1/15492006741476/project/1185117109217413/task/1215512201269422?focus=true)

For the reconstructed alert, we're not filtering down to just Subway platforms before checking if it's a partial closure.

image

### DIFF
--- a/lib/screens/v2/candidate_generator/widgets/reconstructed_alert.ex
+++ b/lib/screens/v2/candidate_generator/widgets/reconstructed_alert.ex
@@ -163,7 +163,8 @@ defmodule Screens.V2.CandidateGenerator.Widgets.ReconstructedAlert do
     # Given informed entities representing an alert at a single station,
     # finds the corresponding platform names for those child stops included.
     with [informed_parent_station] <- Alert.informed_parent_stations(alert),
-         [_ | _] = child_platforms <- informed_parent_station.stop.child_stops,
+         [_ | _] = child_platforms <-
+           informed_parent_station.stop.child_stops |> Stop.filter_subway_platforms(),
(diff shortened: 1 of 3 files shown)


lib/screens/v2/candidate_generator/widgets/reconstructed_alert.ex
```diff
@@ -163,10 +163,11 @@ defmodule Screens.V2.CandidateGenerator.Widgets.ReconstructedAlert do
     # Given informed entities representing an alert at a single station,
     # finds the corresponding platform names for those child stops included.
     with [informed_parent_station] <- Alert.informed_parent_stations(alert),
-         [_ | _] = child_platforms <- informed_parent_station.stop.child_stops,
+         [_ | _] = child_platforms <-
```
rwaskiewicz, line 166: Is there a test that we can write for this change that covers filtering to subway/light rail platforms here before determining the closure type?

rwaskiewicz (approved):

lib/screens/v2/candidate_generator/widgets/reconstructed_alert.ex
robbie-sundstrom: yes, but I wanted to get this fix in quickly since this alert is active. Could add a quick unit test for this case later

robbie-sundstrom: Added modifications to the tests here: https://github.com/mbta/screens/pull/3103

PR #3103, opened 2026-06-10
test: Add non-platform stop to parent station in reconstructed alert test

robbie-sundstrom:
Follow up to a fix made in [this PR](https://github.com/mbta/screens/pull/3102). 

Simply adding a non-platform stop to the parent station that's used in these tests would have caused assertion failures before the fix above. This addition to the test validates that the node stop is filtered out properly when building the alert widget and seeing if all the platform stops are closed.

### DIFF
--- a/test/screens/v2/candidate_generator/widgets/reconstructed_alert_test.exs
+++ b/test/screens/v2/candidate_generator/widgets/reconstructed_alert_test.exs
@@ -5,6 +5,7 @@ defmodule Screens.V2.CandidateGenerator.Widgets.ReconstructedAlertTest do
 
   alias Screens.Alerts.Alert
   alias Screens.LocationContext
+  alias Screens.Stops.Stop
   alias Screens.V2.WidgetInstance.ReconstructedAlert, as: ReconstructedAlertWidget
   alias ScreensConfig.Screen
   alias ScreensConfig.Screen.{Busway, PreFare}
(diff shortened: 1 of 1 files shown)


test/screens/v2/candidate_generator/widgets/reconstructed_alert_test.exs
```diff
@@ -79,11 +79,17 @@ defmodule Screens.V2.CandidateGenerator.Widgets.ReconstructedAlertTest do
         vehicle_type: :subway
       }
 
+      malden_center_non_platform = %Stop{
```
rwaskiewicz, line 82: Out of curiosity, what would a non-platform stop be? Would that be a street-level stop? Or maybe a traffic stopping point to let other trains get on/off the track?

rwaskiewicz (approved):
Thanks for the fast follow on this!

test/screens/v2/candidate_generator/widgets/reconstructed_alert_test.exs
robbie-sundstrom: There are 4 different `location_type` values a stop can have. 0 indicates a distinct boarding location, 1 is the parent complex, 2 designates a station entrance/exit, and 3 indicates a generic node within a station, such as the end of a staircase, elevator, or escalator.

digitalcora: Yeah, the term "stop" in GTFS and thus in the V3 API is overloaded to represent things that are not "stops" at all (in the sense of "place where transit vehicles stop") but in some way relate to stops and have a similar set of properties.

PR #3109, opened 2026-06-11
fix: urgent alerts suppress subway status widget

rwaskiewicz:
**Asana task**: [Investigate: Flex zone alerts interaction with Subway Status](https://app.asana.com/1/15492006741476/project/1213833239363217/task/1211838338299124?focus=true)

Description

Suppress the subway widget on prefare screens when an urgent alert is shown. Previously, alert urgency was computed when the alert was serialized for the client. This commit moves that calculation to widget generation time, so that it may be used for page grouping to exclude the subway status widget.

Urgency calculations have been hoisted from
`lib/screens/v2/widget_instance/reconstructed_alert.ex#flex_zone_fields/2`, using the function clauses and guards to create
`lib/screens/v2/candidate_generator/widgets/reconstructed_alert.ex/urgent?/2`. The function clauses in the latter are meant to align with those in the former to capture the existing logic.

- [x] Tests added?


lib/screens/v2/candidate_generator/widgets/reconstructed_alert.ex
```diff
@@ -106,6 +112,7 @@ defmodule Screens.V2.CandidateGenerator.Widgets.ReconstructedAlert do
               informed_station_names: get_station_names(alert),
               is_terminal_station: is_terminal_station,
               is_priority: is_priority,
+              is_urgent: urgent?(alert, location, all_platforms_names_at_informed_station),
```
digitalcora: I'm not sure this is the right level for this to live at — `alert`, `location`, and `partial_closure_platform_names` are all values already stored in the struct, so from the widget module's perspective we're "telling it something it already knows". This could be a private function on the widget module instead, which the widget protocol can then call with the widget struct to determine `page_groups`.

lib/screens/v2/candidate_generator/widgets/reconstructed_alert.ex
```diff
@@ -285,4 +292,29 @@ defmodule Screens.V2.CandidateGenerator.Widgets.ReconstructedAlert do
...
+    severity >= 7
+  end
+
+  defp urgent?(%Alert{effect: effect}, location, _partial_closure_platform_names)
+       when location in @inside_locations and effect in ~w[shuttle suspension]a do
+    true
+  end
+
+  defp urgent?(%Alert{effect: :station_closure}, location, partial_closure_platform_names)
+       when partial_closure_platform_names != [],
```
digitalcora: This specific part of the condition feels redundant — if we were inside a station closure, and it was _not_ a partial closure, the alert would be a full-screen takeover and there would be no relevance to the flex zone anyway (but if we hypothetically did show such an alert in the flex zone, surely it would qualify as urgent!).

lib/screens/v2/candidate_generator/widgets/reconstructed_alert.ex
```diff
@@ -285,4 +292,29 @@ defmodule Screens.V2.CandidateGenerator.Widgets.ReconstructedAlert do
...
+       when severity <= 1,
+       do: false
+
+  defp urgent?(
+         %Alert{effect: effect, severity: severity},
+         location,
+         _partial_closure_platform_names
+       )
+       when location in @inside_locations and effect == :delay do
+    severity >= 7
```
digitalcora: It seems inconsistent in these clauses whether we use guards or conditions in the body of the function, and indeed whether we use `==` guards or direct value matching; I think picking an approach and sticking with it to the extent possible could make this more readable.

Surveying the conditions, I think we can do this much more simply, in a single clause:

```elixir
eliminates_service? = effect in ~w[shuttle suspension station_closure]a
severity > 1 and location in @inside_locations and (eliminates_service? or severity >= 7)
```

This applies my suggestion from L316 but otherwise should be logically equivalent.

lib/screens/v2/widget_instance/reconstructed_alert.ex
```diff
@@ -1379,9 +1387,10 @@ defmodule Screens.V2.WidgetInstance.ReconstructedAlert do
   end
 
   def page_groups(%__MODULE__{} = t) do
-    case placement(t) do
-      :single_screen -> [:alerts]
-      _ -> []
+    case {placement(t), t.is_urgent} do
+      {:single_screen, _} -> [:alerts]
+      {:flex_zone, true} -> [:alerts]
```
digitalcora: * suggestion: Maybe we could name this something like `urgent_alerts` to make it clear that it only applies in that circumstance.
* As written, this will cause urgent alerts to also exclude flex-zone departures, since those already have a page group defined to make them exclusive with subway status. Is that what we want? The task doesn't say, probably because at the time it was written, flex-zone departures didn't exist yet.

lib/screens/v2/candidate_generator/widgets/reconstructed_alert.ex
rwaskiewicz: Good point - I think I had taken the word "lift" a little too literally in the ticket. I did a mostly mechanical change in cf89e77444ff339bba6745114a73286238754701 to move a good deal of this logic to the widget instance module, remove the stateful vars, and a small refactor to cut down the arity of `urgent?/3` to 2. The plan/goal is to address the rest of your comments atop that one

lib/screens/v2/candidate_generator/widgets/reconstructed_alert.ex
rwaskiewicz: Happy to change this around, I'm just second guessing my current understanding of things 😅 I don't fully understand the example you gave as it relates to the guard. Wouldn't this be checking for a partial station closure if the effect is `:station_closure` and there's >0 platform names associated with a partial closure, rather than a full station closure?

Maybe there's a gap between my understanding of station closures and the code I've written here.... Below are my current understandings/how I arrived at keeping this as it is, LMK if/where I took a wrong turn.

I assumed that `:station_closure` does not always imply a full station closure - I had originally hoisted this particular clause from one of the original implementations of [`serialize_flex_zone/2`](https://github.com/mbta/screens/blob/9a1a98cb6fa6edad01c55681f383f397815e808b/lib/screens/v2/widget_instance/reconstructed_alert.ex#L1086), where I _think_, this combination of matching on the Alert effect and using a guard on `partial_closure_platform_names` was how we expressed partial station closures at the time.

Related: It looks like since then, we've introduced the `:partial_closure` and `:partial_closure_multiple_stops` in `alert.ex` in [95e537d](https://github.com/mbta/screens/commit/95e537d6ca1e776e9f2e105c3fe7109c4c55a780). However, I didn't see them in the `effect` type in `Alert.ex` ([link](https://github.com/mbta/screens/blob/0ae298e66d85c1fc43c613fca57a9064a4ba73f2/lib/screens/alerts/alert.ex#L85-L115)), but perhaps it should be?

lib/screens/v2/candidate_generator/widgets/reconstructed_alert.ex
rwaskiewicz: > Surveying the conditions, I think we can do this much more simply, in a single clause:
  ```elixir
  eliminates_service? = effect in ~w[shuttle suspension station_closure]a
  severity > 1 and location in @inside_locations and (eliminates_service? or severity >= 7)
  ``` 
In the example you gave above, `:delay` is elided in evaluating `eliminates_service?` - that makes sense to me, because a delay doesn't eliminate service. But it doesn't seem to be directly used in the expression below it. Is there something about `:delay`s as a concept or how they interact with the other checks in that second expression that makes evaluating the effect being `:delay` redundant?

> It seems inconsistent in these clauses whether we use guards or conditions in the body of the function, and indeed whether we use == guards or direct value matching; I think picking an approach and sticking with it to the extent possible could make this more readable.

For my own understanding, what is it about picking one or the other makes something more readable in the context of Elixir? I'm still working on building the intuition of the "readability" aspect of Elixir, and would have thought, perhaps incorrectly, that this was readable in the sense that (all of these are opinions):
- The guard denotes whether we should fall in to a function implementation
- The function implementation is disjoint from the guard. That is, in this particular case, the implementation shouldn't be hoisted into the guard because we want either a true/false value from that expression 
    - Alternatively this particular clause could have written this as 2 function clauses, one where each guard evaluates `when ... and severity >= 7, do: true` and `when ... and severity < 7, do: false`, but that had felt less maintainable to me due to two places to maintain the relational check of `severity`

lib/screens/v2/widget_instance/reconstructed_alert.ex
rwaskiewicz: The former is done in 77fd5647dc9a34a385bddb1643a4214ab8ffc57a. The latter....I'm not sure. I'll look into that now/next

rwaskiewicz: > As written, this will cause urgent alerts to also exclude flex-zone departures, since those already have a page group defined to make them exclusive with subway status.

If we wanted to included flex-zone departures, what would the remedy for this be, adding an additional page group to the `Screens.V2.WidgetInstance.Departures#page_groups/1` and `Screens.V2.WidgetInstance.ReconstructedAler#page_groups/1` implementations?

rwaskiewicz: Spoke with Betsy about this:

> I think for now the alert should push the departures off. I will note that this decision may be revisited once departures include alerts in the widget itself.

lib/screens/v2/candidate_generator/widgets/reconstructed_alert.ex
digitalcora: > Wouldn't this be checking for a partial station closure if the effect is `:station_closure` and there's >0 platform names associated with a partial closure, rather than a full station closure?

Correct, that is the effect of this code as you've written it. What I'm saying is that it doesn't seem to make sense conceptually if we treat _partial_ closures as urgent but _full_ closures as non-urgent. My expectation is that the reason the existing logic was this way is because _full_ closures never get a flex-zone presentation at all (they're always a full-screen takeover), so it didn't really matter whether we considered them "urgent", a property that only applies to the flex-zone presentation.

> However, I didn't see them in the `effect` type in `Alert.ex`

They should not be. These are just return values of the `station_closure_type` function.

lib/screens/v2/widget_instance/reconstructed_alert.ex
digitalcora: Cool, then we don't have to do anything else.

lib/screens/v2/candidate_generator/widgets/reconstructed_alert.ex
digitalcora: > In the example you gave above, `:delay` is elided in evaluating `eliminates_service?` - that makes sense to me, because a delay doesn't eliminate service. But it doesn't seem to be directly used in the expression below it. Is there something about `:delay`s as a concept or how they interact with the other checks in that second expression that makes evaluating the effect being `:delay` redundant?

Shuttle, Suspension, Station Closure, and Delay are currently the only effects we create Reconstructed Alerts for. But that aside, I'm not sure why we'd necessarily want to consider delays specially. If we did, and fell through to "not urgent", it would have the same behavior as not checking at all, both currently and if we added new alert types in the future.

Also, I'm noticing we already have this `eliminates_service?` value as a module attribute `@service_eliminating_effects`, because it was already being used in determining relevance/the priority field.

lib/screens/v2/candidate_generator/widgets/reconstructed_alert.ex
rwaskiewicz: Ah okay, that makes sense, Thanks!

lib/screens/v2/candidate_generator/widgets/reconstructed_alert.ex
rwaskiewicz: Ah okay, thanks for the clarification. Lots of new to me topics in this task, appreciate the patience/guidance!

lib/screens/v2/widget_instance/reconstructed_alert.ex
```diff
@@ -26,6 +26,9 @@ defmodule Screens.V2.WidgetInstance.ReconstructedAlert do
   alias ScreensConfig.Departures.Query.Params
   alias ScreensConfig.Screen.PreFare
 
+  @inside_locations ~w[inside boundary_upstream boundary_downstream]a
+  @service_eliminating_effects ~w[shuttle station_closure suspension]a
```
digitalcora: a thought: since this is now duplicated between here and the candidate generator, and several other places in the codebase have similar logic, maybe it's worth extracting to an `Alert.effect_eliminates_service?`. Not essential to this PR.

test/screens/v2/candidate_generator/widgets/reconstructed_alert_test.exs
```diff
@@ -759,5 +763,188 @@ defmodule Screens.V2.CandidateGenerator.Widgets.ReconstructedAlertTest do
                  fetch_location_context_fn
                )
     end
+
+    test "low severity alerts do not create urgent widget instances", context do
```
digitalcora: These tests now seem misplaced — they are not testing any new behavior of this module.

test/screens/v2/widget_instance/reconstructed_alert_test.exs
```diff
@@ -3485,4 +3485,42 @@ defmodule Screens.V2.WidgetInstance.ReconstructedAlertTest do
...
+        |> put_informed_entities([
+          ie(route: "Red", route_type: 1, stop_id: "place-asmnl")
+        ])
+        |> put_effect(:delay)
+        |> put_severity(7)
+
+      assert [:urgent_alerts] == ReconstructedAlert.page_groups(widget)
+    end
+
+    test "returns no page_groups for a non-urgent alert", %{widget: widget} do
```
digitalcora: suggested clarification: `a non-urgent flex zone alert`

test/screens/v2/candidate_generator/widgets/reconstructed_alert_test.exs
rwaskiewicz: Good call - I forgot to remove those in https://github.com/mbta/screens/pull/3109/commits/cf89e77444ff339bba6745114a73286238754701

test/screens/v2/widget_instance/reconstructed_alert_test.exs
rwaskiewicz: 👍 Done in a7ef5e931ed4d8d0a7488dd807516f7e0d846b45

lib/screens/v2/widget_instance/reconstructed_alert.ex
rwaskiewicz: I like that idea. Is there a "good" way to share this logic between a function and a custom guard? ATM, I have something like this:

```diff
diff --git a/lib/screens/alerts/alert.ex b/lib/screens/alerts/alert.ex
index 5b68b8ee..8bfb812e 100644
--- a/lib/screens/alerts/alert.ex
+++ b/lib/screens/alerts/alert.ex
@@ -150,6 +150,8 @@ defmodule Screens.Alerts.Alert do
   @base_includes ~w[facilities stops.child_stops]
   @all_includes ~w[facilities.stop.child_stops facilities.stop.parent_station.child_stops stops]
 
+  @service_eliminating_effects ~w[shuttle station_closure suspension]a
+
   @callback fetch(options()) :: result()
   def fetch(opts \\ [], get_json_fn \\ &V3Api.get_json/2) do
     includes =
@@ -503,4 +505,11 @@ defmodule Screens.Alerts.Alert do
   end
 
   defp whole_route_delay?(_), do: false
+
+  defguard is_service_eliminating_effect(effect) when effect in @service_eliminating_effects
+
+  @spec effect_eliminates_service?(effect()) :: boolean()
+  def effect_eliminates_service?(effect) when is_service_eliminating_effect(effect), do: true
+
+  def effect_eliminates_service?(_effect), do: false
 end
diff --git a/lib/screens/v2/candidate_generator/widgets/reconstructed_alert.ex b/lib/screens/v2/candidate_generator/widgets/reconstructed_alert.ex
index 3cd5b9aa..14645b62 100644
--- a/lib/screens/v2/candidate_generator/widgets/reconstructed_alert.ex
+++ b/lib/screens/v2/candidate_generator/widgets/reconstructed_alert.ex
@@ -1,6 +1,8 @@
 defmodule Screens.V2.CandidateGenerator.Widgets.ReconstructedAlert do
   @moduledoc false
 
+  import Screens.Alerts.Alert
+
   alias Screens.Alerts.Alert
   alias Screens.Alerts.InformedEntity
   alias Screens.LocationContext
@@ -117,7 +119,6 @@ defmodule Screens.V2.CandidateGenerator.Widgets.ReconstructedAlert do
   end
 
   @inside_locations ~w[inside boundary_upstream boundary_downstream]a
-  @service_eliminating_effects ~w[shuttle station_closure suspension]a
 
   # Filter out `elsewhere` alerts (should never happen).
   defp relevance(_alert, :elsewhere, _distance), do: nil
@@ -125,14 +126,14 @@ defmodule Screens.V2.CandidateGenerator.Widgets.ReconstructedAlert do
   # "Immediate disruptions": Service is eliminated in at least one direction at the home stop.
   # Riders may need to take immediate action to continue their trip.
   defp relevance(%Alert{effect: effect}, location, _distance)
-       when effect in @service_eliminating_effects and location in @inside_locations,
+       when is_service_eliminating_effect(effect) and location in @inside_locations,
        do: {0, nil}
 
   # "Downstream disruptions": Service is eliminated starting somewhere downstream of the home
   # stop. Riders may need to take action later to continue their trip. Split into sub-categories
   # based on how close to the home stop the disruption begins (only the closest get "priority").
   defp relevance(%Alert{effect: effect}, _location, distance)
-       when effect in @service_eliminating_effects,
+       when is_service_eliminating_effect(effect),
        do: {1, distance}
 
   # Severe delays are also considered "downstream disruptions".
diff --git a/lib/screens/v2/widget_instance/reconstructed_alert.ex b/lib/screens/v2/widget_instance/reconstructed_alert.ex
index 3b6020af..2f67bba4 100644
--- a/lib/screens/v2/widget_instance/reconstructed_alert.ex
+++ b/lib/screens/v2/widget_instance/reconstructed_alert.ex
@@ -27,7 +27,6 @@ defmodule Screens.V2.WidgetInstance.ReconstructedAlert do
   alias ScreensConfig.Screen.PreFare
 
   @inside_locations ~w[inside boundary_upstream boundary_downstream]a
-  @service_eliminating_effects ~w[shuttle station_closure suspension]a
 
   defstruct screen: nil,
             alert: nil,
@@ -1410,8 +1409,8 @@ defmodule Screens.V2.WidgetInstance.ReconstructedAlert do
 
   @spec urgent?(t(), LocalizedAlert.location()) :: boolean()
   defp urgent?(%__MODULE__{alert: %Alert{effect: effect, severity: severity}}, location) do
-    eliminates_service? = effect in @service_eliminating_effects
-    severity > 1 and location in @inside_locations and (eliminates_service? or severity >= 7)
+    severity > 1 and location in @inside_locations and
+      (Alert.effect_eliminates_service?(effect) or severity >= 7)
   end
 
   def widget_type(%__MODULE__{} = t) do

```
which feels....incorrect. I'll take a look at it with fresh eyes on Monday, but if anything here looks immediately obvious/remediable, LMK!

digitalcora: > Is there a "good" way to share this logic between a function and a custom guard?

There's no need to share at all — guards can be used in function bodies (technically I believe guards are macros).

When using `import` I'd recommend:

```ex
import Screens.Alerts.Alert, only: [is_service_eliminating_effect: 1]
```

Otherwise _everything_ importable from the module gets quietly dumped into the importing module's scope, and it gets harder to work out where a given function (or macro in this case) is coming from.

rwaskiewicz: Sounds good! Done in https://github.com/mbta/screens/pull/3109/commits/706fb060ff7a1a1294cd0fc96ab2a644da9f5427 (and fixed up a dialyzer error in https://github.com/mbta/screens/pull/3109/commits/0fc173576bf89b049f5c77c8ebff9c90fe7a22b6)

lib/screens/alerts/alert.ex
```diff
@@ -503,4 +505,6 @@ defmodule Screens.Alerts.Alert do
   end
 
   defp whole_route_delay?(_), do: false
+
+  defguard is_service_eliminating_effect(effect) when effect in @service_eliminating_effects
```
digitalcora: nit: I'm not sure how strong a convention this really is but I'd "expect" to find guard definitions near the top of a module, before any function definitions (supporting evidence: this is true of all existing modules in this codebase that have both guard and function definitions). It might be intuitive to place it right under the definition of `@service_eliminating_effects`, since it is the only user of that attribute.

digitalcora (approved):

lib/screens/alerts/alert.ex
rwaskiewicz: 👍 done in 732a0eed

PR #3112, opened 2026-06-12
fix: RDS showing cancelled subway departures

digitalcora:
This was a gap in the spec: we specified that stops with headways shouldn't present "scheduled-only" departures, but not that they also shouldn't present skipped/cancelled departures. This resulted in DUPs showing struck-through clock times for skipped subway departures, instead of not displaying them at all, as non-RDS departures did.


robbie-sundstrom (approved):
nice fix!

PR #3116, opened 2026-06-16
feat: Always prioritize station closures over other service disrupting alerts

robbie-sundstrom:
**Asana task**: [Prioritize station closure alert over other alerts ](https://app.asana.com/1/15492006741476/project/1185117109217413/task/1215344534320213?focus=true)

We had a case recently when a station was closed, but there was also a shuttle alert affecting it since it was inside the shuttle boundary. We should prioritize displaying the station closure alert with a higher priority in this case.

### DIFF
--- a/lib/screens/v2/candidate_generator/widgets/reconstructed_alert.ex
+++ b/lib/screens/v2/candidate_generator/widgets/reconstructed_alert.ex
@@ -123,37 +123,42 @@ defmodule Screens.V2.CandidateGenerator.Widgets.ReconstructedAlert do
   # Filter out `elsewhere` alerts (should never happen).
   defp relevance(_alert, :elsewhere, _distance), do: nil
 
+  # "Immediate eliminated service": Service is completely eliminated at the rider's station.
+  defp relevance(%Alert{effect: effect}, location, _distance)
+       when effect == :station_closure and location in @inside_locations,
+       do: {0, nil}
+
   # "Immediate disruptions": Service is eliminated in at least one direction at the home stop.
   # Riders may need to take immediate action to continue their trip.
   defp relevance(%Alert{effect: effect}, location, _distance)
        when is_service_eliminating_effect(effect) and location in @inside_locations,
-       do: {0, nil}
+       do: {1, nil}
 
   # "Downstream disruptions": Service is eliminated starting somewhere downstream of the home
   # stop. Riders may need to take action later to continue their trip. Split into sub-categories
   # based on how close to the home stop the disruption begins (only the closest get "priority").
   defp relevance(%Alert{effect: effect}, _location, distance)
        when is_service_eliminating_effect(effect),
-       do: {1, distance}
+       do: {2, distance}
 
   # Severe delays are also considered "downstream disruptions".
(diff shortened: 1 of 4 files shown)


digitalcora:
The reported issue was on a GL E-ink screen, which doesn't use Reconstructed Alerts.

robbie-sundstrom:
@digitalcora thanks for catching. Still think it makes sense to add this update here in `reconstructed_alert`, but should also be modified in the GL E-ink alert widget

lib/screens/v2/widget_instance/alert.ex
```diff
@@ -82,7 +82,7 @@ defmodule Screens.V2.WidgetInstance.Alert do
 
     cond do
       Enum.any?(tiebreakers, &(&1 == :no_render)) -> :no_render
-      takeover_alert?(t) -> @automated_override_priority
+      takeover_alert?(t) -> @automated_override_priority ++ tiebreakers
```
rwaskiewicz, line 85: For my own understanding, why do we need to concatenate `tiebreakers` here? I commented this concatenation out to see if that would break a test to help (with said understanding), but everything seems to have continued to pass.

rwaskiewicz (approved):
Left one question for my own understanding, O/W looks good to me. I don't have a strong opinion ATM for/against changing the reconstructed alert behavior. I don't _think_ this would push a more relevant alert off in some way/form, but I'm still learning ™️ 🤓

lib/screens/v2/widget_instance/alert.ex
robbie-sundstrom: If we have two alerts with `takeover_alert` true, they will default to `automated_override_priority` and it will be a matter of ordering returned by the API that determines which one is returned. By concatenating `tiebreakers`, they keep the same base priority numbers that help them override other widgets that could go in the same slot, but then have numbers appended to the end of that priority list that allow the takeover alerts to be ranked against each other

rwaskiewicz: Makes sense, thanks!

lib/screens/v2/candidate_generator/widgets/reconstructed_alert.ex
```diff
@@ -123,37 +123,42 @@ defmodule Screens.V2.CandidateGenerator.Widgets.ReconstructedAlert do
   # Filter out `elsewhere` alerts (should never happen).
   defp relevance(_alert, :elsewhere, _distance), do: nil
 
+  # "Immediate eliminated service": Service is completely eliminated at the rider's station.
```
digitalcora: note: The existing terms in quotes here are quoted because they are references to the [Pre-Fare Alert Widget Specification](https://app.notion.com/p/mbta-downtown-crossing/Pre-fare-Alert-v2-Widget-Specification-17cf5d8d11ea8035a708e302208582f5#17cf5d8d11ea81d4bad9c38c41c713ae). "Immediate eliminated service" in quotes makes it sound like it is one of those categories, but really it's kind of an exception/refinement to the "immediate disruptions" category.

robbie-sundstrom: Thanks for pointing this out, clarified in the comment explaining this logic

PR #3117, opened 2026-06-16
feat: RDS data fetches serial, make each section create parallel

deanshi:
- in a Slack conversation, it was discussed that the creation of each section in RDS should be parallel, while the live data fetches (and other cached network calls) could be possibly made serial to simplify the the code
- this makes it so that we call each of the data fetches in serial, with most of the calls hitting cache
- it also moves the `async_stream` to create each RDS section in parallel, setting us up for when we eventually need to create many sections on LCD screens

---
  - https://app.asana.com/0/0/1215780122518154


robbie-sundstrom (approved):
Based off of the Slack conversation, this makes sense to me! Let's obviously just keep an eye on prod when this is deployed to see how this effects things

PR #3122, opened 2026-06-18
fix: incorrect alert filter in RDS logic

digitalcora:
Alert fetching accepts either a `stop_id` option which must be a single stop ID, or a `stop_ids` option which must be a list. RDS logic fetched alerts using `stop_id`, but passing a list. Due to a "fallback" clause in query param building, this filter was quietly dropped. As a result, the part of the RDS logic that checked whether there was an active alert impacting a stop would instead check whether there were *any* active alerts that eliminated service, across the entire system.

* Use the correct spelling `stop_ids` in the RDS logic.

* Eliminate the "fallback" clause in alert fetch option parsing. If we make a mistake here, we probably want to crash and be noisy about it rather than carrying on.

### DIFF
--- a/lib/screens/alerts/alert.ex
+++ b/lib/screens/alerts/alert.ex
@@ -254,8 +254,6 @@ defmodule Screens.Alerts.Alert do
     ]
   end
 
-  defp format_query_param(_), do: []
-
   def happening_now?(%{active_period: aps}, now \\ DateTime.utc_now()) do
     Enum.any?(aps, &in_active_period(&1, now))
   end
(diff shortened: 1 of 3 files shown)


robbie-sundstrom (approved):
This does seem like a weird default case to have ever had for the `stop_ids` query param here

digitalcora:
It's not really a default for the `stop_ids` parameter — you could throw *anything* in the fetch options argument, say `foo: 123`, and it would be quietly rejected. Needless to say I think that is a bad default behavior.

PR #3127, opened 2026-06-22
perf: only fetch stop when needed in DUP alerts

digitalcora:
The DUP Alert widget generator would always make an API call to fetch the name of the screen's stop (if it was configured for alerts and had a `StopId` header), even if no alert widgets were ultimately generated or if the fetched name went unused due to a special case. Move this fetching to be done "just in time" to save a request in the common case where there are no alerts displayed.

### DIFF
--- a/lib/screens/v2/candidate_generator/dup/alerts.ex
+++ b/lib/screens/v2/candidate_generator/dup/alerts.ex
@@ -11,6 +11,7 @@ defmodule Screens.V2.CandidateGenerator.Dup.Alerts do
   alias Screens.V2.LocalizedAlert
   alias Screens.V2.WidgetInstance.{DupAlert, DupSpecialCaseAlert}
   alias ScreensConfig.Alerts, as: AlertsConfig
+  alias ScreensConfig.Header.{StopId, StopName}
   alias ScreensConfig.Screen
   alias ScreensConfig.Screen.Dup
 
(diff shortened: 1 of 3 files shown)


lib/screens/v2/candidate_generator/dup/alerts.ex
```diff
@@ -107,6 +96,16 @@ defmodule Screens.V2.CandidateGenerator.Dup.Alerts do
     end
   end
 
+  defp fetch_stop_name(%Screen{app_params: %Dup{header: %StopName{stop_name: name}}}), do: name
+
+  defp fetch_stop_name(%Screen{app_params: %Dup{header: %StopId{stop_id: id}}}) do
+    case @stop.fetch_stop_name(id) do
+      # Failure to fetch here, though unlikely, shouldn't block the widget from being displayed
+      nil -> "Alert"
```
robbie-sundstrom, line 104: I was thinking this "Alert" might not be the best header, but then I checked this behavior as it currently exists and see that returning an empty list just breaks the widget generation instead. So this is definitely an improvement.

robbie-sundstrom (approved):
nice little improvement that I'm sure adds up!!

lib/screens/v2/candidate_generator/dup/alerts.ex
digitalcora: Yeah it was definitely a bug that this returned `[]`, and `nil` would violate the client typing that this value is always a string. I think the fact that we've never once seen this crash in reality (as far as we know, modulo Sentry rate limits) is representative of how much we need to worry about the exact fallback here.

robbie-sundstrom: exactly, never seeing this failure made me care much less about the header we'd pass here

PR #3128, opened 2026-06-22
refactor: remove include option on `Stop` fetch

digitalcora:
Reducing stop fetching from two possible sets of includes to one improves the potential for cache reuse, in particular between RDS and `StopId` headers in the common case where the screen shows departures filtered by the same parent station as the one used for the header. This will cause higher cache memory use for screens that don't otherwise use Stop fetching, but we currently have memory to spare.


robbie-sundstrom (approved):
makes sense to me that this would improve cache reuse!

PR #3130, opened 2026-06-22
refactor: use "now" consistently for alert display

digitalcora:
The RDS and Elevator Status generators did not pass their "now" value to `Alert.happening_now?`. This would mostly not have any impact on screens currently, but we should be consistent about this, given we'd like to eventually add a "time travel" preview function for upcoming alerts.

Fix these and also:

* make this function require a datetime instead of being optional, so we can't forget to do this in the future
* rename it to better align with this (that there is no assumed "now"; the datetime you pass in could be anything, it might not be "now")

### DIFF
--- a/lib/screens/alerts/alert.ex
+++ b/lib/screens/alerts/alert.ex
@@ -256,9 +256,8 @@ defmodule Screens.Alerts.Alert do
     ]
   end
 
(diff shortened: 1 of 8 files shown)


robbie-sundstrom (approved):
I definitely like the name change to `Alerts.active?`

PR #3131, opened 2026-06-22
feat: Process BE headsign abbreviations for DUP departures

robbie-sundstrom:

To move all screens onto a new shared abbreviation system, I’m planning to do this overall change in three parts to avoid breaking DUP application behavior in production. 
1. Update DUP FE to interpret both BE abbreviations and old system. Once this is merged, deploy the package to DUPs
2. Pass headsign abbreviations from the BE
3. Remove old system from the DUP FE and deploy a final package to DUPs. 

This PR is part 1, and has been tested with part 2, available as a draft PR shortly

See below for an example of the full integrated abbreviation system working on DUPs with more complex headsigns:

https://github.com/user-attachments/assets/4eba9fe8-93c1-438d-9b3c-1d6ced1bde4a

### DIFF
--- a/assets/src/components/departures/destination.tsx
+++ b/assets/src/components/departures/destination.tsx
@@ -7,6 +7,7 @@ const abbreviate = (headsign: string): string => {
 
 type Destination = {
   headsign: string;
+  abbreviations?: string[];
   variation?: string;
 };
 
--- a/assets/src/components/dup/departures/destination.tsx
+++ b/assets/src/components/dup/departures/destination.tsx
@@ -56,6 +56,7 @@ const RenderedDestination = ({ parts, index1, index2, classModifier }) => {
 
 const Destination: ComponentType<DupDestination> = ({
   headsign,
+  abbreviations,
   classModifier,
 }) => {
   const firstLineRef = useRef<HTMLDivElement>(null);
@@ -65,11 +66,16 @@ const Destination: ComponentType<DupDestination> = ({
 
   const [index1, setIndex1] = useState(parts.length);
   const [index2, setIndex2] = useState(parts.length);
-  const [abbreviate, setAbbreviate] = useState(false);
+  const [abbreviationIndex, setAbbreviationIndex] = useState(-1);
   const [phase, setPhase] = useState(PHASES.ONE_LINE_FULL);
 
-  if (abbreviate) {
-    parts = parts.map((p) => ABBREVIATIONS[p] || p);
+  if (abbreviationIndex >= 0) {
+    // abbreviationIndex
+    if (abbreviations && abbreviations.length > abbreviationIndex) {
+      parts = abbreviations[abbreviationIndex].split(" ");
+    } else {
+      parts = parts.map((part) => ABBREVIATIONS[part] || part);
+    }
   }
 
   /* eslint-disable react-hooks/set-state-in-effect --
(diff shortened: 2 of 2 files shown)


assets/src/components/dup/departures/destination.tsx
```diff
@@ -65,11 +66,16 @@ const Destination: ComponentType<DupDestination> = ({
...
   const [index1, setIndex1] = useState(parts.length);
   const [index2, setIndex2] = useState(parts.length);
-  const [abbreviate, setAbbreviate] = useState(false);
+  const [abbreviationIndex, setAbbreviationIndex] = useState(-1);
   const [phase, setPhase] = useState(PHASES.ONE_LINE_FULL);
 
-  if (abbreviate) {
-    parts = parts.map((p) => ABBREVIATIONS[p] || p);
+  if (abbreviationIndex >= 0) {
+    // abbreviationIndex
```
rwaskiewicz, line 73: Nit - unneeded comment?

assets/src/components/dup/departures/destination.tsx
```diff
@@ -102,8 +109,11 @@ const Destination: ComponentType<DupDestination> = ({
           if (firstLineFits) {
             setPhase(PHASES.DONE);
           } else {
-            setAbbreviate(false);
-            setPhase(PHASES.TWO_LINES_FULL);
+            setAbbreviationIndex(abbreviationIndex + 1);
+            if (!abbreviations || abbreviations.length <= abbreviationIndex) {
+              setPhase(PHASES.TWO_LINES_FULL);
+              setAbbreviationIndex(-1);
+            }
```
rwaskiewicz, line 116: Is it possible for us to not enter the body of this `if` statement? IIUC, we never transition to the `DONE` phase if this falls through, but I'm not sure that's possible here

rwaskiewicz (approved):

assets/src/components/departures/destination.tsx
```diff
@@ -7,6 +7,7 @@ const abbreviate = (headsign: string): string => {
 
 type Destination = {
   headsign: string;
+  abbreviations?: string[];
```
digitalcora, line 10: Not really necessary right at the moment, but I can easily envision a world where we want to apply abbreviations to destination `variation`s as well (which are currently not broken out on DUPs, though arguably supporting them could make it easier to apply Betsy's proposed page-breaking rules). Currently we just wrap the `variation` to additional lines if it's too long and don't bother applying any abbreviations to it, but this then takes up vertical space that could be used for more departures.

One way to split the difference here might be to just call this field `headsigns`; then if we introduce variation abbreviations, we can call that field `variations`. (It would probably also be convenient for the implementation/package deploys if each of those fields included the completely un-abbreviated string as the first element; old behavior entirely contained in the old field, new behavior entirely contained in the new.)

digitalcora:
> 1. Update DUP FE to interpret both BE abbreviations and old system. Once this is merged, deploy the package to DUPs
> 2. Pass headsign abbreviations from the BE
> 3. Remove old system from the DUP FE and deploy a final package to DUPs.

This sequence requires two DUP package deploys, but if we "turned it inside-out", it would only require one:

1. Update server to start producing new data
2. Update client to start using the new data (requires DUP package deploy)
3. Update server to remove anything that's now unused

assets/src/components/dup/departures/destination.tsx
robbie-sundstrom: We would only enter this specific `if` statement if there are no additional abbreviations to try. If we don't enter this or move to `PHASES.DONE`, then we have just moved onto a shorter abbreviation to try with the same phase. 

Note that I think this could be less confusing, but requiring it to work with both the old abbreviation system and the new one so we can deploy it to DUPs means that I couldn't change the logic around as much.

robbie-sundstrom:
I decided to flip around the ordering of these PRs as per @digitalcora's suggestion. See the new base PR for this work here: https://github.com/mbta/screens/pull/3156

PR #3133, opened 2026-06-23
feat: suppress alerts five weeks and older

rwaskiewicz:
**Asana task**: [Suppress Stale Alerts on Screens](https://app.asana.com/1/15492006741476/project/1185117109217413/task/1214784349728677?focus=true)

Description

Suppress alerts shown for five weeks or more. Alerts are generally not expected to be this long-lived. However, there are cases where a closure may last for an extended period of time (e.g. Symphony is to close for approximately 3 years). In these scenarios, we wish to suppress alerts to prevent alert fatigue.

- [x] Tests added?

Testing: As luck would have it, Alert 1002417 in AM Staging is 5 weeks and 1 day old. We can see the difference in output when looking at `main` (`a443ae454fc70d23fd24379d6e69645ada055629`) in the first screenshot, against this PR in the second

Screenshot 2026-06-23 at 12 01 21

Screenshot 2026-06-23 at 12 01 35

### DIFF
--- a/lib/screens/alerts/alert.ex
+++ b/lib/screens/alerts/alert.ex
@@ -137,6 +137,7 @@ defmodule Screens.Alerts.Alert do
           activities: [activity()] | :all,
           ids: [id()],
           include_all?: boolean(),
+          now: DateTime.t(),
           route_id: Route.id(),
           route_ids: [Route.id()],
           route_types: RouteType.t() | [RouteType.t()],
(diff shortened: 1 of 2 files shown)


rwaskiewicz:
Closing, as the blanket approach won't work as originally intended. There are long-ish closures on bus lines that will fit this suppression criteria. As a result, einks will display service is normal, when the stations are in fact, still closed. We'll revisit the more nuanced approach as a team in the near future

PR #3143, opened 2026-06-25
fix: update first departure text styling

rwaskiewicz:
**Asana task**: [RDS DUP Polish: First Trip State](https://app.asana.com/1/15492006741476/project/1185117109217413/task/1215451542447463)

Description

Update the departure text "First" styling updating the margin to the right of the text per design's request.

The vertical alignment of the text has been centered with respect to its parent container, effectively vertically centering the text in the row. This option has been selected as opposed to aligning the bottom of the text with the two letter line abbreviation text (e.g. "RL" for "Red Line"). The latter would require a more significant refactor of the DUPs departure row component, as the current structure & CSS rely heavily on `display: inline-block`, making the positioning of "First" to align with "RL" difficult to do without affecting other positioning in its parent.

Screenshots

(Note: I had to do some hardcoding on the backend/frontend to get these to show, so times will be nonsensical relative to when first trips actually begin)

Screenshot 2026-06-25 at 13 04 13
Screenshot 2026-06-25 at 13 04 17
Screenshot 2026-06-25 at 13 04 39
Screenshot 2026-06-25 at 13 05 32
Screenshot 2026-06-25 at 13 06 33

### DIFF
--- a/assets/css/dup/departures/row.scss
+++ b/assets/css/dup/departures/row.scss
@@ -59,6 +59,7 @@
 
 .departure-row__first {
   display: inline-block;
-  margin-right: 32px;
+  margin-right: 0.5ex;
   font-size: 108px;
+  vertical-align: center;
 }


digitalcora (approved):
> (Note: I had to do some hardcoding on the backend/frontend to get these to show, so times will be nonsensical relative to when first trips actually begin)

I assume this is also the reason the times are showing as "live" (with the icon), since this otherwise would never happen and indeed looks a bit weird with the positioning.

rwaskiewicz:
@digitalcora 

> I assume this is also the reason the times are showing as "live" (with the icon), since this otherwise would never happen and indeed looks a bit weird with the positioning.

Yes, that's exactly why. I just hadn't cut that out for the purpose of these screenshots 😅

PR #3145, opened 2026-06-25
chore: remove dependabot groups

rwaskiewicz:
**Asana task**: N/A
Description

Remove package-specific dependency groupings from the Dependabot configuration. Prior to this commit, we were seeing cases where NPM major version updates were falling into the `npm-minor` group. This caused related packages that were peer-dependent on one another's new major version to be split into separate PRs, causing each created PR to fail. The outcome following this PR is the same,
but will require manual intervention from the dev team to fix cases of broken major version updates.

One approach in https://github.com/mbta/screens/pull/3142 was to move `npm-minor` to the bottom of the group list, and explicitly exclude each package in any group from `npm-minor`. This was tested to fix the above issue in a separate repo, but adds maintenance burden. As a result, we choose to simply remove the groups as they exist today

- [ ] Tests added?

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -18,6 +18,11 @@ updates:
     cooldown:
       default-days: 7
     open-pull-requests-limit: 5
+    groups:
+      mix-minor:
+        update-types:
+          - "patch"
+          - "minor"
   - package-ecosystem: npm
     directory: "/assets"
     schedule:
@@ -31,3 +36,8 @@ updates:
       # with the actual Node version used by the project.
       - dependency-name: "@types/node"
         update-types: ["version-update:semver-major"]
+    groups:
+      npm-minor:
+        update-types:
+          - "patch"
+          - "minor"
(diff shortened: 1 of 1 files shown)


digitalcora:
To clarify my previous comment, I meant that we could remove all groups *other* than the `mix-minor` and `npm-minor`, which I think are fairly essential to avoid a flood of individual PRs for tiny updates. I assumed the problem here was being caused by both the "minor" groups and the "named" groups existing *together*, rather than by any groups existing at all. I think this is supported by our experience using just "minor" groups on other repos, but I don't know if you've seen a counter-example.

rwaskiewicz:
Ah, okay - I see what you're saying. Fixed in https://github.com/mbta/screens/pull/3145/commits/698550360df79a4843aea0953a9c0c0b258acd31 (and PR desc updated)

digitalcora (approved):
RIP, my beautiful groups...

rwaskiewicz:
Dependabot wasn't worthy. Renovatebot on the other hand, maybe some day 🌤️

PR #3150, opened 2026-06-26
feat: RDS MVP on LCDs

deanshi:
**Asana task**: [Implement RDS MVP on LCDs](https://app.asana.com/1/15492006741476/project/1185117109217413/task/1214589221132185?focus=true)

Description
Adds the `RealtimeDepartures` candidate generator that uses the previously pulled out shared code and filters sections that we don't support and filters out rows that we don't support within `NormalSections`. 

Depending on if we have a valid `NormalSection` during generation, we either display the `DeparturesNoData` widget, or a `DeparturesWidget` with sections that have `NormalSections`, or a `NormalSection` that displays a `FreeTextLine` with a `No departures currently available` message (adjusted from the sections we don't support). 

It uses the current `no_departures_message()` in the old `Screens.V2.CandidateGenerator.Widgets.Departures`. There should currently be no UI changes from the current pre-RDS impelementation. 

One small difference in this logic is that in the old `Departures` code, we would try and fetch an associated route every time we have no data hitting the route endpoint. In the new code, we depend on the different types of Sections to pass the route and create a `no_departures_message` based on that route, and so I slightly modified HeadwaySections to include the `route` instead of just the `route_id`. 

- [x] Tests added?

### DIFF
--- a/lib/screens/v2/candidate_generator/widgets/realtime_departures.ex
+++ b/lib/screens/v2/candidate_generator/widgets/realtime_departures.ex
@@ -1,0 +1,297 @@
+defmodule Screens.V2.CandidateGenerator.Widgets.RealtimeDepartures do
+  @moduledoc """
+  Candidate generator for LCD RDS Items
+  Takes in the generated sections from Screens.V2.CandidateGenerator.Widgets.RdsDepartures and
+  handles the roll-up and creation of the actual widget that will be serialized and used on
+  the screen itself. 
+  """
+
+  alias Screens.Routes.Route
+  alias Screens.Schedules.Schedule
+  alias Screens.Trips.Trip
+  alias Screens.V2.CandidateGenerator.Widgets.RdsDepartures
+  alias Screens.V2.Departure
+  alias Screens.V2.RDS
+  alias Screens.V2.WidgetInstance.Departures, as: DeparturesWidget
+
+  alias Screens.V2.WidgetInstance.Departures.{
+    HeadwaySection,
+    NoDataSection,
+    NormalSection,
+    NoServiceSection,
+    OvernightSection
+  }
+
+  alias Screens.V2.WidgetInstance.DeparturesNoData
+  alias ScreensConfig.{Departures, FreeTextLine, Screen}
+  alias ScreensConfig.Departures.{Filters, Query, Section}
+  alias ScreensConfig.Departures.Filters.{RouteDirections, RouteDirections.RouteDirection}
+  alias ScreensConfig.Screen.{Busway, PreFare}
+
+  import Screens.Inject
+  @rds injected(RDS)
+
+  @type widget :: DeparturesNoData.t() | DeparturesWidget.t()
+
+  @spec departures_instances(Screen.t(), DateTime.t()) :: [widget()]
+  def departures_instances(%Screen{app_params: app_params} = screen, now) do
+    app_params
+    |> departures_slots()
+    |> Enum.with_index()
+    |> Enum.flat_map(fn {{departures, slots}, index} ->
+      generate_instances(departures, slots, index, screen, now)
+    end)
+  end
+
+  @spec generate_instances(Departures.t(), [atom()], non_neg_integer(), Screen.t(), DateTime.t()) ::
+          [widget()]
+  defp generate_instances(departures, _slot_names, _order, _screen, _now)
+       when is_nil(departures) or departures.sections == [],
+       do: []
+
+  defp generate_instances(
+         %Departures{sections: sections} = departures,
+         slot_names,
+         order,
+         screen,
+         now
+       ) do
+    sections_data =
+      departures
+      |> @rds.get(now)
+      |> RdsDepartures.create_departure_sections(departures, &post_process_rows/4, now)
+
+    [create_departures_instance(sections_data, sections, screen, slot_names, order, now)]
+  end
+
+  @spec create_departures_instance(
+          [DeparturesWidget.section()],
+          Section.t(),
+          Screen.t(),
+          [atom()],
+          non_neg_integer(),
+          DateTime.t()
+        ) :: widget()
+  defp create_departures_instance(
+         sections_data,
+         sections,
+         screen,
+         slot_names,
+         order,
+         now
+       ) do
+    sections_data_with_sections_config = Enum.zip(sections_data, sections)
+
+    # As we begin to support other rows/sections/widgets, add them in here
+    if has_valid_normal_section?(sections_data_with_sections_config) do
+      %DeparturesWidget{
+        screen: screen,
+        sections:
+          sections_data_with_sections_config
+          |> Enum.map(fn
+            {%NormalSection{rows: rows} = normal_section, _section} ->
+              %{
+                normal_section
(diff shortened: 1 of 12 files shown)


lib/screens/v2/candidate_generator/widgets/realtime_departures.ex
```diff
@@ -0,0 +1,269 @@
+defmodule Screens.V2.CandidateGenerator.Widgets.RealtimeDepartures do
+  @moduledoc false
+
```
robbie-sundstrom: Can we add a moduledoc to this and `rds_departures` to explain why they're different and the purpose of each one? I can't think of a better naming scheme, so it would be helpful to have at the top of `rds_departures` that it's a home for shared code between rds candidate generators, and that this is a generator specifically for non-DUPs and shouldn't include shared code.

lib/screens/v2/candidate_generator/widgets/realtime_departures.ex
```diff
@@ -0,0 +1,269 @@
...
+      app_params
+      |> departures_slots()
+      |> Enum.with_index()
+      |> Enum.flat_map(fn {{departures, slots}, index} ->
+        generate_instances(departures, slots, index, screen, now)
+      end)
+    end
+  end
+
+  defp generate_instances(departures, _slot_names, _order, _screen, _now)
```
robbie-sundstrom, line 48: Would be great to add a spec for `generate_instances`

lib/screens/v2/candidate_generator/widgets/realtime_departures.ex
```diff
@@ -0,0 +1,269 @@
...
+       ) do
+    sections_data =
+      departures
+      |> @rds.get(now)
+      |> RdsDepartures.create_departure_sections(departures, &post_process_rows/4, now)
+
+    [create_departures_instance(sections_data, sections, screen, slot_names, order, now)]
+  end
+
+  defp create_departures_instance(
```
robbie-sundstrom, line 75: As with above, a spec for `create_departures_instance`

lib/screens/v2/candidate_generator/widgets/realtime_departures.ex
```diff
@@ -0,0 +1,269 @@
...
+
+  defp filter_rows(
+         rows,
+         %Filters{max_minutes: max_minutes, route_directions: route_directions},
+         now
+       ) do
+    rows
+    |> filter_unsupported_rows()
+    |> filter_by_time(max_minutes, now)
+    |> filter_by_route_direction(route_directions)
```
robbie-sundstrom, line 223: It looks like both `filter_by_time` and `filter_by_route_direction` will fail for the case when `max_minutes` or `route_directions` are respectively set, since they don't handle the `{%Schedule{}, :first_trip}` data shape. A good case for specs to see the potential types that could be passed in 😉

test/screens/v2/candidate_generator/widgets/realtime_departures_test.exs
```diff
@@ -0,0 +1,447 @@
...
+               "stop_id_two",
+               "line_id_two",
+               "test_headsign_two",
+               build_schedule("route_two", "line_id_two")
+             ),
+             headways(
+               "stop_id_three",
+               "line_id_three",
+               "test_headsign_three",
+               "route_id_three",
```
robbie-sundstrom, line 193: This should not be a string based on the way that the helpers function creates a `Headways`. Should either convert that to a route here or within the function

test/screens/v2/candidate_generator/widgets/realtime_departures_test.exs
```diff
@@ -0,0 +1,447 @@
...
+             countdowns("stop_id_one", "line_id_one", "headsign", all_departures)
+           ]}
+        ]
+      end)
+
+      assert [%DeparturesWidget{sections: [%NormalSection{rows: ^included_departures}]}] =
+               RealtimeDepartures.departures_instances(config, @now)
+    end
+
+    test "post process filters departures with included route-directions" do
```
robbie-sundstrom, line 364: Seems like we should also be testing the opposite `:exclude` action, but don't see a test for that

lib/screens/v2/candidate_generator/widgets/realtime_departures.ex
```diff
@@ -0,0 +1,269 @@
...
+                normal_section
+                | rows:
+                    Enum.map(rows, fn
+                      %Departure{} = departure -> departure
+                      # Treat First Trips as Countdowns while we don't support them yet
+                      {%Schedule{} = schedule, :first_trip} -> %Departure{schedule: schedule}
+                    end)
+              }
+
+            {unsupported_section, section} ->
```
robbie-sundstrom, line 103: This is where we'll build other section types in future tasks correct?  Might just be worth adding a comment to that effect.

lib/screens/v2/candidate_generator/widgets/realtime_departures.ex
deanshi: Added a module doc - but agreed. If there's a better naming scheme for this, would be down to rename how the candidate generators.

lib/screens/v2/candidate_generator/widgets/realtime_departures.ex
```diff
@@ -0,0 +1,303 @@
+defmodule Screens.V2.CandidateGenerator.Widgets.RealtimeDepartures do
+  @moduledoc """
+  Candidate generator for LCD RDS Items
+  Takes in the generated sections from Screens.V2.CandidateGenerator.Widets.RdsDepartures and
+  handles the roll-up and creation fo the actual widget that will be serialized and used on
```
robbie-sundstrom: Nit: typos with Widget and of 
```suggestion
  Takes in the generated sections from Screens.V2.CandidateGenerator.Widgets.RdsDepartures and
  handles the roll-up and creation of the actual widget that will be serialized and used on
```

lib/screens/v2/candidate_generator/widgets/realtime_departures.ex
```diff
@@ -0,0 +1,303 @@
...
+
+  defp departures_slots(%PreFare{departures: d, template: :duo}), do: [{d, [:main_content_left]}]
+  defp departures_slots(%PreFare{departures: d, template: :solo}), do: [{d, [:large]}]
+  defp departures_slots(%_app{departures: d}), do: [{d, [:main_content]}]
+
+  # Some screen types are always configured to show departures for one specific transit mode. In
+  # that case, if the mode is devops-disabled, we immediately know the whole screen should display
+  # a "no data" message. Right now, we only handle Bus Shelters
+  defp screen_devops_mode(%Screen{app_id: :bus_shelter_v2}), do: :bus
+  defp screen_devops_mode(_), do: nil
```
robbie-sundstrom: I wonder if you just want to remove this before merging, since this story to fully remove devops disabled mode will be done imminently

robbie-sundstrom (approved):
Overall looks good! As we've talked about, I think we want to hold off on merging this until we have a better handle on the current Screens performance issues, but this is a stamp of approval for once that's done

PR #3162, opened 2026-07-02
chore: update asana permissions

rwaskiewicz:
Updates the permissions for the asana workflow.
When the workflow was updated in 988cd33, we did
not update the permissions set in the workflow in
https://github.com/mbta/workflows/commit/c76e5ccd9556fbea52bc41548f7ceed5985ad775

Prior to this commit, the Asana workflow would fail
to run with the following error:

>  Invalid workflow file: .github/workflows/asana.yml#L9
> The workflow is not valid. .github/workflows/asana.yml (Line: 9, Col: 3): 
> Error calling workflow  'mbta/workflows/.github/workflows/asana.yml@c76e5ccd9556fbea52bc41548f7ceed5985ad775'. 
> The workflow is requesting 'pull-requests: read', but is only allowed 'pull-requests: none'.

Notes for Reviewers
- Link to failed Asana run https://github.com/mbta/screens/actions/runs/28581970522
  - Note this will expire in 90 days, hence the error message being put above/in the commit message when we merge this

**Asana task**: N/A

Description

- [ ] Tests added?

### DIFF
--- a/.github/workflows/asana.yml
+++ b/.github/workflows/asana.yml
@@ -3,7 +3,8 @@ on:
   pull_request:
     types: [edited, opened]
 
-permissions: {}
+permissions:
+  pull-requests: read # Needed to read Asana ticket info from PR description
 
 jobs:
   asana:


panentheos (approved):
Seems fine, but I'm curious why this is needed. We have this workflow in other repos and don't have this explicit permission. Is there a default?

rwaskiewicz:
@panentheos In https://github.com/mbta/workflows/commit/c76e5ccd9556fbea52bc41548f7ceed5985ad775#diff-8f3caa03ab7a2d39e8c55a3e18c6a0887ce935e93549724cae83ee4d0c767333R43, an explicit requirement was added for the `pull-request` permission. We bumped the workflow to v5.1.0 (which included this new permission requirement) via Dependabot this morning in https://github.com/mbta/screens/commit/988cd33da70c6c14e501aeeaa912ec516a8c793c, which caused the workflow to start to fail. However, I either failed to see it fail in the PR or it didn't appear in the checks section (likely the former, as it'd be odd for a workflow run to be omitted like that)

PR #3163, opened 2026-07-02
refactor: RDS items and RDS departure widget logic

deanshi:
Description
18937ccad06408639e63f2304b498d887d7df818 is just a small refactor to clean up how we handle headways compared to other row type

bb25cafe4a6102ff6e02dce7091442a900ca6eee is a refactor to pull out the common RDS Departure Section generation into a common generator. This is currently only used within the DUP Candidate generator, but is a setup for the new candidate generator that will be created for LCDs. It is almost exactly the same code as before, except we added a post_process_rows_fn to allow for the different app types to filter out rows in ways that they want. DUPs can have bidirectional filtering and also have a max number of rows, so we filter and truncated rows based on those values.

[1217ac4](https://github.com/mbta/screens/pull/3163/commits/1217ac4e4e95bf17b4b466f33d77ae8041125c8a) is the refactor from states to items. States will now be potentially rolled up into a list of items, which can contain multiple destinations. Also moves much of the flattening logic into the RDS module now, so in cases where we have exclusively NoService/ServiceEnded/Headways, we roll them up into one state that is passed to the consumer. This allows the consumer to just scan for the single one item when creating the full screen treatments. This code explicitly doesn't roll up Countdowns/sort the states though, due to the need for sorting departures for rows in the consumers anyway. This also introduces the `destinations` field which normally holds a single destination, but in cases where we roll up states will hold onto all of the destinations that item is representing. 

- [x] Tests added?

### DIFF
--- a/lib/screens/v2/candidate_generator/widgets/rds_departures.ex
+++ b/lib/screens/v2/candidate_generator/widgets/rds_departures.ex
@@ -1,0 +1,176 @@
+defmodule Screens.V2.CandidateGenerator.Widgets.RdsDepartures do
+  @moduledoc false
+
+  alias Screens.Routes.Route
+  alias Screens.Schedules.Schedule
+  alias Screens.V2.Departure
+  alias Screens.V2.RDS
+  alias Screens.V2.RDS.{Countdowns, FirstTrip, Headways, NoService, ServiceEnded}
+  alias Screens.V2.WidgetInstance.Departures, as: DeparturesWidget
+
+  alias Screens.V2.WidgetInstance.Departures.{
+    HeadwaySection,
+    NoDataSection,
+    NormalSection,
+    NoServiceSection,
+    OvernightSection
+  }
+
+  alias Screens.V2.WidgetInstance.DeparturesNoData
+  alias ScreensConfig.Departures
+  alias ScreensConfig.Departures.Section
+
+  @type post_process_rows_fn_t ::
+          ([NormalSection.row()], Section.t(), non_neg_integer() -> [NormalSection.row()])
+  @type widget :: DeparturesNoData.t() | DeparturesWidget.t()
+
+  def create_departure_sections(
+        rds_sections,
+        %Departures{sections: departure_sections},
+        post_process_rows_fn \\ fn rows, _section, _total_section_count -> rows end
+      ) do
+    total_section_count = length(rds_sections)
+
+    Enum.zip(rds_sections, departure_sections)
+    |> Enum.map(fn {rds_section, section} ->
+      map_to_departure_section(
+        rds_section,
+        section,
+        total_section_count,
+        post_process_rows_fn
+      )
+    end)
+  end
+
+  @spec map_to_departure_section(
+          RDS.data(),
+          Section.t(),
+          non_neg_integer(),
+          post_process_rows_fn_t()
+        ) ::
+          DeparturesWidget.section()
+
+  defp map_to_departure_section(:error, _, _, _), do: %NoDataSection{}
+
+  defp map_to_departure_section({:ok, []}, _, _, _), do: %NoDataSection{}
+
+  defp map_to_departure_section(
+         {:ok, items},
+         %Section{header: header, layout: layout, grouping_type: grouping_type} = section,
+         section_count,
+         post_process_rows_fn
+       ) do
+    case items do
+      [%NoService{routes: routes}] ->
+        %NoServiceSection{routes: routes}
+
+      [%ServiceEnded{routes: routes}] ->
+        %OvernightSection{routes: routes}
+
+      [
+        %Headways{
+          routes: [%Route{id: route_id} | _rest],
+          range: range,
+          displayed_headsign: displayed_headsign
+        }
+      ] ->
+        %HeadwaySection{route_id: route_id, time_range: range, headsign: displayed_headsign}
+
+      _ ->
+        %NormalSection{
+          rows: items |> create_and_sort_rows() |> post_process_rows_fn.(section, section_count),
+          layout: layout,
+          header: header,
+          grouping_type: grouping_type
+        }
+    end
+  end
+
+  @spec create_and_sort_rows([RDS.item()]) :: [NormalSection.row()]
+  defp create_and_sort_rows(rds_items) do
+    grouped_items =
+      Enum.group_by(rds_items, fn
+        %ServiceEnded{} -> :service_ended
+        %Headways{} -> :headways
+        _ -> :other
+      end)
+
+    service_ended_items = Map.get(grouped_items, :service_ended, [])
+    headway_items = Map.get(grouped_items, :headways, [])
+    other_items = Map.get(grouped_items, :other, [])
+
+    sorted_departures_from_rds(other_items) ++
+      headways_from_rds(headway_items) ++
+      sorted_departures_from_rds(service_ended_items, true)
+  end
+
+  # "Bidirectional" mode: take only the first departure, and the next departure in the opposite
+  # direction from the first, if one exists.
+  @spec maybe_make_bidirectional([Departure.t()], boolean()) :: [Departure.t()]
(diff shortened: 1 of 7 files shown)


digitalcora:
> DUPs can have bidirectional filtering and also have a max number of rows

As a clarification, the `bidirectional` option today works on all screen types, though I'd expect us to only have used it on DUPs. If this change will effectively remove support from non-DUPs that are migrated to RDS, we should check production configuration to confirm this expectation.

lib/screens/v2/rds.ex
```diff
@@ -58,17 +58,24 @@ defmodule Screens.V2.RDS do
...
   defmodule Countdowns do
     @moduledoc """
     State when there is upcoming service to a destination
     and/or alerts which affect service to the destination.
     """
-    @type t :: %__MODULE__{departures: [Departure.t(), ...]}
-    defstruct ~w[departures]a
+    @type t :: %__MODULE__{
+            destinations: [Screens.V2.RDS.destination()],
+            departures: [Departure.t(), ...]
```
digitalcora, line 76: As a clarification on the scope of the proposed refactor, I think we should also flatten the `Countdowns` state into a `Countdown` item, corresponding directly to the same concept in the spec, which only contains (and is guaranteed to contain) a single departure. This would allow `RDS` to take responsibility for sorting, returning a set of items to consumers in the order they should be displayed. My expectation is that this would also allow some other logic simplifications, but I'm open to being told I'm wrong there 🙂

lib/screens/v2/candidate_generator/widgets/rds_departures.ex
```diff
@@ -0,0 +1,243 @@
...
+          displayed_headsign: displayed_headsign
+        }
+      ] ->
+        %HeadwaySection{route_id: route_id, time_range: range, headsign: displayed_headsign}
+
+      _ ->
+        %NormalSection{
+          rows: items |> create_and_sort_rows() |> post_process_rows_fn.(section, section_count),
+          layout: %Layout{},
+          header: %Header{}
```
robbie-sundstrom: This is overwriting the layout and header that are configured for the section. We should pull out these params from the section that's passed into the function

robbie-sundstrom: I think we should also be passing `grouping_type` ahead to `NormalSection` here. Without passing it, we'll always default to grouping by`:time` and ignore that sorting. May want to add tests for this behavior as well.

lib/screens/v2/rds.ex
deanshi: When I was initially implementing this I made this change but for some reason I thought we would _still_ need to have the sorting done on the client side. Re-visting this now, it looks like we can just take the states in order as they go and that will remove the client side sorting/ordering completely. 

Let me make that change here, appreciate the call out.

lib/screens/v2/candidate_generator/widgets/rds_departures.ex
```diff
@@ -0,0 +1,243 @@
...
+        }
+    end
+  end
+
+  def build_instances(
+        slot_names,
+        _departure_sections,
+        true = _all_section_no_data,
+        _all_section_service_ended,
```
robbie-sundstrom: To simplify the arity of this function and prevent multiple booleans getting mixed up, what if we instead passed a single variable to account for the two mutually exclusive booleans. Could be something as simple as a `sections_state` type` where the options are `:all_no_data`, `:all_service_ended`, or a default value. 

It would also be nice to have a `spec` associated with this function, which is the request that this idea sprouted from

lib/screens/v2/candidate_generator/widgets/rds_departures.ex
```diff
@@ -0,0 +1,243 @@
...
+        total_section_count,
+        post_process_rows_fn
+      )
+    end)
+  end
+
+  @spec map_to_departure_section(RDS.data(), boolean(), integer(), ([NormalSection.row()],
+                                                                    Section.t(),
+                                                                    number() ->
+                                                                      [NormalSection.row()])) ::
```
robbie-sundstrom: This spec is incorrect; the second input is `Section.t()`, not a `boolean()`. I think dialyzer didn't flag it because the fourth input should have an additional opening parentheses before the argument list, but I'm surprised this didn't get flagged at all -  `(([NormalSection.row()], Section.t(), non_neg_integer()) -> [NormalSection.row()])`

robbie-sundstrom (changes requested):

lib/screens/v2/candidate_generator/widgets/rds_departures.ex
deanshi: yep, I can do that. I ended up making that change in the LCD PR, but I can make it here.

lib/screens/v2/candidate_generator/widgets/rds_departures.ex
digitalcora: Ignoring your substitution of `non_neg_integer` for `number`, the change you're proposing is:

```elixir
( [NormalSection.row()], Section.t(), number()  -> [NormalSection.row()])
(([NormalSection.row()], Section.t(), number()) -> [NormalSection.row()])
```

but these would be syntactically equivalent (if the latter was accepted as valid syntax; I don't know if parens are even allowed as a "grouping" token in typespecs, such a thing might not be needed given the constraints of the grammar... parens are primarily used to define functions).

lib/screens/v2/candidate_generator/widgets/rds_departures.ex
deanshi: Definitely making this change, but I'm realizing these roll-ups are specifically for DUPs, so I'm going to move these functions back into the DUP client generator as well.

lib/screens/v2/candidate_generator/widgets/rds_departures.ex
deanshi: Auto-formatting actually removed the parens. Making the change for the `non_neg_integer` and changing the spec to be correct.

robbie-sundstrom: I was looking for a reason that the incorrect spec wasn't caught by Dialyzer and I thought that parentheses might be it, but I see now that wasn't the cause. So I'm still a bit confused about that, but mainly care about the `boolean` being changed to `Section.t()`

lib/screens/v2/rds.ex
digitalcora: Going through designs for the upcoming alerts-in-departures work, I'm actually wondering if I should eat my words here because [downstream disruptions](https://www.figma.com/design/mTRbzLUAtUgbw8hmXsNGQJ/Sectional-screens--formerly-Solaris-?node-id=2340-12633&t=tHrxRmJG6cEiCXkN-4) involves "attaching" the alert to a *group* of departures, which assumes all departures for a destination are grouped (we'd be enforcing this automatically instead of with screen configuration). In this vision there basically is no discrete "Alert item" as presented on the screen, just a Countdown grouping that "has" alert information within it. This makes me wonder whether ultimately the grouping should happen within `RDS` as well, which would bring things back around to having a multiple-`Countdowns` item... I'm not sure. Maybe just do what makes the implementation simplest for right now, and we'll rework it later if we have to.

deanshi: I'll keep it the separate `Countdowns` for now - I think if we wanted to change it in the future it wouldn't be the most painful change (I finished up most of it, but was cleaning up tests when I saw this message).

robbie-sundstrom (approved):
Looks great! I haven't tested the full suite of scenarios here, but generally seems to be working as expected. Awesome work on this 🚢

deanshi:
> If this change will effectively remove support from non-DUPs that are migrated to RDS, we should check production configuration to confirm this expectation.

We do have bidirectional filtering for LCDs and it will still be supported in the new work as well. I just failed to clarify that in this comment. Everything will still be supported with this refactor.

PR #3164, opened 2026-07-06
feat: Use headsign abbreviations on DUP departures widget FE

robbie-sundstrom:

Note that this PR won't be merged and deployed to DUPs until the corresponding BE piece is first merged and deployed. 

Two main commits that you can look at to understand the process of this PR.
- I initially just rewrote the logic to work with the list of headsigns on DUPs - [Initial refactor to use headsigns list for DUPs](https://github.com/mbta/screens/commit/bf353257e5c451cffb6012cad78d8c3ebc68865f)
- Broke out step progression in the headsign fitting process so that I could test that logic and make sure I hadn't missed anything. Even though the unit tests don't test the full headsign process, I think it ended up making the whole process easier to understand, and the tests were a good way to validate - [Adjust useLayoutEffect so we can test the logic of sizing state progession](https://github.com/mbta/screens/commit/3f30626731b7e6b95f246d93b79fa4a3e435a706)

I have tested this full process as well with different headsigns on DUPs, including some fake headsigns to force us into fallback scenarios. For example, below was the end state of testing a headsign with a shortest abbreviation of  `"Heath Street Ave Test Station"`

https://github.com/user-attachments/assets/54be55b5-7e35-40b8-b3b3-e9ac516cd6a6


assets/src/components/dup/departures/destination.tsx
```diff
@@ -6,32 +6,84 @@ import { classWithModifier, hasOverflowX } from "Util/utils";
...
-  Government: "Govt",
-  "Medford/Tufts": "Medfd/Tufts",
-};
-
-enum PHASES {
-  ONE_LINE_FULL,
-  ONE_LINE_ABBREV,
-  TWO_LINES_FULL,
-  TWO_LINES_ABBREV,
+export enum PHASES {
```
rwaskiewicz: Non-blocking - `enum` isn't on track to be added to ECMAScript, and transpiles down to an IIFE in JS code. Given our usage, this could probably be replaced with something like:
```ts
export const Phases = {
  OneLine: "ONE_LINE",
  TwoLines: "TWO_LINES",
  Done: "DONE",
} as const; // I used another TS-only construct here...but `Object.freeze` also works

// Equivalent to `type Phases = "ONE_LINE" | "TWO_LINES" | "DONE"`
export type Phases = (typeof Phases)[keyof typeof Phases];
```

which might be generally preferable to an enum.

assets/src/components/dup/departures/destination.tsx
```diff
@@ -55,100 +107,58 @@ const RenderedDestination = ({ parts, index1, index2, classModifier }) => {
...
+        headsignsLength: headsigns.length,
+        firstLineFits: !hasOverflowX(firstLineRef.current),
+        secondLineFits: !hasOverflowX(secondLineRef.current),
+      });
+      if (next) {
+        if (next.headsignIndex !== undefined)
+          setHeadsignIndex(next.headsignIndex);
+        if (next.phase !== undefined) setPhase(next.phase);
+        if (next.partsIndex1 !== undefined) setPartsIndex1(next.partsIndex1);
+        if (next.partsIndex2 !== undefined) setPartsIndex2(next.partsIndex2);
```
rwaskiewicz: Non-blocking - IME/IMO, `if` statements should always have a set of curly brackets surrounding the body (if only to help with long term maintainability). But we're not linting on that and this is purely subjective, hence the non-blocking.

assets/src/components/dup/departures/destination.tsx
```diff
@@ -6,32 +6,84 @@ import { classWithModifier, hasOverflowX } from "Util/utils";
...
+  partsIndex2: number;
+  partsLength: number;
+  phase: PHASES;
+  firstLineFits: boolean;
+  secondLineFits: boolean;
+};
+
+// Returns the next state given current state and line-fit measurements,
+// or null if no update is needed (phase is already DONE or refs are absent).
+export const nextSizingState = (
```
rwaskiewicz, line 49: I do find this much easier to reason about 🎉 

This is all non-blocking, but I tried making the return type a little more explicit than `Partial<SizingState> | null` and came up with the following for the former:
```ts
type PhaseOneResult =
  | {
      partsIndex1: number;
      partsIndex2: number;
      phase: PHASES.DONE;
    }
  | { headsignIndex: number; phase: PHASES.ONE_LINE | PHASES.TWO_LINES };
type PhaseTwoResult =
  | { partsIndex1: number; phase: PHASES.TWO_LINES }
  | {
      partsIndex1: number;
      partsIndex2: number;
      headsignIndex: number;
      phase: PHASES.TWO_LINES;
    }
  | {
      partsIndex2: number;
      phase: PHASES.TWO_LINES;
    }
  | { phase: PHASES.DONE };
```

This was in an attempt to better see/understand the transition states. Note it'll break some existing return types and the usage of the `useState` hooks. 

In the future, if `useAutoSize`(not sure exactly which component the comment is referring to) doesn't eliminate this logic, it might be worth revisiting this with more discrete states to step through this state machine.

assets/tests/components/dup/departures/destination.test.ts
```diff
@@ -0,0 +1,108 @@
...
+const bothOverflow = { firstLineFits: false, secondLineFits: false };
+
+const oneLineBase = {
+  phase: PHASES.ONE_LINE,
+  headsignIndex: 0,
+  partsIndex1: 3,
+  partsIndex2: 3,
+  partsLength: 3,
+  headsignsLength: 1,
+};
```
rwaskiewicz, line 17: Non-blocking, this and `twoLinesBase` case receive `as const` to make them immutable (the values in both are currently mutable) - note this only makes them immutable on the type-level. For object-level immutability, we'd want to `Object.freeze` them


```suggestion
} as const;
```

rwaskiewicz (approved):

assets/src/components/dup/departures/destination.tsx
robbie-sundstrom: Thanks I didn't know this about `enum`! Thanks for this suggestion; I like this improvement a lot!

assets/src/components/dup/departures/destination.tsx
robbie-sundstrom: I actually removed the brackets because I subjectively thought it looked so bad while I was doing a final pass of the code. Maybe my Elixir brain has gotten too used to these single line if statements. Since this isn't even how I initially wrote it, I'll add the curly braces back

digitalcora: My also subjective vote is to keep the inline `if`s, but only when they actually fit on one line.

```js
if (next.phase !== undefined) setPhase(next.phase);
```

⬆️ This looks fine and even desirable to me for readability/conciseness.

```js
if (next.headsignIndex !== undefined)
  setHeadsignIndex(next.headsignIndex);
```

⬆️ This looks cursed. (It's indented, but where are the braces?)

assets/src/components/dup/departures/destination.tsx
robbie-sundstrom: Thanks for the validation that this doesn't just make more sense to me now! I do like the idea of having more explicitly defined states rather than just `Partial<SizingState>` -- I'll play around with that and see if there's any way to neaten up the return types with a variation  of `PhaseOneResult` and `PhaseTwoResult`

assets/src/components/dup/departures/destination.tsx
robbie-sundstrom: I think the `next.headsignIndex !== undefined` was expanded to 2 lines by prettier and I just didn't realize. Agreed that is cursed and should definitely have braces 😅

assets/src/components/dup/departures/destination.tsx
robbie-sundstrom: @rwaskiewicz I realized that there were essentially three different state change types: advancing to the two line phase, advancing to done, and changing indices to try fitting the headsign differently. Made the following types to narrow the scope based on that.

```
// Types representing state changes as we try different headsign variations
type FinishedUpdate = {
  phase: typeof PHASES.Done;
  partsIndex1?: number;
  partsIndex2?: number;
};
type IndexUpdate = {
  headsignIndex?: number;
  partsIndex1?: number;
  partsIndex2?: number;
};
type PhaseUpdate = { phase: typeof PHASES.TwoLines; headsignIndex: number };

export type SizingStateUpdate = FinishedUpdate | PhaseUpdate | IndexUpdate;
```

assets/src/components/dup/departures/destination.tsx
```diff
@@ -23,11 +25,26 @@ type SizingState = {
...
+  partsIndex2?: number;
+};
+type IndexUpdate = {
+  headsignIndex?: number;
+  partsIndex1?: number;
+  partsIndex2?: number;
+};
+type PhaseUpdate = { phase: typeof PHASES.TwoLines; headsignIndex: number };
+
+export type SizingStateUpdate = FinishedUpdate | PhaseUpdate | IndexUpdate;
```
rwaskiewicz: Non-blocking nit-pick: "Phase Update" always reconciles to the "Two Lines" variant. Maybe something like:
```tsx
type FullHeadsignUpdate = { phase: typeof PHASES.TwoLines; headsignIndex: number };

export type SizingStateUpdate = FinishedUpdate | FullHeadsignUpdate | IndexUpdate;
```

assets/src/components/dup/departures/destination.tsx
rwaskiewicz: Love it ❤️ _Much_ more understandable than what I had in the comments!

rwaskiewicz (approved):

assets/src/components/dup/departures/destination.tsx
robbie-sundstrom: Good suggestion -- I think I'll go with `TwoLinesPhaseUpdate` as the type name.

PR #3165, opened 2026-07-07
feat: increase static cache size for RDS LCDs

deanshi:
- from RDS LCD testing in dev-green with Screenplay, we are seeing our cache go beyond our cache limit and seeing an increased number of connection pools that are timing out
- in preparation for RDS on LCDs, increase the cache size to accommodate the increased number of network calls we will be making (and subsequently the cached results)

### DIFF
--- a/config/config.exs
+++ b/config/config.exs
@@ -502,7 +502,7 @@ config :screens,
 
 config :screens,
        Screens.V3Api.Cache.Static,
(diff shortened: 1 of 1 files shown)


robbie-sundstrom (approved):
On the one hand, doubling the cache size seems like a potentially impactful change, but it also doesn't look like our memory allocation has been close to an issue in recent months, so I think this should be fine!

PR #3166, opened 2026-07-08
feat: Use headsign abbreviations on LCDs

robbie-sundstrom:

Uses the list of headsign abbreviations on LCD screens. Some notes
- This should not apply to bus e-inks on Screens Admin because those wrap the overflow onto the next line, and this is specifically looking for horizontal overflow. We specifically look for horizontal overflow because using both horizontal and vertical overflow causes issues with bus e-inks setup, but also makes more sense in the context of headsigns. We have only used `useAutoSize` for text sizing in the past, which is concerned with both horizontal and vertical space; shortening headsigns is only concerned with the horizontal space. 
- Changed the `overflow` type of the headsign name to `clip` because `hidden` would technically create a hidden scroll container that would cause no overflow to register
- Added `user-select: none;` because it really bothered me when testing that interacting would cause the overflow to show. This behavior has always existed and is not technically a problem because our users don't interact with these screens, so I'm fine with removing if there's an argument against. 
- An easy way to see this in action and test it is by overriding the list of headsigns sent by the widget instance or just within the FE component.

### DIFF
--- a/assets/css/lcd/departures/destination.scss
+++ b/assets/css/lcd/departures/destination.scss
@@ -9,7 +9,6 @@
   line-height: 64px;
   color: colors.$cool-black-15;
   white-space: nowrap;
-  user-select: none;
 }
 
 .departure-destination__variation {
--- a/assets/css/lcd/departures/destination.scss
+++ b/assets/css/lcd/departures/destination.scss
@@ -2,13 +2,14 @@
 
 .departure-destination__headsign {
   padding-bottom: 6px;
-  overflow: hidden;
+  overflow: clip;
   text-overflow: ellipsis;
   font-size: 64px;
   font-weight: 600;
   line-height: 64px;
   color: colors.$cool-black-15;
   white-space: nowrap;
+  user-select: none;
 }
 
 .departure-destination__variation {
--- a/assets/src/components/departures/destination.tsx
+++ b/assets/src/components/departures/destination.tsx
@@ -1,6 +1,6 @@
 import type { ComponentType } from "react";
 
-import useAutoSize from "Hooks/use_auto_size";
+import { useHorizontalAutoSize } from "Hooks/use_auto_size";
 
 type Destination = {
   headsigns: string[];
@@ -8,7 +8,7 @@ type Destination = {
 };
 
 const Destination: ComponentType<Destination> = ({ headsigns, variation }) => {
-  const { ref, step: headsign } = useAutoSize(headsigns);
+  const { ref, step: headsign } = useHorizontalAutoSize(headsigns);
 
   return (
     <div className="departure-destination">
--- a/assets/src/components/departures/destination.tsx
+++ b/assets/src/components/departures/destination.tsx
@@ -1,21 +1,19 @@
 import type { ComponentType } from "react";
 
-const abbreviate = (headsign: string): string => {
-  if (headsign === "Government Center") return "Government Ctr";
-  return headsign;
-};
+import useAutoSize from "Hooks/use_auto_size";
 
 type Destination = {
-  headsign: string;
   headsigns: string[];
   variation?: string;
 };
 
-const Destination: ComponentType<Destination> = ({ headsign, variation }) => {
+const Destination: ComponentType<Destination> = ({ headsigns, variation }) => {
+  const { ref, step: headsign } = useAutoSize(headsigns);
+
   return (
     <div className="departure-destination">
-      <div className="departure-destination__headsign">
-        {abbreviate(headsign)}
+      <div className="departure-destination__headsign" ref={ref}>
+        {headsign}
       </div>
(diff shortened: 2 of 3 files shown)


digitalcora:
> it really bothered me when testing that interacting would cause the overflow to show

...why would it do this?

assets/src/components/departures/destination.tsx
```diff
@@ -1,21 +1,19 @@
 import type { ComponentType } from "react";
 
-const abbreviate = (headsign: string): string => {
-  if (headsign === "Government Center") return "Government Ctr";
```
digitalcora, line 4: To confirm that this doesn't change existing behavior, is `Government Ctr` an unconditional abbreviation in the new abbreviations list?

assets/css/lcd/departures/destination.scss
```diff
@@ -2,13 +2,14 @@
...
   padding-bottom: 6px;
-  overflow: hidden;
+  overflow: clip;
   text-overflow: ellipsis;
   font-size: 64px;
   font-weight: 600;
   line-height: 64px;
   color: colors.$cool-black-15;
   white-space: nowrap;
+  user-select: none;
```
digitalcora: > I'm fine with removing if there's an argument against.

I think I'd argue for either:

* removing this, on the grounds that it's not relevant in the "real screen" deployed environment, or that sometimes we legitimately might want to select text inside a screen in order to copy it
* making it universal (i.e. putting it on `#app` or something), on the grounds that sprinkling it on individual elements is duplicative and inconsistent

I admittedly don't yet understand the specific problem you said this fixes.

assets/src/components/departures/destination.tsx
```diff
@@ -1,21 +1,19 @@
...
 
 type Destination = {
-  headsign: string;
   headsigns: string[];
   variation?: string;
 };
 
-const Destination: ComponentType<Destination> = ({ headsign, variation }) => {
+const Destination: ComponentType<Destination> = ({ headsigns, variation }) => {
+  const { ref, step: headsign } = useHorizontalAutoSize(headsigns);
```
digitalcora, line 11: Is there a specific reason we want to allow vertical overflow here? I'm thinking this sounds like a problem to solve by fixing element sizing so there is no vertical overflow (and thus clipping), rather than papering over it by ignoring vertical overflow in the auto-sizing.

robbie-sundstrom:
@digitalcora there was something going on with the way I attached the ref to the element that was causing it to reveal the text hidden behind the ellipses when I would click on it. In my modifications to how the autosizing worked, I must have resolved that issue in a different way because I can't replicate it anymore even with `user-select: none;` removed

assets/src/components/departures/destination.tsx
robbie-sundstrom: Yep, this is the base `Government Center` abbreviation! 
```
    "Government Center" => ["Government Ctr", "Gov’t Center", "Gov't Ctr"],
```

assets/css/lcd/departures/destination.scss
robbie-sundstrom: Commented on the overall PR, but this was a problem with an early version of my code that I seem to have resolved by changing how the `ref` was being used. I'll remove the `user-select` option

assets/src/components/departures/destination.tsx
robbie-sundstrom: I don't think there should ever be a case here where we have vertical overflow to worry about here. Since the text size is fixed for headsigns, it will always fit into the vertical space we have allotted to it. Adding the additional `useHorizontalAutoSize` function just elucidates that we ONLY care about the horizontal space. 

This also prevents breaking our version of the Mercury screens. I believe the elements on the bus e-ink screens could be modified to work with the general `useAutoSize`, but I don't think that's worth the effort given that our FE for those applications isn't actually used.

digitalcora: I think there is at least one current use of `useAutoSize` where we technically only care about one dimension of overflow, but only because we expect there to never be overflow in the other dimension — and if there was, it would technically be a bug in the layout that we should fix. Though I recognize this is partially an argument for spending extra effort to attain an abstract sort of "purity" (bounding boxes never clip their contents according to browser metrics, even when nothing is *visually* clipped) and hence is a bit subjective. If you fall on the other side of the fence, that this doesn't really matter and a tiny bit of extra complexity in the auto-size module is an acceptable fix instead, I can't really say you're wrong!

digitalcora (approved):
Approving regardless of whether you change anything about the horizontal auto-sizing. Good find on `overflow: clip`, this seems more appropriate for elements where we use auto-sizing, though I'm not sure why `hidden` would have prevented it from working since that's what we have used on other elements (which have working auto-sizing) until now.

PR #3167, opened 2026-07-08
chore: Remove unneeded eslint-disables

robbie-sundstrom:
Removing these because they're not needed and running `lint fix` locally removes them


robbie-sundstrom:
Seems like this is an issue with my local Lint being out of date, whoops.

PR #3168, opened 2026-07-08
feat: Hide alerts from flex-zone when there are multiple

robbie-sundstrom:
**Asana task**: [Pre-Fare Alerts - Many Downstream Alerts](https://app.asana.com/1/15492006741476/project/1185117109217413/task/1215510478728893?focus=true)

Added a function to the the `pre_fare` candidate generator to simplify the flex zone once all candidates are generated. Although this doesn't specifically check for the Subway Status, if there's any other widget in the same `:large` slot and multiple alert widgets, we'll filter out the multiple alert widgets. 

Unit tests added and behavior validated . See below for an example of the flex zone with this first enabled, and then disabled:
Park Street
Park Street

### DIFF
--- a/lib/screens/v2/candidate_generator/pre_fare.ex
+++ b/lib/screens/v2/candidate_generator/pre_fare.ex
@@ -7,6 +7,7 @@ defmodule Screens.V2.CandidateGenerator.PreFare do
   alias Screens.V2.Template.Builder
   alias Screens.V2.WidgetInstance
   alias Screens.V2.WidgetInstance.AudioOnly.{AlertsIntro, AlertsOutro, ContentSummary}
+  alias Screens.V2.WidgetInstance.ReconstructedAlert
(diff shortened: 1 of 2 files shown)


deanshi (approved):
:shipit:

digitalcora:
The task specifies that this only applies to *downstream* alerts (i.e. if there is >1 downstream alert, drop all downstream alerts). Is the implementation here logically equivalent? I think there are cases where a "here" or "boundary" alert can appear in the flex zone, in particular on Solo screens.

robbie-sundstrom:
@digitalcora thanks for catching!! Totally missed the solo cases which are adjusted to the flex zone through `adjust_placement`. Adopted a much fuller solution that also checks the location context and only removes down/upstream alerts. Will now keep one or multiple alerts if they fall on a boundary or inside.

deanshi (approved):

PR #3169, opened 2026-07-08
feat: Suppress stale alerts on Screens

rwaskiewicz:
**Asana task**: [Suppress stale alerts on Screens](https://app.asana.com/1/15492006741476/project/1185117109217413/task/1214784349728677?focus=truerl)

**Description**

Suppress stale alerts on Screens. A stale alert is an alert whose
current active period began greater than five weeks ago. Unless
otherwise noted, stale alerts that are suppressed unless they occur at
a stop or the stop is within the alert's boundary.

This commit changes alert filtering logic for several widgets:
- DUP Alerts: Prior to this commit, DUPs only showed alerts when the
stop was inside of, or at the boundary of the alert's location scope.
DUP alert widgets do not have new filtering logic, as we never want to
suppress an alert based on its staleness on a stop that it pertains to.
- Alert Widget (Eink, Bus shelters): Suppress alerts as described above
- Reconstructed Alert Widget (Prefares): Suppress alerts as described
above
- Subway Status: This widget does not have location information, which
keeps it lightweight. For the initial implementation, only filter stale
alerts, per the tasks's engineering notes. In order to introduce the
first test to this widget, `Screens.Alerts.Alert.fetch` has been
replaced with `fetch_alerts_fn` as an argument.

A new `Screens.V2.LocalizedAlert#stop_in_alert_boundary?/3` function
has been extracted from the DUPs Alert widget to for reuse. Regarding
other/future extractions:
1. There are additional cases where this helper can be used. Since the
total LoC of this commit was reaching 500, those will be done in a
follow-up PR.
2. a `suppressable_alert?/3` helper function is repeated in 2 files.
It's current usage didn't merit a common helper (in the author's
opinion)

- [x] Tests added?

**Screenshots**

Note: 
- Since Symphony doesn't have screens, and I didn't filter them out for eink/dups, you may see those in the screenshots from other folks' alerts
- This isn't every combination of all widgets and screens. Please LMK if you'd like me to add additional ones
  - Did not add DUPs, since behavior not changing
 
<details>
<summary>Bus Eink - Stop Closure at Blue Hill Ave at Mattapan Sq</summary>
At the stop:
bus_eink_stale_in_boundary

Not at the stop, same line:
bus_eink_stale_out_boundary
</details>

<details>
<summary>GL Eink - Station Closure at Coolidge Corner</summary>

Station closure at Coolidge Corner, non-stale alert (screen at Govt Center):
gl_eink_not_stale_out_of_boundary_not_stale

Station closure at Coolidge Corner, non-stale alert, subway widget (screen at Govt Center):
gl_eink_not_stale_out_of_boundary

Station closure at Coolidge Corner, non-stale alert, alert shown (screen at Coolidge Corner):
gl_eink_not_stale_station_focus_

Station closure at Coolidge Corner, non-stale alert, subway widget (screen at Coolidge Corner):
gl_eink_not_stale_station_focus

Station closure at Coolidge Corner, stale alert (screen at Govt Center):
gl_eink_stale_not_station_in_boundary

Station closure at Coolidge Corner, stale alert, subway widget (screen at Govt Center):
gl_ink_stale_not_station_in_focus

Station closure at Coolidge Corner, stale alert (screen at Coolidge Corner):
gl_eink_stale_station_focus

Station closure at Coolidge Corner, stale alert, subway widget (screen at Coolidge Corner):
gl_eink_stale_station_focus_

</details>

<details>
<summary>Prefares</summary>
Station closure at Hynes, stale and not stale alert (screen at Hynes):
prefare_stale_and_not_station_focus

Station closure at Hynes, stale alert (screen at Prudential):
prefare_not_stale_station_not_focus

Station closure at Hynes, non-stale alert (screen at Prudential):
prefare_stale_station_not_focus

</details>

### DIFF
--- a/lib/screens/v2/candidate_generator/widgets/alerts.ex
+++ b/lib/screens/v2/candidate_generator/widgets/alerts.ex
@@ -79,7 +79,7 @@ defmodule Screens.V2.CandidateGenerator.Widgets.Alerts do
     |> Stream.filter(&Alert.active?(&1, now))
     |> Stream.filter(&(&1.effect in @relevant_effects))
     |> Stream.filter(&Enum.any?(&1.informed_entities, relevant_ie?))
-    |> Stream.reject(&suppressable_alert?(&1, location_context, now))
+    |> Stream.reject(&LocalizedAlert.suppressable_alert?(&1, location_context, now))
     |> Enum.to_list()
   end
 
@@ -88,10 +88,4 @@ defmodule Screens.V2.CandidateGenerator.Widgets.Alerts do
          home_stop_ids
        ),
        do: home_stop_ids |> MapSet.new() |> MapSet.union(downstream_stops) |> MapSet.to_list()
-
-  @spec suppressable_alert?(Alert.t(), LocationContext.t(), DateTime.t()) :: boolean()
-  defp suppressable_alert?(alert, location_context, now) do
-    Alert.stale?(alert, now) and
-      !LocalizedAlert.stop_in_alert_boundary?(alert, location_context)
-  end
 end
--- a/lib/screens/v2/candidate_generator/widgets/alerts.ex
+++ b/lib/screens/v2/candidate_generator/widgets/alerts.ex
@@ -4,6 +4,8 @@ defmodule Screens.V2.CandidateGenerator.Widgets.Alerts do
   alias Screens.Alerts.Alert
   alias Screens.Alerts.InformedEntity
   alias Screens.LocationContext
+  alias Screens.Routes.Route
+  alias Screens.Stops.Stop
   alias Screens.V2.LocalizedAlert
   alias Screens.V2.WidgetInstance.Alert, as: AlertWidget
   alias ScreensConfig.{Alerts, MultiStopAlerts}
@@ -52,6 +54,8 @@ defmodule Screens.V2.CandidateGenerator.Widgets.Alerts do
 
   (list describes the `relevant_ie?` function clauses in order)
   """
+  @spec relevant_alerts([Alert.t()], [Stop.id()], [Route.id()], LocationContext.t(), DateTime.t()) ::
+          [Alert.t()]
   def relevant_alerts(alerts, stop_ids, route_ids, location_context, now) do
     stop_id_set = MapSet.new(stop_ids)
     route_id_set = MapSet.new(route_ids)
--- a/lib/screens/v2/candidate_generator/widgets/alerts.ex
+++ b/lib/screens/v2/candidate_generator/widgets/alerts.ex
@@ -4,6 +4,7 @@ defmodule Screens.V2.CandidateGenerator.Widgets.Alerts do
   alias Screens.Alerts.Alert
   alias Screens.Alerts.InformedEntity
   alias Screens.LocationContext
+  alias Screens.V2.LocalizedAlert
   alias Screens.V2.WidgetInstance.Alert, as: AlertWidget
   alias ScreensConfig.{Alerts, MultiStopAlerts}
   alias ScreensConfig.Screen
@@ -31,7 +32,7 @@ defmodule Screens.V2.CandidateGenerator.Widgets.Alerts do
          {:ok, alerts} <- fetch_alerts_by_stop_and_route_fn.(reachable_stop_ids, route_ids) do
       alerts
       |> Alert.consolidate_whole_route_delays()
-      |> relevant_alerts(reachable_stop_ids, route_ids, now)
+      |> relevant_alerts(reachable_stop_ids, route_ids, location_context, now)
       |> Enum.map(
         &%AlertWidget{alert: &1, screen: config, location_context: location_context, now: now}
       )
@@ -51,7 +52,7 @@ defmodule Screens.V2.CandidateGenerator.Widgets.Alerts do
 
   (list describes the `relevant_ie?` function clauses in order)
   """
-  def relevant_alerts(alerts, stop_ids, route_ids, now) do
+  def relevant_alerts(alerts, stop_ids, route_ids, location_context, now) do
     stop_id_set = MapSet.new(stop_ids)
     route_id_set = MapSet.new(route_ids)
 
@@ -78,6 +79,7 @@ defmodule Screens.V2.CandidateGenerator.Widgets.Alerts do
     |> Stream.filter(&Alert.active?(&1, now))
     |> Stream.filter(&(&1.effect in @relevant_effects))
     |> Stream.filter(&Enum.any?(&1.informed_entities, relevant_ie?))
+    |> Stream.reject(&suppressable_alert?(&1, location_context, now))
     |> Enum.to_list()
   end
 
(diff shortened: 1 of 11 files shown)


lib/screens/v2/candidate_generator/widgets/alerts.ex
```diff
@@ -51,7 +52,7 @@ defmodule Screens.V2.CandidateGenerator.Widgets.Alerts do
 
   (list describes the `relevant_ie?` function clauses in order)
   """
-  def relevant_alerts(alerts, stop_ids, route_ids, now) do
+  def relevant_alerts(alerts, stop_ids, route_ids, location_context, now) do
```
robbie-sundstrom, line 59: Non-blocking nitpick because you didn't write this function, but a spec for `relevant_alerts` would be great. I'm on the train of adding specs as we touch these old functions

lib/screens/v2/candidate_generator/widgets/reconstructed_alert.ex
```diff
@@ -118,6 +119,12 @@ defmodule Screens.V2.CandidateGenerator.Widgets.ReconstructedAlert do
     end
   end
 
+  @spec suppressable_alert?(Alert.t(), LocationContext.t(), DateTime.t()) :: boolean()
+  defp suppressable_alert?(alert, location_context, now) do
+    Alert.stale?(alert, now) and
+      !LocalizedAlert.stop_in_alert_boundary?(alert, location_context)
+  end
```
robbie-sundstrom: I know you mentioned in your PR summary that you intentionally didn't bundle the two cases of this function up, but I'm wondering about pulling `suppressable_alert` into the `Alert` module. 

One thing that I did have to clarify with the spec is that Subway Status doesn't use this function specifically because it doesn't have the concept of a screen's `LocationContext`. Subway Status and everywhere else we want to mute alerts could all reference the same `suppressable_alert` function that does something like take an optional Location Context. This would allow Subway Status to be handled differently while centralizing where we suppress alerts for any other cases we want to add on in the future.

lib/screens/v2/localized_alert.ex
```diff
@@ -354,4 +354,13 @@ defmodule Screens.V2.LocalizedAlert do
...
   end
+
+  @spec stop_in_alert_boundary?(Alert.t(), LocationContext.t()) :: boolean()
+  def stop_in_alert_boundary?(alert, location_context) do
+    location(%{alert: alert, location_context: location_context}) in [
+      :inside,
+      :boundary_upstream,
+      :boundary_downstream
+    ]
+  end
```
robbie-sundstrom, line 375: Also mentioned in your PR comment that there are other places this could be used, and in fact this function would have come in handy for the code I just wrote for hiding down/upstream alerts from the flex zone. Some follow up clean up with this would definitely be satisfying 😄

test/screens/v2/candidate_generator/widgets/reconstructed_alert_test.exs
```diff
@@ -850,5 +850,127 @@ defmodule Screens.V2.CandidateGenerator.Widgets.ReconstructedAlertTest do
...
+      %{
+        config: config,
+        location_context: location_context,
+        now: now,
+        fetch_stop_name_fn: fetch_stop_name_fn,
+        fetch_location_context_fn: fetch_location_context_fn
+      } = context
+
+      one_week_in_seconds = 604_800
+      ten_weeks_ago = DateTime.add(now, -10 * one_week_in_seconds)
```
robbie-sundstrom: nit, but these are used in multiple tests and seem like just `ten_weeks_ago` should be a part of the context passed to the multiple tests

lib/screens/v2/candidate_generator/widgets/alerts.ex
rwaskiewicz: 👍 Done in 8af1c92b

lib/screens/v2/candidate_generator/widgets/reconstructed_alert.ex
rwaskiewicz: > I know you mentioned in your PR summary that you intentionally didn't bundle the two cases of this function up, but I'm wondering about pulling suppressable_alert into the Alert module.

I considered this, but it felt like `LocalizedAlert` was kind of a superset of `Alert` - that's not _strictly_ true, since a `LocalizedAlert`'s type is the union of the `AlertWidget` type (which itself has an `Alert`) with a few other things: 
https://github.com/mbta/screens/blob/9b20d67220821cc29bd3fc8d474a4f3bb58c790b/lib/screens/v2/localized_alert.ex#L18-L26

But, my general thought was "`LocalizedAlert` as a module knows about `Alert` and relies on it, should `Alert` know about/rely on `LocalizedAlert`?". The conclusion I came to here was "no", as it felt like although we weren't `import`ing these, it felt like a circular dependency still. But, if you don't see an issue with that, let me know and we can revisit that logic.

> One thing that I did have to clarify with the spec is that Subway Status doesn't use this function specifically because it doesn't have the concept of a screen's LocationContext. Subway Status and everywhere else we want to mute alerts could all reference the same suppressable_alert function that does something like take an optional Location Context. This would allow Subway Status to be handled differently while centralizing where we suppress alerts for any other cases we want to add on in the future.

I wrestled with this one a bit during the implementation. What you're saying makes total sense. What I struggled with was where to draw the line in terms of how a suppressable alert may change in the future on a widget-by-widget basis, if at all. Since two of the four use the same suppression logic, I landed on "this might change in the future". In the scenario (which I have no clue have possible it is, FWIW) that we need to add an additional qualifier to see if an alert is suppressable for a subset of these alerts, it felt easier for each to maintain its own function (for now). Anyway, that was long-winded explanation to say, impl in 0907f9c6, LMK what you think

lib/screens/v2/localized_alert.ex
rwaskiewicz: For sure - it's not as much as I thought, but it'll be a nice/quick little PR once this lands (assuming we keep the calling contract the same):
```
$ git diff HEAD~..rw/refactor-rel-loc --stat
 lib/screens/v2/widget_instance/alert.ex     | 21 ++++++++++++++-------
 lib/screens/v2/widget_instance/dup_alert.ex | 10 +++++-----
 2 files changed, 19 insertions(+), 12 deletions(-)
```

test/screens/v2/candidate_generator/widgets/reconstructed_alert_test.exs
rwaskiewicz: Good point - updated in 8c148335c4f3425ea045a3c7bc31cc4dd6b76635

lib/screens/v2/candidate_generator/widgets/reconstructed_alert.ex
robbie-sundstrom: I didn't consider this pseudo-circular module dependency that would create between `Alert` and `LocalizedAlert`, which is a great point. I see how the placement in `LocalizedAlert` is a bit awkward since Subway Status now depends on it, but I also think we should be moving towards consolidating alert handling logic like this between widgets. With that in mind, I kind of see two options:
1. Create some sort of Alert Suppression module that would contain this. This just feels like too much for what currently amounts to a single function
2. Leave it as is.

I would want to go with option 2 and leave it in `LocalizedAlert`. Especially as I think about it more, the location of a screen is foundational to when we want to suppress an alert, and Subway Status is really the only exception to that rule. I'm alright with it having an awkward `nil` case for when we don't have the `location_context` calling this function.

Fine with you changing this slightly, but I like this solution a lot. Thanks for your thoughts here

robbie-sundstrom (approved):
overall looks great!

lib/screens/v2/candidate_generator/widgets/reconstructed_alert.ex
rwaskiewicz: Thanks for the discussion! Seems like we're agreeing to leave it as-is for now, happy to revisit in the future. Something tells me as we have these long-er standing alerts, this won't be the last time we change how all this works

PR #3178, opened 2026-07-15
fix: render green line routes on solo prefares

rwaskiewicz:
**Asana task**: [Prefares Erroring for GL Branches](https://app.asana.com/1/15492006741476/project/1185117109217413/task/1216425679839848?focus=true)

**Description**

Render GL routes in alert banners properly in solo prefare screens.

Prior to this commit, solo prefare screens would fail to properly render
the following full screen alert types:
- station closures: GL route pills would take up half of the screen
- suspensions: GL route pills would take up half of the screen
- shuttles: screen would show "Live updates temporarily unavailable" for
the entirety of the screen

The root cause was twofold:
1. `LocalizedAlert.consolidated_informed_subway_routes/1` returning
multiple lines for alert types that included multiple GL routes, such as
a shuttle buses at Science Center, which can affect the GL-D and GL-E
routes. This resulted in
`(CaseClauseError) no case clause matching: ["Green-D", "Green-E"]`
errors bubbling up.
2. In cases where we would serialize a reconstructed alert wiht multiple
Green Line routes, the serialized result would be rendered such that
they took up half the screen.

In speaking with design, we'd like to collapse all GL routes if all
routes that service the stop are affected by the stop. This PR fixes
the former of the two root causes, and uses the collapsing logic to
work around the latter. Similarly, this commit does not resolve the case
where an alert affects multiple lines (e.g. a station closure at State,
which is serviced by the OL and BL).

- [X] Tests added?

**Notes for Reviewers**

- In the screenshots, it appears that the disruption diagram's view box is smaller than it should be. I didn't dive too far into this as to validate this PR was worth pursuing in its current state. Assuming its generally OK, any thoughts as to how to best work through this without breaking it elsewhere?

**Screenshots**

<details>
<summary>Subway Suspension</summary>
For a subway suspension: 
Screenshot 2026-07-15 at 11 05 49

Before:
Screenshot 2026-07-15 at 11 08 17

After:
I'm unsure why the disruption diagram is off - need to look at that
Screenshot 2026-07-15 at 11 13 04
</details>

<details>
<summary>Shuttles</summary>
For shuttle buses: 
Screenshot 2026-07-15 at 11 15 58

Before:

Screenshot 2026-07-15 at 11 16 46

After:
Screenshot 2026-07-15 at 11 15 22
</details>

<details>
<summary>Station Closure</summary>
For a station closure: 
Screenshot 2026-07-15 at 11 20 07

Before:
Screenshot 2026-07-15 at 11 20 45

After:
I'm unsure why the disruption diagram is off - need to look at that
Screenshot 2026-07-15 at 11 21 43

</details>

### DIFF
--- a/lib/screens/v2/widget_instance/reconstructed_alert.ex
+++ b/lib/screens/v2/widget_instance/reconstructed_alert.ex
@@ -308,14 +308,18 @@ defmodule Screens.V2.WidgetInstance.ReconstructedAlert do
     affected_routes = LocalizedAlert.consolidated_informed_subway_routes(t)
     routes_at_stop = LocalizedAlert.active_routes_at_stop(t)
 
+    collapse_all_green? = all_green_routes_at_stop_affected?(affected_routes, routes_at_stop)
+
     affected_routes
-    # Filter alert-affected routes by which routes are at the current stop
-    # If a green-branch is the affected route, we can generalize it to just "Green-"
-    # because our prefare screens will be on the trunk. Any GL disruption will be
-    # downstream of a GL trunk station.
     |> Enum.filter(fn
       "Green" <> _ -> Enum.find(routes_at_stop, &String.starts_with?(&1, "Green"))
       route -> route in routes_at_stop
+    end)
+    |> Enum.map(fn route ->
+      case route do
+        "Green" <> _ when collapse_all_green? -> "Green"
+        route -> route
+      end
     end)
     |> Enum.flat_map(fn
       route_id ->
@@ -328,6 +332,20 @@ defmodule Screens.V2.WidgetInstance.ReconstructedAlert do
         build_pills_from_headsign(route_id, headsign)
     end)
     |> Enum.uniq()
+  end
+
+  @spec all_green_routes_at_stop_affected?([String.t()], MapSet.t(LocalizedAlert.route_id())) ::
+          boolean()
+  defp all_green_routes_at_stop_affected?(affected_routes, routes_at_stop) do
+    routes_at_stop
+    |> Enum.filter(&String.starts_with?(&1, "Green"))
+    |> MapSet.new()
+    |> then(fn green_routes_or_line_at_stop ->
+      case MapSet.size(green_routes_or_line_at_stop) do
+        0 -> false
+        _ -> MapSet.subset?(green_routes_or_line_at_stop, MapSet.new(affected_routes))
+      end
+    end)
   end
 
   defp build_pills_from_headsign(route_id, nil) do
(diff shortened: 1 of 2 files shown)


deanshi:
> In the screenshots, it appears that the disruption diagram's view box is smaller than it should be.

I think this is a known problem on non-real screens, although looking for a ticket in the backlog, I can't seem to find anything.

lib/screens/v2/widget_instance/reconstructed_alert.ex
```diff
@@ -637,10 +655,16 @@ defmodule Screens.V2.WidgetInstance.ReconstructedAlert do
...
+    routes_at_stop = LocalizedAlert.active_routes_at_stop(t)
+
+    collapse_all_green? = all_green_routes_at_stop_affected?(affected_routes, routes_at_stop)
+
     route_id =
-      case LocalizedAlert.consolidated_informed_subway_routes(t) do
+      case affected_routes do
         ["Green" <> _] -> "Green"
         [route_id] -> route_id
+        _route_ids when collapse_all_green? -> "Green"
```
deanshi, line 667: Just want to make sure I wasn't out of the loop or missing something:

Referencing the Hynes alert issue that you put in the description of the ticket, is there a world where `route_ids` could be two of the Green Line branches, but the full route list at the stop could be three? 

I know this PR explicitly states for solos, but I was wondering if there's something we're doing/we did else where to handle that case.

rwaskiewicz: > Referencing the Hynes alert issue that you put in the description of the ticket, is there a world where route_ids could be two of the Green Line branches, but the full route list at the stop could be three?

Maybe...on duos we show something like (where station closures exist for B & C, but not D):
Screenshot 2026-07-15 at 16 33 52

Let me double check that I haven't glossed over that tomorrow and I'll get back to you

digitalcora:
> In the screenshots, it appears that the disruption diagram's view box is smaller than it should be.

This is a known issue with the Inspector specifically if that's what you're using to look at the screens. The way it does scaling doesn't get along with the disruption diagram code. To work around this, you can set the Inspector scale to 1x, reload the screen page, then reset it to whatever you want afterwards — what matters is that the initial load is at 1x.

rwaskiewicz:
> This is a known issue with the Inspector specifically if that's what you're using to look at the screens. The way it does scaling doesn't get along with the disruption diagram code. To work around this, you can set the Inspector scale to 1x, reload the screen page, then reset it to whatever you want afterwards — what matters is that the initial load is at 1x.

I should have clarified to @deanshi and in the PR summary - the issue persists even after doing this

digitalcora:
The other thing I have seen is that the diagram rendering will inconsistently not work in Firefox, but works consistently in Chrome, which is the browser the real screens use. Some time ago we did have a task to look into this, but decided against messing with it as long as it was still working on real screens. Are you seeing this in Chrome as well?

lib/screens/v2/widget_instance/reconstructed_alert.ex
rwaskiewicz: @deanshi I think this handles it appropriately. I tested locally with Park St, because that let me test GL-B/C/D/E and RL combinations (note: I rebased this branch off of `main` to ensure the changes in https://github.com/mbta/screens/commit/b84e89a7dca980b8c5ffc26e0793207f9b51d89c covered these). There are some CSS issues that I reached out to design about, but those might be a separate PR, as that'll be all frontend styling. At the very least, the server doesn't crash anymore with this PR.

rwaskiewicz:
> The other thing I have seen is that the diagram rendering will inconsistently not work in Firefox, but works consistently in Chrome, which is the browser the real screens use. Some time ago we did have a task to look into this, but decided against messing with it as long as it was still working on real screens. Are you seeing this in Chrome as well?

Confirmed that this was a Firefox issue 😢 Works as intended in Chrome :sigh:

deanshi (approved):
🚢

PR #3180, opened 2026-07-15
chore: Remove `headsign` from non eink departures widgets

robbie-sundstrom:

- Remove the `headsign` passed to the FE now that an updated DUP package has been deployed that removes the use of this value
- Keep `headsign` for einks because that sticks to the "API contract" we have established with Mercury. They expect a single `headsign` value.


docs/mercury_api.md
```diff
@@ -61,6 +61,7 @@ For E-ink screens, each of `sections[].rows[].times_with_crowding[]` has a
 Unix timestamp. This allows the conversion to a number of minutes or "Now" to
 be done "just in time", independent of data fetching. [[#1911]]
 
+We only pass a single headsign for each departure, so we include a `headsign` field instead of the `headsigns` array. [[[#3180]]]
```
rwaskiewicz, line 64: Is this still true? I ask because `serialized_headsign_mercury` is [typed with both `headsign` and `headsigns`](https://github.com/mbta/screens/blob/c3f073cf8b24468923b3a68f8ec6e1cc8bfe0ca8/lib/screens/v2/widget_instance/departures.ex#L125-L129) as of 67832f8adfcdcca238a9d85feddb3169df097df8

rwaskiewicz (approved):

docs/mercury_api.md
robbie-sundstrom: This is a symptom of Mercury's unusual setup for how it generates eink screens. For the departures widget, and the departures widget only, it generates the UI itself. That means they rely on our "widget API" contract remaining the same. However, our Frontend code does rely on the `headsigns` list since its departure widget is tied in with that for LCD screens. It would be too much work for this reason alone to break it out into it's own departures widget on the FE when Mercury builds it's own UI anyway with our data.
I do think switching them to use the list of `headsigns` is possibly the long term solution, but Mercury needs to make a code change on their end to switch to using the list of headsigns instead of an individual headsign. Since I was a bit confused about how this worked initially, I went ahead with this code change to keep `headsign` and `headsigns` for Mercury 

Another task that I picked up which immediately became blocked on design (the XhYYm formatting for departures over an hour away), will also require Mercury code changes, I decided to hold off and batch the `headsign` change request in with that. That will hopefully be the chance for me to actually talk with the Mercury engineers and modify the API usage on their end to use the list of headsigns.

PR #3183, opened 2026-07-16
feat: OvernightSection and Service Ended Rows on LCD

deanshi:
**Asana task**: [Handle RDS ServiceEnded states for LCD screens](https://app.asana.com/1/15492006741476/project/1185117109217413/task/1213921912505438?focus=true)

Description
- Adds the Overnight Section and Service Ended states for LCD screens
- If there's a single common headsign (similar to headways), we display that value along with `No Svc`. In cases where we do not have a common headsign, we display `No Service` as the headsign with the Moon Icon

<details>
Screenshot 2026-07-16 at 1 10 25 PM
Screenshot 2026-07-16 at 1 10 49 PM
Screenshot 2026-07-16 at 1 18 05 PM
Screenshot 2026-09-01 at 10 19 03 AM
Screenshot 2026-09-01 at 10 28 48 AM
Screenshot 2026-09-01 at 10 29 18 AM
Screenshot 2026-07-16 at 1 22 47 PM
Screenshot 2026-07-16 at 1 25 50 PM
Screenshot 2026-07-16 at 1 26 27 PM
</details>

- [x] Tests added?

### DIFF
--- a/lib/screens/v2/rds.ex
+++ b/lib/screens/v2/rds.ex
@@ -158,6 +158,8 @@ defmodule Screens.V2.RDS do
       end)
 
   @spec from_section(Section.t(), DateTime.t()) :: data()
+  defp from_section(%Section{header_only: true}, _now), do: {:ok, []}
+
   defp from_section(
          %Section{query: %Query{params: %Query.Params{stop_ids: stop_ids} = params}},
          now
--- a/lib/screens/v2/rds.ex
+++ b/lib/screens/v2/rds.ex
@@ -30,6 +30,7 @@ defmodule Screens.V2.RDS do
   alias Screens.Util
 
   alias Screens.V2.Departure
+  alias Screens.V2.RDS
 
   alias ScreensConfig.Departures
   alias ScreensConfig.Departures.{Query, Section}
@@ -100,10 +101,11 @@ defmodule Screens.V2.RDS do
     """
     @type t :: %__MODULE__{
             destinations: [Screens.V2.RDS.destination()],
+            displayed_headsign: String.t(),
             routes: [Route.t()],
             last_schedule: Schedule.t() | nil
           }
-    defstruct ~w[destinations routes last_schedule]a
+    defstruct ~w[destinations displayed_headsign routes last_schedule]a
   end
 
   defmodule Headways do
@@ -287,7 +289,7 @@ defmodule Screens.V2.RDS do
           DateTime.t()
         ) :: item() | nil
   defp destination_state(
-         {%Stop{id: stop_id}, _line, _headsign} = destination,
+         {%Stop{id: stop_id}, _line, headsign} = destination,
          departures,
          schedules,
          routes_for_section,
@@ -310,7 +312,10 @@ defmodule Screens.V2.RDS do
 
     cond do
       presented_departures != [] ->
-        %Countdowns{destinations: [destination], departures: presented_departures}
+        %Countdowns{
+          destinations: [destination],
+          departures: presented_departures
+        }
 
       impacted_by_alert? ->
         nil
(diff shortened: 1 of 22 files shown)


lib/screens/v2/widget_instance/departures.ex
```diff
@@ -289,11 +289,35 @@ defmodule Screens.V2.WidgetInstance.Departures do
...
+
+    %{
+      type: :overnight_section,
+      header: Map.put(header, :image_path, Assets.s3_asset_url(image_path)) |> Header.to_json(),
+      text: FreeTextLine.to_json(text),
+      with_headsign: headsign != nil
+    }
   end
 
   def audio_serialize_section(%NormalSection{header: header} = section, screen, now) do
```
robbie-sundstrom, line 323: `audio_serialize` doesn't match on full `OvernightSection`, which causes a crash for this new case

deanshi: This has been fixed and approved by design. Also updated the upcoming PRs for audio as well.

lib/screens/v2/widget_instance/departures.ex
```diff
@@ -476,8 +531,12 @@ defmodule Screens.V2.WidgetInstance.Departures do
...
       times_with_crowding:
         case special_trip_type do
-          :first_trip -> serialize_times_with_crowding(departures, screen, now)
-          :service_ended -> [%{id: id, time: %{type: :overnight, is_live: false}}]
+          # First Trip Types are currently just departures, this will be modified later
+          :first_trip ->
+            serialize_times_with_crowding(departures, screen, now)
+
+          :service_ended ->
+            [%{id: id, time: %{type: :overnight, with_text: app_id != :dup_v2, is_live: false}}]
```
robbie-sundstrom, line 516: NIt: Passing a variable that is essentially encoding "this is a DUP" feels strange when we should have this info on the frontend. I realized accessing this might be messier than expected, after working out how you could do that on the frontend, so I'm leaving this as a comment to see if it sparks any ideas, but not a suggestion. 

The Context in dup_page.tsx defaults to 0, which is indistinguishable from a legitimate "page 0". To detect if you're in a DUP, we could change the context's default to something that can't be a valid page (i.e. null), and expose a small useIsDup() so any shared component can ask "is this a DUP" without a prop drilled down from the backend.

lib/screens/v2/rds.ex
```diff
@@ -505,6 +507,33 @@ defmodule Screens.V2.RDS do
 
   defp maybe_combine_headway_states(headways, _line_direction_pairs), do: headways
 
+  defp get_common_headsign(
```
robbie-sundstrom, line 513: Nit: Add a spec here

lib/screens/v2/rds.ex
```diff
@@ -409,57 +414,54 @@ defmodule Screens.V2.RDS do
...
-           %Headways{
-             destinations: [{_stop, %Line{id: line_id}, first_headsign}],
-             routes: [route],
-             range: range
-           }
-           | _
-         ] = states
-       ) do
-    destinations = Enum.flat_map(states, & &1.destinations)
+  defp combine_service_ended_states_for_section([%ServiceEnded{routes: [route]} | _] = states) do
```
robbie-sundstrom, line 418: Spec seems useful here and on `combine_headway_states_for_section` to guarantee that we're being passed the sections we needed via contract

test/screens/v2/rds_test.exs
```diff
@@ -709,9 +710,14 @@ defmodule Screens.V2.RDSTest do
...
-                    %Route{id: "r1", line: %Line{id: "l1"}, type: :bus}
-                  ])
+                  service_ended(
+                    [{"70061", "l1", "Alewife"}],
+                    nil,
+                    [
+                      %Route{id: "r1", line: %Line{id: "l1"}, type: :bus}
+                    ],
+                    "Alewife"
+                  )
```
robbie-sundstrom, line 829: Seems worth adding additional cases for the following headsign possiblities: 

- multiple destinations all sharing a headsign
- differing headsigns with a shared direction name
- differing headsigns and directions, producing nil

robbie-sundstrom (approved):
Overall looking good! Just small comments. I'm curious to see this tested with some real data instead of faking it myself, so maybe we can leave this in dev for a little bit

PR #3184, opened 2026-07-16
feat: Alert tied to Evergreen Content should be shifted to single_screen

robbie-sundstrom:
**Asana task**: [[extra] Modify Evergreen Content alert_suppression for upcoming RL diversion](https://app.asana.com/1/15492006741476/project/1185117109217413/task/1216636756096278?focus=true)

This case came out of supporting the upcoming RL closure. We would like to show both custom Evergreen Content and the left side of a high priority alert on duo screens when `suppress_alert_widgets` is true. Adds a check on if there is an Evergreen Content Item with the alert ID we're creating a Reconstructed Alert Widget for. If there is, then we bump the placement of the alert down to `:single_screen`. This will not affect any existing behavior for `:solo` screens.

### DIFF
--- a/lib/screens/v2/widget_instance/reconstructed_alert.ex
+++ b/lib/screens/v2/widget_instance/reconstructed_alert.ex
@@ -442,7 +442,7 @@ defmodule Screens.V2.WidgetInstance.ReconstructedAlert do
          %__MODULE__{
            alert: %Alert{id: alert_id, effect: effect} = alert,
            screen: %Screen{
-             app_params: %PreFare{
+             app_params: %_app{
                evergreen_content: evergreen_content
              }
            },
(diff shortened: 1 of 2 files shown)


lib/screens/v2/widget_instance/reconstructed_alert.ex
```diff
@@ -440,7 +440,12 @@ defmodule Screens.V2.WidgetInstance.ReconstructedAlert do
 
   defp base_placement(
          %__MODULE__{
-           alert: %Alert{effect: effect} = alert,
+           alert: %Alert{id: alert_id, effect: effect} = alert,
+           screen: %Screen{
+             app_params: %_app{
+               evergreen_content: evergreen_content
+             }
+           },
```
digitalcora: style nit: this seems like it should all fit on one line

```elixir
           screen: %Screen{app_params: %_app{evergreen_content: evergreen_content}},
```

Including indentation, this is only 84 characters (out of ~98 allowed).

digitalcora (approved):

PR #3185, opened 2026-07-16
fix: attempt to show pills for No Data Sections from Section config

deanshi:
**Asana task**: [Bug: Green Line pill omitted](https://app.asana.com/1/15492006741476/project/1185117109217422/task/1216600856263981?focus=true)

Description
- in the old departures code, we would purposefully do a route call to try and fetch the route from the API to show a pill for the NoDataSections
- in the new code, we omit this extra API call, but here we can attempt to recreate a route depending on the Section configuration. In cases where we have `route_ids` or a `route_type`, we will attempt to create a route to use for `Route.icon/1`.
- In cases where we don't have either, we will omit the route pill.

Question: Should we just attempt to fetch the route like we do in the `post_process_no_data/4`? I opted not to because of how many network calls we're already making for RDS, but an extra route one (that presumably would be cached and only in cases where we have an error or no data at all) wouldn't hurt that bad in my mind. 

Before:
Screenshot 2026-07-16 at 3 17 27 PM
After:
Screenshot 2026-07-16 at 3 17 44 PM

- [x] Tests added?

### DIFF
--- a/lib/screens/v2/candidate_generator/widgets/rds_departures.ex
+++ b/lib/screens/v2/candidate_generator/widgets/rds_departures.ex
@@ -18,7 +18,7 @@ defmodule Screens.V2.CandidateGenerator.Widgets.RdsDepartures do
 
   alias Screens.V2.WidgetInstance.DeparturesNoData
   alias ScreensConfig.Departures
-  alias ScreensConfig.Departures.Section
+  alias ScreensConfig.Departures.{Query, Section}
 
   @type post_process_rows_fn_t ::
           ([NormalSection.row()], Section.t(), non_neg_integer() -> [NormalSection.row()])
@@ -50,9 +50,9 @@ defmodule Screens.V2.CandidateGenerator.Widgets.RdsDepartures do
         ) ::
           DeparturesWidget.section()
 
-  defp map_to_departure_section(:error, _, _, _), do: %NoDataSection{}
-
-  defp map_to_departure_section({:ok, []}, _, _, _), do: %NoDataSection{}
+  defp map_to_departure_section(data, section, _, _)
+       when data == :error or data == {:ok, []},
+       do: %NoDataSection{route: maybe_route_from_section(section)}
 
   defp map_to_departure_section(
          {:ok, items},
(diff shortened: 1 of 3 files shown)


lib/screens/v2/candidate_generator/widgets/rds_departures.ex
```diff
@@ -86,6 +86,32 @@ defmodule Screens.V2.CandidateGenerator.Widgets.RdsDepartures do
...
+              case route_ids do
+                [] -> nil
+                [route_id] -> route_id
+                [first_route_id | _others] -> first_route_id
+              end
+          end,
+        type: route_type
+      }
+    end
+  end
```
robbie-sundstrom: The spec that this returns is incorrect (should return `Route.t() | nil`)
and is also a bit hard to parse. I think something flatter like below, where the different conditions for `route_id` are broken out, would be
```
@spec maybe_route_from_section(Section.t()) :: Route.t() | nil
defp maybe_route_from_section(%Section{
       query: %Query{params: %Params{route_ids: nil, route_type: nil}}
     }),
     do: nil
 
defp maybe_route_from_section(%Section{
       query: %Query{params: %Params{route_ids: route_ids, route_type: route_type}}
     }) do
  %Route{id: route_id(route_ids), type: route_type}
end
 
@spec route_id([String.t()] | nil) :: String.t() | nil
defp route_id(nil), do: nil
defp route_id([]), do: nil
defp route_id(route_ids) do
  if Enum.all?(route_ids, &String.starts_with?(&1, "Green-")) do
    "Green"
  else
    List.first(route_ids)
  end
end
```

deanshi: Made this change but renamed `route_id/1` to `representative_route_id/1` just to be clear that it's not the actual route ID, but one that's for the entire group.

robbie-sundstrom: nice, I like it!

robbie-sundstrom (approved):
looks good!

PR #3188, opened 2026-07-17
refactor: Reduce calls to schedule API in RDS

robbie-sundstrom:
Thought of this improvement while looking through RDS and thinking about the resource usage issues we've been seeing. We currently make two slightly different calls to the Schedules API each time through RDS. By passing the first result of fetching schedules to `Departure.fetch`, we can just use those `Schedule` results to figure out upcoming scheduled `Departures` rather than fetching Schedules a second time.

I don't think this is the cause of any issues with resource usage we've seen, but I think cleaning this up could only help.

### DIFF
--- a/lib/screens/v2/departure.ex
+++ b/lib/screens/v2/departure.ex
@@ -48,7 +48,7 @@ defmodule Screens.V2.Departure do
     fetch_predictions_fn = Keyword.get(opts, :fetch_predictions_fn, &Prediction.fetch/1)
 
     with {:ok, predictions} <- fetch_predictions_fn.(params),
-         {:ok, schedules} <- fetch_schedules(params, opts) do
+         {:ok, schedules} <- get_schedules(params, opts) do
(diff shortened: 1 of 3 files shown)


digitalcora:
> We currently make two slightly different calls to the Schedules API each time through RDS.

I'm curious what the parameters of the different requests are — back in 18e74eb I did some logging of the actual query strings being sent to the API and confirmed that, after that change, we should only be making a single "distinct" schedule request per RDS section, and with a6b5eb0 that should result in just one single request being made to the API.

robbie-sundstrom:
@digitalcora I thought where these calls were differing was the `date` that's passed as a filter. Specifically, I thought we were passing `Util.service_date` when first fetching schedules, and then the second time using `DateTime.utc_now()`, which would lead to double the API calls in at least the hour from 3-4am. I see that I was misreading the code now, and that the default of `Util.service_date` is used in both cases. Confirmed that we should have the same set of params passed in that case, so this is just Friday afternoon brain 🤦

PR #3189, opened 2026-07-20
perf: reduce duplicate self-refresh runs

digitalcora:
The primary change here is that `ScreensByAlert` can now keep track of screen IDs for which a data refresh is in progress. At the *start* of a screen data request or self-refresh, the relevant screen IDs are added to this list, and when displayed alert IDs are updated for a screen ID, it is removed from the list. In case data refreshing does not complete successfully, screen IDs also expire from the list after a short time.

This dramatically reduces the resources used by the self-refresh runner on screens that are already being refreshed by either a data request or another instance's self-refresh runner, though it does not *guarantee* this never happens; doing so would require a lock/transaction mechanism, which Memcache does not have. This is essentially a partial, quick-fix version of the upcoming "resilience" changes, for which we plan to use Redis, which does support these.

Other changes:

* Increase the TTL on alerts data from 40 to 45 seconds, allowing us to increase the buffer time between when the self-refresh runner picks up a screen and its alerts data expires from 5 to 10 seconds. This should improve the reliability of the alerts-on-screens feature, since screen data often does take more than 5 seconds to generate.

* In addition to "normal" screen data requests, also update alerts data on simulation requests. Unclear why we didn't do this before; the data should be identical and equally valid. This mostly cuts resource usage in non-prod environments, where most requests are from Screenplay.

**Asana task**: https://app.asana.com/1/15492006741476/project/1185117109217413/task/1216641767742099?focus=true

### DIFF
--- a/lib/screens/screens_by_alert/memcache.ex
+++ b/lib/screens/screens_by_alert/memcache.ex
@@ -23,6 +23,9 @@ defmodule Screens.ScreensByAlert.Memcache do
 
     # Metadata to make the "self-refresh" mechanism possible
     "screens_last_updated." <> screen_id => timestamp
+
+    # Keeps track of data updates in progress to reduce duplicate work
+    "screens_in_progress." <> screen_id => true
   }
   ```
   """
@@ -37,15 +40,39 @@ defmodule Screens.ScreensByAlert.Memcache do
 
   @screens_by_alert_key_prefix "screens_by_alert."
   @screens_last_updated_key_prefix "screens_last_updated."
+  @screens_in_progress_key_prefix "screens_in_progress."
 
   @config Application.compile_env!(:screens, :screens_by_alert)
   @screens_by_alert_ttl Keyword.fetch!(@config, :screens_by_alert_ttl_seconds)
   @screens_last_updated_ttl Keyword.fetch!(@config, :screens_last_updated_ttl_seconds)
+  @screens_in_progress_ttl Keyword.fetch!(@config, :screens_in_progress_ttl_seconds)
   @screens_ttl Keyword.fetch!(@config, :screens_ttl_seconds)
 
   @impl true
   def start_link(_opts \\ []) do
     Screens.Memcache.start_link(name: @server)
+  end
+
+  @impl true
+  def get_in_progress(screen_ids) do
+    case Memcache.multi_get(@server, Enum.map(screen_ids, &in_progress_key/1)) do
+      {:ok, result} ->
+        result |> Map.keys() |> Enum.map(&key_to_id/1)
+
+      {:error, message} ->
+        log_warning(message, :get_in_progress)
+        []
+    end
+  end
+
+  @impl true
+  def put_in_progress(screen_ids) do
+    key_values = screen_ids |> Enum.map(&in_progress_key/1) |> Map.new(fn key -> {key, true} end)
+
+    case Memcache.multi_set(@server, key_values, ttl: @screens_in_progress_ttl) do
+      {:ok, _result} -> :ok
+      {:error, message} -> log_warning(message, :put_in_progress)
+    end
   end
 
   @impl true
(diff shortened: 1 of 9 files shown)


lib/screens_web/controllers/v2/screen_api_controller.ex
```diff
@@ -120,13 +122,17 @@ defmodule ScreensWeb.V2.ScreenApiController do
     end
   end
 
-  defp simulation_response(screen, "all") do
+  defp simulation_response(screen, "all", _opts) do
```
robbie-sundstrom, line 125: Do we ever used this `simulation_response` for `"all"` variants? Not sure when this would get used in the first place, but I noticed that downstream of this call in `ScreenData.simulation_variants`, we don't use `update_visible_alerts_for_screen_id`

rwaskiewicz (approved):

lib/screens_web/controllers/v2/screen_api_controller.ex
digitalcora: `all` is only used when you enable the variant switcher in the Inspector, which you can do alongside enabling the Screenplay simulation view, so this is reachable (in theory; it is not reachable currently, because no variants are defined). Because we use this pretty rarely and it's an admin-UI-only thing, I thought it was fine to not bother adding the update to it; the performance impact of *not* doing the update would be minimal (probably zero, in production, if you were looking at a "real screen" that was making its own regular refreshes anyway). I would of course like to revisit this with the upcoming changes, including whether we even need to maintain the "variants" feature at all.

lib/screens/screens_by_alert/memcache.ex
```diff
@@ -60,6 +87,7 @@ defmodule Screens.ScreensByAlert.Memcache do
           fn ->
             update_screens_last_updated_key(screen_id, now)
             Enum.each(alert_ids, &update_alert_key(&1, screen_id, now))
+            delete_screens_in_progress_key(screen_id)
```
robbie-sundstrom, line 90: This is likely an over-optimization at this stage, but should we be deleting the screen ID from the in progress list or relying on the TTL? There's a case where these steps happen:
1. Request 1 starts from self-refresh runner
2. Request 2 starts from a screen request (or anything else)
3. Request 1 finishes and then deletes the key from the list
4. An additional self-refresh runner request comes in, but doesn't see that request 2 is still in progress b/c it's screen ID was deleted from the list in Step 3. End up processing an unnecessary request

Putting an additional identifier along with the screen ID for a request would prevent this as well, but falling back to relying on the TTL for cached `screen_id`s feels like the easiest way to avoid this. 

Again, this isn't a huge cost given the Redis work we have planned, so I don't think this absolutely needs to be addressed.

lib/screens_web/controllers/v2/screen_api_controller.ex
robbie-sundstrom: Gotcha, thanks for explaining. I agree that it's fine to leave out updates in this specific code path for now then.

robbie-sundstrom (approved):
Overall looks good! My comment on how we delete IDs definitely doesn't need to be addressed before merging since this will already be a huge improvement as it stands

lib/screens/screens_by_alert/memcache.ex
digitalcora: I think the scenario you're describing might be impossible, because the self-refresh run you introduce in step 4 would not select the screen that was just refreshed by (self-)request 1, because it would not be considered outdated yet. This could only happen if a screen request took longer to process than the self-refresh outdatedness threshold (35s), but that's not possible due to the timeouts we enforce which are universally less than that (≤20s).

digitalcora: That said I do agree it's janky that we might delete an in-progress record despite a refresh (from another source than the one that created the record) still technically being in progress. I think it's *also* janky if we remove the deleting because then we have a record saying a refresh is "in progress" when it maybe actually isn't... idk.

robbie-sundstrom: Ah I agree now that this is even more of an edge case with these numbers in mind! I think it's totally fine to keep the jank of this deletion since it's a temporary solution. The only ways I can foresee this being an issue are extremely contrived

PR #3192, opened 2026-07-21
fix: denote stations that are not accessible

rwaskiewicz:
**Asana task**: [Elevator Widget at Inaccessible Stations](https://app.asana.com/1/15492006741476/project/1185117109217413/task/1214226701648774?focus=true)

Description

Update the elevator status widget to display "This station is not accessible" when there are no elevators at the station. This applies to Bowdoin, Boyleston, and Hynes (Symphony is currently closed to become more accessible).

At an inaccessible station, the widget now shows a QR code that links riders to the MBTA website's trip planner. The URL associated with the QR code includes a `plan` query string, whose value is a base64-encoded value. This value is unique per station. Each base64-encoded value does the following:
1. Prefills the trip origin with the inaccessible station
2. Selects the "Prefer accessible routes" checkbox

These URLs are hardcoded, with the following justification:
1. The number of stations is small. The hope is that this list only changes to remove stations from the list
2. There is additional logic associated with generating these base64-encoded values ([link](https://github.com/mbta/dotcom/blob/062bb95daa3cae7bd4a777582b900fb81f3b6a53/lib/dotcom/trip_plan/anti_corruption_layer.ex#L46-L52)).

The logic for generating these values could have been replicated for Screens. However, even with perfect replication, if how dotcom interpreted these values changes in the future, our generation logic logic may fall out of sync. Regardless of whether we have our own copy of the generation logic or hardcoded URLs, the values sent to dotcom would be broken. We defer to hardcoding for a simpler solution.

A new icon has been added to be shown by the widget for inaccessible stations. The logic for determining the status icon has been refactored to account for the new elevator status variant.

- [X] Tests added?

Screenshot 2026-07-22 at 14 18 49

**New Screenshot**
We updated the QR code size and URL:

**Old Screenshots**
<details>
  <summary>Bowdoin</summary>
Screenshot 2026-07-21 at 14 48 37
</details>
<details>
  <summary>Boylston</summary>
Screenshot 2026-07-21 at 14 52 49
</details>
<details>
  <summary>Hynes</summary>
Screenshot 2026-07-21 at 14 53 29
</details>


digitalcora:
I wonder if we want to increase the size of the QR code in this specific case — those bits are so tiny that they might be hard to resolve, especially at typical screen-viewing distances. The existing design was sized with much shorter URLs in mind. (Or, random idea: QR code leads to a short URL we define at `screens.mbta.com` that then redirects to the trip planner URL?)

rwaskiewicz:
> I wonder if we want to increase the size of the QR code in this specific case — those bits are so tiny that they might be hard to resolve, especially at typical screen-viewing distances. The existing design was sized with much shorter URLs in mind. (Or, random idea: QR code leads to a short URL we define at `screens.mbta.com` that then redirects to the trip planner URL?)

Agreed, it's hit-or-miss in development with a relatively modern phone/camera. I'll ping Design and see if they have thoughts on size/content. That base64 encoded string really adds a lot more data 😅

rwaskiewicz:
Updated the size to 200x200, confirmed by Design to work

lib/screens/v2/widget_instance/elevator_status.ex
```diff
@@ -140,6 +157,27 @@ defmodule Screens.V2.WidgetInstance.ElevatorStatus do
   def serialize_to_map(%__MODULE__{} = widget),
     do: widget |> serialize() |> Map.from_struct() |> Map.delete(:alert_ids)
 
+  # if reached: station does not have any elevators
```
digitalcora: nit: I don't think "if reached" really makes sense here because this is the first function that would be called; it is *always* reached. It would make more sense to me if a comment were added to `closed_here_without_nearby_backups` saying "if reached: station has elevators". (The comments are explaining what must have been *previously* determined to be true if we have reached this function, not what the function itself is going to determine.)

lib/screens/v2/candidate_generator/pre_fare/elevator_status.ex
```diff
@@ -42,8 +44,11 @@ defmodule Screens.V2.CandidateGenerator.PreFare.ElevatorStatus do
           []
       end
 
+    {:ok, elevators} = @facility.fetch(stop_ids: [station_id], types: [:elevator])
+
     [
       %ElevatorWidget{
+        all_station_elevators: elevators,
```
digitalcora: nit: in terms of the "API" between these modules, we are sending more information than the widget needs to do its job — all it ultimately checks for is whether the list is empty, it doesn't even use any of the elevator data. This could be a `home_station_accessible?` or `home_station_has_elevators?` boolean field; that would make setting up the test data easier as well, especially if we also gave the field a default value of `true` in the struct definition (would eliminate a lot of the test diffs in this changeset).

lib/screens/v2/widget_instance/elevator_status.ex
```diff
@@ -101,6 +102,20 @@ defmodule Screens.V2.WidgetInstance.ElevatorStatus do
 
   @app_cta_url "mbta.com/go-access"
   @elevators_url "mbta.com/elevators"
+  @trip_planner_url "mbta.com/trip-planner"
+
+  @utm_campaign_query_string "utm_source=screens&utm_medium=qr&utm_campaign=no_ele&utm_content=[SCREEN_ID]"
```
digitalcora: This contains a placeholder `[SCREEN_ID]` which I assume is meant to be filled out with... the screen ID, but the string is static and we never do anything with this placeholder.

assets/src/components/elevator_status.tsx
```diff
@@ -24,6 +27,17 @@ type Props = {
   qr_code_url: string;
 };
 
+const getStatusIcon = (status: ElevatorStatus): JSX.Element => {
```
digitalcora: nit: "a function that returns an Element" is more conventionally/idiomatically expressed as a Component; I think it's a little confusing/unexpected to have a "plain" function do this unless there's a clear reason it has to be that way.

```tsx
const StatusIcon: ComponentType<{ status: ElevatorStatus }> = ({ status }) => {
  // ...
}

// ...

<div className={className("body")}>
  <StatusIcon status={status} />
```

rwaskiewicz: Done in https://github.com/mbta/screens/pull/3192/commits/e49bf4affa98ba0789c97490a3b939f7cc66a976

lib/screens/v2/widget_instance/elevator_status.ex
rwaskiewicz: Done in https://github.com/mbta/screens/pull/3192/commits/32f32c5504228ae32e96e4010c052c7d1b2f0e06 then updated to use the parent station in https://github.com/mbta/screens/pull/3192/commits/d06e12155b30dfce88c26e5fc802085fa9730cb1, as I don't _think_ we have an easy way to get the screen ID here (and we would want to differentiate between the 2 Boylston prefares). But if that understanding is incorrect, please LMK. In talking to Jim, he's OK with using the station ID in place of screen ID.

lib/screens/v2/candidate_generator/pre_fare/elevator_status.ex
rwaskiewicz: Good point - done in https://github.com/mbta/screens/pull/3192/commits/3f592b686272ed2f12b743622fd2fb8dd0b1a219

lib/screens/v2/widget_instance/elevator_status.ex
rwaskiewicz: https://github.com/mbta/screens/pull/3192/commits/d2f23a724f7352df15a0085f953bb3f9808f8bc8

assets/src/components/elevator_status.tsx
```diff
@@ -24,6 +27,19 @@ type Props = {
   qr_code_url: string;
 };
 
+const StatusIcon: ComponentType<{ status: ElevatorStatus }> = ({
+  status,
+}): JSX.Element => {
```
digitalcora, line 32: the `JSX.Element` type annotation here is redundant — `ComponentType` encompasses the return type.

lib/screens/v2/widget_instance/elevator_status.ex
```diff
@@ -9,11 +9,12 @@ defmodule Screens.V2.WidgetInstance.ElevatorStatus do
   alias Screens.V2.WebLink
   alias ScreensConfig.FreeTextLine
 
-  @enforce_keys ~w[closures home_station_id]a
+  @enforce_keys ~w[ closures home_station_has_elevators? home_station_id]a
```
digitalcora: nit: If we made this field optional instead, like `relevant_station_ids`, and defaulted it to `true`, we could eliminate the test churn in this changeset (no changes would be needed to the existing cases). And I think that is also a reasonable default for the field to have / would not cause confusing behavior, given we always populate it in non-test code.

lib/screens/v2/widget_instance/elevator_status.ex
```diff
@@ -101,6 +102,19 @@ defmodule Screens.V2.WidgetInstance.ElevatorStatus do
...
+    "place-hymnl" =>
+      "https://#{@trip_planner_url}?utm_source=screens&utm_medium=qr&utm_campaign=no_ele&utm_content=place-hymnl&plan=hsQVX3VudXNlZF9kYXRldGltZV90eXBlxADECGRhdGV0aW1lxCAyMDI2LTA3LTE0VDE1OjEwOjM3LjA4MTUyOS0wNDowMMQEZnJvbYTECGxhdGl0dWRly0BFLIeYD1XexAlsb25naXR1ZGXLwFHFoDPnjhnEBG5hbWXEF0h5bmVzIENvbnZlbnRpb24gQ2VudGVyxAdzdG9wX2lkxAtwbGFjZS1oeW1ubMQFbW9kZXOJxANCVVPEBHRydWXEBUZFUlJZxAR0cnVlxARSQUlMxAR0cnVlxAZTVUJXQVnEBHRydWXEDl9wZXJzaXN0ZW50X2lkxAEwxAtfdW51c2VkX0JVU8QAxA1fdW51c2VkX0ZFUlJZxADEDF91bnVzZWRfUkFJTMQAxA5fdW51c2VkX1NVQldBWcQAxAJ0b4TECGxhdGl0dWRlxADECWxvbmdpdHVkZcQAxARuYW1lxADEB3N0b3BfaWTEAMQKd2hlZWxjaGFpcsQEdHJ1ZQ=="
+  }
+  @inaccessible_station_names Map.keys(@inaccessible_station_urls)
```
digitalcora: `_names` feels misleading for this name since it's actually station *IDs*.

digitalcora (approved):
Looking good!

assets/src/components/elevator_status.tsx
rwaskiewicz: While the type can be inferred, it protects us against cases where we grow the type of `ElevatorStatus` (should we do it again). In the event of adding a new variant, the `switch` case won't automatically handle it*. By having an explicit return type, the compiler will yell if we fail to handle one of `ElevatorStatus`'s types (as the inferred return type would change from `JSX.Element` to  `JSX.Element | undefined`). 

\* Unless we add something like:
```ts
   ... 
   default:
     value safisifies never;
```
But I never loved using an expression in TypeScript's value system to prove something in the type system like this.

Something like:
```ts
const throwUnreachable = (val: never): never => {
  throw new Error(`Unreachable code with value '${val}'`)
}

/// in the switch statement
  default:
     return throwUnreachable(status);
```
isn't all that different, but is less awkward (IMO)

lib/screens/v2/widget_instance/elevator_status.ex
rwaskiewicz: 👍 done in 21c234fa

lib/screens/v2/widget_instance/elevator_status.ex
rwaskiewicz: Good point, done in d907d7ad

assets/src/components/elevator_status.tsx
digitalcora: Interesting, I hadn't thought of the extra type annotation as serving that purpose, and was under the impression this was something TypeScript would save us from by default.

Noting that typescript-eslint, which we already use, has a [switch exhaustiveness check](https://typescript-eslint.io/rules/switch-exhaustiveness-check/) available, but it's not part of the default/recommended configuration. Maybe that would be good to enable?

test/screens/v2/widget_instance/elevator_status_test.exs
```diff
@@ -49,7 +49,11 @@ defmodule Screens.V2.WidgetInstance.ElevatorStatusTest do
       # This way we know the AlertsWidget implementation is actually hooked up to the `alert_ids`
       # field of `Serialized`, and tests that assert on this field are meaningful
       closures = [build_closure([stop: %Stop{id: "place-here"}], [redundancy: :in_station], "a1")]
-      widget = %Widget{closures: closures, home_station_id: "place-here"}
+
+      widget = %Widget{
+        closures: closures,
+        home_station_id: "place-here"
+      }
```
digitalcora: There is some lingering "diff noise" here from the extra line breaks added; these changes could be reverted to main.

rwaskiewicz: `git reset` to the rescue! Done in 4eebd68

assets/src/components/elevator_status.tsx
rwaskiewicz: I'd be open to giving it a go - it requires type information, which will require some additional configuration. I'll write up a barebones ticket for us to discuss at next engineering refinement (RE: this check, if we want to switch to use all recommended typed checks (not using them today), etc.)

PR #3193, opened 2026-07-21
perf: increase GC interval for V3 API static cache

digitalcora:
This was originally an experiment, hence the branch name — while it didn't put much of a dent in overall request times, it did dramatically reduce the incidence of timeouts when I tested it in prod, so I'm promoting it to PR.

V3 API static data responses can be unconditionally cached for 30 minutes (and might remain valid for much longer), but they would be evicted from the cache if not accessed in a 60-second period. Despite screen refresh cycles being ≤30 seconds, this is a problem in deployed environments because there is more than one instance of the app, and subsequent data requests from the same screen are unlikely to go to the same instance.

Increase the GC interval to 5 minutes, so even if an instance is "unlucky" (given the current maximum of 5 instances) and receives only 1 in 20 data requests for a given screen, static data will remain available in the cache.

---
  - https://app.asana.com/0/0/1216761382898831

### DIFF
--- a/config/config.exs
+++ b/config/config.exs
@@ -470,26 +470,30 @@ config :screens, Screens.ScreensByAlert.SelfRefreshRunner, batch_size: 20, concu
 
 config :screens, Screens.DeviceMonitor.Store, backend: Screens.DeviceMonitor.Store.Local
 
-v3_api_cache_options = [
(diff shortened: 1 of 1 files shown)


robbie-sundstrom (approved):
great idea!

PR #3194, opened 2026-07-22
feat: filter subway shuttles from departures

deanshi:
Description
- with RDS, we now show schedules for buses but because of this, we are now displaying shuttle schedule times for Subway lines
- filter out subway shuttles because they aren't particularly useful information and we would prefer showing other information if we can

Before: 
Screenshot 2026-07-22 at 10 54 55 AM

After:
Screenshot 2026-07-22 at 10 54 31 AM

(Disregard the visual bug, that will be fixed in an upcoming PR that handles Dual Pill routes for CR)

- [x] Tests modified?

### DIFF
--- a/lib/screens/lines/line.ex
+++ b/lib/screens/lines/line.ex
@@ -11,4 +11,16 @@ defmodule Screens.Lines.Line do
           short_name: String.t(),
           sort_order: integer()
         }
+
+  @spec cr_line?(t()) :: boolean()
+  def cr_line?(%__MODULE__{id: "line-CR-" <> _line}), do: true
+  def cr_line?(_), do: false
+
+  @spec subway_or_light_rail_line?(t()) :: boolean()
+  def subway_or_light_rail_line?(%__MODULE__{id: "line-Blue"}), do: true
+  def subway_or_light_rail_line?(%__MODULE__{id: "line-Orange"}), do: true
+  def subway_or_light_rail_line?(%__MODULE__{id: "line-Red"}), do: true
+  def subway_or_light_rail_line?(%__MODULE__{id: "line-Green"}), do: true
+  def subway_or_light_rail_line?(%__MODULE__{id: "line-Mattapan"}), do: true
+  def subway_or_light_rail_line?(_), do: false
 end
--- a/lib/screens/v2/departure/builder.ex
+++ b/lib/screens/v2/departure/builder.ex
@@ -2,7 +2,9 @@ defmodule Screens.V2.Departure.Builder do
   @moduledoc false
 
   alias Screens.Departures.Departure
+  alias Screens.Lines.Line
   alias Screens.Predictions.Prediction
+  alias Screens.Routes.Route
   alias Screens.Schedules.Schedule
   alias Screens.Stops.Stop
   alias Screens.Trips.Trip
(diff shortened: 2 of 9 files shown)


lib/screens/v2/departure/builder.ex
```diff
@@ -62,6 +65,15 @@ defmodule Screens.V2.Departure.Builder do
...
     }
 
+  defp subway_shuttle?(%Departure{
+         prediction: nil,
+         schedule: %Schedule{route: %Route{line: line} = route}
+       }) do
+    Route.shuttle_route?(route) and Line.subway_line?(line)
+  end
+
+  defp subway_shuttle?(_), do: false
```
robbie-sundstrom, line 75: Shouldn't this also include shuttles for the GL or is there a reason we wouldn't want to filter those out?

deanshi: I think I had it in my head to omit it because of how the branching shuttles might want to be displayed, but I don't think that's quite right. 

I'll add filter them out as well.

deanshi: Actually by that thought, should Mattapan shuttles be on that list too?

robbie-sundstrom: good call! I think we should filter those too. I guess it's all subway and light rail then?

deanshi: yep I think so. only commuter rail will properly show the Dual Pill with scheduled shuttles being created as a Departure and passed all the way through. I guess technically buses too, but I don't think you can have a shuttle for a bus?

deanshi: Okay sorry about that - fixed this up and added a test for this in `builder_test.ex`

lib/screens/lines/line.ex
```diff
@@ -11,4 +11,16 @@ defmodule Screens.Lines.Line do
...
+  def cr_line?(%__MODULE__{id: "line-CR-" <> _line}), do: true
+  def cr_line?(_), do: false
+
+  @spec subway_line?(t()) :: boolean()
+  def subway_line?(%__MODULE__{id: "line-Blue"}), do: true
+  def subway_line?(%__MODULE__{id: "line-Orange"}), do: true
+  def subway_line?(%__MODULE__{id: "line-Red"}), do: true
+  def subway_line?(%__MODULE__{id: "line-Green"}), do: true
+  def subway_line?(%__MODULE__{id: "line-Mattapan"}), do: true
+  def subway_line?(_), do: false
```
robbie-sundstrom: Maybe rename something like `subway_or_light_rail_line`, since subway is technically a `route_type` that GL and Mattapan do not fall under

robbie-sundstrom (approved):
Just a comment on naming, but looks good overall!

lib/screens/lines/line.ex
digitalcora: Perhaps we should think up better terminology for the category of "subway or light rail"... it's unfortunate that GTFS chose to use the unqualified word `subway` for, specifically, heavy-rail subway lines, because usually when we at the T say "subway" we mean all of our rapid-transit rail lines (including, perhaps confusingly, the Mattapan line, which is *not* a "subway" in the sense that it does not travel underground).

PR #3208, opened 2026-07-27
feat: Add Postgres configuration and create screen_configs table

robbie-sundstrom:
**Asana task**: [Create screens table & add Postgres configuration to Screens](https://app.asana.com/1/15492006741476/project/1185117109217413/task/1216318458240930?focus=true)

- Adds the configuration needed for Screens to use Postgres.
- Adds a migration to create the `screen_configs` table.

Have tested this locally pretty thoroughly, but will also deploy this to dev-green before merging for testing.


priv/repo/migrations/20260724202300_create_screen_configs.exs
```diff
@@ -0,0 +1,12 @@
+defmodule Screens.Repo.Migrations.CreateScreenConfigs do
+  use Ecto.Migration
+
+  def change do
+    create table(:screen_configs, primary_key: false) do
+      add :id, :string, primary_key: true
+      add :config, :jsonb
```
rwaskiewicz: Non-blocking - What're your thoughts on making this field `not null`?
```suggestion
      add :config, :jsonb, null: false, default: "{}"
```
Also I still haven't done much work with Ecto, so that suggestion might not be quite right.

.github/workflows/ci.yml
```diff
@@ -18,6 +18,21 @@ concurrency:
...
 env:
   MIX_ENV: test
+  DATABASE_PASSWORD: postgres
+  DATABASE_USER: postgres
+  DATABASE_NAME: screens_test
+  DATABASE_HOST: localhost
+
+services:
+  postgres:
+    image: postgres
```
rwaskiewicz: Nit - can we pin this version using the SHA of the Docker image? Do we need the "full" Docker image, or can we get by with a smaller image (e.g. a bookworm-slim version)?

.github/workflows/ci.yml
```diff
@@ -18,6 +18,21 @@ concurrency:
 
 env:
   MIX_ENV: test
+  DATABASE_PASSWORD: postgres
+  DATABASE_USER: postgres
+  DATABASE_NAME: screens_test
+  DATABASE_HOST: localhost
+
+services:
```
rwaskiewicz: I think this needs to be defined at the job level, rather than the top level of the workflow. ATM, this workflow doesn't run with the following error: 

Image

config/runtime.exs
```diff
@@ -32,6 +32,8 @@ if config_env() == :prod do
     namespace: System.fetch_env!("HOST"),
     hostname: System.fetch_env!("MEMCACHED_HOST")
 
+  config :screens, Screens.Repo, pool_size: 10
```
rwaskiewicz: Double checking my understanding - would this override [the value of `15` for the pool size](https://github.com/mbta/screens/pull/3208/changes#diff-8e72c709985cb6a43cc0ab32cb487b8c17d2b2d003338ea42a4e2ad2829979dcR43) in `prod.exs`?

docs/getting_started.md
```diff
@@ -39,7 +39,8 @@
 
 1. `mix deps.get`
 1. `npm install --prefix assets`
-1. `mix phx.server`
+1. `mix ecto.create` to stand up the DB used for storing screen configs.
```
rwaskiewicz, line 46: Non-blocking - What do you think about adding a Postgres as a prerequisite, [similar to what we do for Screenplay](https://github.com/mbta/screenplay/blob/main/README.md#postgres)?

Dockerfile
```diff
@@ -36,9 +36,14 @@ RUN npm --prefix assets run deploy
...
 FROM elixir-builder AS app-builder
 
+RUN apk add --no-cache --update curl
+
 ENV MIX_ENV="prod"
 WORKDIR /root
 
+RUN curl https://truststore.pki.rds.amazonaws.com/global/global-bundle.pem \
+    -o aws-cert-bundle.pem
+RUN echo "51b107da46717aed974d97464b63f7357b220fe8737969db1492d1cae74b3947  aws-cert-bundle.pem" | sha256sum -c -
```
rwaskiewicz: For my own understanding - is this file not versioned in anyway? How do we determine the SHA? I ask because running these commands below locally exits with a status code of 1 for me:

```
➜  /tmp curl https://truststore.pki.rds.amazonaws.com/global/global-bundle.pem \
    -o aws-cert-bundle.pem
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100  161k  100  161k    0     0   361k      0 --:--:-- --:--:-- --:--:--  361k
➜  /tmp echo "51b107da46717aed974d97464b63f7357b220fe8737969db1492d1cae74b3947  aws-cert-bundle.pem" | sha256sum -c -
aws-cert-bundle.pem: FAILED
sha256sum: WARNING: 1 computed checksum did NOT match
➜  /tmp echo $?
1
```

priv/repo/migrations/20260724202300_create_screen_configs.exs
robbie-sundstrom: I think it could be set to non-nullable, since we'll theoretically never create a screen without specifying the full configuration! Great callout. I'm going back and forth, but think I'd prefer your suggestion of defaulting to an empty `{}` JSON in case we do ever end up in that state

docs/getting_started.md
robbie-sundstrom: good idea! There's gotta be at least some mention of Postgres in the setup readme

.github/workflows/ci.yml
robbie-sundstrom: Pinning the Docker image is a great idea! I think we could get by on a slim version, but I'll dig in a little more and test that out first

config/runtime.exs
robbie-sundstrom: Yes, you're right that this will overwrite the compile time value of 15 for the pool size in prod. This was a miss on my part - this is within the `if config_env() == :prod do` conditional, and should be setting it to 15 in prod. The line in `prod.exs` is then not neede.d The default pool size is 10, so that will be set by default in every other environment.

Dockerfile
robbie-sundstrom: This file is actually not versioned, and apparently it's not infrequently rotated as certificates are added or rotated. I think it might just make sense to get rid of the integrity check in that case since it will start Warning us quickly, but it shouldn't actually be an issue.

priv/repo/migrations/20260724202300_create_screen_configs.exs
robbie-sundstrom: I ended up setting this with no default so that the row won't be created if this column isn't provided.

.github/workflows/ci.yml
robbie-sundstrom: I moved around the workflows a little bit, since both the Elixir build and test steps need access to the test DB. I combined these into a single job, but kept the other Elixir steps separate so that the whole process is mostly run in parallel.

robbie-sundstrom:
@rwaskiewicz this is ready for a re-review. All of the commits that addressed your comments are before the first failed deploy; after that, I was mainly testing things out to debug what was going wrong with connecting to RDS. Context for why the initial deploys didn't work to dev-blue are in [this Slack thread](https://mbta.slack.com/archives/C06Q4V81FS8/p1784749801017879)

.github/workflows/ci.yml
```diff
@@ -20,15 +20,34 @@ env:
   MIX_ENV: test
 
 jobs:
-  ex-compile:
-    name: Elixir Warnings
+  build-elixir:
+    name: Build and Test Elixir
     runs-on: ubuntu-latest
+    env:
+      MIX_ENV: test
```
digitalcora: ✂️ This is applied globally above, and shouldn't need to be repeated here.

.github/workflows/ci.yml
```diff
@@ -20,15 +20,34 @@ env:
...
+      DATABASE_PASSWORD: postgres
+      DATABASE_USER: postgres
+      DATABASE_NAME: screens_test
+      DATABASE_HOST: localhost
+    services:
+      postgres:
+        image: postgres@sha256:1961f96e6029a02c3812d7cb329a3b03a3ac2bb067058dec17b0f5596aca9296
```
digitalcora: Does Dependabot know how to update this? I can't imagine us remembering to do so otherwise.

rwaskiewicz: Non-blocking comment - As I was adding a SHA for Screenplay for zizmor, I realized 3 things for that PR that might apply here:
1. I don't know if / how it will update a version comment (EDIT: Nevermind, this shouldn't apply due to point number 3 in this list)
2. I wonder if we should add comments here and in the Terraform repo that says "Hey, if you ever bump this version, you should change it in <the other repo>"
3. I don't believe Dependabot will try to upgrade this, but if that's the case, that probably works in our favor. If it _does_, we may have to add some config to prevent that.

All that said, the only non-blocking, but actionable comment I have is "Should we add a comment that tells us what image/version this points to?"

.github/workflows/ci.yml
```diff
@@ -20,15 +20,34 @@ env:
...
     steps:
       - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
         with:
           persist-credentials: false
       - uses: ./.github/actions/setup
-      - run: mix compile --warnings-as-errors
+      - name: Elixir Compile (warnings as errors)
+        run: mix compile --warnings-as-errors
+      - name: Elixir Tests
+        run: mix test
```
digitalcora, line 48: Is there a way we can retain the parallelism here without duplicating the `services` block, either by using the new [within-job parallelism](https://github.blog/changelog/2026-06-25-actions-steps-can-now-be-run-in-parallel/) features or composite actions? I suppose it doesn't matter that much because these are two of the fastest steps in the build and they're not the bottleneck anyway, but it rankles that now everything would be in parallel *except* this one somewhat-arbitrary pair of steps.

config/config.exs
```diff
@@ -15,6 +15,19 @@ config :screens, ScreensWeb.Endpoint,
   pubsub_server: ScreensWeb.PubSub,
   live_view: [signing_salt: "cK5v02Jnzmp4C8NjV0wwpD2IdTaZdMvi"]
 
+# Configure the Ecto repository
+config :screens, Screens.Repo,
+  database: "screens",
+  username: System.get_env("DATABASE_USER", ""),
+  password: System.get_env("DATABASE_PASSWORD", ""),
```
digitalcora: Does a blank username/password actually work here in some scenarios, or should these be `fetch_env!`?

.envrc.template
```diff
@@ -13,3 +13,9 @@ export API_V3_URL="https://api-v3.mbta.com"
 # export MERCURY_API_KEY=
 
 export SECRET_KEY_BASE="local_secret_key_base_at_least_64_bytes_________________________________"
+
+## Postgres configuration: username, password, and hostname
+## * Your local Postgres server should go here
+# export DATABASE_USER=
+# export DATABASE_PASSWORD=
+export DATABASE_HOST="localhost"
```
digitalcora: suggestion: since we have a default value of `localhost` for this in `config.exs`, the default does not need to be repeated here; this line could be omitted (we'd probably never change it in development).

config/config.exs
```diff
@@ -15,6 +15,19 @@ config :screens, ScreensWeb.Endpoint,
...
+  hostname: System.get_env("DATABASE_HOST", "localhost"),
+  port: System.get_env("DATABASE_PORT", "5432") |> String.to_integer(),
+  stacktrace: true,
+  show_sensitive_data_on_connection_error: true,
+  backoff_min: 5_000
```
digitalcora: Is this different from the default? If not, it could be removed; if so, maybe there should be a comment explaining why it's different, since it's not entirely self-evident to me.

.envrc.template
rwaskiewicz: Non-blocking comment - probably worth an announcement in the screens-eng channel once this lands with instructions to get this up and running (or just that folks will have to update their local `.envrc` files)

config/prod.exs
```diff
@@ -36,6 +36,12 @@ config :screens, :screens_by_alert, cache_module: Screens.ScreensByAlert.Memcach
 
 config :screens, Screens.ScreensByAlert.SelfRefreshRunner, concurrency: 2
 
+config :screens, Screens.Repo,
+  database: "screens",
```
digitalcora: ✂️ This seems like it could be removed since it is already set to the same value in `config.exs`.

rwaskiewicz (approved):

config/runtime.exs
```diff
@@ -34,6 +34,8 @@ if config_env() == :prod do
     namespace: System.fetch_env!("HOST"),
     hostname: System.fetch_env!("MEMCACHED_HOST")
 
+  config :screens, Screens.Repo, pool_size: 15
```
digitalcora: Curious where this `15` comes from (I see from the earlier discussion that it's different from the default).

docs/getting_started.md
```diff
@@ -39,7 +43,8 @@
...
```
digitalcora, line 46: There are some additional aliases in the [default generated](https://github.com/phoenixframework/phoenix/blob/main/installer/templates/phx_single/mix.exs.eex) Phoenix `mix.exs` when Ecto is enabled, which we might consider adopting, in particular `ecto.setup` and `ecto.reset`. The former could then be used here.

lib/screens/application.ex
```diff
@@ -32,7 +32,9 @@ defmodule Screens.Application do
         Screens.Telemetry,
         {Phoenix.PubSub, name: ScreensWeb.PubSub},
         ScreensWeb.Endpoint,
-        Screens.Health
+        Screens.Health,
+        Screens.Repo,
+        Screens.Migrate
```
digitalcora: ⚠️ I'm not sure it makes sense to put these this late in the startup order; `Repo` needs to start before anything else that will depend on making database queries, which probably will include a good chunk of the app. Since the repo itself will not really depend on any other processes, it might be fine to just put it as the very first one (with `Migrate` immediately after).

lib/screens/repo.ex
```diff
@@ -0,0 +1,43 @@
...
+  require Logger
+
+  use Ecto.Repo,
+    otp_app: :screens,
+    adapter: Ecto.Adapters.Postgres
+
+  def add_prod_credentials(config, auth_token_fn \\ &ExAws.RDS.generate_db_auth_token/4) do
+    host = System.get_env("DATABASE_HOST")
+    port = String.to_integer(System.get_env("DATABASE_PORT", "5432"))
+    user = System.get_env("DATABASE_USER")
```
digitalcora: suggestion: We pass `host` and `user` directly into `generate_db_auth_token`, but these may be `nil` and the function is not expecting that. Since these are required and there's nothing we can really do without them, we could crash earlier and more informatively here by using `fetch_env!`.

lib/screens/repo.ex
```diff
@@ -0,0 +1,43 @@
...
+
+    token =
+      auth_token_fn.(
+        host,
+        user,
+        port,
+        %{}
+      )
```
digitalcora: nit: this definitely all fits on one line.

Dockerfile
```diff
@@ -36,9 +36,13 @@ RUN npm --prefix assets run deploy
 
 FROM elixir-builder AS app-builder
 
+RUN apk add --no-cache --update curl
```
digitalcora: Is there a particular reason to use `--no-cache` here? We don't do so on L15, for instance.

digitalcora (commented):
Apologies for jumping in with a second review, I saw a couple things I wanted to make sure got a mention...

.github/workflows/ci.yml
robbie-sundstrom: I didn't know this within job parallelism was possible, but I'll definitely try that!

config/config.exs
robbie-sundstrom: I don't think the blank username/password would actually work, so fetch_env! to fail startup when it's missing would work here

.github/workflows/ci.yml
rwaskiewicz: I wondered the same - I don't believe it does know how to update it. However, I think this is fine, I think we can add a comment here saying "if/when you bump this, bump the version of RDS in TF as well" (and do the same in TF). 

As an aside - Do we update our Postgres engines often in practice? I seen Screenplay is at Postgres v16 in RDS, but we're using the latest tag in GH Actions

digitalcora: My concern is that this will stay locked to a specific minor/patch version, while RDS and probably also engineers' local development environments will roll on, causing divergence. In the worst case maybe there is a vulnerability in Postgres and CI never gets the patch that fixes it.

robbie-sundstrom: It seems that adding Docker as a `package-ecosystem` in our Dependabot config would allow it to catch updates to a tag and hash version combo: https://docs.github.com/en/code-security/reference/supply-chain-security/supported-ecosystems-and-repositories#docker

I'll add the needed Dependabot changes as an update

rwaskiewicz: > It seems that adding Docker as a package-ecosystem in our Dependabot config would allow it to catch updates to a tag and hash version combo: https://docs.github.com/en/code-security/reference/supply-chain-security/supported-ecosystems-and-repositories#docker

Oh, interesting - I assumed that only applied to Docker files (and compose/k8s manifests). Either way, no harm in a little Dependabot science 🧑‍🔬 

> My concern is that this will stay locked to a specific minor/patch version, while RDS and probably also engineers' local development environments will roll on, causing divergence. In the worst case maybe there is a vulnerability in Postgres and CI never gets the patch that fixes it.

The former is probably true (unless what Robbie said with respect to Dependabot being able to pick this up). There's probably a bigger conversation here about how to keep these aligned - but that's likely something we can tackle outside the scope of this PR. That said, maybe the way to go here is to pin to a Postgres tag instead of a SHA? (EDIT: Should have specified, _major_ version tag)

@robbie-sundstrom This current SHA looks like its resolves to v17 - aren't we running v18 in prod? https://hub.docker.com/layers/library/postgres/bookworm/images/sha256-6d0e8738aeef7ac57bdf5ee9da176001096aa498ed3d0ffcbc44b2a6dc62cf67

Screenshot 2026-07-30 at 16 47 27

config/config.exs
robbie-sundstrom: This is larger than the default, which is just 1 second. I thought 5 seconds would be more reasonable to prevent overwhelming the DB, but I also see that the default backoff approach is random exponential intervals, which is likely enough to prevent that itself

digitalcora: Mm, I think we should go with the defaults here, they are likely well-considered and our usage of the database will probably be *much* lighter than the average database-backed Phoenix app.

Dockerfile
robbie-sundstrom: I don't think it's really needed since this is just a builder with a small image size. I also realize now that the `--update` is redundant with `--no-cache`. I'm removing the no-cache option

config/runtime.exs
robbie-sundstrom: I've removed this pool size increase in prod since I think it's likely not needed given the low load of our application. Starting with the default pool size of 10 in all environments and changing it later if needed seems like the more prudent approach.

.github/workflows/ci.yml
digitalcora: > maybe the way to go here is to pin to a Postgres tag instead of a SHA?

It looks like what we now have is a tag *and* a SHA, which would still pin us to a specific release (?).

robbie-sundstrom: I think including the tag is the best way to go about this. I misunderstood Dependabot's support for this case, and it seems like we would need to set up something specific like the following:
1. Create a dedicated Dockerfile to pull in a postgres image
2. Configure Dependabot to track that file
3. Pull the postgres image in from the Dockerfile within CI

I think this is unfortunately the cleanest way to have Dependabot actually be able to catch upgrades so that it can be pinned to a specific version rather than just track latest.

.github/workflows/ci.yml
digitalcora: I think I will retract my suggestion about using parallelism — from your latest commits it seems to require some workarounds that not only add complexity but also might make the overall job slower anyway (due to duplicatively fetching and compiling the dependencies). The original version seems better, if still not ideal.

robbie-sundstrom: Agreed, that ended up being much more complicated than I was expected. I moved them to just be separate jobs again, but I may combine them again and just have them run sequentially depending on the solution I work out for the Dependabot upgrades on the Postgres image

.github/workflows/ci.yml
robbie-sundstrom: This setup for Dependabot actually wasn't quite as bad as I thought it would be: [a8c496d](https://github.com/mbta/screens/pull/3208/commits/a8c496df33483477e167ebcaa7e06220533e6362).

digitalcora: I can't say this seems worthwhile... it's added complexity just to apply a level of "change caution" *only* to CI that we're (correctly, in my view) not applying to deployed environments, and which isn't going to be enforced in local development either. What do we gain by doing this vs. using the major version 18 tag?

robbie-sundstrom: Since this is CI, I'm pre-disposed to agree that tagging the major version is fine. This setup does give us a deterministic build with an image that will then be kept up to date by Dependabot rather than changed without us realizing, but I don't think that's strictly needed for CI. This approach seemed preferable based on this comment thread.

digitalcora: Oops, I think I was working off a key misconception here, which was that [automatic minor version upgrades](https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/USER_UpgradeDBInstance.PostgreSQL.Minor.html) was the default behavior (or that we had it enabled for this app) — it isn't (and we don't). That means this:

> **RDS** and probably also engineers' local development environments will roll on

...is actually not true of RDS, which will stay locked to whichever version we've set in Tofu (Postgres does not do "patch versions", or it might be more appropriate to say that Postgres "minor versions" *are* patch versions, so e.g. `18.3` describes a single fixed release rather than a prefix for potentially many patch releases).

In this light:

* it seems fine to specify either a major.minor version tag *or* a SHA, as these amount to the same thing
* we probably *don't* want Dependabot bugging us to update this? because if we did that it would become desynced from the version we're using in RDS

Alternatively we could *enable* automatic minor version upgrades, and in that case tagging with just a major version here in CI would be appropriate and would more-or-less match the production behavior.

robbie-sundstrom: I think going ahead with tagging the SHA so it's locked to a major.minor version here. I can make a note that our RDS version is tagged to a specific Postgres release, and that we should update in Terraform and here to be kept in sync if we make changes. Doesn't seem necessary to go back and enable the automatic minor version upgrades in Terraform, so this seems like the simplest option

PR #3211, opened 2026-07-30
feat: added custom handling for Kenmore DUP and PreFare alerts

deanshi:
**Asana task**: [Handling partial suspensions at Kenmore](https://app.asana.com/1/15492006741476/project/1185117109217413/task/1216609610899442?focus=true)

Description
- Updated DUP and Prefare logic to check if the "next stop" in a Green Line branch is informed for an alert. If not, we consider the branch still running
- Added a DUP specific KenmorePartial alerts that are similar to PartialAlerts  but are hardcoded and UI specific for Kenmore. Adjusted TakeoverAlert to include a Kenmore specific flag to display `mbta.com/greenline` rather than the regular alerts URL
- Added specific Kenmore logic for showing a Flex Zone alert displaying which branches are still running during an alert. Handles adjusting the ReconstructedAlert position based on if we are handling the special case or not
- Added a new KenmoreAlertHelper for specific Kenmore handling functions

Few Notes:
1. We will eventually want to refactor this code to handle looking at InformedEntity activities instead of using the "next stop" workaround. Activities seem to be the offical way of handling this sort of edge case.
2. Right now, we have a `FreeText` workaround with Flex Zone alerts (Reconstructed Alerts) by passing `special_text` through with `ReconstructedAlerts`. CSS has been added for `free_text` and associated classes only within Flex Zones. This is to enable some of the pill/icon adding within texts, but we might want to refactor this to have a more general use case than specifically Kenmore. 

Screenshot 2026-07-30 at 1 43 54 PM
Screenshot 2026-07-30 at 8 57 46 AM

- [ ] Tests added?


deanshi:
~Pulling it back into draft as there's a new adjustment from Design about what we want from Prefares.~ 

Reopened after removing the special Prefare design. Now it's the fallback to the header text from the Alert.

lib/screens/alerts/kenmore_alert_helper.ex
```diff
@@ -0,0 +1,94 @@
...
+  end
+
+  @spec reject_branches_if_entity_matches_stop([Alert.t()], [
+          %{branch: String.t(), stop: String.t()}
+        ]) :: [String.t()]
+  def reject_branches_if_entity_matches_stop(alerts, stop_matchers \\ @kenmore_stop_matchers) do
+    alerts
+    |> Enum.flat_map(&branches_if_no_entity_matches_stop(&1, stop_matchers))
+    |> Enum.sort()
+    |> Enum.uniq()
```
robbie-sundstrom: Not a problem for this suspension, but I don't think this would work for multiple alerts that affect different branches. Since `branches_if_no_entity_matches_stop` returns the complement of matched branches. So for example, if the B branch were affected on an alert, it would return `[C,D]`, and the C branch being affected on a separate alert would return `[D,E]`, which gets unioned/uniqed into all three branches. I think a cleaner way to do this would be to first combine the affected set of stops, then find the running branches from that.

lib/screens/alerts/kenmore_alert_helper.ex
```diff
@@ -0,0 +1,94 @@
...
+  ]
+
+  @spec kenmore_only_some_branches_affected?([Alert.t()], String.t()) :: boolean()
+  def kenmore_only_some_branches_affected?(alerts, "place-kencl") do
+    length =
+      alerts
+      |> get_branches_if_entity_matches_stop(@kenmore_stop_matchers)
+      |> length()
+
+    length < 3
```
robbie-sundstrom: I don't think this ends up being an issue because we're only calling this when we know that Kenmore is `:inside` an alert, but it seems potentially risky to be returning true from this function if the list of branches is empty. Is there a problem with having this check that the length is greater than 0 and less than 3?

deanshi: Nope, I don't think so, I can add that in.

lib/screens/alerts/kenmore_alert_helper.ex
deanshi: Ah good call. I've updated the logic so that we get the affected branches and then reject them from the list of `stop_matchers`. 

I've also updated another error in the logic in `dup/alerts.ex` on L242 to be `length(uninformed_suspension_branches) > 0`, in indicate if there are _any_ branches that are still running.

lib/screens/v2/candidate_generator/dup/alerts.ex
```diff
@@ -189,48 +195,78 @@ defmodule Screens.V2.CandidateGenerator.Dup.Alerts do
...
+             special_case: :kenmore_partial_branches_closed
+           },
+           %DupSpecialCaseAlert{
+             alert_ids: alert_ids,
+             widget_type: :kenmore_partial_alert,
+             slot_names: [:full_rotation_two],
+             branches: uninformed_suspension_branches,
+             special_case: :kenmore_partial_branches_closed
+           }
+         ]}
```
robbie-sundstrom, line 264: Nit: This and the three DupSpecialCaseAlerts above seem like they should all be calling a shared helper function

assets/css/dup/alert/kenmore_partial_alert.scss
```diff
@@ -0,0 +1,18 @@
+@use "CSS/colors";
+
+.kenmore-partial-alert__container {
+  background: #e6e4e1;
```
robbie-sundstrom: This is `$warm-neutral-95`

lib/screens/v2/candidate_generator/dup/alerts.ex
```diff
@@ -189,48 +195,78 @@ defmodule Screens.V2.CandidateGenerator.Dup.Alerts do
   defp kenmore_special_case(alerts, true), do: {:normal, alerts}
 
   defp kenmore_special_case(alerts, _inside_location) do
-    branches =
+    # Check shuttle/suspension branches to see if the next stop is also Informed
+    # if it is, assume that the alert properly affects the branch
```
robbie-sundstrom: This comment is a little confusing. Might be helpful to explain that we want to show affected branches for the shuttle special case, but show the branches that are unaffected and still have service for a suspension. Otherwise, at first glance it seems `informed_shuttle_branches` and `uninformed_suspension_branches` are opposites. Maybe it'd be helpful to rename `uninformed_suspension_branches` to something like `active_branches_during_suspension`

robbie-sundstrom (approved):
Small comments, but approving because it looks good overall! Definitely looks alright for testing on DUPs as it currently stands.

lib/screens/v2/candidate_generator/dup/alerts.ex
deanshi: I thought about pulling it out into some helper function/use some shared variables so we don't have this double chunk of `DupSpecialCaseAlerts`, but I decided against it since for really just these two cases, the function/variables I was using was basically just creating all new variables and effectively assigning them again in a similar fashion. 

Maybe if there's a clean way of pulling everything out that I'm missing?

lib/screens/v2/candidate_generator/dup/alerts.ex
deanshi: Renamed it to `active_branches_during_suspension` and renamed the other one to just `shuttle_branches` to be more clear on the intent.

PR #3214, opened 2026-07-31
feat(admin): ability to delete screens

digitalcora:
This was already possible by deleting entries from the raw JSON Editor, but since this will likely be removed soon as a consequence of moving to database storage, there should be at least a basic UI to perform this function.

Committing a config change that will delete one or more screens gets an extra confirmation with some attention-grabbing emoji.

Screenshot 2026-07-31 at 2 36 11 PM

**Asana task**: https://app.asana.com/1/15492006741476/project/1185117109217413/task/1216571034745371?focus=true

Note: I think the original idea of this task was that we'd do this alongside/somewhere in the middle of the database changes, but it seemed reasonable to get out of the way now since nothing about it really "depended on" those changes. Once we migrate the admin UI to be able to use the database in general, this feature would naturally be included.

Open to suggestions on UI wording here while remaining concise; I'd like to somehow make it clearer that you don't have to be super careful around the "Delete selected" button since it just "marks for deletion" or "sends to trash" or some similar concept, and it's the separate Validate/Commit action that *really* deletes screens.

### DIFF
--- a/assets/src/components/admin/editor.tsx
+++ b/assets/src/components/admin/editor.tsx
@@ -22,6 +22,18 @@ const appIdFilters: { id: AppId | null; name: string }[] = [
   ...SCREEN_APP_ENTRIES.map(([id, { name }]) => ({ id, name })),
 ];
 
+const confirmScreenDeletion = (ids: Set<string>): boolean =>
+  window.confirm(
+    `This will permanently delete the following screens:
+
+${Array.from(ids)
+  .sort()
+  .map((id) => `❌ ${id}`)
+  .join("\n")}
+
+This is not easily undone and may break live screens if done in production. Really commit this change?`.trim(),
+  );
+
 const Editor = () => {
   const [didInitialize, setDidInitialize] = useState(false);
   const [isAddingScreen, setIsAddingScreen] = useState(false);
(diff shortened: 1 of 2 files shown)


assets/src/components/admin/editor.tsx
```diff
@@ -80,6 +100,15 @@ const Editor = () => {
     resetKeyFn();
   };
 
+  const deleteScreens = (ids: string[], resetKeyFn: () => void = () => {}) => {
+    setLocalConfig({
+      ...localConfig,
+      screens: _.omit(ids, localConfig.screens),
```
rwaskiewicz, line 106: I'm confused by how this line works - although admittedly I've never used lodash's `omit` for arrays like this. For my own understanding, is this enumerating `localConfig.screens` and removing the `id` that obj is keyed by from `ids`?

rwaskiewicz (approved):
One question for my own understanding, O/W looks good!

assets/src/components/admin/editor.tsx
digitalcora: I _think_ you're describing what it does, though to be clear it ends up as something like this:

```js
_.omit(["SCR-1"], { "SCR-1": { ... }, "SCR-2": { ... } })
// => { "SCR-2": { ... } }
```

> admittedly I've never used lodash's `omit` for arrays like this.

The array is the list of keys to omit from the object. It looks "backwards" because this is [`lodash/fp`](https://gist.github.com/jfmengels/6b973b69c491375117dc#_omitpaths-object), which (to enable currying/partial application) rotates the argument order of most functions so the "main" argument comes last.

rwaskiewicz: Ah gotcha, makes sense. Thanks!

PR #3217, opened 2026-07-31
feat(admin): minor enhancements

digitalcora:
* Improve initialization/in-flight request handling in the Inspector, extracting some code to share with the table editor.

* Add a button to the Inspector to reload the configuration, as in the table editor.

* Put the Inspector's "reload screen page" button next to the "address bar" where it might be more expected to find it.

* When editing only one screen in the table editor, show its ID in the dialog title, instead of "Editing 1 screen".

* Center checkboxes in their columns in the main table view. This also makes them easier to click as the target is expanded to the entire cell horizontally (not vertically, which would be nice but seems more complicated to achieve).

* Fix duplicate "request failed" error dialog in the table editor when e.g. the user needs to reauthenticate.

### DIFF
--- a/assets/src/components/admin/inspector.tsx
+++ b/assets/src/components/admin/inspector.tsx
@@ -14,6 +14,7 @@ import { type AudioConfig } from "Components/screen_container";
 
 import {
   fetch,
+  withInFlight,
   SCREEN_APPS,
   type Config,
   type Screen,
@@ -49,13 +50,17 @@ const buildIframeUrl = (screen: ScreenWithId | null, isSimulation: boolean) => {
 const Inspector: ComponentType = () => {
   const [config, setConfig] = useState<Config | null>(null);
   const [frameLoadedAt, setFrameLoadedAt] = useState(0);
+  const [isLoading, setIsLoading] = useState(false);
 
-  useEffect(() => {
-    fetch
-      .get("/api/admin")
-      .then((response) => JSON.parse(response.config))
-      .then((config) => setConfig(config));
-  }, []);
+  const reloadConfig = () => {
+    withInFlight(setIsLoading, async () => {
+      const response = await fetch.get("/api/admin");
+      const config: Config = JSON.parse(response.config);
+      setConfig(config);
+    });
+  };
+
+  useEffect(reloadConfig, []);
 
   const onScreenUpdated = (screenId, newConfig) => {
     if (config) {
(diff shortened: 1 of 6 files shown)


assets/src/components/admin/inspector.tsx
```diff
@@ -92,17 +96,27 @@ const Inspector: ComponentType = () => {
   const adjustFrame = () => adjustScreenFrame(frameRef, isSimulation, zoom);
   useLayoutEffect(adjustFrame, [isSimulation, zoom]);
 
+  if (!didInitialize) {
+    reloadConfig();
+    setDidInitialize(true);
+  }
+
```
panentheos: Since this mutates state it should be inside an effect. Probably just an "on mount" effect like the previous code used.

assets/src/util/admin.tsx
```diff
@@ -143,6 +143,22 @@ export const useResetKey = (): [Key, () => void] => {
...
+ * Calls an async function and sets a state value according to whether the
+ * function is "in flight" (has not yet resolved).
+ */
+export const withInFlight = async (
+  setFn: (isInFlight: boolean) => void,
+  requestFn: () => Promise<void>,
+) => {
+  setFn(true);
+  try {
+    await requestFn();
```
panentheos, line 156: This extracted version doesn't have the `window.alert` on errors, which seems like something we should keep as a lightweight way of catching unexpected failures.

digitalcora: There is already a (better) `window.alert` in the failure path within the basic admin fetch helpers, which is why I summarized removing this other `window.alert` as

> Fix duplicate "request failed" error dialog in the table editor when e.g. the user needs to reauthenticate.

It does re-throw, but that doesn't seem to cause any problems in the UI.

assets/src/components/admin/inspector.tsx
digitalcora: The sole purpose of this refactor was to remove the "on mount effect", since using effects for one-time initialization logic is discouraged, as being sort of unrelated to the intended purpose of effects (see [You Might Not Need an Effect](https://react.dev/learn/you-might-not-need-an-effect#initializing-the-application) — this section is not exactly describing the same thing, but close).

panentheos: The example code there still uses an effect, it just uses a flag to prevent it from running more than once. Generally I think React doesn't want you setting state or creating side effects directly in the render path. If we want to do more of the right things here, we could implement the cleanup handler by creating an `AbortController` and wiring it up to the underlying fetch. Then handle the initial load + refresh mechanism with an incrementing state counter that's a dependency of the effect. That's how I've typically approached it, anyway.

digitalcora: I don't want this to hold up the other changes so I suppose I'll just set it back how it was before for now.

panentheos (approved):

PR #3219, opened 2026-08-03
fix: Check fonts are loaded before measuring headsign overflow on DUPs

robbie-sundstrom:
See [Slack thread ](https://mbta.slack.com/archives/C02SXM166N8/p1785770961563189).

On initial rendering of the page, the overflow calculation is sometimes done before the font is loaded onto the page. This caused issues with `Heath Street`, which just barely overflows with Inter font, but not with the default font. Adding this check before doing the headsign sizing prevents this issue from occurring.

This behavior is available starting on Chrome 31, and so will work on DUP hardware: https://developer.mozilla.org/en-US/docs/Web/API/Document/fonts

### DIFF
--- a/assets/src/components/dup/departures/destination.tsx
+++ b/assets/src/components/dup/departures/destination.tsx
@@ -147,41 +147,44 @@ const Destination: ComponentType<DupDestination> = ({
    * cause infinite update loops, so we don't need to be warned that it might.
    */
   useLayoutEffect(() => {
-    // First attempt to fit headsign on a single line. Prefer fitting an abbreviated
-    // headsign on a single line than the full headsign across 2 pages.
-    // If that doesn't work, try to fit it on two lines by adjusting
-    // between which words we paginate. Move through abbreviations until we find fit.
-    if (
-      firstLineRef.current &&
-      secondLineRef.current &&
-      phase !== PHASES.Done
-    ) {
-      const next = nextSizingState({
-        phase,
-        headsignIndex,
-        partsIndex1,
-        partsIndex2,
-        partsLength: parts.length,
-        headsigns: headsigns,
-        firstLineFits: !hasOverflowX(firstLineRef.current),
-        secondLineFits: !hasOverflowX(secondLineRef.current),
-      });
-      if (next) {
-        // Update the state so we can re-attempt sizing with updated values
-        if ("headsignIndex" in next && next.headsignIndex !== undefined) {
-          setHeadsignIndex(next.headsignIndex);
-        }
-        if ("phase" in next && next.phase !== undefined) {
-          setPhase(next.phase);
-        }
-        if ("partsIndex1" in next && next.partsIndex1 !== undefined) {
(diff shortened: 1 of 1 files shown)


assets/src/components/dup/departures/destination.tsx
```diff
@@ -147,41 +147,44 @@ const Destination: ComponentType<DupDestination> = ({
...
-        if ("phase" in next && next.phase !== undefined) {
-          setPhase(next.phase);
-        }
-        if ("partsIndex1" in next && next.partsIndex1 !== undefined) {
-          setPartsIndex1(next.partsIndex1);
-        }
-        if ("partsIndex2" in next && next.partsIndex2 !== undefined) {
-          setPartsIndex2(next.partsIndex2);
+    // Wait for fonts to load before measuring text width
+    document.fonts.ready.then(() => {
```
digitalcora, line 151: Should we instead block rendering the client outright until fonts are ready, or something like that? This definitely isn't the only place where we measure something that depends on the font being stable (see every use of `useAutoSize`, for starters) and we could potentially fix a broad swath of inconsistent rendering issues if we made this universal.

robbie-sundstrom: I think that's a great idea so we don't end up with a weird edge case happening like this again. Will update the PR with that change

robbie-sundstrom: Implementing this to wait on all downstream texts to load before rendering the full client didn't work as simply as I expected. Since I'm still multi-threading myself between the screens config and dev-blue work, I'd rather create a small follow up task for this and have this fix just for DUPs get merged and deployed.

digitalcora: OK by me as long as we have that follow-up task.

digitalcora, line 151: suggestion: if it's possible to provide an `async` function to `useLayoutEffect`, we could instead use `await` here and avoid having to change all the other lines.

digitalcora (approved):

assets/src/components/dup/departures/destination.tsx
robbie-sundstrom: Since you can't provide an async funtion to `useLayoutEffect`, I think we'd need to define an `async` function inside of it and then call that within, which would still require indenting the existing lines.

digitalcora: I imagine that's intentional; you probably don't want to encourage blocking the completion of a layout effect on an async function that could potentially take a long time to return. Maybe ideally we wouldn't do this blocking via `then` either, but would either abbreviate right away and then re-do the abbreviation once the browser signals that fonts are done loading, or initially render in an un-abbreviated state and only start the process once that signal arrives (maybe with an optimization — if there's an API to check, without waiting, *whether* fonts are already loaded by the time the component first renders, we can skip the extra work and just do one abbreviated render pass).

PR #3221, opened 2026-08-03
fix: Set env variables for DB in setup action

robbie-sundstrom:
A fix for the setup action since these env variables are now required in `config.exs`. See [successful run](https://github.com/mbta/screens/actions/runs/30855101442)


digitalcora:
This may be obviated by #3222.

PR #3222, opened 2026-08-03
dev database improvements

digitalcora:
See individual commits and their messages for details.


robbie-sundstrom (approved):
Moving the fetch_env calls to runtime does simplify things! Can also remove the setting of `DATABASE_USER` and `DATABASE_PASSWORD` in Dockerfile now.

PR #3223, opened 2026-08-04
feat: Add Screens Admin tool to backfill Configs in Postgres

robbie-sundstrom:
**Asana task**: [DB Backfill Script for Screen Configs](https://app.asana.com/1/15492006741476/project/1185117109217413/task/1216318458240936?focus=true)

- Add a section to the Tools section of Screens Admin to backfill Postgres from the JSON config

### DIFF
--- a/assets/src/components/admin/tools.tsx
+++ b/assets/src/components/admin/tools.tsx
@@ -88,9 +88,12 @@ const ImportConfigs = () => {
 
     setIsImporting(true);
 
-    const { success, upserted, deleted, error } = await fetch.post(IMPORT_PATH, {});
+    const { status, upserted, deleted, error } = await fetch.post(
+      IMPORT_PATH,
+      {},
+    );
 
-    if (success) {
+    if (status === 200) {
       window.alert(`Imported ${upserted} configurations. Deleted ${deleted}.`);
     } else {
       window.alert(`Import failed: ${error || "Unknown error"}`);
--- a/assets/src/components/admin/tools.tsx
+++ b/assets/src/components/admin/tools.tsx
@@ -93,12 +93,13 @@ const ImportConfigs = () => {
       {},
     );
 
+    setIsImporting(false);
+
     if (status === 200) {
       window.alert(`Imported ${upserted} configurations. Deleted ${deleted}.`);
     } else {
       window.alert(`Import failed: ${error || "Unknown error"}`);
     }
-    setIsImporting(false);
   };
 
   return (
--- a/assets/src/components/admin/tools.tsx
+++ b/assets/src/components/admin/tools.tsx
@@ -2,11 +2,13 @@ import { useMemo, useState } from "react";
 import { fetch } from "Util/admin";
 
 const API_PATH = "/api/admin/maintenance";
+const IMPORT_PATH = "/api/admin/import_configs";
 
 const Tools = () => {
   return (
     <main className="admin-page">
       <EvergreenContentCleanup />
+      <ImportConfigs />
     </main>
   );
 };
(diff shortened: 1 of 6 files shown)


assets/src/components/admin/tools.tsx
```diff
@@ -68,4 +70,35 @@ const EvergreenContentCleanup = () => {
...
+  };
+
+  return (
+    <section>
+      <h2>Import Screen Configurations</h2>
+      <p>
+        Load screen configurations from the JSON file into the DB. This will
+        overwrite any existing configurations in Postgres.
+      </p>
+      <button type="button" onClick={importConfigs}>
```
rwaskiewicz: The acceptance criteria mentions a confirmation dialog:

> Popup box to confirm action - "Are you sure you want to run this script" 

Is that not needed anymore?

assets/src/components/admin/tools.tsx
```diff
@@ -68,4 +70,35 @@ const EvergreenContentCleanup = () => {
   );
 };
 
+// This section should be a part of post_config_migration_cleanup
```
rwaskiewicz, line 73: There's a handful of these comments mentioning `post_config_migration_cleanup`. Is this an identifier of sorts to mark what we should remove after the migration?

assets/src/components/admin/tools.tsx
rwaskiewicz: Do we have an idea of how long this operation will take? Should we disable this button until it's completed?

lib/screens_web/controllers/admin_api_controller.ex
```diff
@@ -46,6 +47,20 @@ defmodule ScreensWeb.AdminApiController do
     json(conn, %{images: Image.list()})
   end
 
+  @spec import_configs(Plug.Conn.t(), map()) :: Plug.Conn.t()
+  # This should be a part of post_config_migration_cleanup
+  def import_configs(conn, _params) do
+    case ScreenConfigs.import_from_file() do
+      {:ok, %{upserted: upserted, deleted: deleted}} ->
+        json(conn, %{success: true, upserted: upserted, deleted: deleted})
```
rwaskiewicz: For my own understanding, why use `success` as a field to denote if the operation succeeded or not instead of HTTP status codes?

assets/src/components/admin/tools.tsx
robbie-sundstrom: Oh yes I meant to add this in the PR description! I think this would be helpful to speed up the cleanup a bit, but wanted to get other thoughts. Worth keeping?

lib/screens_web/controllers/admin_api_controller.ex
robbie-sundstrom: status code does probably make more sense as a standard here! I just played around with returning a bunch of different values here, so I just added success to it while playing around and then left it that way. I initially had new, updated, unchanged, and deleted before I realized I was overcomplicating it way more than needed.

assets/src/components/admin/tools.tsx
robbie-sundstrom: Yes I forgot about my own A/C! I think this is still a good thing to have a confirmation window behind, especially since we'll be adding more actions to the Tools page. Don't want to accidentally click this post migration when doing other Admin work

assets/src/components/admin/tools.tsx
rwaskiewicz: Yeah, I think so - in the past, I've had the nicety of JIRA giving identifiers for future tickets (e.g. `SCRNS-123: Do the post config migration cleanup`) that I could create TODOs with (`TODO(SCRNS-123): Some description of the work to be done here`). Since Asana doesn't expose anything like that, the scope is small, and we'll add it to the PR description/commit, I think this is fine 👍

lib/screens_web/controllers/admin_api_controller.ex
robbie-sundstrom: Actually I totally forgot, the `fetch` helper in Screens Admin only returns a parsed JSON, which is why I did this in the first place. Made a small change in [3928d62](https://github.com/mbta/screens/pull/3223/changes/3928d62dca9953de6c309fc1b66cda27bd17ba4f) to parse the status code as well and return that in this function.

assets/src/components/admin/tools.tsx
robbie-sundstrom: Good suggestion! I disabled the button until after the operation  [8c5661a](https://github.com/mbta/screens/pull/3223/commits/8c5661a9e7c94dc0e860a35e68271f9f2ad4037f)

assets/src/components/admin/tools.tsx
```diff
@@ -72,29 +72,40 @@ const EvergreenContentCleanup = () => {
...
 
     if (success) {
-      window.alert(
-        `Imported ${upserted} screen configurations. Deleted ${deleted}.`,
-      );
+      window.alert(`Imported ${upserted} configurations. Deleted ${deleted}.`);
     } else {
       window.alert(`Import failed: ${error || "Unknown error"}`);
     }
+    setIsImporting(false);
```
rwaskiewicz: Non-blocking nit - I would move `setIsImporting` above the conditionals here, as `window.alert` blocks the main thread. I doubt we'll run into issues here with state and UI falling out of sync, but until someone clicks "OK" on the alert, `setIsImporting(false)` won't fire.

rwaskiewicz (approved):

PR #3228, opened 2026-08-06
fix: incorrect audio content summary behavior

digitalcora:
The Content Summary widget generator function checks whether there are any "takeover" widgets in the snapshot it's given, and doesn't generate the summary widget if so. However, this snapshot only included widgets with audio equivalence, which is not true of (say) an evergreen content takeover without any "audio text" configured. Screens in this state would thus incorrectly read out a content summary (and nothing else), and report a warning due to not finding the header widget.

Fix this by passing the full list of widgets to `audio_only_instances`. Only those that are valid audio candidates are included in the readout, as before.

**Asana task**: https://app.asana.com/1/15492006741476/project/1185117109217413/task/1217246407080153?focus=true

### DIFF
--- a/lib/screens/v2/screen_audio_data.ex
+++ b/lib/screens/v2/screen_audio_data.ex
@@ -14,19 +14,11 @@ defmodule Screens.V2.ScreenAudioData do
         now \\ DateTime.utc_now()
       ) do
     if Parameters.audio_enabled?(screen, now) do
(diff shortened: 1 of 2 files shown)


rwaskiewicz (approved):

digitalcora:
Note: merging despite GHA outage causing CI failures. I've run all the Elixir commands that are part of CI locally and confirmed they pass, on my own head be it.

PR #3231, opened 2026-08-11
feat: Screen Config Postgres Client APIs and Admin Logic

robbie-sundstrom:
**Asana task**: [Define Screen Config shared APIs ](https://app.asana.com/1/15492006741476/project/1185117109217413/task/1216318458240942?focus=true)

- Adds APIs for fetching all configs and editing a list of configs that will be called by Screenplay. These are protected behind API authorization]
    - Note that I’ve tested calling the API to fetch all configs from Screenplay locally and it’s in working order
- Modifies Screens Admin. This was initially a separate task, and is split into a separate commit with this work. Including this work allowed me to better understand how I should design use of the feature flag, and what could be shared between the two.
    - The existing `index` Screens Admin endpoint now calls a different function that handles the feature flag logic
    - There is a new `screen_configs` endpoint that handles modifying/deleting screen configs in JSON or Postgres based on the feature flag

Frontend Changes
- Fetching all configs
    - The structure of the JSON we pass could be simplified, but after playing around with it, I think this would be easiest after the migration. Keeping the same nested format for now
- Making updates to configs
    - Since this currently passes back the full JSON to make changes, the FE now passes different data based on the feature flag it receives from the BE. This is handled by the new function `commitScreenConfigChanges`. 
    - `editor.tsx` calls this directly
    - `inspector.tsx` now calls this within the `AdminForm` component.


lib/screens_web/router.ex
```diff
@@ -132,4 +136,11 @@ defmodule ScreensWeb.Router do
 
     get "/screens_by_alert", ScreensByAlertController, :index
   end
+
+  scope "/api", ScreensWeb do
```
rwaskiewicz, line 135: Non-blocking nit - From the perspective of someone who is trying to hit these endpoints for the first time, they'd probably have to start digging through the code to figure out why they'd getting 401's back. It might be worth putting in the README that you need to set an authorization header in requests to these endpoints

assets/src/components/admin/admin_form.tsx
```diff
@@ -44,36 +44,49 @@ const AdminValidateControls = ({
...
-    const config = configRef.current.value;
-    const dataToSubmit = { config };
-    fetch.post(confirmPath, dataToSubmit).then((resultJson) => {
-      if (resultJson.success === true) {
+  const [isLoading, setIsLoading] = useState(false);
+
+  const confirmFn = async () => {
+    setIsLoading(true);
+    try {
+      const config = configRef.current.value;
```
rwaskiewicz: No action required - Can we give `onConfirm` a type? It looks like it should be a ` React.RefObject<HTMLTextAreaElement | null>` based on it's creation [here](https://github.com/mbta/screens/blob/6607c7f8122158e99b04341cc6f130a7f930d85e/assets/src/components/admin/admin_form.tsx#L93), but here we treat it as if it'll never be `null` here. If it ever is `null`, this'll fail at runtime

lib/screens/screen_configs.ex
```diff
@@ -28,8 +30,41 @@ defmodule Screens.ScreenConfigs do
     end
   end
 
-  def list do
-    Repo.all(ScreenConfig)
```
rwaskiewicz, line 32: Double checking - was `list()` simply never used before this?

lib/screens/screen_configs.ex
```diff
@@ -28,8 +30,41 @@ defmodule Screens.ScreenConfigs do
...
+  @spec list_all() :: String.t() | :error
+  def list_all do
+    # Returns all Configs as a JSON to be used by Screens Admin
```
rwaskiewicz: No action required - I think we can drop 'a' - "Returns all Configs as JSON to be used by Screens Admin"

lib/screens/screen_configs.ex
```diff
@@ -28,8 +30,41 @@ defmodule Screens.ScreenConfigs do
...
+        config
+      end
+    end
+  end
+
+  @spec list_screen_configs() :: [ScreenConfig.t()]
+  def list_screen_configs do
+    # Returns all configs as a list of ScreenConfig structs to be used by Screens Admin
+    # The API controller handles the JSON encoding and formatting for the response.
+    # As part of post_config_migration_cleanup, this and above function can be cleaned up
```
rwaskiewicz: No action required - I'd advise against relational comments ("and above function") only because things can change between when we land this code and do the cleanup. Realistically, that _probably_ won't happen though - just pedantry more than anything 😆

lib/screens/screen_configs.ex
```diff
@@ -46,4 +81,91 @@ defmodule Screens.ScreenConfigs do
       conflict_target: :id
     )
   end
+
+  @spec upsert_list([%{:id => String.t(), :config => map()}]) :: :ok | {:error, any()}
```
rwaskiewicz: Are we able to update the `any()` in the tuple for errors a more concise type? It looks like this could be replaced with `Ecto.Changeset.t()`, but (assuming that's correct) I would think we'd want to transform this into a string, different kind of error struct, etc. to decouple what's returned from this function from the API.

That said, it looks like we assume this will succeed [here](https://github.com/mbta/screens/pull/3231/changes/e04192b5b149cbd252ecc1bfa1508992082283f8#diff-67481fcca70d087e0fd1531bbbb987166e9de2dd9f39b675b79a818f94c91513R114). Is there any logging we should do in the event of a Postgres failure that we might not get from a stacktrace? I'm not familiar with the types of errors that Ecto will return, apologies if that's vague!

lib/screens/screen_configs.ex
```diff
@@ -46,4 +81,91 @@ defmodule Screens.ScreenConfigs do
...
+  defp upsert_list(updates) do
+    Enum.reduce_while(updates, :ok, fn update, _acc ->
+      case upsert_screen_config(update) do
+        {:ok, _} -> {:cont, :ok}
+        {:error, reason} -> {:halt, {:error, reason}}
+      end
+    end)
+  end
+
+  @doc """
```
rwaskiewicz, line 121: For my own understanding, we have a mix of `@doc` comments and blocks of comments ([e.g.](https://github.com/mbta/screens/blob/e04192b5b149cbd252ecc1bfa1508992082283f8/lib/screens/screen_configs.ex#L52)). How do we determine what we want to doc with `@doc` vs not?

lib/screens/screen_configs.ex
```diff
@@ -46,4 +81,91 @@ defmodule Screens.ScreenConfigs do
...
+  defp update_to_postgres(updates, deletes) do
+    with :ok <- upsert_list(updates) do
+      perform_deletes(deletes)
+    end
+  end
+
+  @spec perform_deletes([String.t()]) :: :ok | {:error, any()}
+  defp perform_deletes(deletes) do
+    Enum.reduce_while(deletes, :ok, fn id, _acc ->
+      Repo.delete_all(from s in ScreenConfig, where: s.id == ^id)
```
rwaskiewicz: Does this fn ever return `{:error, any()}`? I deleted `{:error, any()}` from the return type, and `mix dialyzer` passed. Maybe something to do with the `reduce_while` and `delete_all` I'm not seeing?

This may be tied to the first question, but for my own understanding, why do we need to use `reduce_while` here? Wouldn't `Enum.each/2` suffice here?

lib/screens/screen_configs.ex
```diff
@@ -46,4 +81,91 @@ defmodule Screens.ScreenConfigs do
...
+    with {:ok, config_json, _version} <- @config_fetcher.fetch_config(),
+         {:ok, decoded_config} <- Jason.decode(config_json),
+         screens when is_map(screens) <- Map.get(decoded_config, "screens", %{}) do
+      updated_screens =
+        Enum.reduce(updates, screens, fn update, acc ->
+          id = extract_id(update)
+          config = extract_config(update)
+
+          existing_config = Map.get(acc, id, %{})
+          merged_config = Map.merge(existing_config, config)
```
rwaskiewicz: No action required - this could also be rewritten as:
```elixir
          merged_config =
            acc
            |> Map.get(id, %{})
            |> Map.merge(extract_config(update))
```

lib/screens/screen_configs.ex
```diff
@@ -46,4 +81,91 @@ defmodule Screens.ScreenConfigs do
...
+    end)
+  end
+
+  @doc """
+  Updates and deletes multiple screen configs.
+  Accepts a list of maps with :id and :config keys for updates, and a list of screen IDs for deletions.
+  """
+  @spec commit_updates([%{:id => String.t(), :config => map()}], [String.t()]) ::
+          :ok | {:error, any()}
+  def commit_updates(updates, deletes \\ []) do
```
rwaskiewicz, line 127: I'm not entirely sure we should combine updates + deletes in a single API call. Maybe I'm being too much of a purist/this reminds me a little bit of SOAP calls though. Is there an advantage of combining these other than one less round trip from the API to the server?

lib/screens_web/controllers/screen_configs_api_controller.ex
```diff
@@ -0,0 +1,38 @@
+defmodule ScreensWeb.ScreenConfigsApiController do
+  use ScreensWeb, :controller
+
+  alias Screens.ScreenConfigs
+
+  def index(conn, _params) do
+    screen_configs =
+      ScreenConfigs.list_screen_configs()
+      |> Enum.map(fn screen_config ->
```
rwaskiewicz: Do we need this `Enum.map` call? Does the `ScreenConfig` list returned by `list_screen_configs()` have everything we need here?

robbie-sundstrom: It actually has more info than is needed, since it includes the timestamp fields (`created_at`, `updated_at`). Since you pointed this out, I do think we could filter these out at the data access layer though

lib/screens/screen_configs.ex
robbie-sundstrom: This is just recreating existing behavior, which calls a single API point for both updates and deletes at the same time. I agree in principle though, and I could update this so that the frontend make two separate API calls

lib/screens/screen_configs.ex
robbie-sundstrom: I do need to be consistent with this 😅 I kind of like the @doc comments because of how they integrate with tools like ElixirLS in VSCode. But I do generally think the `@doc` comments are moreso for API consumers, which is why we generally don't have them throughout our codebase. I'll move towards consistency here, which I'd like to be `@doc` comments for this new file at least

lib/screens/screen_configs.ex
robbie-sundstrom: I actually only used this for testing what values are ingested into Postgres during the previous task to create the backfill script. I just missed deleting it when cleaning that up for PR and eventual merge

lib/screens/screen_configs.ex
robbie-sundstrom: `Enum.each` did originally suffice here, since we're not short circuiting early like in `update_to_postgres` in the event that we do encounter an error with deletion. I'm making some changes to error handling based on your other comments to add more detailed error codes and details, so I have moved this back to a `reduce_while`

lib/screens/screen_configs.ex
robbie-sundstrom: This comment inspired me to do some big improvements in this file. I added a `commit_error` type that correspond to the different type of errors in a more human readable form. For the legacy responses, I didn't iron out the final responses as well since I was focused on getting this up in a workable form. 

`{:upsert_failed, String.t()}` and `{:delete_failed, String.t()}` do process the `Ecto.Changeset.t()` in a clean way to return useful info to consumers. I'm going to add tests and clean up the legacy erroring a little, but I like this general structure much more.

lib/screens/screen_configs.ex
digitalcora: The Elixir convention as far as I've seen is `@doc` for public functions and `#` blocks for private functions, since private functions are treated as "hidden" anyway and there's no way to access docs attached to them beyond looking at the code. Technically the main purpose of `@doc` is in generating ExDocs which are made available on Hex for published libraries; in an application codebase, it's mostly useful for editor LS integration. I usually include them if there's something important for callers to know that isn't covered by the function name, argument names, and typespec, which is all also readily accessible via LSP. (This is not just "for API consumers" in the sense of "consumers of some API external to the whole application", but if you take an expansive view of an "API" as "the set of public functions exported by a module", then that is accurate!)

lib/screens/screen_configs.ex
robbie-sundstrom: This suggestion helped clean things up a bit. Changes where I separate the update and delete endpoints are in this commit: [f12a489](https://github.com/mbta/screens/pull/3231/commits/f12a48903ecd38e0bf967e924940f0c00c619e88)

assets/src/components/admin/admin_form.tsx
robbie-sundstrom: Done in [36fd202](https://github.com/mbta/screens/pull/3231/commits/36fd202b526140310cf5f7c1557c22b60d96e869)

lib/screens_web/router.ex
robbie-sundstrom: Done in [c9a3132](https://github.com/mbta/screens/pull/3231/commits/c9a313204fe44c76bca00aad073383f71e713c45)

assets/src/components/admin/admin_form.tsx
```diff
@@ -1,6 +1,14 @@
-import { useState, useEffect, useRef, type JSX } from "react";
+import { useState, useEffect, useRef, type JSX, type RefObject } from "react";
 import { AUTOLESS_ATTRIBUTES, fetch } from "Util/admin";
 
+type AdminConfirmControlsProps = {
+  onConfirm: (config: any) => Promise<{ success: boolean; error?: string }>;
+  configRef: RefObject<HTMLTextAreaElement | null>;
+  onCancel: () => void;
+  onError: (string) => void;
```
rwaskiewicz, line 8: Non-blocking nit - it looks like  the parameter name (or type, depending how you look at it) here is omitted:

```suggestion
  onError: (errorMsg: string) => void;
```

But looking at it's usage here, maybe its type is more than that of type `string`?

https://github.com/mbta/screens/blob/f12a48903ecd38e0bf967e924940f0c00c619e88/assets/src/components/admin/admin_form.tsx#L77-L81

where in the `else` clause, `result.error` can be `undefined`, and if the `catch` clause, `error` is of type `unknown`. Keeping the API boundary between components the same, one potential type-safe solution could be:
```ts
      if (result.success === true) {
        onSuccess();
      } else if (result.error) {
        onError(result.error);
      } else {
        onError(`An unknown exception occurred updating the config`);
      }
    } catch (error: unknown) {
      if (error instanceof Error) {
        onError(error.message);
      } else if (error && typeof error === 'object' && 'toString' in error) {
        onError(error.toString())
      } else {
        onError(`An unknown exception occurred....we would need to make this distinct from the other exception where we have no idea what happened`);
      }
    }
```

assets/src/util/admin.tsx
```diff
@@ -231,3 +231,47 @@ const doFetch = async (
...
+    if (deletedScreenIds.length > 0) {
+      const deleteResponse = await fetch.post(
+        "/api/admin/screen_configs/delete",
+        {
+          deleted_screen_ids: deletedScreenIds,
+        },
+      );
+
+      if (!deleteResponse.success) return deleteResponse;
+    }
```
rwaskiewicz: No action required - any reason not to fire both of these requests off at the same time and await them with `Promise.all`?

rwaskiewicz (approved):
LGTM! Thanks for making those changes. Two non-blocking comments

assets/src/util/admin.tsx
robbie-sundstrom: Good call! I didn't do this because I wanted to make the calls conditionally when the list of updates and deletes aren't empty, but something like the following would accomplish that in parallel. I think the only downside of this is that we would only show the first error in the case that both updates and deletes fail, but that seems reasonable to me.

```js
const [updateResponse, deleteResponse] = await Promise.all([
  changedConfigs.length > 0
    ? fetch.post("/api/admin/screen_configs/update", {
        screen_configs: changedConfigs,
      })
    : Promise.resolve({ success: true }),

  deletedScreenIds.length > 0
    ? fetch.post("/api/admin/screen_configs/delete", {
        deleted_screen_ids: deletedScreenIds,
      })
    : Promise.resolve({ success: true }),
]);

if (!updateResponse.success) return updateResponse;
if (!deleteResponse.success) return deleteResponse;

return { success: true };
```

assets/src/components/admin/admin_form.tsx
robbie-sundstrom: you're right! Parsing these `unknown`/`undefined` errors into strings seems like a good solution to me

PR #3232, opened 2026-08-11
chore: remove "pending screens"

digitalcora:
All code involving "pending screens" is currently unreachable, since this feature was only used by the now-removed Permanent Config feature in Screenplay. We have no plans to reuse this in the near future, and current tasks such as the database migration and the widget cache can be simplified if they don't have to support pending screens.


robbie-sundstrom (approved):
looks to me like you got everything!

PR #3236, opened 2026-08-13
chore: add exhaustive switch lint check

rwaskiewicz:
**Asana task**: [Screens: Type-Aware Linting](https://app.asana.com/1/15492006741476/project/1185117109217413/task/1216828252543714?focus=true)

Description

Enable `@typescript-eslint/switch-exhaustiveness-check` in the project
to evaluate `switch` statements for completeness with the following
options set:
- `allowDefaultCaseForExhaustiveSwitch` - disabled to detect cases
where `default:` is used in a `switch` that would otherwise be
exhaustive, preventing dead code
- `requireDefaultForNonUnion` - enabled to enforce a `default` when
switching over a type that is not a union/necessarily exhaustive, such
as a `number` or `string`
- `considerDefaultExhaustiveForUnions` - enabled to allow for a single
case in the codebase where a "base" component accepts a wider range of
unioned strings than we provide in practice

Doing so required expanding the `languageOptions` in the eslint config.
`projectService` was used to add type semantics to linting, per the
recommendation of typescript-eslint. A handful of files are not
connected to the project's `tsconfig.json`, which resulted in errors
when linting. They've been added under the `allowDefaultProject`
property to be able to lint them moving forward.

- [ ] Tests added?

### DIFF
--- a/assets/eslint.config.mjs
+++ b/assets/eslint.config.mjs
@@ -7,7 +7,21 @@ import eslintConfigPrettier from "eslint-config-prettier";
 import jestPlugin from "eslint-plugin-jest";
 
 export default [
-  { languageOptions: { globals: globals.browser } },
+  {
+    languageOptions: {
+      globals: globals.browser,
+      parserOptions: {
+        projectService: {
+          allowDefaultProject: [
+            "eslint.config.mjs",
+            "jest.config.js",
+            "stylelint.config.mjs",
+            "webpack.config.js",
+          ],
+        },
+      },
+    },
+  },
   pluginJs.configs.recommended,
   ...tseslint.configs.recommended,
   pluginReact.configs.flat.recommended,
@@ -50,6 +64,14 @@ export default [
           caughtErrorsIgnorePattern: "^_",
           destructuredArrayIgnorePattern: "^_",
           varsIgnorePattern: "^_",
+        },
+      ],
+      "@typescript-eslint/switch-exhaustiveness-check": [
+        "error",
+        {
+          allowDefaultCaseForExhaustiveSwitch: false,
+          requireDefaultForNonUnion: true,
+          considerDefaultExhaustiveForUnions: true,
         },
       ],
     },
--- a/assets/src/components/elevator_status.tsx
+++ b/assets/src/components/elevator_status.tsx
@@ -1,6 +1,6 @@
 import cx from "classnames";
 import { QRCodeSVG as QRCode } from "qrcode.react";
-import { type ComponentType, type JSX } from "react";
+import { type ComponentType } from "react";
 
 import FreeText, { type FreeTextType } from "./free_text";
 
@@ -27,9 +27,7 @@ type Props = {
   qr_code_url: string;
 };
 
-const StatusIcon: ComponentType<{ status: ElevatorStatus }> = ({
-  status,
-}): JSX.Element => {
+const StatusIcon: ComponentType<{ status: ElevatorStatus }> = ({ status }) => {
   switch (status) {
     case "alert":
       return <ElevatorAlertIcon width={280} height={160} />;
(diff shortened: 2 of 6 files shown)


assets/eslint.config.mjs
```diff
@@ -52,6 +66,14 @@ export default [
           varsIgnorePattern: "^_",
         },
       ],
+      "@typescript-eslint/switch-exhaustiveness-check": [
+        "error",
+        {
+          allowDefaultCaseForExhaustiveSwitch: false,
+          requireDefaultForNonUnion: true,
+          considerDefaultExhaustiveForUnions: true,
```
rwaskiewicz, line 74: With `considerDefaultExhaustiveForUnions` enabled, some of the changes in this PR aren't strictly needed:
1. We don't _need_ to replace [the `default` clause of the `switch` statement for the DUP's `nextSizingState` function](https://github.com/mbta/screens/pull/3236/changes#diff-23119b6ebd3917c30b9a802fc36b335adda04a1b2ac9153259c8ed4181c68c1bR102), which would means the return type would remain the same
2. We don't _need_ to replace the `default` clause of the `switch` statement for DUPs [here](https://github.com/mbta/screens/pull/3236/changes#diff-01c57ea0ac05fa28156e9770ba689d7adcff11c5302a3ecfe06402eea66fa56fR23)

However, I think that both of these changes make the codebase a little healthier in terms of type correctness (particularly the first example). 

That said, one might say "why enable this option at all?". Currently, [this function](https://github.com/mbta/screens/blob/49dbc27f8312312d59442e5b97e09cd757457c5d/assets/src/components/bus_shelter/alert.tsx#L108-L119) does not handle "snowflake" for full body alerts in bus shelters. I'd be open to updating the types here (assuming that's possible) to limit `filenameForFullBodyIcon` to not accept "snowflake", but I'd like confirmation from the team that's a valid thing to do (a look at git history suggests that'd be OK, but wanted to double check)

assets/src/components/pre_fare_single_screen_alert.tsx
```diff
@@ -368,89 +368,86 @@ const PreFareSingleScreenAlert: ComponentType<PreFareSingleScreenAlertProps> = (
   } = alert;
 
   /**
-   * This switch statement picks the alert layout
+   * This series of if/else statements picks the alert layout:
```
rwaskiewicz, line 371: I recommend reviewing this file with whitespace hidden, as it'll show the change from `switch` -> `if/else` a little more clearly

assets/src/components/elevator_status.tsx
```diff
@@ -27,9 +27,7 @@ type Props = {
   qr_code_url: string;
 };
 
-const StatusIcon: ComponentType<{ status: ElevatorStatus }> = ({
-  status,
-}): JSX.Element => {
+const StatusIcon: ComponentType<{ status: ElevatorStatus }> = ({ status }) => {
```
rwaskiewicz, line 30: The removed return type is what originally sparked the conversation around this check. By enabling this lint check, we don't need the return type to act as a proxy for switch exhaustiveness (previously, if we added a new variant to `ElevatorStatus` and didn't update this `switch`, the return type would become `JSX.Element | undefined`, which is not assignable to the previously declared `JSX.Element`).

assets/eslint.config.mjs
```diff
@@ -7,7 +7,21 @@ import eslintConfigPrettier from "eslint-config-prettier";
...
 
 export default [
-  { languageOptions: { globals: globals.browser } },
+  {
+    languageOptions: {
+      globals: globals.browser,
+      parserOptions: {
+        projectService: {
+          allowDefaultProject: [
+            "eslint.config.mjs",
```
rwaskiewicz, line 16: I'm not entirely convinced we'll get a ton of value adding type-specific linting to these, but I like the idea of maybe someday helping us catch an lintable error for something like the webpack config.

assets/eslint.config.mjs
digitalcora: I think I agree with this option being enabled since it allows having unions where, for some functions that deal with them, they can have special handling for certain values and a "fallback" for all not-specifically-handled values, using `default`. This feels unlikely to allow through any real correctness issues. Definitely agree with all the "not strictly needed" changes made here, though!

digitalcora (approved):
✨

PR #3237, opened 2026-08-14
deps: update elixir to 1.20.2 

deanshi:
**Asana task**: [Update Screens to Elixir 1.20 / OTP 29](https://app.asana.com/1/15492006741476/project/1185117109217413/task/1215538526595324?focus=true)

Description
- Updated Elixir to 1.20.2 OTP 29 and erlang to 29.0.2
- Cleaned up compiler warnings, removed a test that's now caught by a compiler warning and removed logger backend + added the LoggerBackends helper for Sentry

@robbie-sundstrom - looking at your previous code, I attempted to do the dependency upgrades but ran into an issue with clashing `hackney` versions for `timex` -> `tzdata` and `ex_aws`. I think this is fixable by replacing our few `timex` calls with something else, but after spending some time trying to replace it I decided that it would be best as a follow up task instead of in this PR. ~I'll create a ticket/add that code in after this.~ Ticket [here](https://app.asana.com/1/15492006741476/project/1185117109217413/task/1217467699269071?focus=true). 

- [x] Tests modified?


config/config.exs
```diff
@@ -23,10 +23,6 @@ config :screens, Screens.Repo,
 
 config :screens, ecto_repos: [Screens.Repo]
 
-# Include 2 logger backends
-config :logger,
-  backends: [:console, Sentry.LoggerBackend]
-
 # Configures Elixir's Logger
 config :logger, :console,
```
rwaskiewicz: Minor, I think this should be changed to
```suggestion
config :logger, :default_formatter,
```

When I was converting Screenplay I was looking at the [Backend and backwards compatibility](https://logger.hexdocs.pm/1.20.3/Logger.html#module-backends-and-backwards-compatibility) section of the the Logger docs, which seemed to imply that `:console` will still work, but the mappings changed under the hood - specifically this bullet:

> The :console backend configuration is automatically mapped to the default handler and default formatter. Previously, you would set: 
> ``` 
> config :logger, :console,
>   level: :error,
>   format: "$time $message $metadata"
>   ```
> This is now equivalent to:
> ```
> config :logger, :default_handler,
>   level: :error
> 
> config :logger, :default_formatter,
>   format: "$time $message $metadata"
> ```

mix.lock
```diff
@@ -6,39 +6,40 @@
   "combine": {:hex, :combine, "0.10.0", "eff8224eeb56498a2af13011d142c5e7997a80c8f5b97c499f84c841032e429f", [:mix], [], "hexpm", "1b1dbc1790073076580d0d1d64e42eae2366583e7aecd455d1215b0d16f2451b"},
   "connection": {:hex, :connection, "1.1.0", "ff2a49c4b75b6fb3e674bfc5536451607270aac754ffd1bdfe175abe4a6d7a68", [:mix], [], "hexpm", "722c1eb0a418fbe91ba7bd59a47e28008a189d47e37e0e7bb85585a016b2869c"},
   "corsica": {:hex, :corsica, "2.1.3", "dccd094ffce38178acead9ae743180cdaffa388f35f0461ba1e8151d32e190e6", [:mix], [{:plug, "~> 1.0", [hex: :plug, repo: "hexpm", optional: false]}, {:telemetry, "~> 0.4.0 or ~> 1.0", [hex: :telemetry, repo: "hexpm", optional: false]}], "hexpm", "616c08f61a345780c2cf662ff226816f04d8868e12054e68963e95285b5be8bc"},
-  "cowboy": {:hex, :cowboy, "2.17.0", "059d6ae769214cedf55d115ce5bf38649ff6b32ec693067cf6b185d0ab1b53c3", [:make, :rebar3], [{:cowlib, ">= 2.18.0 and < 3.0.0", [hex: :cowlib, repo: "hexpm", optional: false]}, {:ranch, ">= 1.8.0 and < 3.0.0", [hex: :ranch, repo: "hexpm", optional: false]}], "hexpm", "84f0e4c9d5820342cd506472052b79336d0d9d3624e46303f6d0af9dc94e8d5f"},
```
rwaskiewicz, line 9: Double checking - are these package updates necessary, or are they leftover from the original package update attempts mentioned in the PR summary?

lib/screens/application.ex
```diff
@@ -36,6 +36,8 @@ defmodule Screens.Application do
       ]
       |> Enum.reject(&is_nil/1)
 
+    _ = LoggerBackends.add(Sentry.LoggerBackend)
```
rwaskiewicz, line 39: For my own understanding, why do we need to assign the result of this call? I.E. why not

```suggestion
    LoggerBackends.add(Sentry.LoggerBackend)
```

rwaskiewicz (approved):
LGTM - only non-blocking questions/comments

lib/screens/application.ex
deanshi: We get a dialyzer warning for this. I think it's just warning that we could get an ok/error code from the `LoggerBackends.add/1`, but we ignore it. I think we could also match like we do underneath `:ok = Logster.attach_phoenix_logger()` and crash if we don't match for `{:ok, _}`.

I think in other projects we don't get this warning right?

mix.lock
deanshi: Yep, probably from update attempts. I've reverted it all and it seems to be working no problem.

lib/screens/v2/localized_alert.ex
```diff
@@ -151,21 +151,6 @@ defmodule Screens.V2.LocalizedAlert do
     informed_entity_to_zone(Map.put(entity, :route, nil), context)
   end
 
-  # Both stop and route are not nil (route type ignored)
-  defp informed_entity_to_zone(
```
digitalcora, line 155: This seems like it might actually be the compiler unearthing a bug. On L130 the comment for the following clause says "*only* stop is not nil" (emphasis mine), meaning that the intent was that this clause was only entered if route *is* nil — however, it doesn't actually match on that. This discrepancy between the apparent intent and the actual implementation is what causes _this_ clause to be unreachable.

lib/screens/application.ex
rwaskiewicz: Ah, gotcha. I hadn't seen that in Screenplay 🤔 But that answers my original question, thanks!

lib/screens/v2/localized_alert.ex
deanshi: Good catch - I added `route: nil` to that particular function, added a `not_is_nil(route)` guard to the one on L144 and removed the one on L149 since it overlaps with this removed function. I re-added this function (with a `not is_nil(route)` guard) and also fixed tests (by just adding appropriate routes to the `LocationContext`)

I think this now covers all cases and matches them properly (everything nil, route_type is not nil, stop is not nil, route is not nil, stop and route are not nil).

PR #3241, opened 2026-08-18
deps: remove timex and update other dependencies

deanshi:
**Asana task**: [Screens: dependency upgrades for `ex_aws`](https://app.asana.com/1/15492006741476/project/1185117109217413/task/1217467699269071?focus=true)

Description
- replaced `timex` with `datix` for specifically date parsing
- removed `timex` as a dependency
- updated `ex_aws` and added ~`hackney` 4.0.1 as a dependency~ `ssl_verify_fun` 1.1 as a dependency
- replaced `Tzdata` with `Tz`
- fixed warning by adding `listeners: [Phoenix.CodeReloader]`

The String DateTime parsing was replaced with `[datix](https://datix.hexdocs.pm/api-reference.html)` for our need in S3 fetching and GDS logging. Let me know if there's an alternative to this (maybe just regex?) as opposed to pulling in a separate dependency.


mix.exs
```diff
@@ -61,13 +62,14 @@ defmodule Screens.MixProject do
...
-      {:ex_aws, "~> 2.1"},
+      {:ex_aws, "~> 2.7"},
       {:ex_aws_polly, "~> 0.5.0"},
       {:ex_aws_rds, "~> 2.0.2"},
-      {:ex_aws_s3, "~> 2.1"},
+      {:ex_aws_s3, "~> 2.5"},
       {:ex_cldr_messages, "~> 2.0"},
       {:gettext, "~> 0.26.1"},
       {:guardian, "~> 2.4.0"},
+      {:hackney, "~> 4.0.1"},
```
rwaskiewicz: For my own understanding, why did we add hackney here? Was it for `ex_aws` to use as an HTTP backend, or something else?

If the former, I think we could bump `req` to `0.7.3` and use that (in lieu of hackney).

mix.exs
```diff
@@ -91,8 +93,8 @@ defmodule Screens.MixProject do
       {:telemetry, "~> 1.2"},
       {:telemetry_metrics, "~> 1.0"},
       {:telemetry_poller, "~> 1.0"},
-      {:timex, "~> 3.6"},
-      {:tzdata, "~> 1.1"},
+      {:datix, "~> 0.3"},
```
rwaskiewicz: Non-blocking nit - it looks like we alphabetize this list of deps

lib/screens/v2/widget_instance/line_map.ex
```diff
@@ -275,8 +275,11 @@ defmodule Screens.V2.WidgetInstance.LineMap do
     if prediction_count < 2 and not is_nil(departure) do
       %{name: origin_stop_name} = Enum.at(stops, 0)
 
-      {:ok, timestamp} =
-        departure |> Departure.time() |> Util.to_eastern() |> Timex.format("{h12}:{m}")
+      timestamp =
+        departure
+        |> Departure.time()
+        |> Util.to_eastern()
+        |> Calendar.strftime("%-I:%M")
```
rwaskiewicz, line 282: For my own understanding, what does the `-` in `%-I` do?

deanshi: Let me _triple_ check that it works here, but from what I remember it doesn't pad the leading 0 for the hour (7:20 instead 07:20).

mix.exs
deanshi: To be honest, I took this change from the old Elixir upgrade branch and I didn't question this at all. Let me see if I can even just remove this explicit declaration (and I'll look into the req bump as well). 

From what I remember though, `ex_aws` does use hackney 4.0.0, but I don't remember any conflicts with `req`.

deanshi: Right now I remember - removing the explicit `hackney` dependency gives us this warning when we compile: 

```warning: :ssl_verify_hostname.verify_fun/3 is undefined (module :ssl_verify_hostname is not available or is yet to be defined)```

Since we don't explicitly use `hackney` anywhere, I removed this and included the `ssl_verify_fun` dependency instead.

File: lib/screens/v2/widget_instance/line_map.ex (hunk 3 above)
Comment (deanshi) [reply lib/screens/v2/widget_instance/line_map.ex:282]: Okay verified - so `Timex.format()` above leads to `{:ok, "8:48"}` without the leading zero. If you don't include the `-` when using `Calendar.strftime`, you'll get `"08:48"`.

File: mix.exs (hunk 1 above)
Comment (rwaskiewicz) [reply mix.exs]: Did you push this change? I see `ssl_verify_fun`, but I don't see `hackney` removed

File: lib/screens/v2/widget_instance/line_map.ex (hunk 3 above)
Comment (rwaskiewicz) [reply lib/screens/v2/widget_instance/line_map.ex:282]: Gotcha, thanks! Somehow I missed that in the docs, sorry for the 🪿 chase!

File: mix.exs (hunk 1 above)
Comment (digitalcora) [reply mix.exs]: Warnings like these when compiling *dependencies*, which themselves have optional dependencies, are normal and should not fail CI. We definitely shouldn't have to include `ssl_verify_fun` as an explicit top-level dependency of ours.

File: config/config.exs [hunk 4]
```diff
@@ -60,7 +57,7 @@ config :phoenix, :format_encoders,
 
 config :ex_aws, http_client: ExAws.Request.Req, json_codec: Jason
 
-config :elixir, :time_zone_database, Tzdata.TimeZoneDatabase
+config :elixir, :time_zone_database, Tz.TimeZoneDatabase
```
digitalcora, line 63: Given [the comment](https://app.asana.com/1/15492006741476/task/1217467699269071/comment/1217464440202824?focus=true) I made on this task (and you 👍'd), is there still a notable reason to change which time zone library we're using here?

mix.exs
```diff
@@ -16,7 +16,8 @@ defmodule Screens.MixProject do
           :unmatched_returns
         ]
       ],
-      test_coverage: [tool: LcovEx]
+      test_coverage: [tool: LcovEx],
+      listeners: [Phoenix.CodeReloader]
```
digitalcora, line 20: Why is this included here? More leftovers from the other branch?

lib/screens/util/last_deploy/s3_fetch.ex
```diff
@@ -6,15 +6,19 @@ defmodule Screens.Util.LastDeploy.S3Fetch do
...
 
     with {:request, {:ok, %{headers: headers}}} <- {:request, ExAws.request(operation)},
          {:header, {:ok, value}} <- {:header, headers |> Map.new() |> Map.fetch("last-modified")},
-         {:parse, {:ok, %DateTime{} = datetime}} <- {:parse, Timex.parse(value, @rfc2822)} do
+         # "Mon, 10 Aug 2026 15:11:37 GMT"
+         {:parse, {:ok, %DateTime{} = datetime}} <-
+           {:parse,
+            Datix.DateTime.parse(
+              String.replace_suffix(value, " GMT", ""),
+              "%a, %d %b %Y %H:%M:%S"
```
digitalcora: Rather than replacing a trailing `GMT` here and then doing the parsing, the format string could include it as a literal:

```
%a, %d %b %Y %H:%M:%S GMT
```

We'd want this to fail anyway if the zone was not as we expect, since if we blindly discarded it and interpreted the time as UTC, we'd get the wrong time.

config/config.exs
deanshi: My only thought (post 👍'ing) was that I wasn't sure what the release cadence of `Tzdata` is, so I'm not sure if this will be blocked for a substantial time. If we're okay with waiting, I'm good with holding off on this change.

mix.exs
deanshi: This one seems to be a warning I missed from the [previous upgrade](https://github.com/mbta/screens/pull/3237/changes/a914ba869efde9de6390fbd094e3c6f6852a7558).

mix.exs
deanshi: I believe this is actually a [usage in our code](https://github.com/mbta/screens/blob/ae64fe306cb216897f414762db9f62ecbc989cb2/lib/screens/repo.ex#L28) rather than compiling dependencies. Is there a good way around this without needing to explicitly define the dependency?

rwaskiewicz (approved):
LGTM - last suggestion is that we clean up the lockfile following these package updates. On `main`, running `mix deps.unlock --unused` returns nothing, but on this branch, I think we have room to clean some things up: 

```
➜  screens git:(deanshi/update_deps_remove_timex) mix deps.unlock --unused                     
Unlocked deps:
* certifi
* h2
* hackney
* idna
* metrics
* mimerl
* parse_trans
* quic
* timex
* tzdata
* unicode_util_compat
```

digitalcora (approved):

mix.exs
digitalcora: Oh, good catch, I rescind my comment!

PR #3247, opened 2026-08-20
feat: update departure times independent of screen refresh

rwaskiewicz:
**Asana task**: [Screens client-side countdowns](https://app.asana.com/1/15492006741476/project/1185117109217413/task/1216571034745361?focus=true)

Description

This commit allows departure times to "count down" on screens, without the need for updated data from the server.

This is accomplished by serializing the departure time in the payload sent to the client, using the `time_in_epoch` field similar to responses to Mercury. Note that `time_in_epoch` serializes using a granularity of minutes for Mercury. To keep the data consistent, the same is done for non-Mercury responses.

When the client renders, it maintains a record of the current time that is updated on a one second interval. This triggers rerenders of the `DepartureTimePart` component on a ~1 second interval, rather than the existing screen refresh rate. The difference between the timestamp sent in our last response from the server and "now" is calculated, which informs the client if we can "count down" from the last provided departure time.

- [ ] Tests added?

### DIFF
--- a/assets/src/components/departures/departure_time.tsx
+++ b/assets/src/components/departures/departure_time.tsx
@@ -1,4 +1,4 @@
-import type { ComponentType } from "react";
+import { ComponentType, useEffect, useState } from "react";
 
 import { useCurrentPage } from "Context/dup_page";
 import MoonIcon from "Images/moon.svg";
@@ -14,13 +14,47 @@ type DepartureTime =
 
 interface DepartureTimePartProps {
   time: DepartureTime;
+  timeInEpochMinutes: number;
   currentPage: number;
 }
 
+/**
+ * Given two dates, compute the difference in seconds between `departureTimeSeconds`
+ * and `currentDateTime`.
+ *
+ * If the difference between the two values in minutes results in a float,
+ * round the result (rounding method is considered an implementation detail).
+ * This function will always return a value of 1 or greater - the system relies
+ * on the backend to provide the appropriate text ("ARR", "BRD", etc.) in lieu
+ * of showing '0 min'.
+ *
+ * @param departureTimeSeconds A datetime provided by the backend, representing
+ * the departure time of a route at a stop.
+ * @param currentDateTime The current datetime of the browser
+ */
+const adjustMinute = (
+  departureTimeSeconds: Date,
+  currentDateTime: Date,
+): number => {
+  const timeDifferenceMin =
+    departureTimeSeconds.getTime() - currentDateTime.getTime() / 1000;
+  const timeDifferenceSeconds = Math.floor(timeDifferenceMin / 60);
+  return Math.max(timeDifferenceSeconds, 1);
+};
+
 const DepartureTimePart: ComponentType<DepartureTimePartProps> = ({
   time,
+  timeInEpochMinutes,
   currentPage,
 }) => {
+  const [currentDateTime, setCurrentDateTime] = useState(new Date());
+  useEffect(() => {
+    const intervalTimer = setInterval(() => {
+      setCurrentDateTime(new Date());
+    }, 1000);
+
+    return () => clearInterval(intervalTimer);
+  }, []);
   switch (time.type) {
     case "text":
       return <div className="departure-time__text">{time.text}</div>;
@@ -28,7 +62,9 @@ const DepartureTimePart: ComponentType<DepartureTimePartProps> = ({
     case "minutes":
       return (
         <>
-          <div className="departure-time__minutes">{time.minutes}</div>
+          <div className="departure-time__minutes">
+            {adjustMinute(new Date(timeInEpochMinutes), currentDateTime)}
+          </div>
           <div className="departure-time__minutes-label">m</div>
         </>
       );
@@ -58,12 +94,14 @@ const DepartureTimePart: ComponentType<DepartureTimePartProps> = ({
 
 interface Props {
   time?: DepartureTime;
+  time_in_epoch: number;
   scheduled_time?: DepartureTime;
   is_live: boolean;
 }
 
 const DepartureTime: ComponentType<Props> = ({
   time,
+  time_in_epoch: timeInEpoch,
   scheduled_time,
   is_live: isLive,
 }) => {
@@ -80,7 +118,11 @@ const DepartureTime: ComponentType<Props> = ({
             className="departure-time__live-icon"
           />
         )}
-        <DepartureTimePart currentPage={currentPage} time={time} />
+        <DepartureTimePart
+          currentPage={currentPage}
+          time={time}
+          timeInEpochMinutes={timeInEpoch}
+        />
       </div>
     );
   } else if (scheduled_time && (currentPage === 1 || !time)) {
@@ -91,7 +133,13 @@ const DepartureTime: ComponentType<Props> = ({
           time ? "delayed" : "cancelled",
         ])}
       >
-        <DepartureTimePart {...{ time: scheduled_time, currentPage }} />
+        <DepartureTimePart
+          {...{
+            time: scheduled_time,
+            timeInEpochMinutes: timeInEpoch,
+            currentPage,
+          }}
+        />
       </div>
     );
   } else {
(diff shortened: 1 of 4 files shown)


assets/src/components/departures/departure_time.tsx
```diff
@@ -80,7 +118,11 @@ const DepartureTime: ComponentType<Props> = ({
             className="departure-time__live-icon"
           />
         )}
-        <DepartureTimePart currentPage={currentPage} time={time} />
+        <DepartureTimePart
+          currentPage={currentPage}
+          time={time}
+          timeInEpochMinutes={timeInEpoch}
```
robbie-sundstrom: Does this actually have a minute granularity? From testing, this seems to be a second based time since the numbers will decrement at different times. 

As an aside, the numbers counting down independently of each other makes the predictions feel much more "real". I'm a big fan.

rwaskiewicz: Nope, it's seconds - good catch! Last minute (heh) rename from `timeInEpoch` with the wrong unit 🤦 Fixed in https://github.com/mbta/screens/pull/3247/commits/afe9cb5b4a0380a27495fb2a7fcd405879280039

assets/src/components/departures/departure_time.tsx
```diff
@@ -14,21 +14,57 @@ type DepartureTime =
...
 const DepartureTimePart: ComponentType<DepartureTimePartProps> = ({
   time,
+  timeInEpochSeconds,
   currentPage,
 }) => {
+  const [currentDateTime, setCurrentDateTime] = useState(new Date());
+  useEffect(() => {
+    const intervalTimer = setInterval(() => {
+      setCurrentDateTime(new Date());
+    }, 1000);
```
digitalcora: nit: It seems like we shouldn't need to fire an interval every second, since right from the start we know exactly when the countdown should change; when we receive new data we could `setTimeout` for that point in time, and count down when we reach it (and set a new timeout for the next "countdown point" ~60s in the future, though with current refresh rates it'd be impossible to ever get there).

assets/src/components/departures/departure_time.tsx
```diff
@@ -14,21 +14,57 @@ type DepartureTime =
...
+): number => {
+  const timeDifferenceMin =
+    departureTimeSeconds.getTime() - currentDateTime.getTime() / 1000;
+  const timeDifferenceSeconds = Math.floor(timeDifferenceMin / 60);
+  return Math.max(timeDifferenceSeconds, 1);
+};
+
 const DepartureTimePart: ComponentType<DepartureTimePartProps> = ({
   time,
+  timeInEpochSeconds,
```
digitalcora: nit: It feels potentially confusing that we call this value `timeInEpochSeconds` here and just `timeInEpoch` elsewhere (implying the latter is *not* seconds, but something else... milliseconds?). In my mind the `in_epoch` part already means seconds, so it'd probably be fine not to specify that.

assets/src/components/departures/departure_time.tsx
rwaskiewicz: I'm not sure I follow - are you saying that we should have `setTimeout`'s function parameter invoked every minute, or some other calculated value?

assets/src/components/departures/departure_time.tsx
rwaskiewicz: Good point, done in https://github.com/mbta/screens/pull/3247/commits/b9e211ceeae22a93138a0bcae439bfb240b61d76

assets/src/components/departures/departure_time.tsx
digitalcora: Concrete example: The component renders with a `timeInEpoch` corresponding to 11:35:00. As part of this update (in an effect having `timeInEpoch` as a dependency) we look at the current time and find it's 11:30:20. This means the current displayed time would be `5 min`, but we know according to the display rules that in exactly 11 seconds it should change to `4 min`. Therefore we don't have to wake up every second to check, effectively, "has it been 11 seconds yet?" — we already know how far in the future that is and can set a single timeout for 11 seconds from now. When this timer is up, we both set the countdown display according to the new current time, and set a new timer for the next time we should update it. (This process can't make any assumptions about when it's running, since it's possible the device clock has changed, the timeout was suspended by the browser, etc.)

I haven't sketched out how this would actually look in React code; it's possible doing it this way would be more complex than I think, based on the constraints of the framework, and the approach of "ticking" every second is definitely going to be simpler in any case — I offer this as an alternative and not a strong opinion about which way we should do it.

rwaskiewicz: Ahh I see - thanks for the clarification. I err on the side of not implementing it as described, as I think (IIUC) that'll lead to a less responsive "feeling" countdowns.

I roughly implemented this as follows:
```diff --git a/assets/src/components/departures/departure_time.tsx b/assets/src/components/departures/departure_time.tsx
index 1dc1c934..9befbe99 100644
--- a/assets/src/components/departures/departure_time.tsx
+++ b/assets/src/components/departures/departure_time.tsx
@@ -49,11 +49,14 @@ const DepartureTimePart: ComponentType<DepartureTimePartProps> = ({
 }) => {
   const [currentDateTime, setCurrentDateTime] = useState(new Date());
   useEffect(() => {
-    const intervalTimer = setInterval(() => {
+    // Calculate the number of milliseconds until the next "01" or "31"
+    const nextFireTimeMs = (30 - ((new Date(currentDateTime)).getSeconds() % 30) + 1) * 1_000;
+    console.log("Got",(new Date(currentDateTime)).getSeconds(),"Firing in", nextFireTimeMs)
+    const timer = setTimeout(() => {
       setCurrentDateTime(new Date());
-    }, 1000);
+    }, nextFireTimeMs);
 
-    return () => clearInterval(intervalTimer);
+    return () => clearTimeout(timer);
   }, []);
   switch (time.type) {
     case "text":

```

And found that our rerenders started to fall (roughly) every 16 or 31 seconds. Assuming a refresh rate of 20s, that would mean that the former calculation would only be valid for ~4s, and the latter would be thrown out entirely. But, if I've misunderstood, LMK and I'll take a closer look

robbie-sundstrom (approved):

assets/src/components/departures/departure_time.tsx
digitalcora: > I think (IIUC) that'll lead to a less responsive "feeling" countdowns.

I am not sure if we U each other C, then, since the rider-facing behavior of what I'm trying to describe should be nearly identical to what's currently implemented in this PR, with the main difference being that it would trigger fewer re-renders/wake-ups where ultimately the displayed time doesn't change — in both cases, we should update the time when it "should" update according to the display rules with at least second-level accuracy.

In your example change, it looks like you're setting up a timeline of updating the displayed time at every absolute clock time where the second value is either 01 or 31. This isn't what I described, though the second value of 31 happens to coincide with the example times I used (which I did that way so it would be, hopefully, easier to understand 😅). Here's another possible example: The component renders with `timeInEpoch` corresponding to 11:35:17. The current time is 11:30:39. The initial displayed time should be "5 min", but it should count down to "4 min" exactly 9 seconds from now. The actual clock time at which this occurs is 11:30:48, but that's not important (this may occur at *any* number of "clock seconds"), the point is the *"9 seconds from now"*, which is the moment the actual minutes-until-departure drops below 4.5, and therefore we should display this as "4 min".

rwaskiewicz: Ahh Ok - done in https://github.com/mbta/screens/pull/3247/commits/12e03815f7b5a11fb8402bf882d52b5d14df5596. Thanks for your patience/clarification!

rwaskiewicz: Err...nvm. I have to rework this. I don't have linting properly surfacing in my IDE and missed the circular dep that commit introduced 🤦

rwaskiewicz:
Reverted the changes to calculate the number of seconds for `setTimeout` - I wasn't happy with the level of complexity needed to avoid the circular dependencies. Should be good to review again.

digitalcora:
It was a "nit" in the first place and in any case @robbie-sundstrom's approval stands (I assume). There is not really any issue with correctness as far as I can see, so we could look into this as an optimization later if we have free cycles.

rwaskiewicz:
@digitalcora For sure, just wanted to make sure why I reverted was documented - non-zero chance I'd forget in a month's time

robbie-sundstrom (approved):

PR #3249, opened 2026-08-20
fix: Fallback to mbta.com/alerts for long alert URLs

robbie-sundstrom:


digitalcora (approved):

digitalcora:
Should this revert #3248? It seems like that logic would pretty much never occur after making this change.

robbie-sundstrom:
@digitalcora I think that's reasonable, will add the revert of that safeguard

PR #3250, opened 2026-08-21
feat: Admin Maintenance tool uses Postgres

robbie-sundstrom:
**Asana task**: [Screens Admin: Use new APIs for Configs](https://app.asana.com/1/15492006741476/project/1185117109217413/task/1216329820920785?focus=true)

Most of the other Admin API controller endpoints where completed as part of the initial API implementation, so this `maintenance` endpoint that counts and then removes expired Evergreen Content items from screens is the last one. 

Ended up moving the actual filtering of configs that need to be cleaned up into the SQL query, rather than filtering afterwards. Since this is a complicated query, a `fragment` with just the raw query seems like the simplest way to do this.


digitalcora:
> Ended up moving the actual filtering of configs that need to be cleaned up into the SQL query, rather than filtering afterwards.

I'm not convinced this is the best move and would be interested to hear your reasoning on it. The advantage would seem to be performance, but this feels like premature optimization given the performance of loading only ~500 records should be acceptable for an administrative tool we only use very occasionally. The disadvantage is that this complex query is invisible to the Elixir compiler in terms of the config struct definitions, meaning it can't catch mistakes or divergences before they become a problem at runtime (a manual code search for usages of the structs wouldn't even catch this, since they're just plain maps at the database level). Given this is specifically a deletion/cleanup tool, the impact of such a divergence might not be trivial — though most likely the tool just wouldn't work, and having an end-to-end test for it is a mitigating factor.

rwaskiewicz (commented):
LGTM - my only concern echoes Cora's with respect to using SQL directly and as a `fragment`.

robbie-sundstrom:
@digitalcora @rwaskiewicz my main reasonings for doing this filtering within SQL would not be unfair to categorize as mostly "premature optimization" with a sprinkle of "I think SQL is fun". Unfortunately, these arguments aren't as good as Cora's points. It would also be much easier to maintain this maintenance endpoint in the future if the filtering were done in Elixir, so I'll make that change later today

robbie-sundstrom:
@rwaskiewicz have cleaned up the Git mess I made and implemented the changes we talked about with Cora to remove the complex SQL fragment in this commit: [1a69dfa](https://github.com/mbta/screens/pull/3250/commits/1a69dfafa2a716bfc19494bd942b38fdf4a0c8ed). Also made this small change [2a68c98](https://github.com/mbta/screens/pull/3250/commits/2a68c98cb1ed6b1d558dd73c4b70fb370466b00d) to standardize all of the names in this module

rwaskiewicz (approved):

PR #3252, opened 2026-08-21
chore: partitioned widget caching/generation

digitalcora:
Introduce a partitioned cache for widget instances: Each node in the cluster "owns" a subset of screen IDs and is exclusively responsible for generating the corresponding widgets.

This has the following benefits:

* The memory load of the lower-level V3 API cache becomes distributed across nodes: Instead of every node storing the cached static data required to generate widgets for every screen (since over time every node receives requests for every screen), each node only stores the data needed for its own subset of screens.

* Requests for the same screen ID within the cache TTL can reuse the cached data instead of generating essentially the same widgets from scratch every time (even when these requests are routed to different nodes). This should especially save time for "duo" units, where two independent devices request the data for the same screen ID at nearly the same moment.

Initially the cache TTL is only 5 seconds, but the next phase of this work will add client-side "awareness" of the cache and allow the TTL to be increased to the full length of a screen refresh cycle, resulting in only ever generating a screen's contents once per refresh interval.

### Included changes

* Remove the self-refresh runner's "in progress" pseudo-locking feature; this is no longer needed now that screen data generation (when done via the cache) uses a true cluster-wide lock. Nodes only self-refresh screens they "own" in the cache.

* Lift widget generation out of `Layout` into `ScreenData`, removing an unnecessary layer of indirection and making `Layout` responsible only for the arrangement of widgets into a layout.

* Consolidate `ScreenAudioData` into `ScreenData` to reduce duplication and enable audio requests to reuse cached widget data from non-audio requests (or vice-versa).

**Asana task**: https://app.asana.com/1/15492006741476/project/1185117109217413/task/1216243936471573?focus=true


lib/screens/v2/screen_data/cache.ex
```diff
@@ -0,0 +1,122 @@
...
+          {:ok, result} -> result
+          {:error, error} -> {generate(screen), %Meta{what: {:error, :transaction, error}}}
+        end
+
+      {:ok, node} ->
+        # Another node owns the key; proxy the request to it. The timeout should be sufficient to
+        # wait for the longest possible candidate generation timeout; if we do exceed this, crash
+        # instead of retrying the generation ourself, since the app may be overloaded and at that
+        # point the client has been waiting on us for a long time.
+        case @store.call(node, __MODULE__, :instances, [id, screen], 20_000) do
```
robbie-sundstrom, line 86: This 20 second RPC timeout is less than the 30 second timeout we have in place for the departures candidate generator. We should bump this up to match if we're not confident in raising that timeout.

digitalcora: No candidate generator (as in, top-level `CandidateGenerator` module for a screen type) has a timeout longer than 20 seconds, so in practice that 30-second timeout will never be reached; it could even be infinite and it wouldn't matter.

robbie-sundstrom: Oh true! I guess that departures widget 30 second value is the one that can be adjusted in that case

lib/screens/v2/screen_data/cache.ex
```diff
@@ -0,0 +1,122 @@
...
+        # preserve the timeliness of displayed information, since refresh cycles running on
+        # screen clients are not yet aware they may be receiving cached data. The next phase of
+        # work will address this and enable instances to be cached for longer, up to the full
+        # length of a refresh cycle.
+        case @store.put(id, instances, ttl: :timer.seconds(5)) do
+          :ok -> {instances, %Meta{what: :miss}}
+          {:error, error} -> {instances, %Meta{what: {:error, :put, error}}}
+        end
+
+      {:error, error} ->
```
deanshi: I don't quite know the list of errors for Nebulex, but are there other reasons we might want to actually cache the generation as well?

lib/screens/v2/screen_data/cache.ex
```diff
@@ -0,0 +1,122 @@
...
+
+  import Screens.Inject
+  @parameters injected(Screens.V2.ScreenData.Parameters)
+  @store injected(Store)
+
+  @doc """
+  Fetch or generate the widgets for a given screen.
+
+  This function is "fail-safe" and simply generates widgets on the local node in the event of any
+  error interacting with the cache; the returned `Meta` indicates whether this happened.
```
robbie-sundstrom, line 64: Nit, but this doc is technically incorrect since an error can be raised on an RPC timeout

lib/screens/v2/screen_data/cache.ex
```diff
@@ -0,0 +1,122 @@
+defmodule Screens.V2.ScreenData.Cache do
+  @moduledoc """
+  Distributed partitioned cache that stores the set of widgets generated for a given screen ID.
+
+  Each node in the cluster "owns" a subset of screen IDs and is exclusively responsible for
```
deanshi, line 5: Is this affected by our auto-scaling for AWS? Would this be a problem if we scale down (or I guess even if we scale up?) with which nodes are available?

deanshi (approved):
I think things are looking good from my review, just a a basic knowledge question for my own understanding on how this partitioned cache works.

lib/screens/screens_by_alert/self_refresh_runner.ex
```diff
@@ -117,8 +116,16 @@ defmodule Screens.ScreensByAlert.SelfRefreshRunner do
       end)
   end
 
+  defp should_refresh_on_this_node?(id) do
+    case ScreenData.Cache.Store.find_node(id) do
+      {:ok, n} when n == node() -> true
+      {:ok, _other} -> false
+      {:error, _error} -> true
```
robbie-sundstrom: Nit: Should we record an error here before refreshing?

I was originally commenting on this because I was thinking that transient failures of `find_node` could cause fan outs of screen data to be stored on every node. However, given that the node lookup is deterministic, I don't think this will realistically happen. Instead, it seems like we should definitely be notified if it has returns an error at all.

deanshi: I was also thinking about this too - I think the recording an error here might be useful too potentially for an alert or dashboards? 

Edit: But I guess this is the `self_refresh_runner` so maybe we don't worry about it as much here?

digitalcora: Maybe it's fine if we just use `find_node!` and crash here — I've never seen this fail in reality, and glancing at the code it's possible the only way for it to fail is to misuse the underlying `ExHashRing` library in a way that we can probably trust Nebulex not to do. Crashing would certainly alert us to the problem and would also be the "fail safe" behavior in the case of the self-refresh runner (rather than always self-refreshing all screens locally, which could cause performance issues, we self-refresh *no* screens, and the only consequence is that Screenplay screens-by-alert functionality might be degraded).

lib/screens/v2/screen_data/cache.ex
digitalcora: This is the same error that can be raised by calling the candidate generator normally, so I think of it as not a behavior that using this module "introduces", but rather something a caller would have had to deal with anyway.

lib/screens/screens_by_alert/self_refresh_runner.ex
robbie-sundstrom: Using `find_node!` would also work here! Agreed that this would be a fine spot to raise an Exception and crash

lib/screens/v2/screen_data/cache.ex
digitalcora: This is something the Nebulex Partitioned cache implementation (and the underlying [ExHashRing](https://ex-hash-ring.hexdocs.pm/readme.html) library) handles for us — whatever nodes are online at the moment, the key space is automatically distributed between them. If we scale up or down*, the keys are redistributed (within ~15 seconds at most; DNS entries have a 10s TTL, DNSCluster check interval is 5s). That does mean that for a short time, a node can end up with cached V3 API data that is not needed to fulfill the requests for the screens it currently owns, because it *was* needed for screens it *recently* owned. But the cache GC should clean this up after it hasn't been used for 60 seconds.

_* which is technically something that happens every time we deploy, as well; the new version of the app gradually "scales up" from 0 to whatever instance count is desired_

deanshi: Perfect, appreciate the explanation!

PR #3253, opened 2026-08-21
feat: invalidate widget cache when configs change

digitalcora:
When a screen config is updated, any cached widget data for that screen is potentially no longer "valid", since the generated widgets depend on the config. Ensure the cache is invalidated when this happens. This will become more important for interactive tweaking of screens once the final cache improvements are implemented, since then cache entries will last for a longer time, and we don't want to have to wait around to see the results of our changes.


lib/screens_web/controllers/v2/screen_api_controller.ex
```diff
@@ -65,15 +67,18 @@ defmodule ScreensWeb.V2.ScreenApiController do
...
     Map.merge(%{@base_response | data: data}, %{flex_zone: flex_zone})
   end
 
-  defp screen_response(id, screen) do
-    %{@base_response | data: ScreenData.get(id, screen)}
+  defp screen_response(id, screen, opts) do
+    %{@base_response | data: ScreenData.get(id, screen, opts)}
   end
 
+  defp screen_data_options(%{"force_refresh" => "true"}), do: [refresh?: true]
```
robbie-sundstrom, line 79: This `force_refresh` is in the public API scope, when really it should require admin authentication. There might be a cleaner way to do this, but we could add a conditional plug that only runs when this parameter is set to true to check authentication

digitalcora: Thinking about this, I'm not sure we should even allow admins to force-refresh data arbitrarily often — with the locking changes it's actually a bit more of a footgun than it would have been previously, carrying the potential for a real screen request to fail because it was waiting too long on a lock taken by an admin refresh. Yet I do think we want to be able to force-refresh data sooner than waiting for a full refresh cycle to come around normally (will eventually be the case with the full cache implementation), even in production — with the main use case being editing a screen config.

I'll take this PR back to the drawing board... my initial idea here is:

1. No "force refresh" option for the normal non-admin data request endpoint.
2. Updating a screen's configuration clears the cache entry for that screen. This is appropriate in any case so the cache stays consistent with the configuration.
3. The Refresh button calls a new admin API endpoint specifically to clear the cache entry for a given screen; once this returns, it then signals the screen page to reload data, which because of the cache-busting that just occurred, will be fresh.
    * Doing this multiple times at once would not pile up multiple serialized requests; requests that occur after the first starts, but before it finishes, will re-clear the (already cleared, currently nonexistent) cache entry, wait on the lock, and then immediately retrieve the data that the first request put into the cache when it finished.

Let me know if anything sounds off about this approach!

robbie-sundstrom: I like this approach of splitting it into a separate admin API endpoint! I definitely agree that this Refresh button should force a refresh, and the path you laid out seems like a safer way to handle this.

digitalcora:
@robbie-sundstrom I've repurposed this PR as the core "cache consistency" changes, namely, when a screen config is updated, its cached data should be invalidated. This doesn't include the Inspector's Refresh button "hard refreshing" data, which I will do in a separate PR.

robbie-sundstrom (approved):
looks great to me!

PR #3255, opened 2026-08-21
feat: Replace Screens Cache Usage with Postgres behind FF

robbie-sundstrom:
**Asana task**: [Screens Config Cache: Replace with DB](https://app.asana.com/1/15492006741476/project/1185117109217413/task/1217735139031136?focus=true)

In several places throughout the application, we rely on a cache of Screens configurations to look up details for a specific screen. Since the lookups of screen configurations does not need to be high throughput and only supports ~500 screens currently, a cache won't provide us much (if any) performance improvement over looking things up in Postgres directly. Given that, these changes to conditionally look up in screen configs in Postgres assume that we will remove screen configurations from this cache after migration.


rwaskiewicz:
@robbie-sundstrom - you probably know, but there are merge conflicts on this PR. Can you please TAL?

robbie-sundstrom:
@rwaskiewicz thanks for the reminder! Just did a merge and resolved the conflicts

lib/screens/screen_configs.ex
```diff
@@ -76,6 +91,43 @@ defmodule Screens.ScreenConfigs do
     end
   end
 
+  @doc "Returns screen IDs for screens matching the given app ID."
+  @spec screen_ids_for_app(atom()) :: [screen_id()]
```
deanshi: super nit: could `atom()` be `Screen.app_id()`?

lib/screens/config/screen_config.ex
```diff
@@ -32,4 +32,14 @@ defmodule Screens.Config.ScreenConfig do
     |> changeset(attrs)
     |> apply_action(:insert)
   end
+
+  @spec load_config(term()) :: Screen.t() | term()
```
deanshi: Looking around, I didn't see any usages of this. Is this being used somewhere that I'm missing?

deanshi (approved):
Two tiny questions, otherwise :shipit:

lib/screens/config/screen_config.ex
robbie-sundstrom: Good catch, I removed the usage of this function

lib/screens/screen_configs.ex
robbie-sundstrom: yep, good call!

PR #3256, opened 2026-08-24
chore: added script for populating modes for params and support mode for encoding

deanshi:
**Asana task:** [Screens departures: explicit "section modes"](https://app.asana.com/1/15492006741476/project/1185117109217413/task/1216571034745364?focus=true)

Description
Takes the `local.json` and adds the `mode` parameter to all sections with departures and creates a `local.updated.json` with the idea that we can juse use that JSON to upload/replace any configuration we will eventually want (either locally, prod or any of the dev environments).

First we attempt to get the mode from the parameters that are set for the query. If that doesn't work, we request the departures for that section and actively use the `Route` that is returned to set the mode. If are getting no departures, we'll have to manually intervene and set that mode (by checking which sections have a `mode: null` in the configuration). Note that `header_only` sections and `Wayfinding` only sections will need to have a null mode, since they do not actively request departures. 

Also includes supporting `mode` in some our encoding for grabbing `RoutePattern` and `Departure`. 

Deployed and screens config has been updated in `dev-blue` to have modes populated for all departure sections.


scripts/add_mode_to_departures_params.exs
```diff
@@ -0,0 +1,172 @@
+Logger.configure(level: :warning)
```
rwaskiewicz, line 19: Non-blocking nit - can we add a usage comment similar to [this script](https://github.com/mbta/screens/blob/27d9a2582125d2e5e74b99173cc7b387fddfcca0/scripts/screen_destinations.exs#L1)? It might be useful for cases like understanding that we can use a `dry_run` flag when running the script

scripts/add_mode_to_departures_params.exs
```diff
@@ -0,0 +1,172 @@
...
+  defp mode_from_route_ids(route_ids) do
+    cond do
+      Enum.any?(route_ids, &(&1 == "Red")) ->
+        "rl"
+
+      Enum.any?(route_ids, &(&1 == "Blue")) ->
+        "bl"
+
+      Enum.any?(route_ids, &String.starts_with?(&1, "Green-")) ->
+        "gl"
```
rwaskiewicz, line 140: For my own understanding, why don't we have OL here?

scripts/add_mode_to_departures_params.exs
```diff
@@ -0,0 +1,172 @@
...
+        "bus"
+    end
+  end
+
+  defp mode_from_route_type("rail"), do: "cr"
+  defp mode_from_route_type("ferry"), do: "ferry"
+  defp mode_from_route_type(_), do: "bus"
+
+  defp mode_from_departure(params) do
+    Process.sleep(1_000)
```
rwaskiewicz, line 152: Double checking, is this `sleep` needed?

scripts/add_mode_to_departures_params.exs
```diff
@@ -0,0 +1,172 @@
+Logger.configure(level: :warning)
+
```
rwaskiewicz, line 20: Do we need to declare Jason as a dependency? (or, maybe should we if we plan on using the JSON module at some point in the near future?). When I try to run this as:
```
$ elixir scripts/add_mode_to_departure_params.exs
```
I get:
```
** (UndefinedFunctionError) function Jason.decode!/1 is undefined (module Jason is not available). Make sure the module name is correct and has been specified in full (or that an alias has been defined)
```

If I add the following to the top of the file:
```
Mix.install([{:jason, "~> 1.4"}])
```
I get:
```
** (UndefinedFunctionError) function Screens.V2.Departure.fetch/1 is undefined (module Screens.V2.Departure is not available)
```

So perhaps this is more of a PEBKAC issue?

scripts/add_mode_to_departures_params.exs
```diff
@@ -0,0 +1,172 @@
...
+        "light_rail" -> :light_rail
+        value -> value
+      end)
+      |> Map.new(fn {key, value} -> {String.to_atom(key), value} end)
+      |> Departure.fetch()
+
+    representative_departure = departures |> List.first()
+
+    if representative_departure == nil do
+      # This section will need manual intervention
```
rwaskiewicz, line 170: Is there anything worth noting here as to what we should do here? Not that I expect us to run this more than once, I just don't know what we need to do in this case. Maybe worth a warning log statement?

scripts/add_mode_to_departures_params.exs
deanshi: I think that was a mistake - it must have gotten deleted somehow between script runs/uploading this PR. Let me add this back in.

scripts/add_mode_to_departures_params.exs
deanshi: I thought about this too - initially I was attempting to do the `Mix.install` with `jason` but realized I didn't want to have to add all of the `Screens.V2.Departure` files and dependencies. So instead, I was using:

`mix run scripts/add_mode_to_departures_params.exs` 

to run it as a project and to just run the particular script instead. Let me know if this works out (or if I'm missing a better way to just include Departure and associated code.

scripts/add_mode_to_departures_params.exs
deanshi: Yeah, this was the special case where if we have no departures, we'll have a `null` `mode` and we'll require human eyes to go in and alter the config. Running it across the current production, there were only a handful of Orange Line sections missing departures (I guess because of the current disruption). 

I went through and changed them manually in `dev-blue`. I could probably add a warning log to just notify that we do have these problems.

scripts/add_mode_to_departures_params.exs
deanshi: This was just a quick hack to prevent timeouts - I was seeing retry issues and ultimately timeout failures when I was running it with our current retry behavior.

scripts/add_mode_to_departures_params.exs
deanshi: I'll move the `AddModeToDeparturesConfig` moduledoc here instead (and elaborate on the script since I think it applies more for the script than the module itself.

deanshi: Also, the `dry-run` option made more sense in my head while I was building the script. I think now that it's more or less done, I can remove this option and we can just generate the file.

scripts/add_mode_to_departures_params.exs
rwaskiewicz: Gotcha - that seems to work! Is there a good way to compare the output from before/after? It looks like the order of keys in `screens` (maybe others too, unsure) changes when running this script, which makes it a little trickier to see the before/after on a screen by screen basis

deanshi: Let me see if I can maintain order while doing this work so we can do a proper diff between both.

deanshi: Okay, so I ended up just decoding/encoding the original config in order to match the order and I'm now writing it to a separate file, `priv/local.sorted.json`. You can run a diff on the two files and (hopefully) see that changes are only addding a `mode`. 

Good call though, this also helped me catch and issue where `header_only` sections weren't being re-added back in. That's also been fixed.

rwaskiewicz: Does this preserve order of keys for screens themselves? When I try to diff these locally, I see something like:

Screenshot 2026-08-25 at 13 18 59

Note: This is IntelliJ's diff tool, but I see the same with something like `diff priv/local.json priv/local.sorted.json `

deanshi: That it does not. I can look into trying to keep the order, I think this seemingly is coming from my use of Maps/Jason encoding and decoding so let me see if I can figure that out if we want to cut out the diff noise.

deanshi: Okay - so I spent an embarrassingly long time yesterday trying to figure out why the keys remain the same when we validate on admin versus running it in the script (going down the rabbit hole of `Config.to_json` and `Screen.to_json`). 

Turns out, just switching the `screens-config-lib` to the latest version (with modes added) causes the validation to swap the keys even without changing anything. 

I think maybe the best bet here is just to use `Config.to_json` and `Config.from_json` (similar to how we do it within `admin_api_controller.ex`) and confirm the updated config diff only adds the modes. Unless there's maybe a different way of going about this that I'm missing?

deanshi: Except when we run it through the `dev-blue` validator with my branch deployed, it seems to revert to the original style. Something silly is going on...

rwaskiewicz (approved):
@deanshi Good old `jq` to the rescue for diffing 🎉 

```
cat priv/local.json | jq -S >> local.jq.json
cat priv/local.updated.json | jq -S >> local.updated.jq.json
diff local.jq.json local.updated.jq.json
```

The only non-blocking comment/suggestion is that I think we can stop writing those `local.sorted.json` and `local.updated.json` files now that this is looking good

deanshi:
> @deanshi Good old `jq` to the rescue for diffing 🎉

TIL about `jq`, awesome 🎉 . Let me revert those other files, appreciate the feedback and iteration!

PR #3257, opened 2026-08-24
refactor: Cast config from Postgres as Screen type (retry)

robbie-sundstrom:
I accidentally merged [this PR](https://github.com/mbta/screens/pull/3245) into a branch that has already been merged into main and deleted.

This PR has the same changes as the previous PR, but is now pointed to main:

This refactor will be very helpful for my upcoming PR to use Postgres for configs throughout the application.

Adds ScreenConfigType module that handles casting the column as a ScreenConfig.Screen type.
Consolidates the external API to use the same list_all method that the Admin API controller does. This just returns a the configs as JSON.
Adds a test helper to build configs and integrates into the existing controller tests.


rwaskiewicz (approved):
`git diff`ed this against `9f62eec0f019c10e096ede7590df5d32be4998bd` - results are the same, save for some dependency updates we did outside this PR

PR #3258, opened 2026-08-26
fix: use build identifier to detect deployments

rwaskiewicz:
**Asana task**: [Fix gaps in Screens client "deploy detection"](https://app.asana.com/1/15492006741476/project/1185117109217413/task/1217249394771512?focus=true)

Description

- [X] Tests added?

This commit introduces a new value, a build identifier, to detect when a new version of the backend service has been deployed.

The changes in this commit are intended to replace the existing last_deploy_timestamp mechanism. The existing mechanism has the following issues that we intend to remedy with this change:
1. Reliance on a file pushed to AWS S3 to retrieve the last_deploy_timestamp, which can be brittle if we fail to retrieve the file from S3 or fail to push it to S3 as a part of CI
2. There's additional overhead required to fetch and cache the timestamp from S3
3. There can be a gap in time between the update of the last deploy timestamp in S3 and when the application accepts connections, causing a mismatch in the understanding of the "latest" version of the app.

The mechanism for determining if the frontend's client is stale is similar to the existing last_deploy_timestamp mechanism in place today. The build identifier is sent to the server as a query string parameter, and if the value does not match the server's a reload is forced. A short git SHA is used as the build identifier.

This commit does not remove the existing last_timestamp checks in the code, as we will require deploys to devices such as DUPs to complete the transition.

### DIFF
--- a/assets/src/hooks/use_api_response.ts
+++ b/assets/src/hooks/use_api_response.ts
@@ -111,7 +111,9 @@ const useApiPath = (screenId: string, appendPath?: string): string => {
 
     const params: Record<string, string | null | undefined> = {
       is_real_screen: isRealScreen() ? "true" : null,
+      // TODO: Remove this following a complete rollout of build_identifier code
       last_refresh: getDatasetValue("lastRefresh"),
+      build_identifier: getDatasetValue("buildIdentifier"),
       requestor: getDatasetValue("requestor"),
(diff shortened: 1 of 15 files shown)


rwaskiewicz:
Closing, I'm going to see if we can do this w/o a DUPs + Mercury changes

PR #3260, opened 2026-08-26
chore: remove route_type and use mode

deanshi:
**Asana task**: [Screens departures: explicit "section modes"](https://app.asana.com/1/15492006741476/project/1185117109217413/task/1216571034745364?focus=true)

Description
- removed route_type from parameters and replaced its usage with mode (and the route type from the mode)
- updated tests to use mode in `Query.Params`

Note that this is dependent on https://github.com/mbta/screens-config-lib/pull/109 getting merged. The mix files will be updated once that gets merged, but right now it's just pointing to the branch. 

Note this this has been deployed to dev-blue and the config has been updated as well. 

- [x] Tests modified?


lib/screens/v2/departure.ex
```diff
@@ -66,22 +66,16 @@ defmodule Screens.V2.Departure do
     # route_types to filter on are configured in params AND options, we only include route_types
     # that are set in both.
     all_types = RouteType.all() |> Enum.sort()
-    param_types = (params[:route_type] || all_types) |> List.wrap() |> MapSet.new()
+
+    route_type =
+      Mode.to_route_type(params[:mode])
```
robbie-sundstrom: This could fit on one line

robbie-sundstrom: Looking at the implementation of this in `screens_config_lib`, there's no guard against `nil` being passed in. I've found at least one case of it being called with `nil` from `Screens.V2.CandidateGenerator.GlEink.LineMap.fetch_departures`. 
1. We should fix up that call, which should be fine to do by hardcoding `gl` as the mode. 
2.  I was thinking through how we could record an error if this function is called with nil instead of a valid route. That might need to be something on the `screen_configs_lib` side that returns an error of some sort for us to parse and log. Not sure if this is necessary since most cases should be calling this function from the screen's departure query params, where this actually will be required. Perhaps logging an error when `route_type in option_types` is false and we return `{:ok, []}` would suffice

lib/screens/v2/departure.ex
```diff
@@ -27,7 +27,7 @@ defmodule Screens.V2.Departure do
           optional(:direction_id) => Trip.direction() | :both,
           optional(:mode) => Mode.t(),
           optional(:route_ids) => [Route.id()],
-          optional(:route_type) => nil | RouteType.t() | [RouteType.t()],
+          optional(:mode) => nil | Mode.t(),
```
robbie-sundstrom: Line 30 is a slightly different duplicate of 28. Is there ever a valid case where we would expect and be ok with `mode` being `nil`?

scripts/screen_destinations.exs
robbie-sundstrom, line 1: This script seems to be broken in its current state.

lib/screens/v2/departure.ex
deanshi: Nope, I think this was a rebase error. I'll remove this one and keep the top.

lib/screens/v2/departure.ex
deanshi: I ended up just adding the warning into the `route_type` missing area. If we want more than that, I can look into adding a bit of error handling within the library itself but maybe that's a larger question in the end?

scripts/screen_destinations.exs
deanshi: Fixed up - it will end up using the version of the config where `route_type` doesn't exist and I ran it against the `dev-blue` config where that's currently true.

robbie-sundstrom (approved):

PR #3261, opened 2026-08-26
fix: use build identifier to detect deployments

rwaskiewicz:
**Asana task**: [Fix gaps in Screens client "deploy detection"](https://app.asana.com/1/15492006741476/project/1185117109217413/task/1217249394771512?focus=true)

Description

- [X] Tests added?

This commit introduces a new value, a build identifier, to detect when a new version of the backend service has been deployed.

The changes in this commit are intended to replace the existing last_deploy_timestamp mechanism. The existing mechanism has the following issues that we intend to remedy with this change:
1. Reliance on a file pushed to AWS S3 to retrieve the last_deploy_timestamp, which can be brittle if we fail to retrieve the file from S3 or fail to push it to S3 as a part of CI
2. There's additional overhead required to fetch and cache the timestamp from S3
3. There can be a gap in time between the update of the last deploy timestamp in S3 and when the application accepts connections, causing a mismatch in the understanding of the "latest" version of the app.

The mechanism for determining if the frontend's client is stale is similar to the existing last_deploy_timestamp mechanism in place today. The build identifier is sent to the server as a query string parameter, and if the value does not match the server's a reload is forced. A short git SHA is used as the build identifier.

This commit does not remove the existing last_timestamp checks in the code, as we will require deploys to devices such as DUPs to complete the transition.

**Notes**
- There's a small chance this uncovers a bug or two in how screens refresh - due to the time gap in how long it could be before we refresh the client. It's theoretically possible that shortening the time between deploys and giving clients the updated code could lead to 404's if we're somehow not ready to serve supporting files (HTML, JS). But if we're accepting traffic and we end up in that state, that feels like a separate bug that can be a fast follow (or we revert this and regroup)
- "Build Identifier" is a bit of a misnomer - it's left over from the original impl where I did use a git SHA. However, "Deploy Identifier" doesn't feel right either. Open to better naming suggestions
- This is my biggest concern - I don't fully understand how we route traffic/load balance yet - it's theoretically possible that if ECS instances start up at different times, this could cause some level of refresh thrashing:
    - Instance A: Starts up at 00:00:00.001
    - Screen PRE-123: Request gets routed to Instance A at 00:00:00.002, receives `last_refresh` of 00:00:00.001
    - Instance B: Starts up at 00:00:30.003 (for some reason it takes long enough to spin up)
    - Screen PRE-123: Request gets routed to Instance B at 00:00:31.003, receives `last_refresh` of 00:00:30.003


lib/screens_web/controllers/v2/screen_api_controller.ex
```diff
@@ -102,7 +102,6 @@ defmodule ScreensWeb.V2.ScreenApiController do
 
   # The packaged client can't reload the page to update itself; this would just reload the local
   # copy of the code, resulting in an infinite loop. Compatibility and versioning of the packaged
-  # client is handled manually.
```
deanshi, line 105: qq: was this comment removal intentional?

rwaskiewicz: Nope 😆 I'll put that back 🤦 

(My original impl had a new function clause below this comment block, must have gotten a little too aggressive with CMD+X)

rwaskiewicz: 80a06dd8

deanshi (commented):
Sorry, apologies for the late review. I was just thinking about the second issue you mentioned. 

I think in the ticket itself (and you mentioned it in the description) there was an example of using the Git SHA. Did you end up backing out of that implementation for the `DateTime.utc_now()` timestamp route?

rwaskiewicz:
@deanshi I did - I found that the git SHA was going to require Mercury involvement and a DUP deploy (assuming we add/rename the `last_refresh` field used in the API), an additional PR to transition smoothly from one to the other, and had a bit more complexity to get it into the build. I originally implement it this way, but it felt a little over-engineered for what we need. That said, I do have that branch still. There's one small bug that I need to fix (~30 min of work), but this impl here felt like the more pragmatic approach

deanshi:
> @deanshi I did - I found that the git SHA was going to require Mercury involvement and a DUP deploy (assuming we add/rename the `last_refresh` field used in the API), an additional PR to transition smoothly from one to the other, and had a bit more complexity to get it into the build. I originally implement it this way, but it felt a little over-engineered for what we need. That said, I do have that branch still. There's one small bug that I need to fix (~30 min of work), but this impl here felt like the more pragmatic approach

💯 with all of that I think I agree with this approach. I'm just thinking about how we might be able to get around the instance issue. I'm wondering if the thrashing is tolerable, but wondering also if there's an idea that I'm not thinking of.

rwaskiewicz:
> > @deanshi I did - I found that the git SHA was going to require Mercury involvement and a DUP deploy (assuming we add/rename the `last_refresh` field used in the API), an additional PR to transition smoothly from one to the other, and had a bit more complexity to get it into the build. I originally implement it this way, but it felt a little over-engineered for what we need. That said, I do have that branch still. There's one small bug that I need to fix (~30 min of work), but this impl here felt like the more pragmatic approach
> 
> 💯 with all of that I think I agree with this approach. I'm just thinking about how we might be able to get around the instance issue. I'm wondering if the thrashing is tolerable, but wondering also if there's an idea that I'm not thinking of.

I think the number of refreshes _should_ be limited to N instances at absolute worst. In my testing I only saw one case of this (and that was because the deployments lined up _just right_). With ~250 screens, we'll probably see this for every deploy in some way, but I'm not sure how recognizable it'll actually be.

deanshi (approved):
I think this makes sense and I'm curious if others might have an idea on any of the issues you brought up, but going to approve from my side!

rwaskiewicz:
I'm going to merge this, but if folks wanna continue the conversation, let's start a thread in `#screens-eng` in slack

PR #3262, opened 2026-08-27
fix: add countdown to gl-eink linemaps

rwaskiewicz:
**Asana task**: [Screens client-side countdowns](https://app.asana.com/1/15492006741476/project/1185117109217413/task/1216571034745361?focus=true)

Description

- [x] Tests added?

Add countdowns to the gl-eink linemaps, similar to the departure times as was done in b55b90ab. Common code has been extracted into a separate react hook and utility function.

### DIFF
--- a/assets/src/components/gl_eink/line_map.tsx
+++ b/assets/src/components/gl_eink/line_map.tsx
@@ -483,7 +483,7 @@ type Stop = {
 type Vehicle = {
   id: string;
   index: number;
-  label: { type: string; minutes: number } | { type: string, text: string}
+  label: { type: string; minutes: number } | { type: string; text: string };
   time_in_epoch: number;
 };
 
--- a/assets/src/components/gl_eink/line_map.tsx
+++ b/assets/src/components/gl_eink/line_map.tsx
@@ -483,7 +483,7 @@ type Stop = {
 type Vehicle = {
   id: string;
   index: number;
-  label: { type: string; minutes: number }; // TODO: Double check
+  label: { type: string; minutes: number } | { type: string, text: string}
   time_in_epoch: number;
 };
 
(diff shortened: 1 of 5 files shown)


assets/tests/util/timing.test.ts
```diff
@@ -0,0 +1,37 @@
...
+      new Date(2026, 7, 27, 13, 30, 45, 0).getTime() / 1000,
+    );
+    const currentDateTime = new Date(2026, 7, 27, 13, 30, 45, 0);
+
+    const actual = adjustMinute(departureTimeSeconds, currentDateTime);
+
+    expect(actual).toBe(1);
+  });
+
+  test("returns 1 when the the current datetime is after the departure time are the same", () => {
```
robbie-sundstrom: typo in test name
```suggestion
  test("returns 1 when the the current datetime is after the departure time, () => {
```

assets/src/components/gl_eink/line_map.tsx
```diff
@@ -466,7 +483,8 @@ type Stop = {
 type Vehicle = {
   id: string;
   index: number;
-  label: string;
+  label: { type: string; minutes: number }; // TODO: Double check
```
robbie-sundstrom: typing looks right to me, what are we double checking?

robbie-sundstrom (approved):
Good call pulling this out into a hook! Small comments, but looks good to me.

assets/tests/util/timing.test.ts
rwaskiewicz: Bah, copy pasta 🍝 Nice catch. Fixed in 5e8f59e6

assets/src/components/gl_eink/line_map.tsx
rwaskiewicz: Ah, I needed to double check the possible variations of that type - turns out there's one more variation of `label` i needed to cover - https://github.com/mbta/screens/pull/3262/commits/2183b8104efe4ecb6c46b570cda3797d8db31cf4

PR #3263, opened 2026-08-27
fix: Do not require CSRF for Screens client API

robbie-sundstrom:
Since this is a server-to-server API call, it doesn't really make sense for us to require a CSRF token on the Screens side when we already require bearer token authentication. The `:protect_from_forgery` Plug is specifically what checks for a CSRF token, so do not include that in the `pipeline` for the Screens Config API that will be used by Screenplay.


panentheos (approved):
Makes sense to me

PR #3264, opened 2026-08-28
fix: fixed cancelled prediction with schedules for RDS LCD

deanshi:
Description
- we get a %{time: nil, scheduled_time: nil} when we have a cancelled prediction along with a valid schedule
- we currently don't have a way to display that for LCDs like we do DUPs, so for now just filter them

I think this introduce some edge case problems (e.g. what happens if it's the only departure in the list) but putting this up right now for the crash fix while we discuss what the presentation of the actual cancelled departure should be (maybe just don't show it?)

### DIFF
--- a/lib/screens/v2/widget_instance/departures.ex
+++ b/lib/screens/v2/widget_instance/departures.ex
@@ -599,7 +599,18 @@ defmodule Screens.V2.WidgetInstance.Departures do
   end
 
   def serialize_times_with_crowding(departures, screen, now) do
-    Enum.map(departures, &serialize_time_with_crowding(&1, screen, now))
+    departures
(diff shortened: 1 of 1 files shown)


robbie-sundstrom (approved):
looks good! Just want to make sure we have a follow up ticket to discuss with design/Jim

deanshi:
> looks good! Just want to make sure we have a follow up ticket to discuss with design/Jim

[Ticket's been added](https://app.asana.com/1/15492006741476/project/1185117109217422/task/1217942259654707) onto our board!

digitalcora:
Noting after the fact: We already have a way to address this, the `include_scheduled_cancelled?` option for departure fetching. Screens that don't have a defined presentation for cancelled/skipped departures should not set this option.

PR #3265, opened 2026-08-28
feat: Screen Configs API to list by IDs

robbie-sundstrom:
**Asana task**: [Screenplay: Use shared Screen Config APIs](https://app.asana.com/1/15492006741476/project/1185117109217413/task/1216329820920791?focus=true)

Adds an option to the `api/screen_configs` for `ids` to filter and return a list of configs for the given screen IDs. 

While implementing the linked work for Screenplay, I realized that it would be simpler to just filter down to the screens we want on the API side rather than requesting the entire list of configs when we're modifying a single screen. 

Note that the tests for this were generated using Copilot, reviewed by myself, refactored by Copilot based on my suggestions, and then slightly modified by myself again.

### DIFF
--- a/lib/screens_web/controllers/screen_configs_api_controller.ex
+++ b/lib/screens_web/controllers/screen_configs_api_controller.ex
@@ -32,6 +32,6 @@ defmodule ScreensWeb.ScreenConfigsApiController do
   end
 
   defp parse_query_param_list(param) do
-    param |> String.split(",") |> Enum.map(&String.trim/1)
+    param |> String.split(",") |> Enum.map(&String.trim/1) |> Enum.reject(&(&1 == ""))
   end
 end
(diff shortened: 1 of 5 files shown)


lib/screens_web/controllers/screen_configs_api_controller.ex
```diff
@@ -24,4 +30,8 @@ defmodule ScreensWeb.ScreenConfigsApiController do
     |> put_status(400)
     |> json(%{success: false, error: "screen_configs parameter is required"})
   end
+
+  defp parse_query_param_list(param) do
+    param |> String.split(",") |> Enum.map(&String.trim/1)
```
rwaskiewicz: Should we also filter out empty strings at the end of this pipeline? Just in case we end up with a trailing comma or something (oddly) a bunch of commas back to back in the request?

rwaskiewicz (approved):
One non-blocking comment. Looks good!

lib/screens_web/controllers/screen_configs_api_controller.ex
robbie-sundstrom: good suggestion! Done in [bfd245d](https://github.com/mbta/screens/pull/3265/commits/bfd245d10258b73cba90c774c665ed26dfe52785)

PR #3266, opened 2026-08-28
fix: Crash in RDS caused by String direction_id

robbie-sundstrom:
For this [Sentry failure](https://mbtace.sentry.io/issues/7696007428/?alert_rule_id=9523677&alert_timestamp=1787854733911&alert_type=email&environment=screens-prod&notification_uuid=06bd755d-a826-4eb2-9471-91a55dc2b689&project=6061747&referrer=alert_email) 
that's caused by direction_id being a String. 

```
no function clause matching in Enum.at/3.
arg0:["Southbound", "Northbound"], arg1: "1" arg2:""
```

Ideally, this would be fixed in `screens_config_lib` with the below, but I think that might have downstream effects that shouldn't be pushed out at the end of a Friday. 

```ex
  defp value_from_json("direction_id", direction_id) when is_binary(direction_id) do
    String.to_integer(direction_id)
  end
```


rwaskiewicz (approved):

deanshi:
Just noting that I changed all of the `direction_ids` in production to be integers instead of being Strings. I'm wondering how we can enforce Integers within the config instead of allowing for both (or if we need to end up changing the `direction_id` type altogether.

PR #3273, opened 2026-09-03
feat: Recurring job to backup Screen Configs in S3

robbie-sundstrom:
**Asana task**: [Routinely backup Screen Configs in S3](https://app.asana.com/1/15492006741476/project/1185117109217413/task/1218058354689661?focus=true)

### Summary
Writes Screen Configurations from Postgres into S3 every minutes using a Postgres advisory lock to ensure that only one task will write a backup at a time.

### Notes for Reviewer
- Since this is slightly different than what’s implemented in `Screens.Config.Fetch`, I implemented this separately so that after migration, we can just delete the old module. This approach was simpler than mixing them together and then needing to untangle during cleanup.
- The advisory lock module (Screens.Repo.AdvisoryLock) is implemented generically so that we can use it again in the future for any jobs like this that we would want to run.
- This has been running in dev-blue with good results. See the JSON uploaded [here in S3](https://us-east-1.console.aws.amazon.com/s3/buckets/mbta-ctd-config?region=us-east-1&prefix=screens/latest/&tab=objects), and the image below showing logs that indicate

lock for key 1

### AI Disclosure 
I used Copilot with Claude for this work more generally than I had in months, since the AI pilot began. For transparency of the reviewer, here’s a summary of how that went. 

I gave a general overview of what I was looking for, and it created the general structure. Some things were only slightly modified from the original AI implementation:
- `Screens.Config.Backup.Store` and its associated implementations in `Local`, `S3`, and `Mixed` only had minor changes. This implementation was very effective because I pointed it directly to this existing pattern elsewhere and specified what I wanted different with local being able to specify either a local or S3 environment backup to fetch from, but only write locally. 
- `Screens.Config.Backup` only had slight changes, but the general structure stayed the same. Changes were mainly simplification of the JSON and removing edge cases that weren’t necessary.

Some pieces were completely scrapped and redone:
- Completely rewrote the advisory locking mechanism since it ignored my specifications and overcomplicated it. 
- `Screens.Config.Backup.Runner` was completely rewritten from what was first done. I initially specified this as a GenServer, but then once I realized it was completely stateless, I converted it into a `TaskSupervisor` using AI. I made many general edits after the fact,
- Redid the general naming and scheme of how configuration values were passed around 
- I actually rewrote most of the happy path test

### DIFF
--- a/config/test.exs
+++ b/config/test.exs
@@ -11,6 +11,12 @@ config :screens, Screens.Repo,
   database: "screens_test",
   pool: Ecto.Adapters.SQL.Sandbox
 
+config :screens, Screens.Config.Backup,
+  enabled: false,
+  interval_ms: 10,
+  local_backup_path: "local/config_backups/screens_test.json",
+  store: Screens.Config.Backup.Store.Local
+
 config :screens,
   config_fetcher: Screens.Config.Fetch.Local,
   local_config_file_spec: {:test, "config.json"},
--- a/config/test.exs
+++ b/config/test.exs
@@ -13,9 +13,10 @@ config :screens, Screens.Repo,
 
 config :screens, Screens.Config.Backup,
   enabled: false,
-  interval_ms: 10,
-  local_backup_path: "local/config_backups/screens_test.json",
   store: Screens.Config.Backup.Store.Local
+
+config :screens, Screens.Config.Backup.Store.Local,
+  local_backup_path: "local/config_backups/screens_test.json"
 
 config :screens,
   config_fetcher: Screens.Config.Fetch.Local,
--- a/lib/screens/config/backup/store/s3.ex
+++ b/lib/screens/config/backup/store/s3.ex
@@ -1,0 +1,40 @@
+defmodule Screens.Config.Backup.Store.S3 do
+  @moduledoc """
+  Functions to work with S3-hosted backups of the screen configs.
+  """
+
+  @behaviour Screens.Config.Backup.Store
+
+  @impl true
+  def fetch_backup(environment) do
+    get_operation = ExAws.S3.get_object(bucket(), backup_path(environment))
+
+    case ExAws.request(get_operation) do
+      {:ok, %{body: body, status_code: 200}} ->
+        {:ok, body}
+
+      err ->
+        Logster.warning(["s3_screen_configs_backup_fetch_error", inspect(err)])
+        :error
+    end
+  end
+
+  @impl true
+  def put_backup(file_contents) do
+    path = backup_path(Application.get_env(:screens, :environment_name))
+    put_operation = ExAws.S3.put_object(bucket(), path, file_contents)
+
+    case ExAws.request(put_operation) do
+      {:ok, %{status_code: 200}} ->
+        :ok
+
+      err ->
+        Logster.warning(["s3_screen_configs_backup_put_error", inspect(err)])
+        :error
+    end
+  end
+
+  defp bucket, do: Application.get_env(:screens, :config_s3_bucket)
+
+  defp backup_path(environment), do: "screens/latest/#{environment}.json"
+end
--- a/lib/screens/config/backup/store/s3.ex
+++ b/lib/screens/config/backup/store/s3.ex
@@ -2,6 +2,7 @@ defmodule Screens.Config.Backup.Store.S3 do
   @moduledoc """
   Functions to work with S3-hosted backups of the screen configs.
   """
+  alias Screens.Report
 
   @behaviour Screens.Config.Backup.Store
 
@@ -29,7 +30,7 @@ defmodule Screens.Config.Backup.Store.S3 do
         :ok
 
       err ->
-        Logster.warning(["s3_screen_configs_backup_put_error", inspect(err)])
+        Report.error("s3_screen_configs_backup_put_error", error: inspect(err))
         :error
     end
   end
--- a/lib/screens/repo/advisory_lock.ex
+++ b/lib/screens/repo/advisory_lock.ex
@@ -1,0 +1,45 @@
+defmodule Screens.Repo.AdvisoryLock do
+  @moduledoc """
+  Runs a function while holding a Postgres session-level advisory lock, so that work scheduled on
+  every app instance only runs on one of them at a time.
+
+  Can be reused for any recurring cron job type functions that only needs to run once across all tasks.
+  New uses of this locking mechanism just need to pass an  application-wide unique cron_lock_key
+  and an interval for how frequently the job should run.
+  """
+
+  alias Screens.Repo
+
+  @doc """
+  Runs `func` if the lock identified by `cron_lock_key` can be acquired immediately, returning `:locked` otherwise.
+  `cron_lock_key` must be a unique-per-job 64-bit integer. The lock is held for `interval` ms
+  regardless of how quickly `func` completes, so that another instance can't acquire it early.
+  """
+  @spec with_lock(integer(), non_neg_integer(), (-> result)) :: result
+        when result: term()
+  def with_lock(cron_lock_key, interval, func) do
+    Repo.transact(
+      fn ->
+        started_at = System.monotonic_time(:millisecond)
+
+        case Repo.query!("SELECT pg_try_advisory_xact_lock($1)", [cron_lock_key]) do
+          %{rows: [[true]]} ->
+            result = func.()
+            hold_lock_until(started_at, interval)
+            result
+
+          %{rows: [[false]]} ->
+            {:ok, :locked}
+        end
+      end,
+      timeout: interval
+    )
+  end
+
+  defp hold_lock_until(started_at, interval) do
+    # Sleep for 15 seconds less than the interval for running the locking function
+    # This makes sure that the lock is held for slightly less than the full interval
+    remaining = interval - (System.monotonic_time(:millisecond) - started_at) - :timer.seconds(15)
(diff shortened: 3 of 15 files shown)


config/test.exs
```diff
@@ -11,6 +11,12 @@ config :screens, Screens.Repo,
   database: "screens_test",
   pool: Ecto.Adapters.SQL.Sandbox
 
+config :screens, Screens.Config.Backup,
+  enabled: false,
+  interval_ms: 10,
```
rwaskiewicz: Non-blocking/no action required - I know this won't run since we set `enabled: false` - it might make sense to set this value to a high value if the intent is to have it never run. That way, if we accidentally enable this feature in tests, it doesn't run at this interval.

lib/screens/config/backup/store/s3.ex
```diff
@@ -0,0 +1,40 @@
...
+  """
+
+  @behaviour Screens.Config.Backup.Store
+
+  @impl true
+  def fetch_backup(environment) do
+    get_operation = ExAws.S3.get_object(bucket(), backup_path(environment))
+
+    case ExAws.request(get_operation) do
+      {:ok, %{body: body, status_code: 200}} ->
```
rwaskiewicz, line 14: In https://github.com/mbta/realtime_signs/pull/1069, we had to handle the case of AWS returning strings that were XHTML/empty with a 200 status code. I'm not sure if `ExAws` accounts for this under the hood (I know if has some implicit retry behavior, so it's possible) - if not, it may be worthwhile to expand the clause to check if `body` is a map or not

lib/screens/config/backup/store/s3.ex
```diff
@@ -0,0 +1,40 @@
...
+      {:ok, %{status_code: 200}} ->
+        :ok
+
+      err ->
+        Logster.warning(["s3_screen_configs_backup_put_error", inspect(err)])
+        :error
+    end
+  end
+
+  defp bucket, do: Application.get_env(:screens, :config_s3_bucket)
```
rwaskiewicz, line 38: For my own understanding, is there a reason we use `Application.get_env` here, but [use `compile_env!` in `local.ex`](https://github.com/mbta/screens/pull/3273/changes#diff-3aa6ee68f9ca0b8a8a667b95ca18ce01155439974eee1ca0cc58b68b01c9bf0dR10)? Is there a difference in how these will operate at runtime vs compile time based on how we use these config values?

lib/screens/repo/advisory_lock.ex
```diff
@@ -0,0 +1,45 @@
+defmodule Screens.Repo.AdvisoryLock do
+  @moduledoc """
+  Runs a function while holding a Postgres session-level advisory lock, so that work scheduled on
+  every app instance only runs on one of them at a time.
+
+  Can be reused for any recurring cron job type functions that only needs to run once across all tasks.
+  New uses of this locking mechanism just need to pass an  application-wide unique cron_lock_key
```
rwaskiewicz: The most non-blocking of nits:
```suggestion
  New uses of this locking mechanism just need to pass an application-wide unique cron_lock_key
```

lib/screens/repo/advisory_lock.ex
```diff
@@ -0,0 +1,45 @@
...
+          %{rows: [[false]]} ->
+            {:ok, :locked}
+        end
+      end,
+      timeout: interval
+    )
+  end
+
+  defp hold_lock_until(started_at, interval) do
+    # Sleep for 15 seconds less than the interval for running the locking function
```
rwaskiewicz: Prior to reading this comment, I would have assumed that the lock was guaranteed(-ish) to have been held by the whole `interval` passed to `with_lock`. For my own understanding, why doe we want to hold the lock for less that the full interval?

lib/screens/repo/advisory_lock.ex
```diff
@@ -0,0 +1,45 @@
...
+  Runs `func` if the lock identified by `cron_lock_key` can be acquired immediately, returning `:locked` otherwise.
+  `cron_lock_key` must be a unique-per-job 64-bit integer. The lock is held for `interval` ms
+  regardless of how quickly `func` completes, so that another instance can't acquire it early.
+  """
+  @spec with_lock(integer(), non_neg_integer(), (-> result)) :: result
+        when result: term()
+  def with_lock(cron_lock_key, interval, func) do
+    Repo.transact(
+      fn ->
+        started_at = System.monotonic_time(:millisecond)
```
rwaskiewicz, line 27: Non-blocking - It might be worthwhile to log attempts to acquire a lock, success/failures, and relinquishing it (even if we do this on a `DEBUG` level). That way, we can reason about backup behaviors if we run into a situation where a lock is held for a long time - otherwise we end up having to connect to the DB to run SQL against it directly

rwaskiewicz (commented):
I wasn't able to get this running locally. After pulling the latest and running `mix deps.get && npm ci --prefix assets`, I got:
```
** (Mix) Could not start application screens: Screens.Application.start(:normal, []) returned an error: shutdown: failed to start child: :screens_config_cache_owner
    ** (EXIT) an exception was raised:
        ** (ArgumentError) the following keys must also be given when building struct ScreensConfig.Departures.Query.Params: [:mode]
```
It's not immediately obvious what's wrong here - it looks like it's related to the `mode`/`route` work that Dean recently, but I'm not sure why that's failing (your branch is based atop all of his work).

config/test.exs
robbie-sundstrom: I set this to a low value because the `interval_ms` is used to sleep before `Backup.run` returns a response (but after it writes the config backup). That test then waits for the `interval_ms` amount of time before completing. This logic could probably be rearranged so that it doesn't interfere in this way

lib/screens/repo/advisory_lock.ex
robbie-sundstrom: I just wanted this to be slightly less than the interval so that the lock is 100% guaranteed to be released by the time the next task begins to run. But now that I'm explaining it, I realize that `runner.ex` won't even run the next backup run until it receives a response. So this can definitely be removed without a problem

lib/screens/repo/advisory_lock.ex
robbie-sundstrom: I was thinking about this myself. It feels like logging all of those situations is just cluttering the logs more than necessary, but maybe just logging when the lock is acquired and relinquished would be the right way to go about it. That should give us the info we need, since we'll know that other tasks are finding it's locked if we don't get either of those messages

robbie-sundstrom:
@rwaskiewicz are you trying to run it with the config feature flag set to true? I'm wondering if having old config data in your Postgres would cause that issue on startup. I haven't seen this failure before though

lib/screens/config/backup/store/s3.ex
robbie-sundstrom: No this really should be `get_env` since it's within a function at runtime. I was playing around with the environment values in `local.ex` a lot, and I think I was mainly looking for the crash which I could've gotten with `fetch_env!` instead which is better for runtime. So basically this is just incorrect and I'll change to `get_env` with a default for the `local` setup

rwaskiewicz:
> @rwaskiewicz are you trying to run it with the config feature flag set to true? I'm wondering if having old config data in your Postgres would cause that issue on startup. I haven't seen this failure before though

I am (or, I'm running it without changing any config). I'll dig into that now and see what I can find

EDIT: Updating `local.json` seems to have done the trick

config/test.exs
robbie-sundstrom: I ended up mocking the module to sleep, so that it's not called in tests. See that change (with some other small ones) in this commit: [af586ed](https://github.com/mbta/screens/pull/3273/commits/af586ed6f9b1e4919e262bcc1552487b9930c7e2)

lib/screens/config/backup/store/s3.ex
robbie-sundstrom: I've looked into this, and if `ExAws`received an empty vody with a 200 status code, it would return that as `{:ok, body}` where body is still a binary string of an empty JSON. Checking for a map would not catch that, and actually `Jason.decode(body)` would return an error downstream in this case, which we do handle and log

lib/screens/repo/advisory_lock.ex
robbie-sundstrom: Added logs in [6313535](https://github.com/mbta/screens/pull/3273/commits/6313535372bfce98d5078b5697affa2cf4f240b9)

lib/screens/config/backup/store/s3.ex
```diff
@@ -0,0 +1,40 @@
...
+  def put_backup(file_contents) do
+    path = backup_path(Application.get_env(:screens, :environment_name))
+    put_operation = ExAws.S3.put_object(bucket(), path, file_contents)
+
+    case ExAws.request(put_operation) do
+      {:ok, %{status_code: 200}} ->
+        :ok
+
+      err ->
+        Logster.warning(["s3_screen_configs_backup_put_error", inspect(err)])
```
deanshi: Should this be a `Report` so we get a Sentry event associated with this?

deanshi (approved):
I _think_ everything's running fine locally, and if things are looking on `dev-blue` I'd be confident on that as well. One tiny question, but otherwise looks good to me.

lib/screens/config/backup/store/s3.ex
robbie-sundstrom: Good call, that makes sense! Will update once I finish my current work before merging

PR #3276, opened 2026-09-09
feat: Always show default mbta.com/alerts as alert URL

robbie-sundstrom:
**Asana task**: [Alert URL Overflow](https://app.asana.com/1/15492006741476/project/1185117109217413/task/1217722875659778?focus=true)

Revise Alert URL display override to no longer display the content of the Alert URL field.  Instead, [mbta.com/alerts](https://mbta.com/alerts) should always be shown.  This is applied to both the `Alerts` and `ReconstructedAlerts` widgets. 

`url` is also removed as a parsed field for the `Alert` struct, since we no longer use it anywhere. 

- [X] Tests REMOVED?


lib/screens/v2/web_link.ex
```diff
@@ -14,20 +14,6 @@ defmodule Screens.V2.WebLink do
...
-      |> String.replace(~r/^https?:\/\//i, "")
-      |> String.replace(~r/^www\./i, "")
-
-    if byte_size(cleaned_url) > 20 do
-      "mbta.com/alerts"
-    else
-      cleaned_url
-    end
-  end
+  # Default to alerts page instead of using the URL from Alert metadata
```
digitalcora, line 17: nit: this comment doesn't really make sense except in the context of code that is removed in this change; with the current/new state of the code, nothing *else* uses the URL from alert metadata, and it's not even parsed. I don't think it adds much to the understanding of this code and could be omitted.

digitalcora (approved):
Nice cleanup!

PR #3280, opened 2026-09-10
feat: No Service UI for LCD departures

deanshi:
**Asana task**: [RDS on LCDs: "No Service Today" Departure Rows](https://app.asana.com/1/15492006741476/project/1185117109217413/task/1213973277663098?focus=true)

Description
- added No Service UI for handling RDS No Service state

Screenshot 2026-09-10 at 11 03 49 AM
Screenshot 2026-09-10 at 11 04 09 AM
Screenshot 2026-09-10 at 11 44 51 AM

- [x] Tests modified?


lib/screens/v2/rds.ex
```diff
@@ -414,6 +409,28 @@ defmodule Screens.V2.RDS do
     end
   end
 
+  @spec combine_no_service_states_for_section([RDS.NoService.t()]) :: [RDS.NoService.t()]
+  defp combine_no_service_states_for_section([%NoService{} | _] = states) do
+    {state_destinations, routes, direction_id} =
+      get_combined_destinations_routes_direction_ids(states)
+
+    representative_route = List.first(routes)
```
robbie-sundstrom, line 417: I know this is a bit of an edge case, but shouldn't we be pulling out the shared headsign here, as I believe you did for the Service Ended work, instead of just picking the first route from the list? Why not pass the full list into `get_common_headsign`?

assets/css/lcd/departures/section.scss
```diff
@@ -38,26 +38,37 @@
...
 }
 
+.no-service-section,
 .overnight-section {
   & > .departures__notice-row {
     display: flex;
-    align-items: end;
+    align-items: flex-end;
     justify-content: space-between;
     margin-right: 24px;
```
robbie-sundstrom, line 48: I know this is from your previous PR, but I was trying to understand this CSS more fully, and it seems like `gap` would make more sense here now. `gap` would give a guaranteed minimum separation, while with `space-between` and `justify-content`, the elements are pushed to opposite edges, so the spacing depends on the available width. Since `.free-text` having `flex: 1` already pushes the status label rightward, `gap` should be all that's necessary

assets/css/lcd/departures/section.scss
```diff
@@ -38,26 +38,37 @@
...
+      font-weight: 600;
     }
-  }
 
-  & > .departures__notice-row--with-headsign {
-    .free-text {
-      font-weight: 600;
+    .free-text__line-container {
+      display: flex;
+      width: 100%;
```
robbie-sundstrom: isn't `width: 100%` redundant here because of the parent's flex?

lib/screens/v2/widget_instance/departures.ex
```diff
@@ -186,25 +190,33 @@ defmodule Screens.V2.WidgetInstance.Departures do
...
       |> Enum.map(&Route.icon(&1))
       |> Enum.uniq()
-
-    route_icon =
-      if length(route_pills) == 1 do
-        Enum.at(route_pills, 0)
-      else
-        nil
-      end
+      |> Enum.at(0)
```
robbie-sundstrom, line 207: Similar question to your handling of the headsign in the `RDS` module. Should this instead by an icon that represents the mode if we have more than one unique route? Or does that introduce more issues (i.e. 2/3 buses in a section have no service, but there is still an active route running)

test/screens/v2/candidate_generator/widgets/realtime_departures_test.exs
```diff
@@ -301,14 +324,37 @@ defmodule Screens.V2.CandidateGenerator.Widgets.RealtimeDeparturesTest do
...
-                     layout: %Layout{},
-                     rows: [
-                       %ScreensConfig.FreeTextLine{
-                         icon: :bus,
-                         text: ["No departures currently available"]
+                   %Screens.V2.WidgetInstance.Departures.NoServiceSection{
+                     header: %ScreensConfig.Departures.Header{},
+                     headsign: nil,
+                     routes: [
+                       %Screens.Routes.Route{
```
robbie-sundstrom: You could use the `RouteBuilder` helper to shorten these tests and make them easier to maintain in the future.

lib/screens/v2/rds.ex
deanshi: We actually get the first route for `ServiceEnded` as well, `[%ServiceEnded{routes: [route]} | _] = states` is what we use instead of getting the combined routes list and then getting the first from the list. 

But it's still a good question, let me double check that this is still valid but I think the idea here is that when we did `get_combined_destinations_routes_direction_ids`, we see if all of the routes have the same `direction_id` and then further down in `get_common_headsign`, if that direction ID exists AND all of the lines are the same, then we safely just use the first represented route to use `direction_name` and the `direction_id`.

assets/css/lcd/departures/section.scss
deanshi: Added gap, and it definitely added some space between the headsign/the end text which was well needed. Appreciate the callout!

lib/screens/v2/widget_instance/departures.ex
deanshi: Nope, this is actually a good call out. Currently none of the No Service work uses modes, as I was going to save them for a slightly larger refactor across all files a little bit later.

That being said, `Route.icon` actually already handles the mode generalization for us. So something like Green Line will just be a GL pill and Bus routes are always just the Bus icon rather than the pill. `RoutePill.serialize_for_departure` I think handles the Route specific serialization, but we don't use that.

robbie-sundstrom (approved):
looks good!

PR #3281, opened 2026-09-11
feat: Abbreviate alert text when overflow

robbie-sundstrom:
**Asana task**: [Apply station abbreviations to alert text](https://app.asana.com/1/15492006741476/project/1185117109217413/task/1217384386756785?focus=true)

Applies our existing automatic abbreviation rules to station names in the alert text when we detect the message overrunning its container by passing an abbreviated message from the backend in the field that contains text from OIOs. 

### Summary
- Creates new `useAutoSizeWithAbbreviation` hook, an extension of `useAutoSize`, that also takes a possible list of abbreviations
- Note that `remedy` and `issue` are used differently on the Front End for single screen alerts and flex takeover alerts. The input OIO text is used as `remedy` for the full screen alert, while `issue` is used for the flex zone.
  - I also created a new `Props` interface for the flex zone `ReconstructedAlert` to pull this out from the full page alert props, since these already had a few differences.
- I applied the header abbreviation to some possible alert text that is not from OIOs but I could hypothetically imagine overflowing.

**Abbreviation Logic**
- Only applies entire station name abbreviations; does not apply universal word abbreviations! We don't want to abbreviate something like "Street next to the park" to "St next to the pk".
- Abbreviate every station name that's found to its shortest version. 

Below is the alert in question that overflowed without this logic. On the full size alert page, I had to create very arbitrary mocked alerts to get it to overflow and need abbreviation. 
GREEN LINE

### AI Disclosure
I used AI to consolidate the logic in this commit: [Consolidate abbreviation logic into hook](https://github.com/mbta/screens/commit/32eac2a388f470c91d5a5f126f3264977f68b6ae). This only required small changes from what Copilot generated. 
I selected the three snippets of code I wanted to pull out the logic for and used the following prompt: "In all three of these places, we have similar logic to find if there is an abbreviation for a specific field (either issue or remedy) and then abbreviate as the final sizing step. Can we pull this out into a shared hook?"


assets/src/hooks/use_auto_size.ts
```diff
@@ -96,4 +96,38 @@ export const useHorizontalAutoSize = <T extends Value>(
   return useAutoSize(steps, key, hasOverflowX);
 };
 
+/**
+ * Like `useAutoSize`, but for content that may have an abbreviated fallback
```
rwaskiewicz: Non-blocking/no-action-required - the [JSDoc annotation `@link`](https://jsdoc.app/tags-inline-link) might be a better in terms of being able to right/control-click on `useAutoSize` to navigate to it. Generally, I find it's fine to only link the first instance (instead of littering the JSDoc with the same `@link` annotation over and over):

```suggestion
 * Like {@link useAutoSize}, but for content that may have an abbreviated fallback
```

assets/src/hooks/use_auto_size.ts
```diff
@@ -96,4 +96,38 @@ export const useHorizontalAutoSize = <T extends Value>(
...
+ * value. `values` is either a single (non-abbreviated) value, or a
```
rwaskiewicz: Non-blocking/no-action-required - I would consider using the `@param` annotation here, but in the case of this hook, they won't buy us much other than formatting in JSDoc renderers in editors.

Before:
Image

After:
Image

assets/src/components/reconstructed_takeover.tsx
```diff
@@ -6,8 +6,8 @@ import DisruptionDiagram, {
 import FreeText, { FreeTextType } from "./free_text";
 import { QRCodeSVG as QRCode } from "qrcode.react";
 
-interface ReconAlertProps {
-  issue: string | any; // shouldn't be "any"
+interface Props {
+  issue: string;
```
rwaskiewicz, line 10: 💙

assets/src/components/reconstructed_alert.tsx
```diff
@@ -1,10 +1,20 @@
...
 import { classWithModifier, classWithModifiers, imagePath } from "Util/utils";
 
-import { ReconAlertProps } from "./reconstructed_takeover";
+interface Props {
+  issue: string | string[];
+  location: string;
+  cause: string;
+  remedy: string;
+  show_alternate_route_text: boolean;
+  routes: any[]; // shouldn't be "any"
```
rwaskiewicz: Can we make this of type `Pill[]`? Or maybe even `ReadonlyArray<Pill>`? I'm more concerned about shrinking the type from `any` here though:

```suggestion
  routes: Pill[];
```
where we'd import `Pill` like so:
```diff
- import RoutePill, { routePillKey } from "Components/departures/route_pill";
+ import RoutePill, { Pill, routePillKey } from "Components/departures/route_pill";
`

File: assets/src/components/pre_fare_single_screen_alert.tsx [hunk 5]
```diff
@@ -122,7 +126,7 @@ const DownstreamLayout: ComponentType<DownstreamLayoutProps> = ({
     <DownstreamIssueSection endpoints={endpoints} />
     <RemedySection
       effect={effect}
-      remedy={remedy}
+      remedy={remedy[0]}
```
rwaskiewicz: Potential bug - `remedy` is typed as a `string | string[0]` here. Using `remedy[0]` when `remedy` is a string will result in the first character being displayed

lib/screens/v2/widget_instance/reconstructed_alert.ex
```diff
@@ -90,14 +91,14 @@ defmodule Screens.V2.WidgetInstance.ReconstructedAlert do
           # Unique to station closures
           optional(:unaffected_routes) => list(enriched_route()),
           optional(:location) => String.t() | nil,
-          optional(:remedy) => String.t() | nil,
+          optional(:remedy) => String.t() | list(String.t()) | nil,
```
rwaskiewicz, line 94: For my own understanding, does `single_screen_serialized_response` ever get used for audio output? I'm trying to trace how we get to 
https://github.com/mbta/screens/blob/78c0b3cceb5016fec1abc6d6727d4f413424cb84/lib/screens_web/views/v2/audio/reconstructed_alert_single_screen_view.ex#L177-L183

where I'm thinking about how `remedy` would "render" if it was a list. I can trace that back to the [`render` fn](https://github.com/mbta/screens/blob/78c0b3cceb5016fec1abc6d6727d4f413424cb84/lib/screens_web/views/v2/audio/reconstructed_alert_single_screen_view.ex#L7) in that same module and see [it's usage in this module](https://github.com/mbta/screens/blob/9411cab5de143db6835893b84851bea090afcc62/lib/screens/v2/widget_instance/reconstructed_alert.ex#L1467-L1472) and how we'd use [it in the `WidgetInstance` impl](https://github.com/mbta/screens/blob/9411cab5de143db6835893b84851bea090afcc62/lib/screens/v2/widget_instance/reconstructed_alert.ex#L1584), and so on, but I'm not sure what that `t` in the call to `audio_view` is.

robbie-sundstrom: I completely forgot about the audio view here, so thanks so much for catching. This does use the `serialized_response` from the widget, so we need to account for the possibility of lists here. Added changes to both views in [79a520f](https://github.com/mbta/screens/pull/3281/commits/79a520f82b9b1bb532aa19cd6d06e8cd50b443a8)

assets/src/components/pre_fare_single_screen_alert.tsx
robbie-sundstrom: good catch! fixed in [24eb3e7](https://github.com/mbta/screens/pull/3281/commits/24eb3e79147c39f288e5831a5fb105218a49e6fe)

assets/src/components/reconstructed_alert.tsx
robbie-sundstrom: I updated this and I believe it should work for every case. That said, I can't see why it was set to `any` in the first place, so hopefully I'm not missing something

assets/src/hooks/use_auto_size.ts
robbie-sundstrom: this is cleaner! Since we don't really use this format for JS docs in our codebase, I just didn't even think about the fact that it's definitely helpful here where we're describing each of the params

rwaskiewicz (approved):

PR #3282, opened 2026-09-11
fix: Config Sync - Add to DB connection timeout, consitent local env naming

robbie-sundstrom:


config/runtime.exs
```diff
@@ -18,11 +18,13 @@ unless config_env() == :test do
     api_v3_key: System.get_env("API_V3_KEY")
 end
 
+eb_env_name = System.fetch_env!("ENVIRONMENT_NAME")
```
digitalcora: We should add this to the `.envrc.template` with a value of `screens-local` already filled in, so running the app locally works out of the box.

robbie-sundstrom: whoops didn't hit save on that file before the commit. will add

config/runtime.exs
```diff
@@ -18,11 +18,13 @@ unless config_env() == :test do
     api_v3_key: System.get_env("API_V3_KEY")
 end
 
+eb_env_name = System.get_env("ENVIRONMENT_NAME", "screens-local")
```
digitalcora: nit: This was probably added long enough ago that the `eb` stands for [Elastic Beanstalk](https://aws.amazon.com/elasticbeanstalk/) — in any case, it's not very accurate anymore since this is now always set, even in non-AWS environments. It could probably just be `env_name` or something.

digitalcora (approved):

config/runtime.exs
robbie-sundstrom: I was trying to figure out what that eb stood for - will remove it

PR #3284, opened 2026-09-14
feat: update single platform alert text

deanshi:
**Asana task**: [Screens: update wording for single platform closures](https://app.asana.com/1/15492006741476/project/1185117109217413/task/1216970939148400?focus=true)

Description
- updated the single platform alert text from "`Station Name`: Trains skip `Platform Name`" platform to "Stop Skipped: `Station Name` (`Direction Name`)" (if direction name exists)

Screenshot 2026-09-14 at 12 00 04 PM
Screenshot 2026-09-14 at 12 42 23 PM

- [x] Tests modified?


lib/screens/v2/widget_instance/reconstructed_alert.ex
```diff
@@ -133,6 +133,18 @@ defmodule Screens.V2.WidgetInstance.ReconstructedAlert do
...
+  @route_direction_names %{
+    "Blue" => ["Westbound", "Eastbound"],
+    "Orange" => ["Southbound", "Northbound"],
+    "Red" => ["Southbound", "Northbound"],
+    "Green-B" => ["Westbound", "Eastbound"],
+    "Green-C" => ["Westbound", "Eastbound"],
+    "Green-D" => ["Westbound", "Eastbound"],
+    "Green-E" => ["Westbound", "Eastbound"],
+    "Green" => ["Westbound", "Eastbound"],
+    "Mattapan" => ["Outbound", "Inbound"]
```
digitalcora, line 145: Things like `Copley & West` currently exist nowhere in our shared transit data and so it's appropriate to hard-code them for now, but this new addition is a copy of information that does already exist in transit data (`Route.normalized_direction_names/1`) and we should use that if at all possible.

deanshi: I think the main reason I put this here (actually lifted from `subway_status.ex`) was because there didn't seem to be a clear way to get a full route from an Alert (without making a network call). I might have missed something, but there seemed to only be a `route_id` associated with the Alerts/Informed Entities, and so there was no easy way to use `Route.normalized_direction_names/1`.

Is there a way for us to get a route from an alert without needing to actually make a network call? I'd love to be wrong so I can use previously established code, I just didn't find anything.

digitalcora: > Is there a way for us to get a route from an alert without needing to actually make a network call?

Yes, we'd have to `include` routes when fetching alerts and then populate the `route` field of `InformedEntity`s as a full `Route` struct using the included data, similar to [what we previously did for stops](https://github.com/mbta/screens/pull/2885). This would require a larger refactor first so existing code that assumed the value was just an ID would (at minimum) instead access the `id` field on the struct.

The fact that this is actually the third time we've duplicated this exact data (previously in `WidgetInstance.ReconstructedAlert` and `WidgetInstance.SubwayStatus.Serialize.Utils`) strongly suggests we should centralize it as part of this work, maybe as a lookup function in `Routes.Route`. The IE refactor would be a substantial effort (though probably not *as* much as the stop refactor?) so we could push that off, but I think we should create a task for it at this point.

deanshi: Pulled into `Routes.Route` and created [this ticket](https://app.asana.com/1/15492006741476/project/1185117109217422/task/1218519976835648) to encompass the refactor work.

digitalcora (approved):

PR #3287, opened 2026-09-15
feat: Use Postgres conditionally for Screens Admin refresh action

robbie-sundstrom:
**Asana task**: [Screens Admin: Use Postgres within refresh action ](https://app.asana.com/1/15492006741476/project/1185117109217413/task/1218468880635241?focus=true)

When making all of the Screens Admin API endpoints conditionally use Postgres for the config based on the feature flag, I missed the `refresh` action triggered by clicking the "Schedule Page Reload" button. 

I've tested this locally with both the feature flag disabled and enabled.

- [X] Tests added?


lib/screens_web/controllers/admin_api_controller.ex
```diff
@@ -72,9 +72,8 @@ defmodule ScreensWeb.AdminApiController do
   end
 
   def refresh(conn, %{"screen_ids" => screen_ids}) do
-    fetch_config()
-    |> Config.schedule_refresh_for_screen_ids(screen_ids)
```
rwaskiewicz, line 76: Non-blocking - with the removal of this fn call, it looks like there won't be any more usages of it (at least according to https://github.com/search?q=org%3Ambta+schedule_refresh_for_screen_ids&type=code). Should we mark this in the config lib to be deleted? Admittedly I didn't go through the remaining Asana tasks, feel free to ignore if it's already marked

rwaskiewicz (approved):

lib/screens_web/controllers/admin_api_controller.ex
robbie-sundstrom: I was thinking of a follow up PR to `screens_config_lib` to remove once this is merged! Didn't make anything to track since it can essentially be wrapped into this story

robbie-sundstrom: https://github.com/mbta/screens-config-lib/pull/112

PR #3294, opened 2026-09-17
feat: Headways on LCDs

deanshi:
**Asana task**: [Display RDS Headways on all LCD screens](https://app.asana.com/1/15492006741476/project/1185117109217413/task/1213921912505441?focus=true)

Description
- add support for headways for LCDs using RDS
- created a new time type :headways due to the CSS differences between :status and headways
- slight refactor to `RoutePill.serialize_for_departure/3` to fix `serialize_for_audio_departure` line/route discrepancy

Screenshot 2026-09-16 at 3 28 20 PM
Screenshot 2026-09-16 at 4 19 01 PM

- [x] Tests modified?


lib/screens/v2/widget_instance/departures.ex
```diff
@@ -560,23 +571,43 @@ defmodule Screens.V2.WidgetInstance.Departures do
...
   defp serialize_departure_group(
-         [{line, direction_id, {lo, hi}, headsign} | _],
-         screen,
+         [{route, direction_id, {lo, hi}, headsign} | _],
+         %Screen{app_id: :dup_v2} = screen,
          _now,
          route_pill_serializer
        ) do
-    id = hash_and_encode(line.id <> headsign)
+    id = hash_and_encode(route.id <> headsign)
```
robbie-sundstrom: Should this be the same as the below case for LCDs: `id = hash_and_encode(route.id <> (headsign || ""))`

Or will this never be `nil` because we won't have a headway without a headsign on DUPs?

lib/screens/v2/candidate_generator/widgets/realtime_departures.ex
```diff
@@ -94,6 +94,20 @@ defmodule Screens.V2.CandidateGenerator.Widgets.RealtimeDepartures do
             {%NoServiceSection{} = no_service_section, _section} ->
               no_service_section
 
+            # The current headway section presentation is just a row with headways,
+            # we can reuse the NormalSection with the Headway row
```
robbie-sundstrom: Nit: but this is really that `The current headway section presentation is just one row with a headway value, so we can reuse a NormalSection with a single Headway row"?

lib/screens/v2/widget_instance/departures.ex
```diff
@@ -560,23 +571,43 @@ defmodule Screens.V2.WidgetInstance.Departures do
...
 
+  defp serialize_departure_group(
+         [{route, direction_id, {lo, hi}, headsign} | _],
+         screen,
+         _now,
+         route_pill_serializer
+       ) do
+    id = hash_and_encode(route.id <> (headsign || ""))
+
+    {base_headsign, variation} = headsign |> headsign_with_variation()
```
robbie-sundstrom: `headsign_with_variation` doesn't handle the case when `headsign` is `nil`, which is technically allowed by the typing.

lib/screens/v2/candidate_generator/widgets/realtime_departures.ex
```diff
@@ -153,36 +168,8 @@ defmodule Screens.V2.CandidateGenerator.Widgets.RealtimeDepartures do
...
-    |> filter_by_route_direction(route_directions)
-  end
-
-  defp filter_unsupported_rows(rows) do
-    Enum.filter(rows, fn
-      %Departure{} -> true
-      {%Schedule{}, _special_trip_type} -> true
-      _ -> false
-    end)
+    rows |> filter_by_route_direction(route_directions)
```
robbie-sundstrom: It doesn't look like `route_direction` will handle a headway row, which could cause a crash on mixed headway/countdown rows if direction filtering is enabled

lib/screens/v2/widget_instance/departures.ex
deanshi: Nope, I think we can add that fall back for DUPs as well just in case.

lib/screens/v2/candidate_generator/widgets/realtime_departures.ex
deanshi: Updated so we handle these now. The one issue is that it also shows the combined mode instead of the specific route (for example, it'll show GL instead of GL-E for something like at Kenmore). 

I'm cleaning this up in a separate bug ticket that I have in the sprint, and I'll add this to the list since it's requiring a slightly different refactor to allow for specific route pills.

robbie-sundstrom (approved):

PR #3295, opened 2026-09-18
fix: RDS bug fixes (overnight pills, showing route pills instead of combined pills, fixing bus pills always showing)

deanshi:
Description
Reviewer can look at each commit, as they should be separated by the bug they fix.

1. ~edf0e031d9f9492f3568712382d4f30c87607706 - Fixed a bug where if we were showing a combined `OvernightDepartures` for DUPs, we weren't displaying the route pills associated with the sections.~ This is actually the design spec, removed it/reverted it from the list. 
2. 0473a3c124b53c9b3e8384482ffc30c9dd108050 - Fixed an issue where we were showing a combined mode pill instead of a route pill for routes with finished service. This was because as we have it right now, `FreeText` doesn't support the `RoutePill`, only an `Icon`. This has been swapped for just using the NormalSection with an overnight row to show a specific route being in the Service Ended state, but we have also kept the OvernightSection in cases where we are combine multiple routes in the service ended state.
3. e0cadc27db7b84e8eb39dfd3de666f9d72ca75c3 - Fixed an issue where we were always showing the Bus mode pill for cases where we have No Data. This was due to our use of `route_ids` and `route_type` in the params to attempt to guess the correct route in cases where we have no data at all. Passing `nil` value to `Route.icon/1` ultimately forces the end result to be a Bus icon, causing the issue. This has been fixed by the usage of `mode`, which is now a required param, always selecting the correct icon now. 

- [x] Tests updated?


lib/screens/v2/candidate_generator/widgets/realtime_departures.ex
```diff
@@ -89,9 +89,18 @@ defmodule Screens.V2.CandidateGenerator.Widgets.RealtimeDepartures do
         sections:
           sections_data_with_sections_config
           |> Enum.map(fn
-            {%OvernightSection{} = overnight_section, _section} ->
+            {%OvernightSection{last_schedule: nil} = overnight_section, _section} ->
```
digitalcora, line 91: question: When is `last_schedule` `nil`?

test/screens/v2/candidate_generator/dup/departures_test.exs
```diff
@@ -143,10 +143,22 @@ defmodule Screens.V2.CandidateGenerator.Dup.DeparturesTest do
...
+        screen: config,
+        slot_names: [:main_content_reduced_one]
+      },
       %OvernightDepartures{routes: [:red], screen: config, slot_names: [:main_content_two]},
-      %OvernightDepartures{routes: [:red], screen: config, slot_names: [:main_content_reduced_two]}
+      %OvernightDepartures{
+        routes: [:red],
+        screen: config,
+        slot_names: [:main_content_reduced_two]
+      }
```
digitalcora, line 161: This and all above changes in this file seem to be formatting-only, and could be reverted to reduce diff noise.

lib/screens/routes/route.ex
```diff
@@ -151,6 +152,17 @@ defmodule Screens.Routes.Route do
...
+  @spec icon_from_mode(Mode.t()) :: icon()
+  def icon_from_mode(:bl), do: :blue
+  def icon_from_mode(:ol), do: :orange
+  def icon_from_mode(:rl), do: :red
+  def icon_from_mode(:m), do: :mattapan
+  def icon_from_mode(:gl), do: :green
+  def icon_from_mode(:sl), do: :silver
+  def icon_from_mode(:bus), do: :bus
+  def icon_from_mode(:cr), do: :cr
+  def icon_from_mode(:ferry), do: :ferry
```
digitalcora, line 164: I'm not sure if this ultimately belongs in `Route`, since this "mode" concept only tangentially involves routes (and route types). Perhaps ideally it would go in `screens-config` since that is where `Mode` is defined, but doing that would be a bit more involved and would thus hold up this bug fix from going out. Also there's some weirdness with `icon()` being defined here even though *it* doesn't directly relate to routes. This feels like a larger refactor waiting to happen but I don't know what it would look like yet. Ignore me, basically.

digitalcora (commented):
Could you explain the fix in edf0e031? It seems like it has some refactoring mixed in, or perhaps the fix somehow is part of the refactoring, but it's not trivial for me to tell what's happening.

deanshi:
> Could you explain the fix in [edf0e03](https://github.com/mbta/screens/commit/edf0e031d9f9492f3568712382d4f30c87607706)? It seems like it has some refactoring mixed in, or perhaps the fix somehow is part of the refactoring, but it's not trivial for me to tell what's happening.

The main fix here was to include routes in the `OvernightDepartures` in cases where all screens and sections were in the Service Ended state, but actually reviewing the design specs again leaving out the route pills is intended. I'm going to drop this commit and keep the others.

lib/screens/v2/candidate_generator/widgets/realtime_departures.ex
deanshi: The only time this happens is when we are combining multiple `ServiceEnded` states into one. It might make more sense to change this to `last_schedules` and just be a list of the last schedules for each of the destinations though?

lib/screens/routes/route.ex
deanshi: Yeah, it didn't feel quite right here and I feel like it doesn't feel quite right in `Mode` either. I know we have `route_pill.ex` that handles some serialization as well, but this doesn't feel like it should fit in there either with how we handle icons in `FreeTextLine` versus the full `RoutePill`, so I opted for this instead. Maybe this should just be inherently handled by `FreeTextLine` knowing how to serialize a `Mode`?

deanshi:
I'm going to close this PR and open two separate ones because of a comment Brooke made about the No Service having the same issue. I think this might require a slightly larger refactor that I'll handle in one PR rather than putting up commits here.

PR #3297, opened 2026-09-18
feat: Admin Tool to Sync Configs & Assets from Source Environment

robbie-sundstrom:
**Asana task**: [Admin Tool for Syncing Environments](https://app.asana.com/1/15492006741476/project/1185117109217413/task/1218058354689663?focus=true)

- A new tool on Screens Admins that will copy the snapshot of screen configurations from the chosen source environment to the current environment. 
- When running in a deployed environment, this will also copy S3 assets. 
- This has been deployed and tested in dev-green.

image


assets/src/components/admin/tools.tsx
```diff
@@ -116,4 +119,80 @@ const ImportConfigs = () => {
...
+      setEnvironment(environments[0] ?? "");
+    });
+  }, []);
+
+  const restore = async () => {
+    if (isRestoring || environment === "") return;
+
+    if (
+      !window.confirm(
+        `Overwrite all screen configurations and assets with the latest backup from "${environment}"? ` +
```
deanshi: super nit: could these two strings just be combined to begin with?

lib/screens/config/backup/assets.ex
```diff
@@ -0,0 +1,80 @@
+defmodule Screens.Config.Backup.Assets do
+  @moduledoc """
+  Copies screen assets (images, videos, etc.) from a source environment into the environment the
+  app is running in. When running locally, assets are written to a local directory instead of S3.
```
deanshi: Just to be clear - when we run it locally there are no assets written right?

assets/src/components/admin/tools.tsx
```diff
@@ -116,4 +119,80 @@ const ImportConfigs = () => {
...
+
+  // There will be no environments to restore from in production, so don't show this section.
+  if (environments.length === 0) return null;
+
+  // Display different prompt locally where we don't have assets in S3 to copy over
+  const configSyncPrompt =
+    "Replace all screen configurations in Postgres with the latest snapshot of the selected environment. ";
+  const assetSyncPrompt = ` Assets from the source environment will also be copied over in S3.`;
+  const fullBackupPrompt =
+    configSyncPrompt + (environments[0] === "local" ? "" : assetSyncPrompt);
```
deanshi: What's the use case for syncing from `screens-local`?

deanshi (commented):
Looks good - few small PR questions and also: 
1. Do we need tests for any of these specific cases? or do you think we're okay? 
2. The title of the PR is `Recurring job to backup Screens Configs in S3` whereas the ticket/work seems to be the Admin Tool. Is this right or am I missing code somewhere?

robbie-sundstrom:
@deanshi the PR title is incorrect! I did some rebasing and squashing of commits while the PR was open, so I must've named it after the incorrect commit. That's fixed now

assets/src/components/admin/tools.tsx
robbie-sundstrom: I really just included local for testing, but this will make even less sense to have for testing after the config storage migration is completed. It allows for testing the specific workflow of syncing from a latest snapshot that's done locally, but after the migration is complete, those snapshots will always pull from Postgres, so we'd just be writing the existing configs back to Postgres

I think it's worth removing just to remove that chance of confusion in the future

assets/src/components/admin/tools.tsx
robbie-sundstrom: Slightly modified this window prompt to add a `\n` new line for readability. Since adding that, I actually did keep this as two strings so the code matches what's displayed: [4797c14](https://github.com/mbta/screens/pull/3297/commits/4797c147d045edacef3f92c8a1497b03d3f7332a)

lib/screens/config/backup/assets.ex
robbie-sundstrom: Made the string conditional based on the `environmentName` [7795e2d](https://github.com/mbta/screens/pull/3297/commits/7795e2d23173f5e77d56095e93048449d1b99f5d)

deanshi: Sorry, just wanted to make sure, was this for this comment? 

Also I only ask because the module doc says `assets are written to a local directory instead of S3`.

deanshi (approved):
🚢

PR #3298, opened 2026-09-21
fix: route pill always showing bus for no data section

deanshi:
Description
- prior to modes, we were attempting to use the route_id/route_type to get the pill to show in cases where we were getting no data. this was leading to cases where if route_id was missing from params, we would show an incorrect bus pill
- this has been updated to use the new `mode` value in params, which is required and will always show the correct pill for the section

- [x] Tests updated?


digitalcora (approved):

digitalcora:
@deanshi Let's go ahead and deploy this to prod before EOD given the high visibility of this issue.

